PR #3, opened 2017-12-20
#88 #89 #90 set up the basic workflow classes with demo

todd-dembrey:
Includes a refactor of the code originally laid out in #2 (now closed) adds a demo of the basic app behaviour at `/apply/demo/`.


opentech/apply/views.py
```diff
@@ -1,3 +1,60 @@
...
+    workflow = workflow_class([BasicSubmissionForm] * wf)
+
+    current_phase = request.POST.get('current')
+    current = workflow.current(current_phase)
+
+    if request.POST:
+        if current.stage.name not in submission:
+            submitted_form = current.stage.form(request.POST)
+            if submitted_form.is_valid():
+                submission[current.stage.name] = submitted_form
```
kaedroho: I think ``session``/``logs`` would be better put on ``request.session`` rather than global variables

kaedroho:
I don't know much about this "workflows" feature, there doesn't seem to be any information on codebase about it. But the code generally looks fine to me (except for few issues I have with the ``demo_workflow`` view, but I assume that's only temporary).

PR #5, opened 2018-01-02
Feature/20 user accounts and groups

zerolab:
- Initial user model,
- groups and permissions pre-creation.
- FE login views stubs


opentech/settings/base.py
```diff
@@ -172,6 +173,9 @@
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
 MEDIA_URL = '/media/'
 
+AUTH_USER_MODEL = 'users.User'
```
todd-dembrey: This should point to `opentech.User`

opentech/users/migrations/0002_initial_data.py
```diff
@@ -0,0 +1,56 @@
...
+
+from opentech.users.groups import GROUPS
+
+
+def add_groups(apps, schema_editor):
+    # Workaround for https://code.djangoproject.com/ticket/23422
+    db_alias = schema_editor.connection.alias
+    try:
+        emit_post_migrate_signal(2, False, db_alias)
+    except TypeError:
```
todd-dembrey: We have control over the django version and will not be targeting django < 1.11 so dont need this code.

opentech/users/urls.py
```diff
@@ -0,0 +1,15 @@
+from django.conf.urls import url
+
+from django.contrib.auth import views as auth_views
+
+urlpatterns = [
+    url(r'^$', auth_views.login, {
+        'template_name': 'users/login.html'
+    }, name='login'),
```
todd-dembrey: You cant have two urls with the same name. any attempt to reverse will error·

opentech/urls.py
```diff
@@ -25,6 +25,7 @@
     url('^sitemap\.xml$', sitemap),
 
     url(r'^apply/', include(apply_urls)),
+    url(r'^user/', include(users_urls))
```
todd-dembrey: Should this be `account` instead of `user`. `user` seems developer focused.

opentech/users/groups.py
```diff
@@ -0,0 +1,35 @@
...
+    {
+        'name': 'Staff',
+        'permissions': [],
+    },
+    {
+        'name': 'Manager',
+        'permissions': [
+            'add_image', 'change_image', 'delete_image',
+            'add_document', 'change_document', 'delete_document',
+            'access_admin'
```
todd-dembrey: Can you split this so that there is one permission per line. make life easier in the long run.

opentech/users/urls.py
todd-dembrey: Also I'm not sure we need the first url at the moment, just the `/login` is enough.

zerolab: This is /user and /user/login, both requiring login. Will replace the first one with the dashboard view with the `login_required` when we start building the dashboard

opentech/users/migrations/0002_initial_data.py
zerolab: 👍

opentech/settings/base.py
zerolab: it is pointing to `app_label.ModelName` as required

opentech/urls.py
zerolab: let's discuss

opentech/settings/base.py
todd-dembrey: ah, my mistake in typing. I meant to say: `opentech.users.User`, `opentech.users` is what is in installed apps as the app_label. 

Does the above auth model definitely point at the custom auth model table?

zerolab: :nod:

opentech/urls.py
zerolab: Raised with Dan.

PR #7, opened 2018-01-04
features: 18, 19, 25, 26. Setup of Fund type and application forms

todd-dembrey:
Adds the fundtype page and allows admins to create forms which are associated with the fund type


zerolab:
Travis failing with `RuntimeError: Model class opentech.apply.models.ApplyHomePage doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`

todd-dembrey:
@zerolab Fixed that yesterday. All passing now.

opentech/apply/models.py
```diff
@@ -1,3 +1,111 @@
...
+    ]
+
+    promote_panels = (
+        Page.promote_panels +
+        SocialFields.promote_panels +
+        ListingFields.promote_panels
+    )
+
+
+class FundPage(AbstractStreamForm):
```
zerolab: Can we call this fund type since it is a "meta" class?

opentech/apply/models.py
```diff
@@ -1,3 +1,111 @@
...
+    def __str__(self):
+        return self.name
+
+
+class Option(Orderable):
+    value = models.CharField(max_length=255)
+    category = ParentalKey('Category', related_name='options')
+
+
+class Category(ClusterableModel):
```
zerolab: Can the Category model and blocks go into a different app (taxonomy or stream forms)?

opentech/apply/models.py
```diff
@@ -1,3 +1,111 @@
...
+    def workflow_class(self):
+        return WORKFLOW_CLASS[self.get_workflow_display()]
+
+    content_panels = AbstractStreamForm.content_panels + [
+        FieldPanel('workflow'),
+        InlinePanel('forms', label="Forms"),
+    ]
+
+
+class FundPageForm(Orderable):
```
zerolab: Currently one can create Form pages under Funds. As well as creating Application Forms. Should this be happening?

opentech/apply/models.py
```diff
@@ -1,3 +1,111 @@
...
+    base_form_class = WorkflowFormAdminForm
+    WORKFLOWS = {
+        'single': SingleStage.name,
+        'double': DoubleStage.name,
+    }
+
+    workflow = models.CharField(choices=WORKFLOWS.items(), max_length=100, default='single')
+
+    def get_defined_fields(self):
+        return self.forms.all()[0].fields
```
zerolab: Adding as a note -- this currently returns the first form fields for a double stage WF fund

opentech/apply/models.py
todd-dembrey: I want to keep them out of taxonomies (I'm imagining this is more public facing data). This is not to say that I'm against grouping them elsewhere, but they do belong to the apply app in my mind.

stream forms is the generic implementation of streamfield based forms and would ideally end up back in wagtail core, so keeping it as unadulterated as possible is preferable to me.

opentech/apply/models.py
todd-dembrey: Yeah, I've been thinking of changing the name. Will do.

opentech/apply/models.py
todd-dembrey: No, i'll remove sub pages from Fund type

opentech/apply/models.py
zerolab: They are Taxonomy. Could add additional restrictions to hide from the public site. It is odd to have both "Categories" and "Taxonomy" menu items...

opentech/apply/blocks.py
```diff
@@ -21,6 +21,17 @@ def target_model(self):
 
 
 class CategoryQuestionBlock(FormFieldBlock):
+    # Overwrite field label and help text so we can defer to the category
+    # as required
+    field_label = CharBlock(
+        label=_('Label'),
+        required=False,
+        help_text=_('Leave blank to use the default on the category'),
```
zerolab: 'Leave blank to use the default Category label' has a nicer sound to it. And it feels category should be uppercase

opentech/apply/blocks.py
```diff
@@ -21,6 +21,17 @@ def target_model(self):
...
+    )
+    help_text = TextBlock(
+        required=False,
+        label=_('Leave blank to use the default on the category'),
```
zerolab: as above. 'Leave blank to use the default Category help text'

opentech/apply/models.py
```diff
@@ -100,7 +100,17 @@ class Option(Orderable):
 
 
 class Category(ClusterableModel):
+    """Used to manage the global select questions used in most of the application form
+    Also used in the front end by editors when writing about projects.
```
zerolab: This indicates Category lives in a grey area as it is not 100% in the apply app if used in the front-end by editors :)

zerolab:
Almost there. 👍

opentech/apply/models.py
todd-dembrey: As discussed. Agreed it does cross a boundary, however in my mind this is the home of the category (it is where is it configured). Other uses will just be views on this data.

zerolab: 👍

opentech/apply/models.py
zerolab: discussed offline. the additional changes work. so no need for further work here.

PR #8, opened 2018-01-04
Add doc strings into the workflow and fix the bug in single stage

todd-dembrey:
Updates to the workflow as per your comments @zerolab

### DIFF
--- a/opentech/apply/workflow.py
+++ b/opentech/apply/workflow.py
@@ -190,7 +190,7 @@ class Action:
     """
     Base Action class.
 
-    Actions return the Uhase within the current Stage which the workflow should progress to next.
+    Actions return the Phase within the current Stage which the workflow should progress to next.
     A value of `None` will allow the Stage to progress.
     """
     def __init__(self, name: str) -> None:
--- a/opentech/apply/workflow.py
+++ b/opentech/apply/workflow.py
@@ -6,7 +6,7 @@ from django.forms import Form
 from django.utils.text import slugify
 
 """
-This file defines classes which allow you to compose workflows based on the following strucutre:
+This file defines classes which allow you to compose workflows based on the following structure:
 
 Workflow -> Stage -> Phase -> Action
 """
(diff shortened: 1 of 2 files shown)


opentech/apply/workflow.py
```diff
@@ -5,8 +5,18 @@
 from django.forms import Form
 from django.utils.text import slugify
 
+"""
+This file defines classes which allow you to compose workflows based on the following strucutre:
```
zerolab: `s/strucutre/structure/`

opentech/apply/workflow.py
```diff
@@ -5,8 +5,18 @@
...
+
+Workflow -> Stage -> Phase -> Action
```
zerolab: Having read through the comments, it makes a lot of sense to start the definitions from Action and work upwards.

Stage/Phase/Action is used with mixed case in docblocks. Let's keep it consistent.

opentech/apply/workflow.py
```diff
@@ -5,8 +5,18 @@
...
+"""
+
 
 class Workflow:
+    """
+    Holds the stages of a workflow. When a stage finishes will return the next stage in the workflow
```
zerolab: nitpicking: 
> A Workflow is a collection of Stages an application goes through. When a Stage is complete, it will return the next Stage in the list or `None` if no such Stage exists.

opentech/apply/workflow.py
```diff
@@ -177,6 +201,7 @@ def process(self, phase: 'Phase') -> Union['Phase', None]:
 
 
 class NextPhaseAction(Action):
+    # Change to the next action in the current Stage
```
zerolab: ChangePhase/NextPhaseAction are part of the generic WF, as far as I can see it. Let's clearly demarcate the WF base and the actual OTF WF definitions

todd-dembrey: I moved the visual indicator down a few lines

opentech/apply/workflow.py
```diff
@@ -5,8 +5,28 @@
...
 from django.utils.text import slugify
 
+"""
+This file defines classes which allow you to compose workflows based on the following structure:
+
+Workflow -> Stage -> Phase -> Action
+"""
+
+
+def phase_name(stage: 'Stage', phase: Union['Phase', str], occurance: int) -> str:
```
zerolab: occurence :)

PR #10, opened 2018-01-05
Update cookiecutter build for final changes

todd-dembrey:
This should be the final changes from the cookiecutter template.


.travis.yml
```diff
@@ -32,9 +32,25 @@ before_script:
...
+  - nvm install 8
+
+  # Move into the static_src folder where we will compile the FE
+  - cd ./opentech/static_src
+
+  # Install node dependencies
+  - npm install --quiet
+
+  # Build the static files
+  - npm run build:prod
```
zerolab: what does that accomplish on Travis? we are not doing JS/visual regression tests

todd-dembrey: Currently it will confirm that it compiles successfully. 

We should probably add the collectstatic as part of the test phase to ensure we have all the static files available.

PR #13, opened 2018-01-05
feature:100 Add the optional header image to all pages and update quote block

todd-dembrey:
I've had to update all pages inheriting from BasePage to ensure they get updated panels


opentech/templates/base.html
```diff
@@ -69,7 +69,8 @@
         {% wagtailuserbar %}
 
         {# {% if page.header_image %} #}
-            <!-- <header class="header header--standard header--has-bg-image" style="background-image:url('')"> -->
+            {% image page.header_image original as header_image %}
```
chris-lawton: what do you think to using a resizing method here to save on bandwidth? (incase the user uploads a high-res img for example) - something along the lines of `fill-2560x410` (large screen x header height)

todd-dembrey: Great, was meant to ask you about that :+1:

PR #20, opened 2018-01-09
User: login, password change and reset

zerolab:


zerolab:
rebased and ready

opentech/apply/users/templates/users/change_password.html
```diff
@@ -0,0 +1,36 @@
...
+                {% for error in form.non_field_errors %}
+                <li class="error">{{ error }}</li>
+                {% endfor %}
+            </ul>
+        </div>
+    {% endif %}
+
+    {% if form.errors %}
+        <div class="messages">
+            {% if form.errors.items|length == 1 %}{% trans "Please correct the error below." %}{% else %}{% trans "Please correct the errors below." %}{% endif %}
```
todd-dembrey: Blocktrans provides a way to add pluralisation. Prefer to use this.

https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#blocktrans-template-tag

```
{% blocktrans count counter=list|length %}
There is only one {{ name }} object.
{% plural %}
There are {{ counter }} {{ name }} objects.
{% endblocktrans %}
```

zerolab: done

opentech/apply/users/urls.py
```diff
@@ -2,15 +2,60 @@
...
+        ),
+        name='password_change',
+    ),
+
+    # Password reset
+    url(r'^reset/$', auth_views.PasswordResetView.as_view(
+        template_name='users/password_reset/form.html',
+        email_template_name='users/password_reset/email.txt',
+        success_url=reverse_lazy('users:password_reset_done'),
+        form_class=PasswordResetForm
```
todd-dembrey: I'm not a fan of the wagtail password reset form, it leaks information about whether the user is registered with the site.

I would prefer that we accept all emails and redirect to a page with "email sent to: {{ email }} with further instructions."  which i think is what the standard django form does. 
https://docs.djangoproject.com/en/2.0/topics/auth/default/#django.contrib.auth.forms.PasswordResetForm 

https://github.com/django/django/blob/master/django/contrib/auth/forms.py#L285 Only checks you enter something that validates as an email address.

zerolab: 👍 
worth raising a Wagtail issue too

PR #21, opened 2018-01-09
WIP: Switch to pipenv

zerolab:


zerolab:
This now includes a better provisioning script

todd-dembrey:
Do we need to update the deployment script as well?

PR #22, opened 2018-01-09
Restructure project

todd-dembrey:
Totally restructured. adds a nice bit of namespacing and the apply folder has no dependencies on the public section of the project.


zerolab:
LGTM 👍

PR #23, opened 2018-01-10
Fund category block and template tweaks

zerolab:


zerolab:
you are right, sorted

PR #25, opened 2018-01-11
removing erroneous timing parameter from from mobile menu transition

chris-lawton:


zerolab:
Just a note -- we are not going to merge staging into master, so would be good to fold this into one into the nav branch.

PR #27, opened 2018-01-11
adding account button to mobile menu

chris-lawton:


opentech/public/navigation/templates/navigation/primarynav.html
```diff
@@ -17,5 +17,9 @@
         {% endfor %}
     </ul>
 </nav>
+<a href="/apply" class="link link--button-transparent link--mobile-standout">
```
zerolab: Could this be a reusable include? link with icon (or button with icon?) that would take the icon name and possibly additional classes as parameters?

zerolab: @todd-dembrey we'll need to discuss and agree where this links to.

chris-lawton: Yup, we can definitely look in to making it an include. Let's add it to the next sprint.

Re. the apply link - I should have left it as href="#" until you've agreed on link structure. I can change it to avoid confusion if you'd like?

todd-dembrey: Leaving as `#` works for now. 

We will be wanting to point to the apply homepage on all pages of the public site. 

If it is in the primary nav tempalte we should update the context here: https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/public/navigation/templatetags/navigation_tags.py#L16

else we will need a context processor to pass the page/url into all pages.

PR #33, opened 2018-01-11
feature/24-oauth

zerolab:
Add the ability to use Google OAuth for logins


zerolab:
note to self: this needs a test for the custom view decorator. Other than that, ready to go, IMHO

zerolab:
Ready for review

opentech/apply/users/templatetags/users_tags.py
```diff
@@ -0,0 +1,40 @@
+from django import template
+from django.conf import settings
+
+register = template.Library()
+
+
+@register.filter
+def backend_name(name):
```
todd-dembrey: Is there much value in listing the backends we aren't supporting at this point in time?

opentech/apply/users/views.py
```diff
@@ -12,3 +15,13 @@ def account(request):
     return render(request, 'users/account.html', {
         'show_change_password': password_management_enabled() and request.user.has_usable_password(),
     })
+
+
+@login_required(login_url=reverse_lazy('users:login'))
+@require_oauth_whitelist
```
todd-dembrey: You duplicate the logic for the oauth white listing in the decorator and in the tags. I would prefer to just use the version you have in the tags (moved to utils?) and combine that with the view decorator `user_passes_test`

https://docs.djangoproject.com/en/2.0/topics/auth/default/#django.contrib.auth.decorators.user_passes_test

opentech/apply/users/templatetags/users_tags.py
```diff
@@ -0,0 +1,40 @@
...
+def backend_class(backend):
+    return backend.replace('-', ' ')
+
+
+@register.simple_tag(takes_context=True)
+def can_use_oauth(context):
+    user = context.get('user')
+
+    try:
+        if settings.SOCIAL_AUTH_GOOGLE_OAUTH2_WHITELISTED_DOMAINS:
```
todd-dembrey: You don't need this check. if the `SOCIAL_AUTH_GOOGLE_OAUTH2_WHITELISTED_DOMAINS` isn't set then the `AttributeError` will catch it.

opentech/apply/users/templatetags/users_tags.py
```diff
@@ -0,0 +1,40 @@
...
+            for domain in settings.SOCIAL_AUTH_GOOGLE_OAUTH2_WHITELISTED_DOMAINS:
+                if user.email.endswith(f'@{domain}'):
+                    return True
```
todd-dembrey: alternative:
```python
try:
    domain = user.email.split('@')[-1]
    return domain in settings.SOCIAL_AUTH_GOOGLE_OAUTH2_WHITELISTED_DOMAINS
except AttributeError:
    # anonymous user or setting not defined
    pass
return False
```

opentech/settings/base.py
```diff
@@ -185,6 +191,12 @@
 LOGIN_URL = 'users:login'
 LOGIN_REDIRECT_URL = '/'
 
+AUTHENTICATION_BACKENDS = (
+    'social_core.backends.google.GoogleOAuth2',
+
+    'django.contrib.auth.backends.ModelBackend',
```
todd-dembrey: minor: extra spacing probably not required

opentech/apply/users/templatetags/users_tags.py
zerolab: Probably not, but I like to future-proof things.
Will remove all but Google/Facebook/Twitter

opentech/apply/users/views.py
zerolab: 👍 
was thinking about that.

zerolab: actually, cannot use `user_passes_test` as this is two-fold:
1. needs to be logged-in, hence `@login_required`
2. logged-in users that cannot use OAuth should see a 403, not a login page

opentech/apply/users/templatetags/users_tags.py
zerolab: 👍

opentech/apply/users/templatetags/users_tags.py
zerolab: ok, removed all

opentech/apply/users/views.py
todd-dembrey: Valid points. I think the changes you have made alleviate most of my concerns.

`login_required` is already pointing at settings.LOGIN_URL (`'users:login'`)

PR #44, opened 2018-01-17
Enable overwritting of django template forms and update apps to not clash

todd-dembrey:
We need to change one of the base form templates. django.forms will then conflict so i have updated the app labels to fix this.

https://docs.djangoproject.com/en/1.11/ref/forms/renderers/#overriding-built-in-widget-templates

@chris-lawton For reference


zerolab:
LGTM

@chris-lawton this allows you to customise the form element widgets in `opentech/public/forms/templates/django/forms/widgets/` or globally in `opentech/templates/django/forms/widgets/`

zerolab:
@todd-dembrey see tests fail

todd-dembrey:
I was going to recommend `opentech/templates/django/forms/widgets/` for this point in time.

zerolab:
`form_lanading_page.html`

todd-dembrey:
This should create a new table under the new app_name for the staging server. Seeing as we currently have no content in this table i'm happy for that to happen here. It may be worth clearing content types if we get issue.

PR #46, opened 2018-01-17
Allow categories to be required on application forms

zerolab:
All wagtailform fields have the option


opentech/apply/categories/blocks.py
```diff
@@ -34,6 +34,7 @@ class CategoryQuestionBlock(FormFieldBlock):
     )
     category = ModelChooserBlock('categories.Category')
     multi = BooleanBlock(label='Multi select', required=False)
+    required = BooleanBlock(label='Required', required=False)
```
todd-dembrey: Good spot, would prefer change is base class.

https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/apply/stream_forms/blocks.py#L47

zerolab: sorted. 704ada7

PR #47, opened 2018-01-17
Feature/106 create project page

todd-dembrey:


opentech/public/utils/models.py
```diff
@@ -259,3 +260,39 @@ class Meta:
...
+        ]),
+        PageChooserPanel('source', ['public_funds.FundPage', 'public_funds.LabPage']),
+    ]
+
+    class Meta(Orderable.Meta):
+        abstract = True
+
+
+class FundingMixin(models.Model):
+    '''Impliments the funding total calulation
```
zerolab: Impliments -> Implements
calulation -> calculation

opentech/public/utils/models.py
```diff
@@ -259,3 +260,39 @@ class Meta:
...
+
+    You still need to include the InlinePanel('funding') in the child class
```
zerolab: Is there no way of making this so one just needs to do `panels + FundingMixin.panels` ?

opentech/public/people/models.py
```diff
@@ -13,13 +13,12 @@
...
     InlinePanel,
     MultiFieldPanel,
-    PageChooserPanel,
     StreamFieldPanel
 )
 from wagtail.wagtailimages.edit_handlers import ImageChooserPanel
 
 from opentech.public.utils.blocks import StoryBlock
-from opentech.public.utils.models import BasePage
+from opentech.public.utils.models import BasePage, BaseFunding, FundingMixin
```
zerolab: since this is shared, should utils be... on the same level of apply and public?

todd-dembrey: It is only shared in the public section of the site so didn't want it higher. equally it seems a bit weird in the utils folder.

opentech/public/utils/models.py
todd-dembrey: That is the correct pattern. Thank you.

opentech/public/utils/models.py
```diff
@@ -259,3 +261,40 @@ class Meta:
...
+    class Meta(Orderable.Meta):
+        abstract = True
+
+
+class FundingMixin(models.Model):
+    '''Implements the funding total calculation
+
+    You still need to include the content panel in the child class
+    '''
+    content_panels = InlinePanel('funding', label="Funding"),
```
zerolab: needs to be `[InlinePanel('funding', label="Funding")]`

opentech/public/projects/models.py
```diff
@@ -0,0 +1,152 @@
...
+    panels = [
+        PageChooserPanel('page', 'projects.ProjectPage'),
+    ]
+
+
+class ProjectFunding(BaseFunding):
+    page = ParentalKey('ProjectPage', related_name='funding')
+
+
+class ProjectPage(FundingMixin, BasePage):
```
zerolab: This is missing the categories:
Core issues (select)
Objectives (select)
Beneficiaries (select)
Addressed problems (select)
Technology focus (select)
Regions (select)
Countries (select)

opentech/public/projects/widgets.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self, queryset, display):
+        self.queryset = queryset
+        self.display = display
+
+    def __iter__(self):
+        for choice in self.queryset.values_list(*self.display):
+            yield choice
+
+
+class LazyWidgets:
```
zerolab: These will need some tests.

todd-dembrey: Done.

opentech/public/projects/templates/projects/project_page.html
```diff
@@ -0,0 +1,47 @@
...
+                    {% endfor %}
+
+                {% endif %}
+            {% endwith %}
+
+        </div>
+    </div>
+    {% include "utils/includes/funding.html" %}
+
+    {% regroup page.category_options by category as categories %}
```
zerolab: So, the widget shows the options in the correct order when editing the project 👍 
The order is not preserved here. 👎

opentech/public/projects/models.py
```diff
@@ -0,0 +1,166 @@
...
+
+    content_panels = BasePage.content_panels + [
+        ImageChooserPanel('icon'),
+        FieldPanel('introduction'),
+        FieldPanel('status'),
+        StreamFieldPanel('body'),
+        InlinePanel('contact_details', label="Contact Details"),
+        InlinePanel('related_pages', label="Related Projects"),
+    ] + FundingMixin.content_panels + [
+        FieldPanel('categories', widget=CategoriesWidget),
```
zerolab: Make this collapsible and possibly collapsed by default?

Other than this/ FE order, :shipit:

todd-dembrey: image

todd-dembrey: Looks a little weird with the duplicate name, but not terrible.

opentech/public/projects/templates/projects/project_page.html
todd-dembrey: `regroup` preserves order. I have updated the query to ensure the correct order is passed into the template

PR #49, opened 2018-01-19
Feature/36 can associate a round with a date

todd-dembrey:


opentech/apply/funds/models.py
```diff
@@ -67,3 +96,73 @@ class ApplicationForm(models.Model):
...
+            error_message = mark_safe('Overlaps with the following rounds:<br> {}'.format(
+                '<br>'.join([f'{round.start_date} - {round.end_date}' for round in conflicting_rounds])
+            ))
+            error = {
+                'start_date': error_message,
+            }
+            if self.end_date:
+                error['end_date'] = error_message
+
+            raise ValidationError(error)
```
zerolab: Would be great to display this as an error on the model form, rather than 500. (unless I am missing something, this happens regardless of the `DEBUG` value)

opentech/apply/funds/models.py
```diff
@@ -41,11 +51,30 @@ def get_defined_fields(self):
...
         InlinePanel('forms', label="Forms"),
     ]
 
+    def serve(self, request):
+        if hasattr(request, 'is_preview'):
+            return super().serve(request)
+
+        # delegate to the open_round to use the latest form instances
+        request.show_page = True
+        return self.open_round.serve(request)
```
zerolab: The result of this is that after one submits the application form, they are taken to the round page which returns a 404

opentech/apply/funds/models.py
```diff
@@ -67,3 +96,73 @@ class ApplicationForm(models.Model):
...
+        MultiFieldPanel([
+            FieldRowPanel([
+                FieldPanel('start_date'),
+                FieldPanel('end_date'),
+            ]),
+        ], heading="Dates")
+    ]
+
+    def get_defined_fields(self):
+        # Only return the first form, will need updating for when working with 2 stage WF
```
zerolab: Make note that it does not handle changing forms per round. i.e. it inherits the form from the fund type.

todd-dembrey: I'm not too worried about this. The following tickets will look at start storing form data on the round object and this will be totally reworked.

opentech/apply/funds/models.py
todd-dembrey: It should show the form with an error. Can you provide a specific example where it doesnt.

opentech/apply/funds/models.py
zerolab: happy this was sorted by e77d60b37

opentech/apply/funds/models.py
```diff
@@ -143,7 +143,7 @@ def clean(self):
                 Q(end_date__gte=self.start_date)
             )
 
-        conflicting_rounds = Round.objects.sibling_of(self).filter(
+        conflicting_rounds = Round.objects.child_of(self.parent_page).filter(
```
zerolab: nice one!

opentech/apply/funds/models.py
```diff
@@ -67,3 +96,73 @@ class ApplicationForm(models.Model):
...
+                Q(end_date__gte=self.start_date)
+            )
+
+        conflicting_rounds = Round.objects.child_of(self.parent_page).filter(
+            conflict_query
+        ).exclude(id=self.id)
+
+        if conflicting_rounds.exists():
+            error_message = mark_safe('Overlaps with the following rounds:<br> {}'.format(
+                '<br>'.join([f'{round.start_date} - {round.end_date}' for round in conflicting_rounds])
```
zerolab: missed it the first time, but in testing it becomes apparent that this should include the round title as well. Or tweak the "Overlaps with the following rounds" message

PR #50, opened 2018-01-19
Add Tinymce as a richtext editor

todd-dembrey:
I will be implementing a bleaching process when i update the submission model.

https://django-bleach.readthedocs.io/en/latest/


zerolab:
the migration will need renumbering after #49 (or the other way around. let's discuss)

zerolab:
needs rebase / migration renumbering

PR #52, opened 2018-01-22
Feature/73 news page

chris-lawton:


opentech/public/news/templates/news/news_page.html
```diff
@@ -4,20 +4,33 @@
...
-    <h1>{{ page.title }}</h1>
-
-    {% for author in page.authors.all %}
-        <p><a href="{% pageurl author.author %}">{{ author.author }}</a><p>
-    {% endfor %}
-
-    <p>{{ page.introduction }}</p>
+    <h5 class="heading heading--author">
+        {{ page.display_date|date:"D, Y-m-d H:i" }}
+        {% if page.authors.all %}
```
todd-dembrey: I think this should be ok here, the QS is only small. For your reference: 
https://docs.djangoproject.com/en/2.0/ref/models/querysets/#when-querysets-are-evaluated
https://docs.djangoproject.com/en/2.0/ref/models/querysets/#django.db.models.query.QuerySet.exists

chris-lawton: cool, thanks Todd

PR #64, opened 2018-01-24
Feature/28 fix date default for fe

todd-dembrey:
It _should_ have been doing this anyway.  Perhaps it was trying to localise to US under the hood.

@chris-lawton There is a broader question of what date format we should be using. I'm guessing YYYY-MM-DD is the least likely to cause confusion generally, but we should probably discuss.

### DIFF
--- a/opentech/apply/stream_forms/blocks.py
+++ b/opentech/apply/stream_forms/blocks.py
@@ -158,6 +158,7 @@ class DatePickerInput(forms.DateInput):
             'data-date-format': 'yyyy-mm-dd',
         })
         kwargs['attrs'] = attrs
+        kwargs['format'] = '%Y-%m-%d'
         super().__init__(*args, **kwargs)
 
 


opentech/apply/stream_forms/blocks.py
```diff
@@ -158,6 +158,7 @@ def __init__(self, *args, **kwargs):
             'data-date-format': 'yyyy-mm-dd',
         })
         kwargs['attrs'] = attrs
+        kwargs['format'] = '%Y-%m-%d'
```
zerolab: Why not make this site-wide using `WAGTAIL_DATE_FORMAT` 
 - http://docs.wagtail.io/en/v1.13.1/topics/streamfield.html#dateblock

zerolab: could be useful to configure both date and datetime as per http://docs.wagtail.io/en/v1.13.1/advanced_topics/settings.html#date-and-datetime-inputs

todd-dembrey: Because we aren't in wagtail anymore

`Specifies the date and datetime format to be used in input fields in the Wagtail admin`

The mechanism we are actually using is:
https://github.com/django/django/blob/1.11.9/django/forms/widgets.py#L468
which _should_ fall through to using this value: https://github.com/django/django/blob/1.11.9/django/utils/formats.py#L22
(which is the same as the value I included above)

As i said i think the `formats.localize_input` is messing with this but didn't get to the bottom of it. 

I'll spend a bit more time looking at this when we know what the format required from  OTF is.

zerolab: fair enough. What I am trying to get at is this should be consistent throughout, so we need to tie it in with the wagtail date formats one way or another when we get to it

zerolab:
Let's get this in and revisit with further clarification on date formats

PR #65, opened 2018-01-24
Bugfix/ensure submission class on round

todd-dembrey:
Since the previous work was done based on submission being on the fund, submissions now occur on the round. This updates the submission to correctly create be created.


zerolab:
linted and merged manually

PR #66, opened 2018-01-25
30/31 Creates an account from a submission + sends confirmation email

zerolab:


opentech/apply/funds/blocks.py
```diff
@@ -140,4 +140,15 @@ class ValueBlock(MustIncludeFieldBlock):
     widget = forms.NumberInput
 
 
+class EmailBlock(MustIncludeFieldBlock):
+    name = 'email'
+    description = 'The applicant email address'
+    widget = forms.EmailInput
+
```
todd-dembrey:
```
    class Meta:
        icon = 'email'
```

opentech/apply/funds/blocks.py
```diff
@@ -140,4 +140,15 @@ class ValueBlock(MustIncludeFieldBlock):
...
+
+class FullNameBlock(MustIncludeFieldBlock):
+    name = 'full_name'
+    description = 'Full name'
```
todd-dembrey: Maybe...?
```
    class Meta:
        icon = 'user'
```

opentech/apply/funds/models.py
```diff
@@ -74,6 +77,18 @@ def next_deadline(self):
...
+                FieldRowPanel([
+                    FieldPanel('from_address', classname="col6"),
+                    FieldPanel('to_address', classname="col6"),
+                ]),
+                FieldPanel('subject'),
+                FieldPanel('confirmation_text_extra'),
+            ],
+            heading="Confirmation email",
+            classname="collapsible collapsed"
+        ),
```
todd-dembrey: Maybe a new tab for this? "email settings" 

Not essential, but my preference.

http://docs.wagtail.io/en/v1.13.1/advanced_topics/customisation/page_editing_interface.html#customising-the-tabbed-interface

opentech/apply/funds/models.py
```diff
@@ -143,12 +158,60 @@ def process_form_submission(self, form):
                 response = cleaned_data.pop(field.id)
                 cleaned_data[field.block.name] = response
 
+        if cleaned_data.get('email'):
```
todd-dembrey: If we are here then the required field "email" will exist so we don't need to check if an email is provided. 

I think the rest of this section should be a new method `create_user`.

opentech/apply/funds/models.py
```diff
@@ -143,12 +158,60 @@ def process_form_submission(self, form):
...
+                user_data = {
+                    'username': full_name if full_name else email,
+                    'email': email,
+                    'is_active': False,  # User needs to activate account
+                }
+
+                if full_name and hasattr(User, 'full_name'):
+                    user_data['full_name'] = full_name
+
+                user = User.objects.create_user(**user_data)
```
todd-dembrey: This should be the only part wrapped in the try except. 

```
try:
    user = User.objects.create_user(**user_data)
except IntegrityError:
    user = User.objects.get(email=email)
else:
    # send the activation
```

I think I would be happy to move the `send_activation_email` as part of our `create_user` method on the custom User.

Can you think of a reason not to?

opentech/apply/funds/models.py
```diff
@@ -143,12 +158,60 @@ def process_form_submission(self, form):
...
+
+                if user:
+                    # TODO: move to method on User?
+                    from opentech.apply.users.utils import send_activation_email
+                    send_activation_email(user)
+
+            except IntegrityError:
+                pass
+
+            self.send_confirmation_email(form, cleaned_data)
```
todd-dembrey: pass in `user` ?

opentech/apply/funds/models.py
```diff
@@ -143,12 +158,60 @@ def process_form_submission(self, form):
...
+    def send_confirmation_email(self, form, cleaned_data):
+        if not cleaned_data.get('email'):
+            return
+
+        from django.template.loader import render_to_string
+        from wagtail.wagtailadmin.utils import send_mail
+
+        email = cleaned_data.get('email')
+        fund = self.get_parent().specific
+        context = {
```
todd-dembrey: `to_address` not used?

opentech/apply/funds/models.py
```diff
@@ -143,12 +158,60 @@ def process_form_submission(self, form):
...
+
+            self.send_confirmation_email(form, cleaned_data)
+
         return self.get_submission_class().objects.create(
             form_data=cleaned_data,
             page=self.get_parent(),
             round=self,
         )
 
+    def send_confirmation_email(self, form, cleaned_data):
```
todd-dembrey: might be worth making this generic as we will need to apply this to labs. Can do this at a later date.

opentech/apply/users/views.py
```diff
@@ -20,8 +25,62 @@ def account(request):
...
+
+class ActivationView(TemplateView):
     """
-    Generic, empty view for the OAuth associations.
+    Inspired by https://github.com/ubernostrum/django-registration
     """
 
-    return TemplateResponse(request, 'users/oauth.html', {})
+    def get(self, request, *args, **kwargs):
+        user = self.activate(*args, **kwargs)
```
todd-dembrey: Not too happy with the way this reads. Would prefer something like below which means you don't need to deconstruct the class to understand.

```
user = self.get_user(uidb64)
if self.valid(user, token):
    self.activate(user)
    login(user)
    return redirect()
return redirect('invalid')
```

opentech/apply/funds/models.py
zerolab: the only reason against it is that the only use case for this is when the user is created through submission

todd-dembrey: I'm not sure i understand. You should still tighten up the try/except even if the except doesnt set the user.

opentech/apply/funds/models.py
zerolab: see `(email,)` in `send_mail()`

opentech/apply/users/views.py
zerolab: 👍

opentech/apply/funds/models.py
todd-dembrey: I'm talking about `fund.to_address` I'm assuming that is for cc/bcc?

zerolab: yes. that is handled by `AbstractEmailForm`

opentech/apply/funds/blocks.py
zerolab: done

opentech/apply/funds/blocks.py
zerolab: done

zerolab:
Superseded by #76

PR #67, opened 2018-01-25
Feature/123 create lab type page

todd-dembrey:
Labs will submit to the core application submission but doesnt require rounds

@zerolab You might be interested in the change to the way data gets built for submission


opentech/public/funds/models.py
```diff
@@ -135,6 +137,16 @@ def clean(self):
                 'lab_link': 'Please provide a way for applicants to apply',
             })
 
+        if self.lab_type and self.link_text:
+            raise ValidationError({
+                'link_text': 'You cant customise the text for internal lab pages, leave blank',
```
zerolab: "cant" -> "cannot" or "can\'t"

zerolab:
minor text change needed. otherwise LGTM

todd-dembrey:
@chris-lawton Can you please review the templates for feature 71.

opentech/public/funds/templates/public_funds/lab_page.html
```diff
@@ -3,14 +3,27 @@
 
 {% block content %}
     {% include "public_funds/includes/lab_apply_cta.html" %}
-    <div class="wrapper wrapper--flex">
+    <div class="wrapper">
         <section class="section section--main">
-            <h1>{{ page.icon }}{{ page.title }}</h1>
-            <h5>{{ page.introduction }}</h5>
+            <div class="media-box media-box--reverse">
```
zerolab: When no icon or intro are set, we'll get a funky empty set of divs. I say less markup, the better

todd-dembrey: Yeah, Chris should hopefully clear that up. It came from a stale branch so probably isnt correct.

PR #73, opened 2018-01-31
Feature/75 person page

chris-lawton:


opentech/public/people/templates/people/person_page.html
```diff
@@ -1,60 +1,61 @@
 {% extends "base.html" %}
-
+{% block page_title %}People{% endblock %}
```
todd-dembrey: If this is the page title it should probably show the persons name so users can identify different tabs.

opentech/public/utils/templates/utils/includes/funding.html
```diff
@@ -1,15 +1,16 @@
...
+        {% for funding in page.funding.all %}
+        <div class="fund-box__row">
+            <span>{{ funding.year }}</span>
+            <span>${{ funding.value }}</span>
+            <span>{{ funding.duration }} months</span>
+        </div>
+
+        <h4 class="heading heading--top-space heading--total-funding">Total Funding: <span>${{ page.total_funding }}</span></h4>
+        <a href="{% pageurl funding.source %}">{{ funding.source }}</a>
+        {% endfor %}
```
todd-dembrey: You should only show the total funding once. Rather than for each funding they have received.

opentech/public/people/templates/people/person_page.html
chris-lawton: The `<title>` for the page should still show the persons name as `{% block title %}` hasn't changed. See - https://otf-dev.torchboxapps.com/people/person/

Perhaps the naming of the blocks should be reviewed as we have `{% block title %}` and `{% block page_title %}`?

opentech/public/utils/templates/utils/includes/funding.html
chris-lawton: yup, I got you. Will correct now

martycoote:
Just code reviewed from an FE point of view. Looking good to me.

PR #75, opened 2018-01-31
Feature/124 login frontend

chris-lawton:


opentech/templates/base.html
```diff
@@ -120,9 +120,9 @@
                 </section>
 
                 <div class="header__button-container">
-                    <a href="#" class="button button--transparent button--contains-icons">
+                    <a href="{% url 'users:login' %}" class="button button--transparent button--contains-icons">
                         <svg class="icon icon--person"><use xlink:href="#person-icon"></use></svg>
-                        Log in
```
zerolab: could use wire the logout button to `{% url 'users:logout' %}` since that exists as well

todd-dembrey: For the public site, no, we wont be reasoning about the User's authentication for now. When we are working through the dashboard header we will add a logout.

opentech/static_src/src/sass/components/_form.scss
```diff
@@ -4,7 +4,7 @@
         border-bottom: 1px solid $color--white;
     }
 
-    &--login {
+    &--with-p-tags {
```
zerolab: could this be made general. i.e. `form p { ... }` ?

opentech/apply/users/templates/users/password_reset/form.html
```diff
@@ -17,7 +17,7 @@
             </div>
         {% endif %}
 
-        <form class="form form--login" method="post">
+        <form class="form form--with-p-tags" method="post">
             {% csrf_token %}
             <p>{{ form.email.label_tag }}</p>
```
zerolab: label_tag is `<label>` and it is best practice to use them for forms.

chris-lawton: Agreed, however the form is output with surrounding <p> tags [here](https://github.com/OpenTechFund/opentech.fund/pull/75/files/fa7f48d97b6578b2f1506bb21202bc9fae7604e1#diff-094070c446ed733c0fcc633f442f50bdR18) which are used for styling

opentech/static_src/src/sass/components/_form.scss
chris-lawton: Yes, potentially. We'd need to check what affect this would have where other forms are used - search, funds, footer signup etc

PR #76, opened 2018-02-01
Feature/30 add tests

todd-dembrey:
Rebased to include the lab pages. 

probably should close #66 in favour of this.


opentech/apply/funds/models.py
```diff
@@ -55,16 +61,31 @@ def process_form_submission(self, form):
...
+            user = form.user
+            cleaned_data['email'] = user.email
+            cleaned_data['full_name'] = user.get_full_name()
+        else:
+            User = get_user_model()
+            email = cleaned_data.get('email')
+            full_name = cleaned_data.get('full_name')
+            user = User.get_or_create_new(
+                email=email,
+                defaults={'full_name': full_name}
```
zerolab: 👍

opentech/apply/funds/models.py
```diff
@@ -55,16 +61,31 @@ def process_form_submission(self, form):
...
 
+        if form.user.is_authenticated():
+            user = form.user
+            cleaned_data['email'] = user.email
+            cleaned_data['full_name'] = user.get_full_name()
+        else:
+            User = get_user_model()
+            email = cleaned_data.get('email')
+            full_name = cleaned_data.get('full_name')
+            user = User.get_or_create_new(
```
zerolab: Is it worth adding a check that the user model has a `get_or_create_new` method? The test will fail if the user model is changed at a later point, so that's a decent safeguard. But something tells me ought to be defensive

opentech/apply/funds/models.py
```diff
@@ -275,6 +339,7 @@ def build_json_order_by(field):
 class ApplicationSubmission(AbstractFormSubmission):
     form_data = JSONField(encoder=DjangoJSONEncoder)
     round = models.ForeignKey('wagtailcore.Page', on_delete=models.CASCADE, related_name='submissions', null=True)
+    user = models.ForeignKey(get_user_model(), on_delete=models.SET_NULL, null=True)
```
zerolab: Should we enforce the user fk? I had it like that initially, but changed it just before the demo to allow empty values.

opentech/apply/users/forms.py
```diff
@@ -0,0 +1,12 @@
+from django import forms
+from django.utils.translation import ugettext_lazy as _
+
+from wagtail.wagtailusers.forms import UserEditForm, UserCreationForm
+
+
+class CustomUserEditForm(UserEditForm):
+    full_name = forms.CharField(label=_("Full name"), required=False, help_text=_("Leave blank to generate from First and Last name values."))
```
zerolab: I realise the help text is misleading. The full name will not be generated. From first/last. Maybe remove? Will submit a PR at a later point that enforces email as username and makes full_name required

zerolab:
Needs linting. 
As an aside, it is interesting that the tests have intermittent fails on one of the workflow tests.

todd-dembrey:
ooo, intermittent failures. Can you provide an example of this? 

The good/annoying side effect of using faker data.

zerolab:
I restarted the build, so can't point to anything. Next will show before restrting

opentech/apply/funds/models.py
todd-dembrey: You want to ensure there is no chance the anonymous user could fall through to this stage?

Perhaps we should be doing something more like this: https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.SET

opentech/apply/users/models.py
```diff
@@ -1,31 +1,31 @@
...
         user, created = super().get_or_create(defaults=defaults, **kwargs)
         if created:
             send_activation_email(user)
         return user
 
 
 class User(AbstractUser):
-    full_name = models.CharField(verbose_name='Full name', max_length=255, blank=True)
+    USERNAME_FIELD = 'email'
+    email = models.EmailField(_('email address'), unique=True)
```
zerolab: Is this needed? `AbstractUser` already defines the email field

opentech/apply/funds/models.py
```diff
@@ -275,6 +345,7 @@ def build_json_order_by(field):
 class ApplicationSubmission(AbstractFormSubmission):
     form_data = JSONField(encoder=DjangoJSONEncoder)
     round = models.ForeignKey('wagtailcore.Page', on_delete=models.CASCADE, related_name='submissions', null=True)
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, null=True)
```
zerolab: why not use `get_user_model()` as it is used in at least one other place above.

opentech/apply/users/models.py
```diff
@@ -1,5 +1,31 @@
...
+
+class UserManager(BaseUserManager):
+    def get_or_create(self, defaults, **kwargs):
+        defaults.update(is_active=False)
+
+        first_name, last_name = convert_full_name_to_parts(defaults.pop('full_name', ''))
+        defaults.update(first_name=first_name, last_name=last_name)
+
+        user, created = super().get_or_create(defaults=defaults, **kwargs)
+        if created:
```
zerolab: As discussed move this to a `create_and_notify()` method that defers to `get_or_create()`. The activation email is only sent to new users that come from submissions

opentech/apply/users/models.py
todd-dembrey: Adds the unique attribute. Prefer to the unique together trick you used previously.

zerolab: Fair enough. Being more explicit.

opentech/apply/funds/models.py
todd-dembrey: > When you define a foreign key or many-to-many relations to the user model, you should specify the custom model using the AUTH_USER_MODEL setting. 

We should be consistent. FK should use the settings and model references should use the `get_user model`.

`get_user model` is basically a nice wrapper for:
```
from django.utils.module_loading import import_string
import_string(settings.AUTH_USER_MODEL)
```
https://docs.djangoproject.com/en/2.0/topics/auth/customizing/#django.contrib.auth.get_user_model

zerolab: 👍 missed that from the docs.

zerolab:
LGTM

PR #87, opened 2018-02-08
adding project icon markup in a conditional so it won't render if the…

chris-lawton:
…re isn't one

### DIFF
--- a/opentech/public/funds/templates/public_funds/includes/project_listing.html
+++ b/opentech/public/funds/templates/public_funds/includes/project_listing.html
@@ -1,10 +1,13 @@
 {% load wagtailcore_tags wagtailimages_tags %}
 <a class="media-box" href="{% pageurl project %}">
-    <div class="media-box__images">
-        {% image project.icon max-210x235 as project_icon %}
-        <img class="media-box__image media-box__image--large" src="{{ project_icon.url }}" alt="{{project_icon.alt}}">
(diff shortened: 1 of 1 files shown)


opentech/public/funds/templates/public_funds/includes/project_listing.html
```diff
@@ -1,10 +1,13 @@
 {% load wagtailcore_tags wagtailimages_tags %}
 <a class="media-box" href="{% pageurl project %}">
-    <div class="media-box__images">
-        {% image project.icon max-210x235 as project_icon %}
-        <img class="media-box__image media-box__image--large" src="{{ project_icon.url }}" alt="{{project_icon.alt}}">
-        <img class="media-box__image media-box__image--small" src="{{ project_icon.url }}" alt="{{project_icon.alt}}">
-    </div>
+    {% if project.icon %}
```
zerolab: should this not display the [default icon](https://www.opentech.fund/sites/default/files/styles/otf_project_icon/public/default_images/project_fallback_icon.png?itok=nlmTpin8) if none is specified?

chris-lawton: I'm not sure to be honest - it's not covered in the design but would make sense if that's what shows when you click through to the project. I'll double check with everyone.

chris-lawton: we're going to add the default image in

PR #89, opened 2018-02-09
refactoring media boxes

chris-lawton:
refactoring media boxes so the large image is used as a background image, rather than an image tag, therefore eliminating the image wobble on hover and the need for an extra img tag

### DIFF
--- a/opentech/public/funds/templates/public_funds/includes/project_listing.html
+++ b/opentech/public/funds/templates/public_funds/includes/project_listing.html
@@ -1,17 +1,7 @@
 {% load wagtailcore_tags wagtailimages_tags %}
 <a class="media-box" href="{% pageurl project %}">
-    {% if project.icon %}
-        {% image project.icon max-210x235 as project_icon %}
-        <div class="media-box__image-container" style="background-image:url('{{ project_icon.url }}')">
-            <img class="media-box__image media-box__image--small" src="{{ project_icon.url }}" alt="{{project_icon.alt}}">
-        </div>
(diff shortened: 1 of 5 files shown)


opentech/public/funds/templates/public_funds/includes/project_listing.html
```diff
@@ -1,10 +1,18 @@
...
-    <div class="media-box__images">
+    {% if project.icon %}
         {% image project.icon max-210x235 as project_icon %}
-        <img class="media-box__image media-box__image--large" src="{{ project_icon.url }}" alt="{{project_icon.alt}}">
-        <img class="media-box__image media-box__image--small" src="{{ project_icon.url }}" alt="{{project_icon.alt}}">
-    </div>
+        <div class="media-box__image-container" style="background-image:url('{{ project_icon.url }}')">
+            <img class="media-box__image media-box__image--small" src="{{ project_icon.url }}" alt="{{project_icon.alt}}">
+        </div>
+    {% else %}
```
zerolab: this should probably be an include that you pass the icon and additional classes to. there are 3 places where it is included with a slight variation

chris-lawton: Yup, good shout. In `templates/includes`?

zerolab: sounds good

todd-dembrey:
@zerolab It all works fine. we tested locally

PR #90, opened 2018-02-09
Bugfix/sprint 4 review

todd-dembrey:
Some general tidy ups based on feedback from demoing locally


zerolab:
Test fail because of conflicting migrations, by the looks of it. :shipit: once sorted

PR #95, opened 2018-02-13
Feature/45 create submission detail page

todd-dembrey:
Render the user's response into the submission page.

All custom user responses are bleached on render. I've also rolled the rich text changes into this PR.


opentech/apply/funds/models.py
```diff
@@ -413,13 +414,42 @@ def build_json_order_by(field):
 
 
 class ApplicationSubmission(AbstractFormSubmission):
+    field_template = 'funds/includes/submission_field.html'
+
     form_data = JSONField(encoder=DjangoJSONEncoder)
+    form_fields = StreamField(CustomFormFieldsBlock())
```
todd-dembrey: Would appreciate a discussion around this. 

It is the easiest in terms of rendering the form later but does rely on a definition for the field remaining in the repo. We could perhaps fall back to a Charfield if no field can be found.

opentech/apply/funds/models.py
```diff
@@ -251,6 +252,7 @@ class Round(WorkflowStreamForm, SubmittableStreamForm):  # type: ignore
     parent_page_types = ['funds.FundType']
     subpage_types = []  # type: ignore
 
+    lead = models.ForeignKey(settings.AUTH_USER_MODEL, limit_choices_to={'is_staff': True})
```
zerolab: The round should inherit the lead from the Fund type. Is that covered somewhere else (i.e. another user story)?

opentech/apply/funds/models.py
zerolab: Just to confirm, we discussed having the fallback to Charfield and keeping things simple.

opentech/apply/funds/models.py
```diff
@@ -413,13 +416,42 @@ def build_json_order_by(field):
...
 
+    def render_answers(self):
+        context = {'fields': list()}  # type: ignore
+        for field in self.form_fields:
+            try:
+                data = self.form_data[field.id]
+            except KeyError:
+                pass  # It was a named field or a paragraph
+            else:
+                form_field = field.block.get_field(field.value)
```
zerolab: Move this block to separate method

opentech/apply/users/piplines.py
```diff
@@ -0,0 +1,8 @@
+from django.conf import settings
+
+
+def make_otf_staff(backend, user, response, *args, **kwargs):
```
zerolab: 👍

opentech/apply/users/piplines.py
```diff
@@ -0,0 +1,8 @@
...
+    _, email_domain = user.email.split('@')
+    if email_domain in settings.STAFF_EMAIL_DOMAINS:
+        user.is_staff = True
```
zerolab: `is_staff` grants users access to the Django admin. Do we want that? I see 2 alternatives:
1. flag on the user model `is_apply_staff` (to abstract it a bit. `is_otf_staff` works too)
2. add the users to the "Staff" group

opentech/apply/funds/blocks.py
```diff
@@ -39,6 +39,7 @@ class RichTextFieldBlock(TextFieldBlock):
     widget = TinyMCE(mce_attrs={
         'elementpath': False,
         'branding': False,
+        'toolbar1': 'undo redo | styleselect | bold italic | bullist numlist | link'
```
zerolab: the style select still includes all the possible options.
May be worth using https://www.tinymce.com/docs/configure/content-formatting/#style_formats

opentech/apply/funds/templates/funds/applicationsubmission_detail.html
```diff
@@ -0,0 +1,32 @@
+{% extends "base-admin.html" %}
```
zerolab: This needs to be `base-apply.html` now that #94 is merged

opentech/apply/funds/models.py
todd-dembrey: Yes, there is another story for this.

opentech/apply/funds/blocks.py
todd-dembrey: Great link, thanks. Couldn't find that before.

opentech/apply/dashboard/tables.py
```diff
@@ -9,14 +12,17 @@
 
 
 class DashboardTable(tables.Table):
+    title = tables.LinkColumn('funds:submission', args=[A('pk')], orderable=True)
     submit_time = tables.DateColumn(verbose_name="Submitted")
     status_name = tables.Column(verbose_name="Status")
     stage = tables.Column(verbose_name="Type")
     page = tables.Column(verbose_name="Fund")
+    lead = tables.Column(accessor='round.specific.lead', verbose_name='Lead')
```
zerolab: This fails with `Cannot query "LEAD NAME HERE": Must be "Page" instance.`

PR #97, opened 2018-02-14
Feature/156 add status to submission

todd-dembrey:
Adds the workflow and current status onto the application. Doesn't include the machinery to update it.

I've freed up the `workflow` attribute to return an instantiated workflow object which felt more natural. I may have overthought the workflow classes initially and how they contain forms. When we come to the progression of stages i will need to look into if this is even needed.


todd-dembrey:
There are migration conflicts with #95

opentech/apply/funds/models.py
```diff
@@ -412,14 +430,32 @@ def build_json_order_by(field):
...
-class ApplicationSubmission(AbstractFormSubmission):
+class ApplicationSubmission(WorkflowHelpers, AbstractFormSubmission):
     form_data = JSONField(encoder=DjangoJSONEncoder)
     page = models.ForeignKey('wagtailcore.Page', on_delete=models.PROTECT)
     round = models.ForeignKey('wagtailcore.Page', on_delete=models.PROTECT, related_name='submissions', null=True)
     user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, null=True)
 
+    # Workflow inherited from WorkflowHelpers
+    status = models.CharField(max_length=254)
+    status_name = models.CharField(max_length=50)
```
zerolab: Should this be an attribute? It is a display value. `def get_status_display` ?

todd-dembrey: I was wondering about this overnight. The reason I added the `status_name` was because of querying/filtering. We will likely have the same `status_name` e.g. "in review" but vastly different status e.g:
`<workflow 1>__<stage 5>__in_review`
`<workflow 2>__<stage 1>__in_review`

So i guess partly premature optimising. I have got some ideas about alternatives to this that i could try when actually implementing filters so happy to remove.

zerolab: That makes sense, but let's make the change when it is needed ;)

zerolab:
👍

PR #100, opened 2018-02-14
Limit the search to only pages on the public site

todd-dembrey:
Hides the fund forms and the round pages

@chris-lawton Should fix the issue we had the other day.


zerolab:
LGTM

PR #101, opened 2018-02-15
Feature/153 filter dashboard

todd-dembrey:
Adds basic filters for the dashboard. 

Based on django-select2, but expanded to use the multi-checkboxes from https://github.com/wasikuss/select2-multi-checkboxes (example: https://jsfiddle.net/wasikuss/7ak9skbb/)

Depending on merger order I will add other filters (see #95 #97)

### DIFF
--- a/opentech/apply/dashboard/tables.py
+++ b/opentech/apply/dashboard/tables.py
@@ -1,7 +1,7 @@
 import django_filters as filters
 import django_tables2 as tables
 
-from opentech.apply.funds.models import ApplicationSubmission
+from opentech.apply.funds.models import ApplicationSubmission, Round
 
 
 class DashboardTable(tables.Table):
@@ -19,7 +19,13 @@ class DashboardTable(tables.Table):
         return value.get_full_name()
 
 
+def get_used_rounds(request):
+    return Round.objects.filter(submissions__isnull=False).distinct()
+
+
 class SubmissionFilter(filters.FilterSet):
+    round = filters.ModelMultipleChoiceFilter(queryset=get_used_rounds)
+
     class Meta:
         model = ApplicationSubmission
         fields = ('round',)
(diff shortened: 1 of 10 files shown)


opentech/apply/dashboard/tables.py
```diff
@@ -13,3 +18,28 @@ class Meta:
...
+class Select2ModelMultipleChoiceFilter(filters.ModelMultipleChoiceFilter):
+    def __init__(self, *args, **kwargs):
+        label = kwargs.get('label')
+        kwargs.setdefault('widget', Select2MultiCheckboxesWidget(attrs={'data-placeholder': label}))
+        super().__init__(*args, **kwargs)
+
+
+class SubmissionFilter(filters.FilterSet):
+    round = Select2ModelMultipleChoiceFilter(queryset=get_used_rounds, label="Rounds")
+    page = Select2ModelMultipleChoiceFilter(queryset=get_used_funds, label='Funds')
```
zerolab: I see the reasoning behind using `page`, but `fund` makes a lot more sense especially when it comes to query parameters

opentech/apply/dashboard/widgets.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def build_attrs(self, *args, **kwargs):
+        attrs = super().build_attrs(*args, **kwargs)
+        attrs['class'] = attrs['class'].replace('django-select2', 'django-select2-checkboxes')
+        return attrs
+
+    @property
+    def media(self):
+        media = super().media
+        media.add_js([static('js/select2.multi-checkboxes.js'), static('js/django_select2-checkboxes.js')])
```
zerolab: Would be good to use the minified versions

opentech/apply/dashboard/tables.py
zerolab: missing lead filter

opentech/apply/dashboard/widgets.py
todd-dembrey: using https://jscompress.com/ i get the followin stats, so dont really feel a desperate need to minify.

`Stats: 47.9% compression, saving 1.22 kb`

PR #104, opened 2018-02-16
Feature/159 apply homepage fe

chris-lawton:

PR #105, opened 2018-02-16
Update the listing to only show funds/labs with open rounds

todd-dembrey:
@chris-lawton for reference as this might affect your styling work in #104


zerolab:
LGTM

zerolab:
LGTM

PR #106, opened 2018-02-16
Feature/148 submissions frontend

chris-lawton:


opentech/static_src/src/sass/apply/components/_table.scss
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        @include media-query($table-breakpoint) {
+            border-top: 0;
+            border-right: 0;
+            border-bottom: 2px solid $color--light-grey;
+            border-left: 0;
+        }
+
```
martycoote: Could the above two includes be combined into one?

chris-lawton: yup - good spot. Thanks!

PR #107, opened 2018-02-19
Feature/154 add search of submissions

todd-dembrey:
@zerolab I've managed to pull too much into the PR. Because of needing to build the searchable field i refactored some code surrounding it. This PR covers the following:

1. Add a searchable field to the submission and the relevant views. This was driven by search over the JSON provides unexpected results (e.g. searching for "email" returns everything because it is a key)  and we can prepare/limit data for searching better this way.
2. Move the creation of the user into the submission. The was driven by the fact that the submission should know about how it stores its own data internally and shouldn't rely on the form page to change the data passed to it. This meant that the email field doesn't exist for the form page to work with.
3. Use the block's render method to control how the block is displayed. This stops me messing around with prepare value and delegates this to the blocks who know about how they should display.


opentech/apply/stream_forms/templates/stream_forms/render_field.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    <h5>{{ value.field_label }}</h5>
+    <div>{% block data_display %}{{ data|default:"No response" }}{% endblock %}</div>
+</div>
```
todd-dembrey: @chris-lawton These are the changes to the field i mentioned. See also `render_list_field` and `render_unsafe_field`

opentech/apply/dashboard/views.py
```diff
@@ -16,5 +14,12 @@ class DashboardView(SingleTableMixin, FilterView):
     filterset_class = SubmissionFilter
 
 
-class SubmissionDetailView(DetailView):
```
zerolab: Should this replace the submission detail view?

opentech/apply/funds/models.py
```diff
@@ -482,44 +486,36 @@ def save(self, *args, **kwargs):
...
+            field.render(context={'data': data})
+            for data, field in self.data_and_fields()
+        ]
+        return mark_safe(''.join(fields))
+
+    def prepare_search_values(self):
+        excluded_fields = ['data', 'time', 'datetime', 'value', 'category', 'number']
+        for data, stream in self.data_and_fields():
+            if stream.block_type not in excluded_fields:
+                value = stream.block.get_searchable_content(stream.value, data)
```
zerolab: My gut reaction here was "we should handle the edge case where the block doesn't implement this", but then again it is defined on FormFieldBlock, so that's ok

opentech/apply/dashboard/tables.py
```diff
@@ -67,3 +70,34 @@ class SubmissionFilter(filters.FilterSet):
     class Meta:
         model = ApplicationSubmission
         fields = ('funds', 'round', 'status')
+
+
+class JSONSearchFilter(filters.CharFilter):
```
zerolab: Given you store the submission as text, is the JSONFilter needed?
What about using an [HStoreField](https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/fields/#django.contrib.postgres.fields.HStoreField) which has the benefit of a [values()](https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/fields/#values) lookup ?

opentech/apply/dashboard/views.py
todd-dembrey: SubmissionDetailView exists here. This shouldnt have been here.
https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/apply/funds/views.py#L9-L16

opentech/apply/dashboard/tables.py
todd-dembrey: Been considering it. Most of the data is going to be in a single hierarchy. There is a small element of nesting but i don't think we will really need to query over it.

opentech/apply/dashboard/views.py
zerolab: 👍 I thought it was odd

PR #110, opened 2018-02-21
Feature/152 submission page fe

chris-lawton:


opentech/apply/funds/templates/funds/applicationsubmission_detail.html
```diff
@@ -1,35 +1,65 @@
 {% extends "base-apply.html" %}
-
+{% block title %}{{ object.title }}{% endblock %}
+{% block body_class %}{% endblock %}
```
martycoote: Does this empty body class block need including?

opentech/static_src/src/sass/apply/components/_status-bar.scss
```diff
@@ -0,0 +1,178 @@
...
+            left: 0;
+            width: 20px;
+            height: 20px;
+            background: $color--dark-grey;
+            border: 5px solid $color--mid-grey;
+            border-radius: 50%;
+            content: '';
+        }
+
+        // last items dont have a dot
```
martycoote: Is this comment is opposite of the case? Should it be last item displays dot only?

PR #112, opened 2018-02-21
Share the navigation settings across the public part of apply

todd-dembrey:
This add the navigation settings to the top of the apply site for public pages


opentech/public/navigation/templatetags/navigation_tags.py
```diff
@@ -13,8 +13,9 @@
 @esi_inclusion_tag('navigation/primarynav.html')
 def primarynav(context):
     request = context['request']
+    site = context['PUBLIC_SITE']
```
zerolab: I am a big fan of graceful degradation. if `PUBLIC_SITE` is not set for any reason, can this defer to `request.site` ?

PR #117, opened 2018-02-23
Add a base class to uploadable media to reduce errors

todd-dembrey:
Image field wasnt being processed correctly.


zerolab:
How interesting. Any way to test this?

PR #126, opened 2018-02-28
Drop first/last names in favour of a single full_name field

zerolab:


opentech/apply/users/models.py
```diff
@@ -63,15 +46,46 @@ def get_or_create_and_notify(self, defaults=dict(), site=None, **kwargs):
...
+    is_staff = models.BooleanField(
+        verbose_name='staff status',
+        default=False,
+        help_text='Designates whether the user can log into this admin site.',
+    )
+    is_active = models.BooleanField(
+        verbose_name='active',
+        default=True,
+        help_text='Designates whether this user should be treated as active. '
+                  'Unselect this instead of deleting accounts.',
```
todd-dembrey: I'm not sure why you have copied over these two fields?

opentech/apply/users/models.py
```diff
@@ -63,15 +46,46 @@ def get_or_create_and_notify(self, defaults=dict(), site=None, **kwargs):
...
+        super().clean()
+        self.email = self.__class__.objects.normalize_email(self.email)
+
+    def get_full_name(self):
+        return self.full_name.strip()
+
+    def get_short_name(self):
+        return self.email
+
+    def email_user(self, subject, message, from_email=None, **kwargs):
```
todd-dembrey: This exists on the base class, no need to copy through.

opentech/apply/users/models.py
```diff
@@ -63,15 +46,46 @@ def get_or_create_and_notify(self, defaults=dict(), site=None, **kwargs):
...
 
+    class Meta:
+        ordering = ['id']
+
     def __str__(self):
         return self.get_full_name()
+
+    def clean(self):
+        super().clean()
+        self.email = self.__class__.objects.normalize_email(self.email)
```
todd-dembrey: a) You don't need to access this through the class. 

b) Do you need to do this? The email is normalised on `_create_user`.

opentech/apply/users/models.py
```diff
@@ -63,15 +46,46 @@ def get_or_create_and_notify(self, defaults=dict(), site=None, **kwargs):
...
-    email = models.EmailField(_('email address'), unique=True)
     REQUIRED_FIELDS = []
 
     # Remove the username field which is no longer used
     username = None
 
     objects = UserManager()
 
+    class Meta:
+        ordering = ['id']
```
todd-dembrey: Where was the ordering an issue?

opentech/apply/users/templates/wagtailusers/users/create.html
```diff
@@ -0,0 +1,26 @@
...
+        {% include "wagtailadmin/shared/field_as_li.html" with field=form.username_field %}
+    {% endif %}
+    {% include "wagtailadmin/shared/field_as_li.html" with field=form.email %}
+    {% include "wagtailadmin/shared/field_as_li.html" with field=form.full_name %}
+
+    {% comment %}
+        First/last name hidden input with dummy values because.. Wagtail admin
+        See opentech.apply.users.forms.CustomUserCreationForm
+    {% endcomment %}
+    {{ form.first_name }}
```
todd-dembrey: Does `{% include "wagtailadmin/shared/field_as_li.html" with field=form.first_name %}` render something?

Using  `extra_fields` block would be preferable to copying everything in.

https://github.com/wagtail/wagtail/blob/76975525f7a3757b280572165f8717cc4c13bba6/wagtail/users/templates/wagtailusers/users/create.html#L27

opentech/apply/users/forms.py
```diff
@@ -0,0 +1,26 @@
...
+        # HACK: Wagtail admin doesn't work with custom User models that do not have first/last name.
+        self.fields['first_name'].widget = forms.HiddenInput(attrs={'value': f"fn{self.instance.pk}"})
+        self.fields['last_name'].widget = forms.HiddenInput(attrs={'value': f"ln{self.instance.pk}"})
+
+
+class CustomUserCreationForm(UserCreationForm):
+    full_name = forms.CharField(label=_("Full name"), required=True)
+
+    def __init__(self, *args, **kwargs):
+        super(CustomUserCreationForm, self).__init__(*args, **kwargs)
```
todd-dembrey: python 3: we dont need the super with class and self.

opentech/apply/users/forms.py
```diff
@@ -0,0 +1,26 @@
+from django import forms
+from django.utils.translation import ugettext_lazy as _
+
+from wagtail.wagtailusers.forms import UserEditForm, UserCreationForm
+
+
+class CustomUserEditForm(UserEditForm):
```
todd-dembrey: Could you not just pull this out into a common class and then both forms inherit from it. 

would also prefer to be explicit about the content of the input: "Not used - see full_name," rather than something a little cryptic

opentech/apply/users/models.py
zerolab: base class is `AbstractBaseUser`

opentech/apply/users/models.py
zerolab: Wagtail admin

opentech/apply/users/templates/wagtailusers/users/create.html
zerolab: Yes. The field label. 
screen shot 2018-02-28 at 14 26 01

opentech/apply/users/forms.py
zerolab: 👍

opentech/apply/users/forms.py
zerolab: changed the contents of the input.

need to have the init in each class. Doing it on a common class doesn't make the change

zerolab: sorted, thanks!

opentech/apply/users/templates/wagtailusers/users/create.html
todd-dembrey: Could be pulled into a shared include, but not too fussed at this point now.

todd-dembrey:
Code looks good. 2 things, otherwise LGTM:

 Use an include for the fields

Are we happy this wont be too much of a pain on dev & staging (i.e no user will have a `full_name`)

zerolab:
If that is a pain, this can be split into 2 steps:
1. add full_name + run migration to populate it
2. drop the fname/lname fields

will sort it out

zerolab:
@todd-dembrey as discussed, we don't need to do split migration so this is ready

PR #127, opened 2018-02-28
191 user migration

zerolab:
This builds on #126.

Adds a migration command to migrate user accounts. The JSON is in the following format:

```json
{
  "UID":{
    "uid":"UID",
    "name":"NAME",
    "mail":"NAME@example.com",
    "roles":{
      "2":"authenticated user",
      "4":"proposer"
    },
    "field_otf_real_name":[]
  },
  "UID1":{
    "uid":"UID1",
    "name":"NAME",
    "mail":"NAME@example.com",
    "roles":{
      "2":"authenticated user",
      "4":"administrator"
    },
    "field_otf_real_name": {
      "und":[{
        "value":"First Last",
        "format":null,
        "safe_value":"First Last"
      }]
    }
  }
}
```


opentech/apply/users/management/commands/migrate_users.py
```diff
@@ -0,0 +1,78 @@
...
+        with options['source'] as json_data:
+            User = get_user_model()
+            users = json.load(json_data)
+
+            for uid in users:
+                user = users[uid]
+
+                full_name = self.get_full_name(user)
+                user_object, created = User.objects.get_or_create(
+                    email=user.get('mail'),
```
todd-dembrey: Wouldn't have reached to `get` here. I know that `None` would not validate but the `KeyError` might be informative.

opentech/apply/users/management/commands/migrate_users.py
```diff
@@ -0,0 +1,78 @@
...
+                operation = "Imported" if created else "Processed"
+
+                for group in self.get_user_groups(user):
+                    user_object.groups.add(group)
+
+                # Ensure uid is set
+                user_object.drupal_id = uid
+                user_object.save()
+
+                print(f"{operation} user {uid} ({full_name})")
```
todd-dembrey: > When you are using management commands and wish to provide console output, you should write to self.stdout and self.stderr, instead of printing to stdout and stderr directly. By using these proxies, it becomes much easier to test your custom command. 

https://docs.djangoproject.com/en/2.0/howto/custom-management-commands/#module-django.core.management

opentech/apply/users/management/commands/migrate_users.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def get_full_name(self, user):
+        full_name = user.get('field_otf_real_name', None)
+        try:
+            full_name = full_name['und'][0]['safe_value']
```
todd-dembrey: Can you add a comment for non-drupalers

opentech/apply/users/management/commands/migrate_users.py
zerolab: yep. keep doing this.

opentech/apply/users/management/commands/migrate_users.py
zerolab: user mail is always set...

opentech/apply/users/management/commands/migrate_users.py
```diff
@@ -0,0 +1,78 @@
...
+
+from opentech.apply.users.groups import STAFF_GROUP_NAME
+
+
+class Command(BaseCommand):
+    help = "User migration script. Requires a source JSON file."
+    groups = Group.objects.all()
+
+    def add_arguments(self, parser):
+        parser.add_argument('source', nargs='?', type=argparse.FileType('r'), help='Migration source JSON file')
```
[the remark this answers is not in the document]
zerolab: 👍

opentech/apply/users/management/commands/migrate_users.py
```diff
@@ -0,0 +1,78 @@
...
+                    defaults={
+                        'full_name': full_name,
+                        'drupal_id': uid,
+                    }
+                )
+
+                operation = "Imported" if created else "Processed"
+
+                for group in self.get_user_groups(user):
+                    user_object.groups.add(group)
```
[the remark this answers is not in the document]
zerolab: 👍

opentech/apply/users/management/commands/migrate_users.py
todd-dembrey: which means get needed even less :)

zerolab:
Rebased and ready

PR #128, opened 2018-02-28
Feature/179 add comment model and form

todd-dembrey:
This is going to be a little interesting as a review.

I've perhaps overstepped the boundary of this ticket by thinking through the need to have multiple forms easily submitting to a single endpoint.

The `DelegatedCreateView` class will basically return the parent view with it's context but updates the form with what ever data you just submitted so we can add errors etc. 

`SubmissionDetailView` can be extended to allow for multiple forms being submitted through the same end point. we delegate to different forms as required.


zerolab:
Other than the linter complaining, 👍 
As discussed in person, neither of us can think of a more elegant solution and this does the job

opentech/apply/funds/views.py
```diff
@@ -40,15 +49,70 @@ def get_context_data(self, **kwargs):
...
+
+class UpdateLeadView(DelegatedViewMixin, UpdateView):
+    model = ApplicationSubmission
+    form_class = UpdateSubmissionLeadForm
+    context_name = 'lead_form'
+
+    def form_valid(self, form):
+        old_lead = form.instance.lead
+        response = super().form_valid(form)
+        new_lead = form.instance.lead
```
zerolab: old lead and new lead are the same.

opentech/apply/funds/views.py
```diff
@@ -40,15 +49,70 @@ def get_context_data(self, **kwargs):
...
-class SubmissionDetailView(DetailView):
+class ProgressSubmissionView(DelegatedViewMixin, UpdateView):
+    model = ApplicationSubmission
+    form_class = ProgressSubmissionForm
+    context_name = 'progress_form'
+
+    def form_valid(self, form):
+        old_phase = form.instance.phase.name
+        response = super().form_valid(form)
+        new_phase = form.instance.phase.name
```
zerolab: I believe this is the same as the lead change

opentech/apply/funds/templates/funds/includes/update_lead_form.html
```diff
@@ -0,0 +1,5 @@
+<form method="post" id="update-lead-form">
+    {% csrf_token %}
+    {{ lead_form }}
```
zerolab: Would be good set the default value to the current lead, if any

todd-dembrey: I explicitly removed the current lead from the list. This was to prevent issues where you reassign the project back to current lead.

Perhaps we could update the label to be: "Update lead from {{ current lead }} to"

opentech/apply/funds/views.py
todd-dembrey: yes, i need to get the original object not the instance.

opentech/apply/funds/views.py
todd-dembrey: Save updates the status which is not changed on the instance until the save method happens, so `old_phase` =/= `new_phase` here.

https://github.com/OpenTechFund/opentech.fund/pull/128/files#diff-9635ffc975a463e0b4204eae3936ff9cR22

opentech/apply/funds/templates/funds/includes/status_bar.html
```diff
@@ -1,17 +1,26 @@
...
+                        status-bar__item--is-current
+                    {% elif phase.step < status.step %}
+                        status-bar__item--is-complete
+                    {% endif %}">
+            <span class="status-bar__tooltip"
+                {% if phase == status %}
+                    data-title="{{ status.name }}" aria-label="{{ status.name }}"
+                {% else %}
+                    data-title="{{ phase.name }}" aria-label="{{ phase.name }}"
+                {% endif %}
```
zerolab: this if feels redundant. when `phase == status` one can simply output `phase.name` as it doesn't make a difference.

opentech/apply/funds/views.py
zerolab: 👍

opentech/apply/funds/templates/funds/includes/status_bar.html
todd-dembrey: well... 

the follow proves equality (`phase==status`)
https://github.com/OpenTechFund/opentech.fund/pull/128/files#diff-75e081f2231c82e6d73565e9d6bca75aR192 

But this then displays the name as "outcome"
https://github.com/OpenTechFund/opentech.fund/pull/128/files#diff-75e081f2231c82e6d73565e9d6bca75aR187

so unless you use the actual phase rather than the iterated phase you never get the actual phase name. hence the switcheroo.

opentech/apply/funds/workflow.py
```diff
@@ -152,6 +169,45 @@ def next(self, current_phase: 'Phase'=None) -> 'Phase':
...
+
+        @property
+        def step(self) -> int:
+            return self.phases[0].step
+
+        @property
+        def name(self) -> str:
+            # Hardcode a name for multi-phased step - always outcome at the moment
+            if len(self.phases) > 1:
+                return 'Outcome'
```
zerolab: Worth making this a constant? e.g. `MULTISTEP_PHASE_NAME = 'Outcome'`

opentech/apply/funds/templates/funds/includes/status_bar.html
zerolab: make a note of it ;)

PR #129, opened 2018-03-01
Feature/171 progress application

todd-dembrey:
builds off of #128


opentech/apply/funds/workflow.py
```diff
@@ -152,6 +169,41 @@ def next(self, current_phase: 'Phase'=None) -> 'Phase':
...
+            self.phases = phases
+
+        @property
+        def step(self) -> int:
+            return self.phases[0].step
+
+        @property
+        def name(self) -> str:
+            if len(self.phases) > 1:
+                return 'Outcome'
```
zerolab: It is worth documenting what this is

PR #131, opened 2018-03-01
Feature/update django 2.0

todd-dembrey:
~The only thing i have found broken in the styling of filter elements on the submissions page. Will look into that this week.~


opentech/apply/users/migrations/0004_django2_update.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.0.2 on 2018-03-01 21:46
```
zerolab: This migration looks out of place.

requirements.txt
```diff
@@ -1,10 +1,11 @@
-Django==1.11.8
+Django==2.0.2
+djangorestframework==3.7.4
```
zerolab: are we using it yet?

todd-dembrey: No, but wagtail allows a broad range of versions (`djangorestframework>=3.1.3,<4.0`). This pins it to the django 2.0 compat version.
http://www.django-rest-framework.org/topics/release-notes/#374 

Might be an issue to raise with wagtail core.

opentech/apply/users/migrations/0004_django2_update.py
todd-dembrey: You're right...

requirements.txt
zerolab: I think it is worth it. I sense a minor release coming in not too far of a future containing all the little tweaks like that

zerolab:
Other than the rogue migration, LGTM

PR #133, opened 2018-03-05
Create the lead update form

todd-dembrey:
I needed to add lead to a few objects as it didn't exist everywhere it needed to. There is also a little bit of a refactor in the way forms are contributed to make it less repetitive.

@chris-lawton FYI


zerolab:
@todd-dembrey happy with the approach 👍

PR #134, opened 2018-03-05
Feature/48 172 status lead modal

chris-lawton:


opentech/static_src/src/sass/apply/components/_button.scss
```diff
@@ -42,6 +42,17 @@
         }
     }
 
+    &--white {
+        @include button($color--white, $color--light-blue);
+        display: inline-block;
+        color: $color--light-blue;
+        border: 1px solid $color--mid-grey;
+
+        &:hover {
```
martycoote: add '&:focus' to this block?

PR #135, opened 2018-03-05
Feature/179 180 182 187 comments actvity feed

chris-lawton:
contains FE work for comments styling and activity feed

### DIFF
--- a/opentech/apply/activity/templates/activity/include/listing_base.html
+++ b/opentech/apply/activity/templates/activity/include/listing_base.html
@@ -5,7 +5,7 @@
     <div class="feed__content">
         <div class="feed__meta">
             <p class="feed__label feed__label--{{ activity.type }} feed__label--mobile">{{ activity.type|capfirst }}</p>
-            <p class="feed__meta-item">{{ activity.timestamp|date:"m.d.y P e" }}</p>
+            <p class="feed__meta-item">{{ activity.timestamp|date:"m.d.y h:iA e" }}</p>
         </div>
         <div class="feed__heading">
             <h4 class="feed__name">{{ activity.user }}</h4>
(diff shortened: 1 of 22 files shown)


opentech/static_src/src/sass/apply/components/_activity-feed.scss
```diff
@@ -0,0 +1,40 @@
...
+        background: $color--white;
+        opacity: 1;
+    }
+
+    &__header {
+        padding: 0 20px;
+        background-color: $color--default;
+
+        @include media-query(tablet-portrait) {
+            padding: 0;
```
martycoote: This padding removal means the close icon is right up against the window edge until the wrapper--medium max-width applies at 1020px. screen grab attached.

otf-dev torchboxapps com_apply_submissions_

chris-lawton: Good spot - addressed here: https://github.com/OpenTechFund/opentech.fund/pull/135/commits/886dc8015b458ea87bb0d14075e50995d1eeb4bc

opentech/apply/activity/templates/activity/include/listing_base.html
```diff
@@ -10,7 +10,7 @@
         <div class="feed__heading">
             <h4 class="feed__name">{{ activity.user }}</h4>
             {% if submission_title %}
-                <h6>for</h6>
+                <h6>udpated</h6>
                 <h6><a href="{{ activity.submission.get_absolute_url }}">{{ activity.submission.title }}</a></h6>
```
zerolab: should this not be a single h6 tag?

zerolab: also, udpated -> updated

PR #140, opened 2018-03-06
Feature/183 add private comment

todd-dembrey:
Hiding of private comments for applicants will be handled in another PR


opentech/apply/activity/models.py
```diff
@@ -9,6 +9,14 @@
...
 }
 
+ALL = 'all'
+PRIVATE = 'private'
+
+VISIBILITY = {
+    ALL: 'All',
+    PRIVATE: 'Internal',
+}
+
```
zerolab: As discussed:

`PUBLIC: 'Public'`, `PRIVATE: 'Private'` and then `INTERNAL: 'Internal'` (for OTF-only comms in a followup feature)

todd-dembrey: `PUBLIC: 'Public', REVIEWERS: 'Reviewers' and then INTERNAL: 'Internal'`

PR #142, opened 2018-03-07
Add anonymisation option to user migration

zerolab:
was faster than anonymising the JSON

### DIFF
--- a/opentech/apply/users/management/commands/migrate_users.py
+++ b/opentech/apply/users/management/commands/migrate_users.py
@@ -84,8 +84,7 @@ class Command(BaseCommand):
         if not anonymize or self.is_staff(email):
             return email
 
-        uid = user['uid']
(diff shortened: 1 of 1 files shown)


opentech/apply/users/management/commands/migrate_users.py
```diff
@@ -78,3 +78,14 @@ def get_user_groups(self, user):
                 groups.append(self.groups.filter(name=group_name).first())
 
         return groups
+
+    def get_email(self, user, anonymize=False):
+        email = user['mail']
+        if not anonymize or self.is_staff(email):
+            return email
+
+        return "aeon+%s@torchbox.com" % user['uid']
```
todd-dembrey: prefer f string of format. `%` is old style.

zerolab: didn't want an extra assignment :)

PR #147, opened 2018-03-07
Correctly inherit the meta for the child tables

todd-dembrey:
@chris-lawton 

This should allow the child table to correctly display the status on the submissions table


zerolab:
:shipit:

PR #151, opened 2018-03-08
Feature/s6 demo prep fixes

todd-dembrey:
@zerolab Updates the display on the progress form

@chris-lawton Add the redirect to the communications tab. Tidies up the js as the tab was jumping on load. The page also fall back to the submissions detail tab if a dodgy href is passed.


todd-dembrey:
Discussed with Chris. Tab runs on everypage, so pages without tabs will error when trying to set the default. So no merge yet.

zerolab:
Can the tabs be specifically added to dashboard/submissions list/submission templates only?

PR #154, opened 2018-03-08
Feature/form styling

chris-lawton:
tidying up apply form elements


chris-lawton:
@todd-dembrey - can you cast an eye over the template changes?

PR #188, opened 2018-03-16
Feature/linked concept notes

todd-dembrey:
- Adds the ability to progress applications from one stage to the next. 
- Adds the linked application to the nested table on the submissions overview page.
- Adds the new form to the users application for the next stage
- Allow users to edit their application if they have been progressed to the next stage.
- Basic implementation of phase based permissions.
- Bugfix to addressfield for not storing data (probably should be cherry picked out)


opentech/apply/funds/views.py
```diff
@@ -125,6 +138,48 @@ class SubmissionDetailView(ViewDispatcher):
...
+@method_decorator(login_required, name='dispatch')
+class SubmissionEditView(UpdateView):
+    """
+    Converts the data held on the submission into an editable format and knows how to save
+    that back to the object. Shortcuts the normal update view save approach
+    """
+    model = ApplicationSubmission
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.user != self.get_object().user:
```
zerolab: Admins should be able to edit the submission, imho

opentech/apply/funds/models.py
```diff
@@ -111,9 +109,12 @@ class WorkflowStreamForm(WorkflowHelpers, AbstractStreamForm):  # type: ignore
     class Meta:
         abstract = True
 
-    def get_defined_fields(self):
-        # Only return the first form, will need updating for when working with 2 stage WF
-        return self.forms.all()[0].fields
+    def get_defined_fields(self, stage=None):
```
zerolab: inviting for proposal results in `'Page' object has no attribute 'get_defined_fields`

todd-dembrey: fixed in 7ecfc1d

opentech/apply/funds/views.py
todd-dembrey: There is another ticket that will cover editing to a greater extent. I want to cover those needs there. This treads on the toes of that ticket. But the requirement of submitting a a proposal when requested is what drove me to including the basic editing part in this.

PR #189, opened 2018-03-16
Feature/203 reviews backend

todd-dembrey:
@zerolab Updates slightly the work you did. including the titles onto the template and a slight refactor of the form code.


opentech/apply/review/forms.py
```diff
@@ -84,7 +84,10 @@ def validate_unique(self):
         self.instance.author = self.request.user
         self.instance.review = self.cleaned_data
 
-        super().validate_unique()
+        try:
+            self.instance.validate_unique()
+        except ValidationError as e:
+            self._update_errors(e)
```
zerolab: :)

I went through the same path, first 59e5d21791c9 then this

opentech/apply/review/forms.py
```diff
@@ -78,45 +79,29 @@ def __init__(self, *args, **kwargs):
...
-            self.instance.recommendation = self.cleaned_data['recommendation']
-        except KeyError:
-            pass
+        for field in self.get_score_fields():
+            score = int(self.cleaned_data.get(field, NA))
+            if score != NA:
+                scores.append(score)
 
-        super().save()
+        return sum(scores) / len(scores)
```
zerolab: so what happens when there are no score fields, thus `len(scores)` is 0?

opentech/apply/review/forms.py
```diff
@@ -78,45 +79,29 @@ def __init__(self, *args, **kwargs):
...
-            self.instance.validate_unique()
-        except ValidationError as e:
-            self._update_errors(e)
+        super().validate_unique()
 
     def save(self, commit=True):
+        self.instance.score = self.calculate_score()
 
-        score_fields = self.get_score_fields()
+        self.instance.recommendation = self.cleaned_data['recommendation']
```
zerolab: I like to be defensive.

opentech/apply/review/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+from opentech.apply.funds.models import ApplicationSubmission
+from .forms import ConceptReviewForm, ProposalReviewForm
+from .models import Review
+
+
+class ReviewCreateView(CreateView):
+    model = Review
+
+    def dispatch(self, request, *args, **kwargs):
```
zerolab: 👍

opentech/apply/review/forms.py
todd-dembrey: `recommendation` is required. if it isnt in `cleaned_data` we have bigger problems

opentech/apply/review/forms.py
todd-dembrey: I was missed it out. good spot.

PR #191, opened 2018-03-19
Feature/206 reviewers block

todd-dembrey:


opentech/apply/review/models.py
```diff
@@ -1,17 +1,59 @@
...
+from opentech.apply.users.models import User
+
+NO = 0
+MAYBE = 1
+YES = 2
 
 RECOMMENDATION_CHOICES = (
-    (0, 'No'),
-    (1, 'Yes'),
-    (2, 'Maybe')
```
zerolab: Ugh, this becomes yet another value map to migrate. In the old site "maybe" is 2

opentech/apply/review/models.py
```diff
@@ -1,17 +1,59 @@
...
+    (NO, 'No'),
+    (MAYBE, 'Maybe'),
+    (YES, 'Yes'),
 )
 
 
+class ReviewQuerySet(models.QuerySet):
+    def by_staff(self):
+        return self.filter(author__in=User.objects.staff())
```
zerolab: staff or admin

opentech/apply/users/models.py
```diff
@@ -2,11 +2,19 @@
...
 from .utils import send_activation_email
 
 
-class UserManager(BaseUserManager):
+class UserQuerySet(models.QuerySet):
+    def staff(self):
+        return self.filter(groups__name=STAFF_GROUP_NAME)
+
+    def reviewers(self):
+        return self.filter(groups__name=REVIEWER_GROUP_NAME)
```
zerolab: I am increasingly convinced this should be permission-drive. One to discuss in person

opentech/apply/funds/templates/funds/includes/review_table.html
```diff
@@ -0,0 +1,22 @@
...
+    </tr>
+    {% include 'funds/includes/review_table_row.html' with reviews=staff_reviews %}
+    {% include 'funds/includes/review_table_row.html' with reviews=object.missing_staff_reviews missing=True %}
+</table>
+
+<table>
+    <tr>
+        <th>Reviewers</th>
+        <th>{{ object.reviews.reviewers_reccomendation|traffic_light }}</th>
+        <th>{{ object.reviews.reviewers_score|default_if_none:'-' }}</th>
```
zerolab: If the user is added as reviewer, but is also staff, the block will display both...

opentech/apply/review/models.py
todd-dembrey: True, but it feels more reasonably to do it this way. If we are migrating it is there another approach you would rather do that might work better?

zerolab: had a thought and I can't find a better way, so let's leave it as is.

PR #194, opened 2018-03-20
Feature/164 submissions expanded info

zerolab:
Includes last update data based on activity, reviewers and outcomes
and some tweaks to allow showing/hiding

### DIFF
--- a/opentech/apply/funds/templates/funds/tables/table.html
+++ b/opentech/apply/funds/templates/funds/tables/table.html
@@ -1,5 +1,5 @@
 {% extends 'django_tables2/table.html' %}
-{% load table_tags django_tables2 %}
+{% load django_tables2 table_tags review_tags %}
 
 {# example of how to extend the table template #}
 {% block table.tbody.row %}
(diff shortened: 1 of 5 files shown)


opentech/apply/funds/templates/funds/tables/table.html
```diff
@@ -1,29 +1,73 @@
...
+                        <td>{{ submission.full_name }}</td>
+                        <td>
+                            {% if row.cells.last_update != '—' %}
+                                <strong>by {{ row.cells.last_update.user }}</strong><br/>
+                                {{ row.cells.last_update.timestamp|date:"m.d.y \a\t H:i" }}
+                            {% else %}
+                                {{ row.cells.last_update }}
+                            {% endif %}
+                        </td>
+                        <td>&mdash;</td>
```
todd-dembrey: Do we want to be adding deadlines when this will not show anything yet and is potentially confusing.

opentech/apply/funds/templates/funds/tables/table.html
```diff
@@ -1,29 +1,73 @@
...
+{% load django_tables2 table_tags review_tags %}
 
 {# example of how to extend the table template #}
 {% block table.tbody.row %}
-    {{ block.super }}
+    <tr {{ row.attrs.as_html }} data-record-id="{{ row.record.id }}">
+        {% for column, cell in row.items %}
+            <td {{ column.attrs.td.as_html }}>{% if column.localize == None %}{{ cell }}{% else %}{% if column.localize %}{{ cell|localize }}{% else %}{{ cell|unlocalize }}{% endif %}{% endif %}</td>
+        {% endfor %}
+    </tr>
```
todd-dembrey: Would prefer to keep the standard tables mechanism. Use the `row_attrs` meta option to add `data-record-id`, Would add extra data attribute to the child rows but means we dont need to copy in the row templating.

zerolab: 👍

PR #195, opened 2018-03-20
Ensure that the score is formatted in a reasonable way

todd-dembrey:
Prevent many trailing numbers after the decimal point.

rounds to int if `x.0` else 1 decimal place.


zerolab:
Best to round to 2 decimal places, IMHO

PR #196, opened 2018-03-20
Feature/207 view all reviews

todd-dembrey:


opentech/apply/review/templates/review/review_list.html
```diff
@@ -0,0 +1,22 @@
...
+        <h5>For <em>{{ submission.title }}</em></h5>
+    </div>
+</div>
+
+<table>
+{% for question, answers in review_data.items %}
+    <tr>
+        <th>{{ question }}</th>
+        {% for answer in answers %}
+            <td>{{ answer|bleach }}</td>
```
zerolab: the answers seem to contain some boolean values, leading to an `argument cannot be of 'bool' type, must be of text type` TypeError

todd-dembrey: fixed in 60ead48

PR #197, opened 2018-03-20
Feature/203 add permissions to reviews

todd-dembrey:


opentech/apply/review/views.py
```diff
@@ -34,6 +35,10 @@ class ReviewCreateView(CreateView):
 
     def dispatch(self, request, *args, **kwargs):
         self.submission = get_object_or_404(ApplicationSubmission, id=self.kwargs['submission_pk'])
+
+        if not self.submission.phase.has_perm('reivew', request.user):
```
zerolab: `s/reivew/review/g`

zerolab: what happens when the user is a reviewer, but is not assigned to review the submission?

zerolab:
Looking good with the additional tweaks and fixes. As discussed, the perm logic needs tests.

PR #200, opened 2018-03-21
Add "active" reviewers filter

zerolab:


zerolab:
OK, this should be ready. Let's discuss

opentech/apply/review/models.py
```diff
@@ -72,3 +74,25 @@ def __str__(self):
...
         return f'<{self.__class__.__name__}: {str(self.review)}>'
+
+
+@receiver(post_save, sender=Review)
+def update_submission_reviewers_list(sender, **kwargs):
+    review = kwargs.get('instance')
+
+    if not review.submission.reviewers.filter(id=review.author.id).exists():
+        review.submission.reviewers.add(review.author)
+        review.submission.save()
```
todd-dembrey: You dont need to save after an m2m add. Unless there is something else you are trying to trigger?

opentech/apply/review/models.py
```diff
@@ -72,3 +74,25 @@ def __str__(self):
...
+
+        Activity.actions.create(
+            user=review.author,
+            submission=review.submission,
+            message=f'Added themselves to the Reviewers list.'
+        )
```
todd-dembrey: I don't think we want to track this. It feel like it should be an implicit part of submitting a review if they aren't in the list.

`Dan added himself to reviewers`
`Dan submitted a review`

opentech/apply/review/models.py
```diff
@@ -72,3 +74,25 @@ def __str__(self):
...
+
+    if not review.id:
```
todd-dembrey: Dont you want the created Boolean from kwargs?

PR #202, opened 2018-03-21
Feature/211 expanded submissions

chris-lawton:
front end work for styling the reviewers table row and linked concept note table tow


opentech/apply/funds/templates/funds/tables/table.html
```diff
@@ -28,14 +27,12 @@
                                 {% endif %}
                             {% endwith %}
                         </td>
-                        <td>&mdash;</td>
                         <td>
-                            <dl>
+                            <ul class="list lsit--no-margin">
```
zerolab: `list-no-margin`?

PR #206, opened 2018-03-22
adding reviewers block template and corresponding styles

chris-lawton:
Front end work for the reviewers block used on the fund page.
Changed from using the `person_listing.html` partial as this is used on the person index page.

### DIFF
--- a/opentech/public/funds/templates/public_funds/includes/reviewer_listing.html
+++ b/opentech/public/funds/templates/public_funds/includes/reviewer_listing.html
@@ -4,13 +4,14 @@
     {% include "utils/includes/media_box_icon.html" with page_icon=person_photo listing=True %}
 
     <div class="media-box__content">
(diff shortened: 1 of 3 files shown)


opentech/public/funds/templates/public_funds/includes/reviewer_listing.html
```diff
@@ -0,0 +1,16 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+<a class="media-box" href="{% pageurl person %}">
+    {% image person.photo max-210x235 as person_photo %}
+    {% include "utils/includes/media_box_icon.html" with page_icon=person_photo listing=True %}
+
+    <div class="media-box__content">
+        {% for type in person.person_types.all %}
```
zerolab: wrap in `{% if person.person_types.exists %}

zerolab:
👍

PR #210, opened 2018-03-22
Change the name of the Concept note and only use the first word in db

todd-dembrey:


zerolab:
Feels a bit hacky, but let's get it in

PR #211, opened 2018-03-22
Feature/54 review draft

zerolab:
builds on top of #197

- adds review detail view
- review draft update view
- tweaks url structure 
- updates reviews queryset and filtering to use published reviews


opentech/apply/review/forms.py
```diff
@@ -78,6 +79,10 @@ def __init__(self, *args, **kwargs):
         self.submission = kwargs.pop('submission')
         super().__init__(*args, **kwargs)
 
+        if "save_draft" in self.data:
```
todd-dembrey: Can we add this as an attribute to the form. Then reference it in the template. Removes the magic strings

zerolab: 👍

opentech/apply/review/templates/review/review_form.html
```diff
@@ -29,7 +29,10 @@ <h2>{{ value }}</h2>
                 {{ field }}
             {% endif %}
         {% endfor %}
-        <input class="button button--primary" type="submit" value="Submit Review" />
+        <input class="button button--primary" type="submit" value="Submit Review" name="submit" />
+        {% if not object.id or object.is_draft %}
+            <input class="button button--secondary button--white" type="submit" value="Save Draft" name="{{ draft_button_name }}" />
```
todd-dembrey: `form.draft_button_name`

PR #213, opened 2018-03-22
Fix/205 combined score

zerolab:
@chris-lawton will require your FE eye


chris-lawton:
That's fine - let's get it merged and then I'll style separately.

PR #214, opened 2018-03-22
Fix/201 tidy up reviewers assignment

todd-dembrey:


zerolab:
Let's merge after #211 with the corresponding tweaks

opentech/apply/funds/models.py
```diff
@@ -662,16 +662,15 @@ def save(self, *args, **kwargs):
             submission_in_db.save()
 
     @property
+    def missing_reviewers(self):
+        return self.reviewers.exclude(id__in=self.reviews.values('author'))
```
zerolab: `self.reviews.submitted().values('author')`

zerolab:
:shipit:

PR #223, opened 2018-03-22
Feature/206 reviewers sidebar

todd-dembrey:
@chris-lawton Please review as on dev not on master.


todd-dembrey:
@chris-lawton I rebased, can you pelase confirm you are still happy with the code

PR #366, opened 2018-06-04
219: Functionality to show/hide reviewers

zerolab:


opentech/apply/funds/templates/funds/applicationsubmission_admin_detail.html
```diff
@@ -17,6 +17,7 @@
         <h6 class="heading heading--light-grey heading--small heading--uppercase">Reviews</h6>
 
         {% include 'funds/includes/review_table.html' %}
+        <a class="link link--bold link--underlined link--toggle-reviewers js-toggle-reviewers" href="#">all assigned advisors</a>
```
zerolab: Should only show if there are advisors assigned

PR #373, opened 2018-06-12
Feature/234 reviewer block tidy

chris-lawton:


opentech/apply/funds/templates/funds/includes/review_table.html
```diff
@@ -1,16 +1,19 @@
...
-
 <table class="reviews-sidebar">
-    <tr class="tr tr--subchild light-grey-bg">
-        <th colspan="2"></th>
-        <th>{{ object.reviews.submitted.recommendation|traffic_light }}</th>
-        <th>{{ object.reviews.submitted.score|floatformat|default_if_none:'-' }}</th>
-    </tr>
-    {% include 'funds/includes/review_table_row.html' with reviews=staff_reviews %}
-    {% include 'funds/includes/review_table_row.html' with reviews=object.staff_not_reviewed missing=True %}
+    {% if staff_reviews or object.staff_not_reviewed %}
```
chris-lawton: @todd-dembrey - is the logic here ok? The idea being we only want to show the following table markup when a there's a staff review/missing staff review.

todd-dembrey: That looks like it should work

PR #375, opened 2018-06-14
Feature/47 change to django fsm

todd-dembrey:
This includes only the FSM changes for master


zerolab:
Needs linting and mypy love

PR #381, opened 2018-06-19
Determinations (57)

zerolab:
Contains:

- model
- forms
- detail view
- permission logic
- status changes


opentech/apply/determinations/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from .views import DeterminationCreateOrUpdateView, DeterminationDetailView
+
+app_name = 'determinations'
+
+urlpatterns = [
+    path('determination/', DeterminationCreateOrUpdateView.as_view(), name="form"),
+    path('determination/<int:pk>', DeterminationDetailView.as_view(), name="detail"),
```
todd-dembrey: Determination is a one to one with the submission. So should the urls be `/determination/create` and `/determination`

opentech/apply/determinations/views.py
```diff
@@ -0,0 +1,134 @@
...
+
+class DeterminationCreateOrUpdateView(CreateOrUpdateView):
+    model = Determination
+    template_name = 'determinations/determination_form.html'
+
+    def get_object(self, queryset=None):
+        return self.model.objects.get(submission=self.submission, author=self.request.user)
+
+    def dispatch(self, request, *args, **kwargs):
+        self.submission = get_object_or_404(ApplicationSubmission, id=self.kwargs['submission_pk'])
```
todd-dembrey: Maybe worth looking at a mixin, we use this pattern on reviews as well

opentech/apply/determinations/views.py
```diff
@@ -0,0 +1,134 @@
...
+
+        if self.submission.phase not in DETERMINATION_PHASES \
+                and not self.submission.user_lead_or_admin(request.user):
+            raise PermissionDenied()
+
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        try:
+            has_determination_response = self.submission.determination.determination != UNDETERMINED \
```
todd-dembrey: This looks a bit janky... perhaps change the determination field to `outcome` and add a method on the determination:
```python
def  submitted(self):
    return self.outcome != UNDETERMINED and not self.is_draft
```

opentech/apply/funds/models.py
```diff
@@ -767,6 +767,9 @@ def can_review(self, user):
 
         return self.has_permission_to_review(user)
 
+    def user_lead_or_admin(self, user):
+        return user.is_superuser or self.lead.id == user.id
```
todd-dembrey: Can't say i'm a massive fan. can we keep it descriptive like the `has_permission_to_review` methods?

also `self.lead == user` is sufficient

opentech/apply/determinations/views.py
```diff
@@ -0,0 +1,134 @@
...
+            except AttributeError:
+                pass
+
+            try:
+                value = determination_data[name]
+                form_determination_data.setdefault(field.label, str(value))
+            except KeyError:
+                pass
+
+        lead_or_admin = self.request.user.is_superuser or self.request.user == self.get_object().submission.lead
```
todd-dembrey: You've used this in multiple locations in slightly different ways, can you please simplify it.

opentech/apply/determinations/templatetags/determination_tags.py
```diff
@@ -0,0 +1,60 @@
...
+register = template.Library()
+
+
+@register.filter
+def is_lead_or_admin(user, submission):
+    return submission.user_lead_or_admin(user)
+
+
+@register.filter
+def can_add_determination(user, submission):
```
todd-dembrey: I think there is too much going on here. These functions are taking on multiple responsibilities and  duplicating checks that are being done elsewhere.

e.g. `has_determination_draft` does a permission check and a draft check

We share a lot of this functionality and the views check similar things. we are going to get quite confused if functionality changes and the view checks one thing and the template tags check another.

opentech/apply/determinations/templates/determinations/includes/determination_button.html
```diff
@@ -0,0 +1,12 @@
+{% load determination_tags workflow_tags %}
+{% if request.user|is_lead_or_admin:submission %}
+    {% if request.user|can_add_determination:submission %}
```
todd-dembrey: this looks like we are performing two checks which do the same thing. I dont kknow the difference form looking at it.

opentech/apply/determinations/templates/determinations/determination_detail.html
```diff
@@ -0,0 +1,28 @@
...
+
+<div class="grid">
+    <div>
+        <h5>Determination</h5>
+        <p>{{ determination.get_determination_display }}</p>
+        {{ determination.determination_message|bleach }}
+    </div>
+</div>
+
+    {% if can_view_extended_data %}
```
todd-dembrey: If we are using the same pattern as earlier we should probably be providing two views. One for admin and one for applicant and non privileged staff.

opentech/apply/determinations/models.py
```diff
@@ -0,0 +1,69 @@
...
+            user=determination.author,
+            submission=determination.submission,
+            message=f'Created a determination for {determination.submission.title}'
+        )
+
+    if not kwargs.get('is_draft', False):
+        Activity.actions.create(
+            user=determination.author,
+            submission=determination.submission,
+            message=f'Sent the determination for {determination.submission.title}'
```
todd-dembrey: Shouldn't this be the content of the determination?

opentech/apply/determinations/models.py
```diff
@@ -0,0 +1,69 @@
...
+        'funds.ApplicationSubmission',
+        on_delete=models.CASCADE,
+        related_name='determination'
+    )
+    author = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.PROTECT,
+    )
+
+    determination = models.IntegerField(verbose_name=_("Determination"), choices=DETERMINATION_CHOICES, default=1)
```
todd-dembrey: See other comment. `outcome`

opentech/apply/determinations/models.py
```diff
@@ -0,0 +1,69 @@
...
+    determination_message = models.TextField(verbose_name=_("Determination message"), blank=True)
+    determination_data = JSONField(blank=True)
```
todd-dembrey: duplicating determination is not nice, why not message and data?

opentech/apply/determinations/forms.py
```diff
@@ -0,0 +1,260 @@
...
+                    action = self.submission.phase.transitions[action_name]
+                    raise forms.ValidationError(f'You do not have permission to "{ action }"')
+
+                self.transition = transition
+
+        return cleaned_data
+
+    def save(self, commit=True):
+        self.instance.determination = int(self.cleaned_data['determination'])
+        self.instance.determination_message = self.cleaned_data['determination_message']
```
todd-dembrey: Model form should be doing this if you define fields i think

opentech/apply/determinations/forms.py
```diff
@@ -0,0 +1,260 @@
...
+        return super().get_initial_for_field(field, field_name)
+
+    def validate_unique(self):
+        # update the instance data prior to validating uniqueness
+        self.instance.submission = self.submission
+        self.instance.author = self.request.user
+        self.instance.determination_data = {key: value for key, value in self.cleaned_data.items()
+                                            if key not in ['determination', 'determination_message']}
+
+        try:
```
todd-dembrey: Probably just super() it

opentech/apply/determinations/views.py
```diff
@@ -0,0 +1,134 @@
...
+            title="Update Determination draft" if self.object else 'Add Determination',
+            **kwargs
+        )
+
+    def get_form_class(self):
+        return get_form_for_stage(self.submission)
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs['request'] = self.request
```
todd-dembrey: We're not interested in the request, only the user.

opentech/apply/determinations/templates/determinations/includes/determination_button.html
zerolab: simplified.

opentech/apply/determinations/urls.py
zerolab: 👍

opentech/apply/funds/models.py
zerolab: changed to `has_permission_to_add_determination`, yet I am not a fan of that as it couples things even further

opentech/apply/determinations/views.py
```diff
@@ -0,0 +1,134 @@
...
+        kwargs['submission'] = self.submission
+
+        if self.object:
+            kwargs['initial'] = self.object.determination_data
```
todd-dembrey: we should just be passing in the instance and letting the form handle the rest.

opentech/apply/determinations/tests/test_views.py
```diff
@@ -0,0 +1,104 @@
...
+    def get_kwargs(self, instance):
+        return {'submission_pk': instance.submission.id}
+
+    def test_can_access_determination(self):
+        submission = ApplicationSubmissionFactory(status='in_discussion')
+        determination = DeterminationFactory(submission=submission, author=self.user, not_draft=True)
+        response = self.get_page(determination)
+        self.assertContains(response, determination.submission.title)
+        self.assertContains(response, self.user.full_name)
+        self.assertContains(response, reverse('funds:submissions:detail', kwargs={'pk': submission.id}))
```
todd-dembrey: Should this have a not contains for the extra data?

todd-dembrey: plus lead should see data

zerolab: sorted

opentech/apply/determinations/views.py
zerolab: we are. e.g. in `get_initial_for_field()`

PR #382, opened 2018-06-19
Allow superusers to log in as other users

todd-dembrey:
This allows super admins to assume the role of another user, we may end up removing this when the site is live, but it will make testing the site during MVP much easier.


zerolab:
LGTM. will merge when green

PR #383, opened 2018-06-19
57 determination canned messages

zerolab:
@chris-lawton this is the branch for determination message templates


zerolab:
@todd-dembrey added request message templates + moved missing view stuff

PR #390, opened 2018-06-20
Update state of menu as active for any child page

todd-dembrey:


zerolab:
LGTM.
My gut does say that maybe it should be a starts with comparison, but this works

PR #391, opened 2018-06-20
Make sure the restart stage cannot be used on accepted/rejected

todd-dembrey:


zerolab:
I think this could use a test

zerolab:
I'm afraid we will need to come back to this one.
The changes in #397 break this. I thought I rebased the branch to run the tests, but looks like I missed this :(

PR #396, opened 2018-06-21
Determination tweaks

zerolab:
- filters out outcome choices that are not possible with the available transitions
- treats `invited_to_proposal` as an accepted determination response


zerolab:
Ben D reviewed it.

PR #401, opened 2018-06-22
Wrap diffs in span

zerolab:
Checked against various diffs. Doesn't seem to have an impact on style for large diffs, but improves the look of small diffs, especially in the required fields area.


zerolab:
Discussed with Maya. Merging and can tweak later

PR #402, opened 2018-07-02
Attempt to make the migration use django-fsm to set state on migrated…

frjo:
I found out that the migrate_rr_applications.py is not longer working. I believe I have traced it to the switch to django-fsm.

This is me attempt att fixing it but I got stuck on "PermissionDenied: Invalid "accepted" transition", see full traceback below.

If I skip running "perform_transition()" the import seems to work as it should.

```
$ m migrate_rr_applications var/otf_export_2018-06-20/otf_rapid_response_application_export.json
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/__init__.py", line 371, in execute_from_command_line
    utility.execute()
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/__init__.py", line 365, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/base.py", line 335, in execute
    output = self.handle(*args, **options)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "/Users/frjo/Sites/otf-django/opentech/apply/funds/management/commands/migrate_rr_applications.py", line 245, in handle
    self.process(id)
  File "/Users/frjo/Sites/otf-django/opentech/apply/funds/management/commands/migrate_rr_applications.py", line 291, in process
    submission.perform_transition(status, lead)
  File "/Users/frjo/Sites/otf-django/opentech/apply/funds/models.py", line 589, in perform_transition
    raise PermissionDenied(f'Invalid "{ action }" transition')
django.core.exceptions.PermissionDenied: Invalid "accepted" transition
```


opentech/apply/funds/management/commands/migrate_rr_applications.py
```diff
@@ -281,6 +288,7 @@ def process(self, id):
 
         try:
             submission.save()
+            submission.perform_transition(status, lead)
```
frjo: Commenting out this line makes the migration run.

zerolab: At this very point the submission does not have a state, so `perform_transition()` throws the permission error.
In the context of migrations, setting `submission.status = self.get_workflow_state(node)` should suffice

opentech/apply/funds/management/commands/migrate_rr_applications.py
```diff
@@ -244,8 +245,14 @@ def handle(self, *args, **options):
                 self.process(id)
 
     def process(self, id):
+        User = get_user_model()
         node = self.data[id]
 
+        try:
+            lead = User.objects.get(full_name="Lindsay Beck")
```
zerolab: This migration depends on https://github.com/OpenTechFund/opentech.fund/blob/fa10f297856d4358eeeda87033a848a293447522/opentech/apply/users/management/commands/migrate_users.py

But a good idea to wrap it in try/except. 

Note that the submission will inherit the lead from the Round - https://github.com/OpenTechFund/opentech.fund/blob/fa10f297856d4358eeeda87033a848a293447522/opentech/apply/funds/management/commands/migrate_rr_applications.py#L221

opentech/apply/funds/management/commands/migrate_rr_applications.py
frjo: The original error I got was "Direct status modification is not allowed". My conclusion was that this was caused by setting "protected=True" for the FSMField. Does this sound plausible?

```
m migrate_rr_applications var/otf_export_2018-06-20/otf_rapid_response_application_export.json
Traceback (most recent call last):
  File "manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/__init__.py", line 371, in execute_from_command_line
    utility.execute()
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/__init__.py", line 365, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django/core/management/base.py", line 335, in execute
    output = self.handle(*args, **options)
  File "/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "/Users/frjo/Sites/otf-django/opentech/apply/funds/management/commands/migrate_rr_applications.py", line 244, in handle
    self.process(id)
  File "/Users/frjo/Sites/otf-django/opentech/apply/funds/management/commands/migrate_rr_applications.py", line 262, in process
    submission.status = self.get_workflow_state(node)
  File "/Users/frjo/.virtualenvs/otf-django/lib/python3.6/site-packages/django_fsm/__init__.py", line 230, in __set__
    raise AttributeError('Direct {0} modification is not allowed'.format(self.field.name))
AttributeError: Direct status modification is not allowed
```

zerolab: Ah, I see. Yes, that is the root of the issue. There is a way to prevent this attribute error from happening, will try to get a solution back to you before the end of the day

zerolab:
@frjo I've pushed some changes that fix the issue. Can you test locally?

frjo:
Works very well locally, thank you.

frjo:
Will create a new branch for each content type based on this branch.

PR #406, opened 2018-07-19
Bugfix/allow progression of application

todd-dembrey:
It looks like the old test that covered this was fixed in a way that meant this was no longer checked progression of the phase. TBH it was a bad test as it covered multiple behaviours.

Add a new test which tests the progression explicitly.


zerolab:
Nice catch! Merge when green.

PR #407, opened 2018-07-19
240 streamform reviews

zerolab:
As discussed. Contains the barebones review stream form.
Assignment, submission handling and display coming next.

screen shot 2018-07-19 at 12 46 53

Score field rendered in form:
screen shot 2018-07-19 at 12 59 50


opentech/apply/review/tests/factories/blocks.py
```diff
@@ -0,0 +1,82 @@
...
+import random
+import uuid
+
+import factory
+from wagtail.core.blocks import CharBlock
+import wagtail_factories
+
+from opentech.apply.stream_forms import blocks as stream_blocks
+from opentech.apply.review import blocks
+from opentech.apply.utils import blocks as utils_blocks
```
todd-dembrey: I would prefer to include these in the stream form factories and import rather than copy them across from the funds factories. (which i think has happened here).

opentech/apply/review/tests/test_views.py
```diff
@@ -3,7 +3,7 @@
 from opentech.apply.funds.tests.factories import ApplicationSubmissionFactory
 from opentech.apply.users.tests.factories import StaffFactory, UserFactory
 from opentech.apply.utils.tests import BaseViewTestCase
-from .factories import ReviewFactory
+from .factories.models import ReviewFactory
```
todd-dembrey: You can still import from `.factories` due to the `__init__`

opentech/apply/utils/blocks.py
```diff
@@ -0,0 +1,163 @@
...
+    class Meta:
+        label = _('Rich text field')
+        icon = 'form'
+
+    def get_searchable_content(self, value, data):
+        return bleach.clean(data, tags=[], strip=True)
+
+
+class CustomFormFieldsBlock(StreamBlock):
+    rich_text = RichTextFieldBlock(group=_('Fields'))
```
todd-dembrey: I probably should have done so when i created this, but now it is no longer in the funds file can you add some text to say what this does. We might need to update the name as well to reflect the required block aspect.

opentech/apply/utils/blocks.py
```diff
@@ -0,0 +1,163 @@
...
+    required_blocks = None
```
todd-dembrey: `required_blocks = None` should be an interable by default e.g. `[]` otherwise `[block.name for block in self.required_blocks]` will error if you dont define it.

opentech/apply/utils/blocks.py
```diff
@@ -0,0 +1,163 @@
...
+    required_block_names = None
```
todd-dembrey: Would prefer not to define this on the class as that implies it is configurable (in my mind). When actually we just blast it in the init.

opentech/apply/utils/blocks.py
```diff
@@ -0,0 +1,163 @@
...
+        new_error = ErrorList([message])
+        try:
+            errors[child_number].data[0].params[field] = new_error
+        except KeyError:
+            errors[child_number] = ErrorList(
+                [ValidationError('Error', params={field: new_error})]
+            )
+
+    def to_python(self, value):
+        # If the data type is missing, fallback to a CharField
```
todd-dembrey: Can we add a docstring to this to say that this allows historic data to still be accessible even if a custom field type is removed from the code in the future.

(I should have done that originally)

opentech/apply/review/forms.py
```diff
@@ -1,21 +1,11 @@
...
-    (0, '0. Need more info'),
-    (1, '1. Poor'),
-    (2, '2. Not so good'),
-    (3, '3. Is o.k.'),
-    (4, '4. Good'),
-    (5, '5. Excellent'),
-    (NA, 'n/a - choose not to answer'),
-)
 
 YES_NO_CHOICES = (
```
todd-dembrey: Does this need removing as well? see `opentech/apply/review/options.py`

todd-dembrey:
Also, are there tests which do cover this or do you need to add a couple?

opentech/apply/funds/admin.py
```diff
@@ -67,7 +68,12 @@ def used_by(self, obj):
         return mark_safe('<br>'.join(rows))
 
 
+class ReviewFormAdmin(ModelAdmin):
+    model = ReviewForm
+    menu_icon = 'form'
+
+
 class ApplyAdminGroup(ModelAdminGroup):
```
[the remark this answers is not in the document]
zerolab: I've been considering that, but couldn't think of a good place for it. suggestions are welcome

opentech/apply/dashboard/tests/test_views.py
```diff
@@ -1,6 +1,6 @@
 from opentech.apply.funds.tests.factories import ApplicationSubmissionFactory, ApplicationRevisionFactory
 from opentech.apply.users.tests.factories import UserFactory, StaffFactory
-from opentech.apply.utils.tests import BaseViewTestCase
+from opentech.apply.utils.tests.tests import BaseViewTestCase
```
todd-dembrey: ?

opentech/apply/review/blocks.py
```diff
@@ -0,0 +1,123 @@
...
+                widget_attrs['id'] = '%s_%s' % (id_, i)
+            else:
+                widget_attrs = final_attrs
+
+            rendered.append(widget.render(widget_name, widget_value, widget_attrs, renderer))
+
+        return ''.join([mark_safe(item) for item in rendered])
+
+
+class ScoredAnswerField(forms.MultiValueField):
```
todd-dembrey: Can you move these the field and widget into a separate file (at a minimum `fields.py` probably dont need `widgets.py`)

opentech/apply/dashboard/tests/test_views.py
todd-dembrey: Ah i've got it. This is the test utils  rather than the tests for the utils. 

Maybe it should be `from opentech.apply.utils.testing import BaseViewTestCase` ? But not happy with that.

regardless, this naming isnt working

todd-dembrey:
Looking really good now. Just two minor things.

zerolab:
All sorted

todd-dembrey:
Merge when ready and approved

PR #410, opened 2018-07-27
Feature/251 create slack adapter

todd-dembrey:
Create the Slack adapter and re-purpose the messages adapter to present messages to the user when not sending them.

### DIFF
--- a/opentech/apply/activity/messaging.py
+++ b/opentech/apply/activity/messaging.py
@@ -38,7 +38,11 @@ class AdapterBase:
             message = self.message(message_type, **kwargs)
         except KeyError:
             return
-        self.send_message(message, **kwargs)
+        if settings.SEND_MESSAGES:
+            self.send_message(message, **kwargs)
+        else:
+            messages.add_message(kwargs['request'], messages.INFO, message)
+
 
     def send_message(self, message, **kwargs):
         raise NotImplementedError()
--- a/opentech/apply/activity/messaging.py
+++ b/opentech/apply/activity/messaging.py
@@ -1,4 +1,8 @@
 from enum import Enum
+
+import requests
+
+from django.conf import settings
 from django.contrib import messages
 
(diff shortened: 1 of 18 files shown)


opentech/apply/activity/messaging.py
```diff
@@ -82,6 +89,56 @@ def send_message(self, message, user, submission, **kwargs):
...
+            slack_target = ''
+
+        message = ' '.join([slack_target, message]).strip()
+        return message
+
+    def slack_id(self, user):
+        return f'<{user.slack}>'
+
+    def send_message(self, message, **kwargs):
+        if not self.destination and not self.target_room:
```
zerolab: Should this not return when either destination or target_room is not set?
i.e. `if not self.destination or not self.target_room:`

opentech/apply/determinations/forms.py
```diff
@@ -1,7 +1,7 @@
 from django import forms
 from django.core.exceptions import NON_FIELD_ERRORS
 
-from opentech.apply.funds.workflow import DETERMINATION_RESPONSE_TRANSITIONS
+from opentech.apply.funds.workflow import DETERMINATION_OUTCOMES
```
zerolab: 👍

opentech/apply/funds/models.py
```diff
@@ -46,13 +46,14 @@
 from .edit_handlers import FilteredFieldPanel, ReadOnlyPanel, ReadOnlyInlinePanel
 from .workflow import (
     active_statuses,
+    DETERMINATION_OUTCOMES,
```
zerolab: imported but not used?

todd-dembrey:
Need a better way to turn off slack notifications during tests. Will add another commit

todd-dembrey:
It was due to my local.py settings rather than the CI tests which was my concern. I'm not too worried if that happens for now.

PR #414, opened 2018-08-01
Create the RFP interface and basic sealed round objects

todd-dembrey:
This allows the admin to create both rounds and sealed rounds as children of the RFP page.


todd-dembrey:
I've just done a refactor to improve how the code works now we have RFPs and Funds. 

There is now a common page model for both Applications and Round. Forms now point at the base model as well so no additional form models required. I migrated the existing data as follows:
1. renamed the existing models to the base model.
2. created the new models
3. Using the data from the renamed models recreate the Fund, Round and Lab pages

opentech/apply/funds/migrations/0038_recreate_objects_that_exist.py
```diff
@@ -0,0 +1,51 @@
...
+                'numchild': obj.numchild,
+                'url_path': obj.url_path,
+            }
+            try:
+                kwargs.update(lead=obj.lead)
+            except:
+                pass
+
+            new_obj = model(**kwargs)
+            new_obj.save()
```
zerolab: This whole thing is rather neat!

zerolab:
@frjo #402 will need updating for these changes. Should be mostly imports

opentech/apply/funds/migrations/0038_recreate_objects_that_exist.py
todd-dembrey: go go wagtail!

PR #415, opened 2018-08-02
Feature/254 limit access to sealed rounds

todd-dembrey:

PR #416, opened 2018-08-03
Feature/255 rfp public

todd-dembrey:
Requires #415


opentech/public/funds/migrations/0008_recreate_pages.py
```diff
@@ -0,0 +1,50 @@
...
+    # We removed the old definition of these objects, need to create
+    # a new object with a pointer back to that object, the underlying
+    # data structure was unaffected
+    ContentType = apps.get_model('contenttypes.ContentType')
+
+
+    for model_name, new_model_name in [
+            ('FundPage', 'BaseApplicationPage'),
+    ]:
+        content_type, _ = ContentType.objects.get_or_create(model=model_name, app_label='public_funds')
```
zerolab: model_name.lower() ?

todd-dembrey: yeah... where did that go...

PR #449, opened 2018-08-15
Feature/288 submit review at any time

todd-dembrey:


opentech/apply/review/models.py
```diff
@@ -100,24 +101,16 @@ def __str__(self):
...
     @property
-    def fields(self):
-        return [
-            field.render(context={'data': data})
-            for data, field in self.data_and_fields()
-        ]
-
-    def render_answers(self):
-        return mark_safe(''.join(self.fields))
+    def for_latest(self):
```
zerolab: make this more explicit? i.e. `for_latest_revision`

opentech/apply/review/models.py
```diff
@@ -100,24 +101,16 @@ def __str__(self):
...
+        return self.revision == self.submission.live_revision
+
+    def get_compare_url(self):
```
zerolab: This is another one of those cross-over moments.
Can this be renamed to be more verbose? again, this is the submission compare url, not the review one and it reads   as the latter

opentech/apply/review/templates/review/review_detail.html
```diff
@@ -18,9 +18,17 @@ <h5>Recommendation</h5>
         <h5>Score</h5>
         <p>{{ review.score }}</p>
     </div>
+    {% if not review.for_latest %}
+        <div>
+            <h5>Review was not against the lasted version:</h5>
```
zerolab: "The Review was not added against the latest submission version"?
(or at least `s/lasted/latest`)

opentech/apply/review/views.py
```diff
@@ -130,25 +129,28 @@ def get_context_data(self, **kwargs):
...
-
-                    if isinstance(field.block, ScoreFieldBlock):
-                        value = json.loads(data)
-                        rating_value = int(value[1])
-                        rating = RATE_CHOICES_DICT.get(rating_value, RATE_CHOICE_NA)
-                        comment = str(value[0])
-                        review_data[field.id]['answers'].append(rating + comment)
-                    else:
-                        review_data[field.id]['answers'].append(str(data))
+                    # If we haven't see the question before we backfill with '-'
```
zerolab: backfilling with ''. no?

todd-dembrey: and that is why i dont like comments..

zerolab: :)

opentech/apply/review/templates/review/review_detail.html
```diff
@@ -20,7 +20,7 @@ <h5>Score</h5>
     </div>
     {% if not review.for_latest %}
         <div>
-            <h5>Review was not against the lasted version:</h5>
+            <h5>Review was not against the lastest version:</h5>
```
zerolab: "latest" ;)

PR #452, opened 2018-08-17
Feature/287 edit at any time

todd-dembrey:


zerolab:
LGTM. I know this builds on #449, but in case this gets merged first, it is missing 69a3cd3, b9e6405 and 763a868

PR #477, opened 2018-08-31
Handle the files being imported into the application for historic data

todd-dembrey:


opentech/apply/funds/management/commands/migration_base.py
```diff
@@ -260,3 +283,17 @@ def get_workflow_state(self, node):
...
         return value.replace('\r\n', '<br>\n')
+
+    def process_file(self, value):
+        if isinstance(value, dict):
+            value = [value]
+
+        files = []
+
+        for file_data in value:
+            file_path = 'files/private/application_upload/' + file_data['filename']
```
frjo: Application uploads are in a number of different directories unfortunately. The code need to look in file_data['uri'] and parse it for path and filename.

opentech/apply/funds/management/commands/migration_base.py
```diff
@@ -260,3 +283,17 @@ def get_workflow_state(self, node):
...
+            saved_file = migration_storage.open(file_path)
+            saved_file.name = file_data['filename']
```
frjo: Drupal store the original file name in "filename" but the actual file can be renamed so not to overwrite files, Drupal adds "_0", "_1" etc. to the name. This also happens when you simply upload a new version with the same filename so there are a few of them.

todd-dembrey: We are doing the same in django. Fixed in 050f1de

opentech/apply/funds/management/commands/migration_base.py
todd-dembrey: Fixed in 050f1de

PR #506, opened 2018-09-04
Remove the username step from the social auth

todd-dembrey:
Our user model uses email and doesnt have an username field


zerolab:
LGTM.  Social Auth [returns early](https://github.com/python-social-auth/social-core/blob/1d809941ab8a99af9e1bdf12bae548202c94eaa2/social_core/pipeline/user.py#L10) for user models without a username, but all improvements are welcome.

todd-dembrey:
We weren't using those settings so wasn't aware of that. Seemed the most reasonable to remove it as we font have the username on the model

PR #507, opened 2018-09-04
Bugfix/allow presigned urls for private files

### LINKED ISSUE #504: 11 upload files show up on a newly submitted concept note to the IFFund even though no files were uploaded
See https://apply.opentech.fund/apply/submissions/15/, they weirdly all link back to the submission :-/

todd-dembrey:
Fixes the following three issues

* Allow presigned urls for custom domains
* Allow dynamic signing of urls on objects
* Render "No response" for multi file so that it doesn't display 11 file boxes... fixes #504


todd-dembrey:
This needs testing properly on the private files setup before going live.

opentech/storage_backends.py
```diff
@@ -24,4 +27,16 @@ class PrivateMediaStorage(S3Boto3Storage):
     encryption = True
     file_overwrite = False
     querystring_auth = True
-    url_protocol = 'https:'
+    secure_urls = True
```
frjo: This change give the same result but it looks like url_protocol is the current version of this setting.

https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto3.py#L251

todd-dembrey: OK, i guess i prefer the secure urls argument. But you are current, `url_protocol` is the correct one to use with boto3

PR #518, opened 2018-09-07
Disable auto_* on date fields so imported dates are used.

frjo:
Notices that all dates got set to NOW on import.


opentech/apply/funds/management/commands/migration_base.py
```diff
@@ -80,6 +80,11 @@ def process(self, id):
         except ApplicationSubmission.DoesNotExist:
             submission = ApplicationSubmission(drupal_id=node['nid'])
 
+        # Disable auto_* on date fields so imported dates are used.
+        for field in submission._meta.local_fields:
+            if field.name == "submit_time":
```
frjo: This field is not part of the ApplicationSubmission model in submissions.py. I find it only in 0005_applicationsubmission.py?

frjo:
There are no dates on the Review model? Using only the date from events?

opentech/apply/funds/management/commands/migration_base.py
todd-dembrey: Ably demonstrating why inheritance/mixins is a pain

It comes from AbstractFormSubmission: https://github.com/gasman/wagtail/blob/7034cd131774b8971ff3c7424999a28164480f29/wagtail/contrib/forms/models.py#L46

frjo:
Then I will add date fields for reviews, good to have there before migration and then we can figure out how to use them.

PR #520, opened 2018-09-10
Make CategoryQuestionBlock use Select2MultipleWidget if more than 32 …

frjo:
…option to maintain usability.

### DIFF
--- a/opentech/apply/categories/blocks.py
+++ b/opentech/apply/categories/blocks.py
@@ -6,6 +6,7 @@ from wagtail.core.blocks import BooleanBlock, CharBlock, ChooserBlock, TextBlock
 from wagtail.core.utils import resolve_model_string
 
 from opentech.apply.stream_forms.blocks import OptionalFormFieldBlock
+from django_select2.forms import Select2MultipleWidget
 
 
 class ModelChooserBlock(ChooserBlock):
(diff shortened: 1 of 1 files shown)


opentech/apply/categories/blocks.py
```diff
@@ -63,7 +64,13 @@ def get_field_kwargs(self, struct_value):
 
     def get_widget(self, struct_value):
         if struct_value['multi']:
-            return forms.CheckboxSelectMultiple
+            category = struct_value['category']
+            category_size = len(category.options.values_list('id'))
```
todd-dembrey: `categroy.options.count()` should probably be fine here.

opentech/apply/categories/blocks.py
```diff
@@ -63,7 +64,13 @@ def get_field_kwargs(self, struct_value):
...
+            # Pick widget according to number of options to maintain good usability.
+            if category_size < 32:
+                return forms.CheckboxSelectMultiple
+            else:
+                return Select2MultipleWidget()
```
todd-dembrey: I would remove the call to the widget. I don't think we are currently passing arguments in when we are instantiating the objects, but it would be consistent with what the other widgets are doing.

PR #532, opened 2018-09-11
Issue: #531 Make it possible to mark CheckboxField as required.

frjo:


opentech/apply/stream_forms/blocks.py
```diff
@@ -127,6 +127,7 @@ def get_searchable_content(self, value, data):
 
 class CheckboxFieldBlock(FormFieldBlock):
     default_value = BooleanBlock(required=False)
+    required = BooleanBlock(label=_('Required'), required=False)
```
todd-dembrey: Previously we've been inheriting from `OptionalFieldBlock` if we are adding the required status.

https://github.com/OpenTechFund/opentech.fund/blob/35ed8f1d2123a6828ced971ba51a0f739db640a6/opentech/apply/stream_forms/blocks.py#L68-L69

PR #542, opened 2018-09-13
Allow existing files on imported submissions to get correct url so th…

frjo:
…ey can be downloaded.

### DIFF
--- a/opentech/apply/funds/models/mixins.py
+++ b/opentech/apply/funds/models/mixins.py
@@ -38,6 +38,9 @@ class AccessFormData:
 
     @classmethod
     def stream_file(cls, file):
+        if 'path' in file:
+            file['filename'] = file['name']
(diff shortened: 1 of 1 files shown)


todd-dembrey:
This is causing quite a few issues for multi file case.

The `process_file` case is relying on the `TypeError` to identify when the `file` attribute is a list. However you trigger an `AtrributeError` for when it is checking if any of the things in the list is `'path'` using the equality check on the file.

See `OTF-WEBAPP-T` error in sentry.

Either we should be checking the for a list in the `process_file` method or catching the `AtrributeError` in the `process_file`. I'm leaning towards the first as broad exception catching is a pain.

I should have made sure #507 was backwards compatible, but it  would probably have been easier to have just updated the stored values than pushing this fix.

frjo:
Do you think I simply should revert this?

PR #556, opened 2018-09-14
Rebuilding the front end stack with gulp 4.

### LINKED ISSUE #508: Rebuilding the front end stack with gulp 4, move all scripts except jQuery lib to footer and more
The main.js file is loaded in the header. Find out if this is really needed for some reason or if it can be moved to the footer for improved front end performance.

```
<script type="text/javascript" src="{% static 'js/main.js' %}"></script>
```

### LINKED ISSUE #601: When looking at the all reviews page in Chrome, the cursor is a move cursor rater than a pointer and each row pops up as an object that can be moved.
Can we undo those properties that make that?

### LINKED ISSUE #607: On apply pages, radio button text gets cut off from other buttons when there are more than two
https://apply.opentech.fund/red-team-lab/

### LINKED ISSUE #662: Truncate text on listing items
When long introductions are input into news items the text flows out of the listing item container. See 'RESEARCH CORNER: MASSBROWSER' news item on https://www.opentech.fund/news/. Suggest either truncating the text on display or setting a character limit on the introduction field.

frjo:
Trying to get the build time down, front end takes a lot of time on each deploy to Heroku now.

With HTTP/2 you also do not gain anything by concatenating css and js files together. Better to add each script only where they are needed.

Using Cloudflare cdn for jQuery and add that to the header on the apply site so Tinymce and the adress field work but move all other scripts to the footer.

<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>

Fixes #508
Fixes #662
Fixes #601 
Fixes #607


frjo:
This shaves about a minut off the time to deploy on Heroku.

frjo:
CircleCI 2 min 28 sec
TravisCI 4 min 29 sec

PR #558, opened 2018-09-15
Short circuit the save for the transition

todd-dembrey:
This allows us to skip the file handling if we are just updating the status field. Should avoid some of the issues we've been seeing on sending determinations.


todd-dembrey:
This should fix the following issues:
 
* Error when accessing files during transitions
* Error when accessing old files at any point
* Ability to transition submission when a determination exists and it had errored.

todd-dembrey:
I'm a little concerned by the CI failure. Might want to look at what is happening locally.

PR #559, opened 2018-09-15
Revert "Allow existing files on imported submissions to get correct u…

frjo:
…rl so they can be downloaded."

This reverts commit 65750f412c735b39123a66fc9fb5f2b69ca43ec6.


todd-dembrey:
I think we should keep the changes with some slight additions. see #558

PR #592, opened 2018-09-19
Disable checking for required fields on application forms for now. La…

frjo:
…bs do not need them always.

This is a hack but there is a point in keeping these fields separate under a "Required for most applications" header since that is true. There are a few labs that do not need them.

Maybe should add a checkbox to the application form "Disregard missing required fields" or something like that. That would make sure staff do not forget the fields by mistake but allow them to omit them when needed.


todd-dembrey:
@frjo I'm going to have a look at this today

PR #623, opened 2018-09-26
Make sure we display the linebreaks the user provided

todd-dembrey:
This will tidy up the applications so that new lines are honoured from the richtext editor.

### DIFF
--- a/opentech/apply/stream_forms/templates/stream_forms/render_unsafe_field.html
+++ b/opentech/apply/stream_forms/templates/stream_forms/render_unsafe_field.html
@@ -2,7 +2,7 @@
 {% load bleach_tags %}
 {% block data_display %}
     {% if data %}
-        {{ data|bleach }}
+        {{ data|bleach|linebreaks }}
     {% else %}
         {{ block.super }}
     {% endif %}


frjo:
I do not think this works well.

It turns this:

~~~~ html
<div>
    <h5>If yes, which application and what was the outcome?</h5>
    <div>
        <p>gfkgfdkljgfd</p>
        <p>lkgjfdjglkfdjglfd</p>
        <p>gfdgdfgfdgf</p>
    </div>
</div>
~~~~

In to this:

~~~~ html
<div>
    <h5>If yes, which application and what was the outcome?</h5>
    <div>
        <p></p>
        <p>gfkgfdkljgfd</p>
        <br>
        <p>lkgjfdjglkfdjglfd</p>
        <br>
        <p>gfdgdfgfdgf</p>
        <p></p>
    </div>
</div>~~~~

todd-dembrey:
It only adds paragraphs where there are two linebreaks `\n` otherwise it adds a `<br>`. So your example above would not change the underlying html.
e.g. 
```html
<p>some\ncontent\n\nwith new lines</p>
```
becomes
```html
<p>some<br>content</p>
<p>with new lines</p>
```

TinyMCE seems to be picking up the double newline case and making it a new paragraph, but  because users are coping in the text to the richtext editor we are getting the single lines which renders poorly.

This doesn't change any of the underlying data and should make the worse looking responses better and in the worse case add a small amount of space into other content.

I could also use `linebreaksbr` which does not add the paragraph tag ever.

todd-dembrey:
As discussed it isn't doing anything that i said it was... Will need to look into this a bit more.

todd-dembrey:
Closing as this doesn't work.

PR #624, opened 2018-09-26
Feature/slacken required block behaviour

### LINKED ISSUE #568: Lab applications are are forced to include the same required fields as funds, but those are not relevant/appropriate for all lab applications.
Let's undo making them actually required but show them in the add new field as distinct/important fields for now

todd-dembrey:
This refactors the required blocks to allow users to not include required blocks.

I have added the concept of a Named Block. This is any block that can only be included once. A special case of this is the Required block, which has to be included only once. This means that the refactor is not that onerous

Keeping the optional named blocks under the same banner as the required blocks gives us the flexibility to still have the queryable data for value even when some forms don't provide it. The data must be stored with the specific name, not the field id, so that this works.

For cases where there is no key in the `form_data`, postgres will return a `null`. I have added a `nulls_last` to the queyset ordering to accommodate this.

I cant say i'm 100% happy with the template rending as we still show all the Named Block fields in the proposal header, but with a `-` where not included (see below). We could probably get a nicer solution in time.

image

fixes #592
fixes #568


todd-dembrey:
fixes #592 #568

todd-dembrey:
This does need a little testing on the admin side. It should all just work but i want to check that the form creation is still working and that existing forms load just fine.

opentech/apply/utils/blocks.py
```diff
@@ -76,7 +79,7 @@ def clean(self, value):
 
         if duplicates:
             all_errors.append(
-                'You have duplicates of the following required fields: {}'.format(', '.join(prettify_names(duplicates)))
+                'You have duplicates of the following non duplicate fields: {}'.format(', '.join(prettify_names(duplicates)))
```
zerolab: What about `The following fields must be included only once: {}`?

todd-dembrey: that would work

frjo:
@todd-dembrey I have tested this and it works well as far as I can see. Any problems deploing this?

todd-dembrey:
@frjo I'm happy for this to go if the existing forms are unaffected by the changes above.

PR #636, opened 2018-10-04
WIP: Set random password when user gets created

frjo:
This makes the password reset form work for all accounts.

### DIFF
--- a/opentech/apply/funds/models/submissions.py
+++ b/opentech/apply/funds/models/submissions.py
@@ -378,16 +378,22 @@ class ApplicationSubmission(
             User = get_user_model()
             if 'skip_account_creation_notification' in self.form_data:
                 self.form_data.pop('skip_account_creation_notification', None)
-                self.user, _ = User.objects.get_or_create(
+                self.user, created = User.objects.get_or_create(
                     email=email,
                     defaults={'full_name': full_name}
                 )
             else:
-                self.user, _ = User.objects.get_or_create_and_notify(
+                self.user, created = User.objects.get_or_create_and_notify(
                     email=email,
                     site=self.page.get_site(),
                     defaults={'full_name': full_name}
                 )
+
+            # Set random password on new accounts.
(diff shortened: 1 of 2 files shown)


opentech/apply/funds/models/submissions.py
```diff
@@ -378,17 +378,23 @@ def ensure_user_has_account(self):
...
-                self.user, _ = User.objects.get_or_create_and_notify(
+                self.user, created = User.objects.get_or_create_and_notify(
                     email=email,
                     site=self.page.get_site(),
                     defaults={'full_name': full_name}
                 )
 
+            # Set random password on new accounts.
+            if created:
+                password = User.objects.make_random_password(length=16)
```
todd-dembrey: This short circuits the activation process.

Reviewing the code it looks like we use the same tokenisation for security purposes.So not too concerned there. It looks like the only issue is that user's do not have their status set to `active=True`. 

I do not want password resets to make the user active, as that removes the control from the admins in disabling accounts. Users would just need to reset password to re-activate.

It might make more sense to resend the activation email if the user is not active and has no password and move them through that workflow correct.

Throughts?

opentech/apply/users/management/commands/migrate_users.py
```diff
@@ -47,6 +47,11 @@ def handle(self, *args, **options):
                 if not created:
                     user_object.slack = slack_name
 
+                # Set random password on new accounts.
```
todd-dembrey: The workflow described above would also be valid for this case as well.

opentech/apply/funds/models/submissions.py
frjo: Good arguments. This seems to be less of a problem now than it was at the start. Lets wait and see on this.

frjo: A easy way for staff to resend the activation link is a nice idea.

PR #646, opened 2018-10-08
Allow staff to reject draft proposals.

frjo:


todd-dembrey:
Happy with the changes. If you have tested the behaviour and are happy then merge.

PR #652, opened 2018-10-09
Add a new SingleStageExternalDefinition workflow.

### LINKED ISSUE #650: We need an additional workflow available, a 1-stage application with external review, for things like Labs and Rapid Response, who have AC review requirements
Right now only the 2-stage application workflow allows external review for the 2nd application. We have a 1-stage application (for RFP's) but it doesn't allow external review.

frjo:
Fixes #650


opentech/apply/funds/workflow.py
```diff
@@ -137,6 +137,8 @@ def can_staff_edit(self, user: 'User') -> bool:
 
 Request = Stage('Request', False)
 
+RequestExt = Stage('RequestExt', True)
```
frjo: Ideas for a better name then "RequestExt"?

todd-dembrey: Does this need to be different? We display it on the submission detail page as part of the meta information. The fact it has an extra stage for the External Review is covered by the status bar?

image

Thus:
```suggestion
RequestExt = Stage('Request', True)
```

PR #671, opened 2018-10-24
Make flake8 work again on travisci.

frjo:
Travis seems to have updated its version of pycodestyle and added new checks. Adding them to ”ignore” in setup.cfg.

Also removed one unused variable ”e” in mixins.py.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ install:
 # Run the tests
 script:
   # Run python code style checks
-  - flake8
+  - flake8 ./opentech
 
   # Collect static
   - python manage.py collectstatic --noinput --verbosity=0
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -48,7 +48,7 @@ install:
 # Run the tests
 script:
   # Run python code style checks
-  - flake8
+  - flake8 ./opentech
```
frjo: Not really needed but does no harme either.

opentech/apply/funds/models/mixins.py
```diff
@@ -97,7 +97,7 @@ def data(self, id):
         definitive_id = self.get_definitive_id(id)
         try:
             return self.raw_data[definitive_id]
-        except KeyError as e:
+        except KeyError:
```
frjo: Error F841 from Flake8, better to fix this than to ignore since only one instance of it.

todd-dembrey:
Would like to resolve the F821 issues in worflow.py but otherwise happy with updating the settings as specified.

PR #681, opened 2018-10-26
Rename some of the custom and requeried fields when creating/editing …

### LINKED ISSUE #586: Rename some streamfield blocks when creating/editing an application
On application form:

- "Other" is a bit weird (maybe come up with something other than Other or just move this in with all the other elements or rename it from Paragraph to Section header or something).
- Title: rename to Application Title
- Value: rename to Requested Amount
- Add 1, 2, 3 to beginning of streamfield blocks sections in order to control ordering of display. E.g. required fields at top.

frjo:
…application forms. Remove the other section and merge that with custom.

Fixes #586


opentech/apply/utils/blocks.py
```diff
@@ -48,7 +48,7 @@ class CustomFormFieldsBlock(StreamBlock):
     single_blocks = []
 
     def __init__(self, *args, **kwargs):
-        child_blocks = [(block.name, block(group=_('Required'))) for block in self.required_blocks]
+        child_blocks = [(block.name, block(group=_(' Required'))) for block in self.required_blocks]
```
todd-dembrey:
```suggestion
        child_blocks = [(block.name, block(group=_('Required'))) for block in self.required_blocks]
```

frjo: This is a ugly hack to get it to come first, and it works even.

frjo: I think it’s a lot more logical for the required fields to come first. Is there a correct way to do this?

todd-dembrey: Ah, i see what you are doing. 

So... `CustomFormFieldsBlock` inhertis from `StreamBlock` which defines `sorted_child_blocks`

https://github.com/OpenTechFund/opentech.fund/blob/201039c44c247c7062a8119a2831b993bb829914/opentech/apply/utils/blocks.py#L45
https://github.com/wagtail/wagtail/blob/a671f19298874e1a0753b903694d8aab439df802/wagtail/core/blocks/stream_block.py#L60-L63

I think defining the following the following method would solve the issue:
```python
    def sorted_child_blocks(self):
        order = {
              'Required': '1',
              'Custom': '2',
         }
        return sorted(
            self.child_blocks.values(),
            key=lambda child_block: 
                order.get(child_block.meta.group, child_block.meta.group)
        )
```

But hey thats also hacky...

This is what does the actual grouping based on the initial order passed in...
https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/admin/templates/wagtailadmin/block_forms/stream_menu.html#L4

Issue #330, opened 2018-04-04
As an applicant/reviewer/manager, I want a clear list of to-dos assigned to me and to be returned there when I've completed something assigned to me


danblah:

danblah:
I think this is covered by the waiting for your review section.

PR #693, opened 2018-11-13
 Add markdown filter and markdown editor

### LINKED ISSUE #333: As an applicant/reviewer, ability to submit forms with richtext or markdown editors

### LINKED ISSUE #567: Ability to attach files and format text (wysiwyg editor) for communication messages (comments in the communications tab)

### LINKED ISSUE #580: Wysiwyg vs markdown?
Should we go back to markdown, both?

frjo:
This adds a markdown stream field for use in application forms and adds markdown support to the comments form on submissions.

The Markdown editor from Stack Exchange, PageDown, is used. The mistune library is used to convert Markdown to HTML, very easy to replace but it seems fast and popular.

PageDown displayes a HTML preview by default. Handy I believe, but easy to turn of.

After the text is run through the markdown filterit’s also run through the bleach filter. Bleach is used in most output to strip bad stuff, limit the allowed html tags etc.

The following Bleach Settings are in use on the site now, and has been from the start.

BLEACH_ALLOWED_TAGS = ['h2', 'h3', 'p', 'b', 'i', 'em', 'strong', 'a', 'ul', 'ol', 'li', 'br']
BLEACH_ALLOWED_ATTRIBUTES = ['href', 'title', 'style']
BLEACH_ALLOWED_STYLES = ['font-family', 'font-weight', 'text-decoration', 'font-variant']
BLEACH_STRIP_TAGS = True
BLEACH_STRIP_COMMENTS = True

P.s. Should we really allow ”style”?

Fixes #333  
Fixes #567 
Fixes #580


frjo:
@danblah Please test this at https://test-apply.opentech.fund/apply/submissions/10/#communications

danblah:
The list function (which is super key) doesnt seem to be working ofr me. Here's a screen share: https://www.useloom.com/share/83192f1284454424993d9de8dc3c6f6c

Can we show/hide the html preview? 

This still leaves the file attachment request open in #567 and im not sure completes the spirit of #330 but is a huge leap forward.

frjo:
The editor is not as nice as the one we had for Drupal. As you found out you can not make multiple lines in to a list, only line by line. Will continue looking for a better one, not hard to replace.

It should be #333 and not #330.

The HTML preview can easily be disabled if you want. I think it should be on for commentes sice you can’t edit them but maybe of for submissions since the forms is really long as they are.

danblah:
Ok, let's go with this one and investigate other options. Is def a huge improvement.

PR #751, opened 2018-11-28
Make the e-mail stream field validate e-mail addresses.

frjo:


zerolab:
LGTM, but you probably want a quick test for this, in case anything changes in the future.

PR #757, opened 2018-11-30
Add slack channel field to Funds and Labs.

### LINKED ISSUE #756: Give fund and lab mangers ability to send notifications to specific Slack channels [Medium/New feature]
-=Sensitive=-
No

-=Reporter=-
Blah

-=Ideal deadline=-

-=Who's affected=-
OTF team

-=Description=-
In the previous system, activity for specific funds, (like the Rapid Response Fund)  was able to be sent to a specific Slack channel, in addition to the all notifications channel. Let's bring that back. For all fund and lab pages, let's introduce a field to specify a Slack channel that all activity notification specific to fund or lab also goes to.

-=Objective=-
Make it easier for the fund/lab managers to see the things important to them

-=Additiional info=-

-=Timestamp=-
2018-11-29T17:11:54.648Z

frjo:
Fixes #756


frjo:
@todd-dembrey This fails on makemigration for some reason. Can not replicate it locally, can you see what I did wrong?

todd-dembrey:
1. We should probably increase the verbosity of the make migrations command.
2. I think that the setting varies between environments: `settings.SLACK_DESTINATION_ROOM`. so that when you come to make the migrations the help text updates.

Will need to have a think about the best way to handle this.

opentech/apply/funds/models/applications.py
```diff
@@ -48,6 +49,8 @@ class ApplicationBase(EmailForm, WorkflowStreamForm):  # type: ignore
         blank=True,
     )
 
+    slack_channel = models.TextField(blank=True, help_text=_('The slack #channel for notifications. Default channel is {slack_default}').format(slack_default=settings.SLACK_DESTINATION_ROOM))
```
Parbhat: @frjo We can use CharField here as the slack channel will be shorter in most cases.

Parbhat: As Todd mentioned setting varies between environments. It requires a new migration in a different environment as `help_text` changes. We can hard-code the slack channel in help text as it will not cause any harm if the channel name change.

PR #760, opened 2018-12-06
Make submission shortcodes in the format #1234 expand to links to the…

### LINKED ISSUE #732: Ability to reference other submissions in reviews or comments like you can with github issue comments

frjo:
… submission in reviews and comments.

Fixes #732

### DIFF
--- a/opentech/apply/funds/templatetags/submission_tags.py
+++ b/opentech/apply/funds/templatetags/submission_tags.py
@@ -10,18 +10,15 @@ register = template.Library()
 
 @register.filter
 def submission_links(value):
-    regex = re.compile('[^\w]\#(\d+)[^\w]')
+    # Match tags in the format #123 that is not preceeded and/or followed by a word character.
+    matches = re.findall('(?<!\w)\#(\d+)(?!\w)', value)
     links = {}
-    for match in regex.finditer(value):
-        try:
-            submission = ApplicationSubmission.objects.get(id=match[1])
-        except ApplicationSubmission.DoesNotExist:
-            pass
-        else:
-            links[f'#{submission.id}'] = f'<a href="{submission.get_absolute_url()}">{submission.title} <span class="mid-grey-text">#{submission.id}</span></a>'
+    if matches:
+        for submission in ApplicationSubmission.objects.filter(id__in=matches):
+            links[f'\#{submission.id}'] = f'<a href="{submission.get_absolute_url()}">{submission.title} <span class="mid-grey-text">#{submission.id}</span></a>'
 
     if links:
         for sid, link in links.items():
-            value = value.replace(sid, link)
+            value = re.sub(f'(?<!\w){sid}(?!\w)', link, value)
 
     return mark_safe(value)
--- a/opentech/apply/funds/templatetags/submission_tags.py
+++ b/opentech/apply/funds/templatetags/submission_tags.py
@@ -1,0 +1,27 @@
+import re
+
+from django import template
+from django.utils.safestring import mark_safe
+
+from opentech.apply.funds.models import ApplicationSubmission
+
+register = template.Library()
+
+
+@register.filter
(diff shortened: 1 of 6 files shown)


opentech/apply/funds/templatetags/submission_tags.py
```diff
@@ -0,0 +1,27 @@
...
+        try:
+            submission = ApplicationSubmission.objects.get(id=match[1])
+        except ApplicationSubmission.DoesNotExist:
+            pass
+        else:
+            links[f'#{submission.id}'] = f'<a href="{submission.get_absolute_url()}">{submission.title} <span class="mid-grey-text">#{submission.id}</span></a>'
+
+    if links:
+        for sid, link in links.items():
+            value = value.replace(sid, link)
```
todd-dembrey: this should probably be `re.sub()` because otherwise you risk replacing all `#1234` with the link to `#123`

e.g. sub pattern  `f'[^\w]({sid})[^\w]`

https://docs.python.org/3/library/re.html#re.sub

opentech/apply/funds/templatetags/submission_tags.py
```diff
@@ -0,0 +1,27 @@
...
+register = template.Library()
+
+
+@register.filter
+def submission_links(value):
+    regex = re.compile('[^\w]\#(\d+)[^\w]')
+    links = {}
+    for match in regex.finditer(value):
+        try:
+            submission = ApplicationSubmission.objects.get(id=match[1])
```
todd-dembrey: Maybe we should be looking at using `findall` instead of `finditer` and then doing:

```
referenced_submissions = ApplicationSubmission.objects.filter(id__in=list_of_matches).only('id', 'title)
links = {
    f'#{submission.id}': f'<a href="{submission.get_absolute_url()}">{submission.title} <span class="mid-grey-text">
    for submission in referenced_submissions
}
``` 

in order to get a smaller, single, query for the searches. At a minimum adding `.only` to the existing code is recommended.

todd-dembrey:
Also worth throwing a test in as well.

opentech/apply/funds/templatetags/submission_tags.py
frjo: Good catch, this is a very real possibility. Changed now and worth the performance loss.

opentech/apply/funds/templatetags/submission_tags.py
frjo: One query for better than many, fixed that.

Did look up "only" but title is not a real field so not possible to use only('id', 'title'). Title is in form_data and when you have loaded that you have loaded more or less everything so not much to save?

opentech/apply/funds/templatetags/submission_tags.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@register.filter
+def submission_links(value):
+    # Match tags in the format #123 that is not preceeded and/or followed by a word character.
+    matches = re.findall('(?<!\w)\#(\d+)(?!\w)', value)
+    links = {}
+    if matches:
+        for submission in ApplicationSubmission.objects.filter(id__in=matches):
+            links[f'\#{submission.id}'] = f'<a href="/apply/submissions/{submission.id}/">{submission.title} <span class="mid-grey-text">#{submission.id}</span></a>'
```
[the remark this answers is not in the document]
todd-dembrey: You need to make sure the correct urlconf if used. This is a result of only enabling the urls for the apply site and not the public.  

See here: https://github.com/OpenTechFund/opentech.fund/blob/4b878dc87440127e15f6afc40b50a0cbd119277a/opentech/apply/utils/testing/tests.py#L20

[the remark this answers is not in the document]
frjo: Thanks, that fixes the problem!

PR #763, opened 2018-12-11
Use count query instead of retrieving list and counting list length.

stdevteam:


todd-dembrey:
As per the docs, if we are reusing the queryset (as we are in the table here: https://github.com/OpenTechFund/opentech.fund/blob/b6880d1127d0fd9a61fd0779d2dfc7fbcb2f4a30/opentech/apply/funds/templates/funds/submissions.html#L46) then it should be quicker and cheaper to just load the queryset using `len`  https://docs.djangoproject.com/en/2.1/ref/models/querysets/#count

However because this may be two different querysets in which case we are making no savings. Can you do a benchmark to prove this is inefficient or look at reusing the data from the table so we only do the one query.

frjo:
I can confirm fewer queries and faster load, clearly noticeable improvement. The view is still to slow but we are getting there.

PR #769, opened 2018-12-17
Use Mailgun via javascript to validate e-mail addresses in submission forms

frjo:


frjo:
@Parbhat Does this look like a good solution? It's client side so possible to circumvent for a user but that is not a problem, the backend have a valid email address check.

The goal is to help users avoid misspelled addresses when they submit applications.

PR #775, opened 2018-12-19
Add a partner group who's members can be added to submissions and edit as well as review them

### LINKED ISSUE #608: We need a new role / user type called Service Partner, that can be assigned (added to) submissions from labs and funds to review and refine like an applicant
We get submissions from applicants who's requested work will be completed by contracted service partners. We'd like them to be able to review and refine submissions with applications before OTF's review,  AC review, and final review.

### LINKED ISSUE #771: We need a new role / user type called Host Organization, that can be assigned (added to) submissions from funds/fellows, that can submit reviews and refine submission like an applicant

frjo:
Fixes #608 
Fixes #771


opentech/apply/users/migrations/0008_add_partner_group.py
```diff
@@ -0,0 +1,44 @@
...
+                group.permissions.add(Permission.objects.get(codename=permission))
+            except ObjectDoesNotExist:
+                print("Could not find the '%s' permission" % permission)
+
+
+def remove_groups(apps, schema_editor):
+    Group = apps.get_model('auth.Group')
+
+    groups = [group_data['name'] for group_data in GROUPS]
+    Group.objects.filter(name__in=groups).delete()
```
zerolab: the trouble here is this will remove all groups should you revert just this migration. You only want to remove `TEAMADMIN_GROUP_NAME` and `PARTNER_GROUP_NAME`

frjo: Thanks for finding this bug. Hopefully fixed now.

frjo:
update_partners

I'm missing something so the form does not populate. Any tips?

frjo:
That worked perfectly, thanks!

opentech/apply/funds/views.py
```diff
@@ -316,7 +342,7 @@ class SubmissionDetailView(ViewDispatcher):
     applicant_view = ApplicantSubmissionDetailView
 
     def admin_check(self, request):
-        if request.user.is_reviewer:
+        if request.user.is_reviewer or request.user.is_partner:
```
frjo: @Parbhat I need to check not only that the user is a partner but also that they are assigned as a partner on this specific submission.

How do you recommend I do that best? The submission object does not seem to be available at this stage.

frjo:
@Parbhat I can you take on completing this PR. It's the permissions that are left to do, the complicated part.

Partners should have the same permissions as Reviewers, but only on the submissions that they have been added too.

Partners should also be able to edit submissions at the same stages that applicants can edit them.

Parbhat:
@frjo Will go through the requirements and work on it.

opentech/apply/funds/views.py
todd-dembrey: I don't think you want to give partners the admin view of the application. They should have the same access rights as a normal applicant?

frjo: OTF has requested a mayor change to what reviewers can see but I think that is outside of this ticket.

This access solution is needed either way so solve it in this pr and then we take on separating the views etc. in another pr.

Parbhat:
@frjo PR updated to only allow partners to view the submissions they are added as partners. What are the next steps?

danblah:
@frjo, @Parbhat, and @todd-dembrey bump bump. Need this tested/accepted/deployed soonest so we can properly run the ICFP round that's closed yesterday.

PR #788, opened 2019-01-04
Add Kosovo to the address field.

### LINKED ISSUE #782: Kosovo not in the Address field
The Internet Freedom and Core Infrastructure application form's address section does not include Kosovo in the list of countries.

frjo:
Fixes #782


frjo:
Why is this not working for me?

* Kosovo is added to the addressfield.min.json
* Have run collectstatic and confired that the updated file is preset at /static/addressfield.min.json

But Kosovo does not show up as an option in the address field on existing application forms where the addressfiled is present.

frjo:
Excellent, now it works! Thanks for fixing this.

Issue #768, opened 2018-12-14
API spec | Winter 2019


todd-dembrey:
Below is the API spec I think will be required in order to complete the work outlined in the functional spec #765.

It looks like some of the queries on submissions (status and active) will need to be added after we some work in Sprint 1, but otherwise the rest looks to be independent.

Would appreciate some feedback/comments on this.

https://docs.google.com/document/d/1rh2q7VCxUFu9wem06Z5rutspEJlFBuzgBIUoZGNPM78/edit?usp=sharing

frjo:
I have added the following domains to Cloudflare and Heroku:

* api.opentech.fund
* test-api.opentech.fund
* dev-api.opentech.fund
* sandbox-api.opentech.fund

PR #794, opened 2019-01-07
WIP: API spec implementation

Parbhat:
link #768 

Initial work: DRF integration (like default authentication), staff permissions and submissions list endpoint.

Example endpoint: `/apply/api/submissions/`. Normal submissions URL is `/apply/submissions/`. It seems better to contain API endpoints under `***/api/***`.

### DIFF
--- a/opentech/apply/funds/api_views.py
+++ b/opentech/apply/funds/api_views.py
@@ -1,0 +1,25 @@
+from rest_framework import generics
+from rest_framework import permissions
+from django_filters.rest_framework import DjangoFilterBackend
+
+from .models import ApplicationSubmission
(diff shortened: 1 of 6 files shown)


todd-dembrey:
Can we prefix the api with `/v1/`? A nice to have. Namespace versioning should probably be enough.
https://www.django-rest-framework.org/api-guide/versioning/#namespaceversioning

@frjo did you want to serve this under the same host name (www.opentech.fund) or would `api.opentech.fund` be better in the long run?

frjo:
Both the "/vi" prefix and a separate domain api.opentech.fund is very good ideas that should be implemented. Work on this has moved to the STDev team so they will pick this up.

PR #797, opened 2019-01-08
789-filters-behaviour

chris-lawton:


chris-lawton:
Hi @hippogriffic - would you be able to review this for me please? Thanks!

opentech/apply/funds/templates/funds/base_submissions_table.html
```diff
@@ -8,25 +8,29 @@
 
 {% block content %}
     {% block table %}
+    <div class="wrapper wrapper--table-actions">
         {% if table.data or active_filters %}
-        <div class="button button--filters button--contains-icons js-open-filters">Filter By</div>
+            <div class="button button--filters button--contains-icons js-toggle-filters">Filters</div>
```
hippogriffic: is there a reason you're going with a `div` and not `button`?

opentech/apply/funds/templates/funds/base_submissions_table.html
```diff
@@ -8,25 +8,29 @@
...
-        <div class="filters js-filter-wrapper">
-            <div class="filters__header">
-                <div class="js-clear-filters">Clear</div>
-                <div>Filter by</div>
-                <div class="js-close-filters">Close</div>
-            </div>
+        {% if table.data or active_filters %}
+            <div class="filters">
+                <div class="filters__header">
+                    <div class="js-clear-filters">Clear</div>
```
hippogriffic: same here for `clear` and `close`, seems like they could be `button` elements

opentech/apply/funds/templates/funds/base_submissions_table.html
chris-lawton: Think this was from the first phase of work, but agree that they should all be buttons. Will swap.

PR #825, opened 2019-01-09
adds lvh.me domains to allowed hosts

chris-lawton:
- Adds `lvh.me` and `apply.lvh.me` domains to `ALLOWED_HOSTS` 

Browsers other than Chrome won't handle `*.localhost` subdomains without modifying hosts files - when browser testing this isn't practical so the addition of these domains will remedy this.

### DIFF
--- a/opentech/settings/dev.py
+++ b/opentech/settings/dev.py
@@ -8,7 +8,7 @@ SECRET_KEY = 'CHANGEME!!!'
 
 INTERNAL_IPS = ('127.0.0.1', '10.0.2.2')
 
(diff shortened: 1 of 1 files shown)


frjo:
I use the "test" tld for all local env, but that needs a local dns or edits to the host file.

I see a point in avoiding that for quicker setup but would it not be better for OTF to set up their own local dev domains pointed to 127.0.0.1 (i.e. localhost)?

Something like:

* dev.otf.is
* dev-apply.otf.is

zerolab:
During phase one I used otf.test / apply.otf.test, added to local settings, so should not be an issue.

@frjo your suggestion for local dev domains is a very good one and would be consistent for everyone 👍

danblah:
set dev.otf.is and dev-apply.otf.is to 127.0.0.1

PR #828, opened 2019-01-10
Give the staff group access to the view message log on submissions.

### LINKED ISSUE #590: Create a permission to expose the "VIEW MESSAGE LOG" ability to users in a group (staff)
Right now this is only available to admins. This is becoming super important for round leads to confirm what's been sent and for us to figure out what edits to the msgs templates we need to make.

frjo:
Fixes #590


opentech/apply/funds/templates/funds/applicationsubmission_detail.html
```diff
@@ -32,7 +32,7 @@ <h5 class="heading heading--meta">
                 <a class="tab__item" href="#activity-feed" data-tab="tab-3">
                     Activity Feed
                 </a>
-                {% if request.user.is_superuser %}
+                {% if request.user.is_staff %}
```
Parbhat: [is_apply_staff](https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/apply/users/models.py#L90) property seems better here.

frjo: Good catch, have fixed that now.

PR #830, opened 2019-01-10
WIP: Feature/809 rounds block fe

chris-lawton:
Frontend styles for the rounds block on the submissions page
Contains hardcoded data for BE to pick up


todd-dembrey:
Closed in favour of #881

Issue #806, opened 2019-01-08
Screening: Filters and display


gibbsmaya:
**Acceptance criteria**

- [x] Submission detail page will be updated to include the outcome of the screening in the sidebar (as per determination block, placed above determination block)
- [x] Screening block on submission detail page only visible to OTF staff and admin
- [x] 'Screening' to be added as a column in submissions table to display outcome of screening (currently called remit in designs) 
- [x] 'Screening' to be added as filter on submissions table 
- [x] Screening activity will be visible in the activity feed for OTF Staff
- [x] This taxonomy is not visible to the applicant at any time

Design: https://projects.invisionapp.com/share/KZPIO1NENQJ#/screens/335805636

**QA criteria**
Dev:

- [x] checked feature meets acceptance criteria/conforms exactly to the specification.
- [x] provided good unit test coverage (if this is non-trivial behaviour).
- [x] checked all tests for the project pass with this feature enabled.
- [x] checked code conforms to the project coding standards.
- [x] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end user of the website/app (Can I get to it? Is it useable? Can I break it? Does it work in an end-to-end context?)
- [x] checked that this feature works on the server/s I am deploying it to.

QA:

 * [x] tested this feature as a front end user and it meets the acceptance criteria/conforms to the specification and design.
 * [x] checked that the feature works on the server/s deployed to

todd-dembrey:
@gibbsmaya This is now deployed to dev for testing

gibbsmaya:
This all looks good to me. Marking as done.

gibbsmaya:
@todd-dembrey please can you deploy this to test?

gibbsmaya:
Hi @fourthletter, @danblah

This work is now ready for testing. This work follows on from #805 , and consists of displaying the screening action on the submission page, the submissions listing and the activity feed, as well as adding a filter for the screening outcome to the submissions listing.

Please test this against the acceptance criteria noted in the initial issue description.

Please let us know if anything is unclear or if you have any questions. Otherwise please provide feedback of any issues you find or if everything is working as expected, please confirm and we can get this work deployed to the live site and issue closed off.

Thanks,
Maya

danblah:
See https://test-apply.opentech.fund/apply/submissions/4/#activity-feed, when applying screen term, the activity feed fails to load, shows nothing for team member. Aside from that, all other criteria is met.

[...]

PR #831, opened 2019-01-10
Feature/805 Screening: Wagtail interface and action on submission page 

emullaney:
See https://github.com/OpenTechFund/opentech.fund/issues/805 (copied here)

- New taxonomy added to Wagtail 'Screening'
- Taxonomy controlled through Wagtail by admin role only
- New action “Screen application” added to the submissions detail page for roles OTF staff and admin
- "Screen" Action allows selection of one taxonomy term
- Record activity when screening action is taken
- Screening cannot be changed except by admin if the submission has received a final determination (Accepted/Dismissed/Invited to Proposal).
- Tests added to test screening form shows up appropriately per user type


opentech/apply/funds/models/submissions.py
```diff
@@ -309,6 +309,14 @@ class ApplicationSubmission(
     # Workflow inherited from WorkflowHelpers
     status = FSMField(default=INITIAL_STATE, protected=True)
 
+    screening_status = models.ForeignKey(
+        'funds.ScreeningStatus',
+        related_name='+',
+        on_delete=models.CASCADE,
```
todd-dembrey: This should probably be `SET_NULL` or `PROTECT`. We don't want to loose submissions if a status is removed.

opentech/apply/funds/forms.py
```diff
@@ -22,6 +22,24 @@ def __init__(self, *args, **kwargs):
         self.should_show = bool(choices)
 
 
+class ScreeningSubmissionForm(forms.ModelForm):
+    action = forms.ChoiceField(label='Update screening status')
```
todd-dembrey: this should be: `screening_status` within the Meta fields. Then everything works as a model choice field without the setup you have in the init. Or am i missing something?

opentech/apply/funds/views.py
```diff
@@ -90,6 +90,33 @@ def form_valid(self, form):
...
 
 
+@method_decorator(staff_required, name='dispatch')
+class ScreeningSubmissionView(DelegatedViewMixin, UpdateView):
+    model = ApplicationSubmission
+    form_class = ScreeningSubmissionForm
+    context_name = 'screening_form'
+
+    def form_valid(self, form):
+        action = form.cleaned_data.get('action')
```
todd-dembrey: If you are making the changes above to the form then this simplifies as per the `UpdateLeadView` (i think).


1. get old value before save - no need to do the copy as you should have the screening object not the submission instance
2. save using super
3. send message

I would rather the messaging handles any of the logic around what is shown (i.e. set/set from), so just passing None or the object is the way to go.

opentech/apply/funds/forms.py
```diff
@@ -22,6 +22,24 @@ def __init__(self, *args, **kwargs):
...
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop('user')
+        super().__init__(*args, **kwargs)
+        choices = [(status.pk, status.title) for status in ScreeningStatus.objects.all()]
+        action_field = self.fields['action']
+        action_field.choices = choices
+        self.should_show = False
+        if (self.instance.active and self.user.is_apply_staff) or self.user.is_superuser:
+            self.should_show = True
```
todd-dembrey: this looks good.

opentech/apply/funds/views.py
todd-dembrey: You might have some fun with the generic relationship on the activity, but will address that in #806

opentech/apply/funds/models/submissions.py
emullaney: Of course! Agreed.

opentech/apply/funds/forms.py
emullaney: You're not missing anything, I messed this up during my copy/pasting from another existing form.

opentech/apply/funds/views.py
emullaney: Yes, this all makes total sense! The messaging was a little tricky, but then I realized I could send in a single character if screening status is None ("-") so that the messaging would make sense and not error out on getting a None value.

todd-dembrey: OK, lets just confirm when looking at #806 that this still makes sense.

opentech/apply/funds/tests/test_views.py
```diff
@@ -608,3 +632,36 @@ def test_cant_access_normal_page(self):
...
+        self.refresh(self.submission)
+
+    def test_can_screen_submission(self):
+        screening_outcome = ScreeningStatusFactory()
+        self.post_page(self.submission, {'form-submitted-screening_form': '', 'screening_status': screening_outcome.id})
+        submission = self.refresh(self.submission)
+        self.assertEqual(submission.screening_status, screening_outcome)
+
+    def test_cant_screen_submission(self):
+        submission = ApplicationSubmissionFactory(lead=self.user)
```
todd-dembrey: `ApplicationSubmissionFactory(rejected=True)` should be suitable for this test. We dont want the user to be lead as this might confuse the tests - especially if we decide to give the lead special permissions later.

opentech/apply/funds/tests/test_views.py
```diff
@@ -608,3 +632,36 @@ def test_cant_access_normal_page(self):
...
+        DeterminationFactory(submission=submission, rejected=True, submitted=True)
```
todd-dembrey: You shouldnt need a determination - we are just interested in the status

opentech/apply/funds/tests/test_views.py
```diff
@@ -608,3 +632,36 @@ def test_cant_access_normal_page(self):
...
+    def __setUp__(self):
+        self.refresh(self.submission)
+
+    def test_can_screen_submission(self):
+        screening_outcome = ScreeningStatusFactory()
+        self.post_page(self.submission, {'form-submitted-screening_form': '', 'screening_status': screening_outcome.id})
+        submission = self.refresh(self.submission)
+        self.assertEqual(submission.screening_status, screening_outcome)
+
+    def test_cant_screen_submission(self):
```
todd-dembrey: Name is incorrect - `test_can_screen_applications_in_final_status`?

opentech/apply/funds/tests/test_views.py
```diff
@@ -361,6 +381,10 @@ def test_cant_edit_other_submission(self):
         response = self.get_page(submission, 'edit')
         self.assertEqual(response.status_code, 403)
 
+    def test_cant_screen_submission(self):
+        response = self.post_page(self.submission)
+        self.assertNotIn('screening_form', response.context_data)
```
todd-dembrey: We also want to confirm they cant submit the form (just to be sure)

opentech/apply/funds/tests/test_views.py
```diff
@@ -182,6 +183,25 @@ def test_not_included_fields_render(self):
...
         self.assertNotContains(response, 'Value')
 
+    def test_can_screen_submission(self):
+        screening_outcome = ScreeningStatusFactory()
+        self.post_page(self.submission, {'form-submitted-screening_form': '', 'screening_status': screening_outcome.id})
+        submission = self.refresh(self.submission)
+        self.assertEqual(submission.screening_status, screening_outcome)
+
+    def test_cant_screen_submission(self):
+        submission = ApplicationSubmissionFactory(lead=self.user)
```
todd-dembrey: See the Commented about `rejected=True` and `lead` for the SuperUser case.

todd-dembrey:
The rebase has caused a conflict for the migration. Could you push a fix for that as well please.

opentech/apply/funds/tests/test_views.py
emullaney: Oh nice, I didn't notice that was an option for ApplicationSubmissionFactory(), that will be helpful for testing.

opentech/apply/funds/tests/test_views.py
emullaney: Makes sense

opentech/apply/funds/tests/test_views.py
emullaney: ... copy paste error strikes again, thanks for the catch!

opentech/apply/funds/admin.py
```diff
@@ -75,6 +97,7 @@ class ApplyAdminGroup(ModelAdminGroup):
     menu_icon = 'folder-open-inverse'
     items = (
         RoundAdmin,
+        ScreeningStatusAdmin,
```
todd-dembrey: I think putting it at the top or bottom of the list is probably more reasonable.

opentech/apply/funds/admin.py
```diff
@@ -75,6 +98,7 @@ class ApplyAdminGroup(ModelAdminGroup):
...
```
todd-dembrey: I think putting this lower down the order is more reasonable. Before or after `CategoryAdmin` i think

todd-dembrey: I have done this.

PR #838, opened 2019-01-14
Feature/792 update statuses

todd-dembrey:
This implements the changes to statuses to enable the applicants to see different statuses than the staff.

Includes
* Updates to notifications so that:
  * applicants dont see activity items internal changes
  * applicants see different text to staff where nessecary
  * applicants dont get emails for internal items
* Updates to the status bar to reflect:
  * Internal names to staff
  * public names to applicants
  * Hidden phases from applicants
  * Future statuses shown where they haven't yet reached a phase
* Updates to workflow in order to reflect all of the above.


opentech/apply/funds/workflow.py
```diff
@@ -79,60 +113,48 @@ def __str__(self):
...
 
 
-class BasePermissions:
-    def can_edit(self, user: 'User') -> bool:
-        if user.is_apply_staff:
-            return self.can_staff_edit(user)
-
-        if user.is_applicant:
-            return self.can_applicant_edit(user)
+class Permissions:
```
todd-dembrey: I like the changes I have made to permissions generally here. they are easier to compose together.

However I think the addition of the can_view permission is not helpful. Probably adding a `publicly_visible` property to the phase might be more reasonable.

`hidden_from_applicant_permissions` just feels so clunky. especially as it inherits the default and overrides it.

emullaney: What don't you like about `hidden_from_applicant_permissions`?

PR #857, opened 2019-01-17
Feature/811 all round page

todd-dembrey:


opentech/apply/funds/tables.py
```diff
@@ -150,3 +152,102 @@ def __init__(self, *args, exclude=list(), **kwargs):
...
+    def order_progress(self, qs, desc):
+        return qs.order_by(self._field_order('progress', desc)), True
+
+
+# TODO remove in django 2.1 where this is fixed
+F.relabeled_clone = lambda self, relabels: self
+
+
+# TODO remove in django 2.1 where this is added
+class Left(Func):
```
todd-dembrey: @frjo These are the fixes i was hoping to have from 2.1. Not much to include here but worth remembering for future reference.

Left Function
https://docs.djangoproject.com/en/2.1/ref/models/database-functions/#left

Fixes an inssue with `F` and `OuterRef`
https://code.djangoproject.com/ticket/29142

PR #862, opened 2019-01-17
Feature/drf integretion stdev

frjo:
Link to spec here: https://docs.google.com/document/d/1rh2q7VCxUFu9wem06Z5rutspEJlFBuzgBIUoZGNPM78/

### DIFF
--- a/opentech/apply/funds/serializers.py
+++ b/opentech/apply/funds/serializers.py
@@ -1,15 +1,31 @@
 from rest_framework import serializers
 
-from .models import ApplicationSubmission
+from opentech.apply.activity.models import Activity
+from .models import ApplicationSubmission, Round
 
 
 class SubmissionListSerializer(serializers.ModelSerializer):
+    title = serializers.CharField(source='page.title')
+
     class Meta:
         model = ApplicationSubmission
-        fields = ('id',)
+        fields = ('id', 'title', 'status')
 
 
 class SubmissionDetailSerializer(serializers.ModelSerializer):
+    actions = serializers.ListField(
+        child=serializers.PrimaryKeyRelatedField(
+            queryset=Activity.actions.all()
+        )
+    )
+
     class Meta:
         model = ApplicationSubmission
-        fields = ('id', 'title',)
+        fields = ('id', 'title', 'actions')
+
+
+class RoundDetailSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Round
+        fields = ('title', )
+        # TODO check what statuses are and add statuses list in fields
--- a/opentech/apply/funds/serializers.py
+++ b/opentech/apply/funds/serializers.py
@@ -1,0 +1,15 @@
+from rest_framework import serializers
+
+from .models import ApplicationSubmission
+
+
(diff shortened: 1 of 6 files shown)


opentech/apply/funds/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+
+class SubmissionListSerializer(serializers.ModelSerializer):
+    title = serializers.CharField(source='page.title')
+
+    class Meta:
+        model = ApplicationSubmission
+        fields = ('id', 'title', 'status')
+
+
+class SubmissionDetailSerializer(serializers.ModelSerializer):
```
todd-dembrey: We have ha to write some of this for the work we have been doing.

See below for how we serialised the questions/meta

https://github.com/OpenTechFund/opentech.fund/pull/852/files#diff-7ba15010bea2ee21506ce1d29441cafe

todd-dembrey:
1. What kind of object is the "Status" ?

They are states set on submissions to create a workflow. Done with <https://github.com/viewflow/django-fsm>. 

You can access the statuses from the workflow object attached to the round. `object.workflow`. The workflow is then iterable for phases which have both a `name` and a `display_name` that we are interested in.
https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/apply/funds/workflow.py

I perhaps confused the issue calling it a status.

 2. "Query" - as we understood from the comment it doesn't exist: Should we develop it? If yes - should it be in the submission? (Is it a submission attribute?)

The "Query" is any combination of lookups. e.g. `?round=23&?active=true`

The `active` query which i had highlighted didnt exist previously. I have implemented it here 
https://github.com/OpenTechFund/opentech.fund/pull/857/files#diff-1970bcbbc085947392119b95e3303063R191
and 
https://github.com/OpenTechFund/opentech.fund/pull/857/files#diff-69bf3b7f321569af3a40118621f7c5f3R522

Which may help your understanding. I think i need to make this a bit more generic, but you should get the idea.

3. "Required questions - (value, duration, name, email, address)_
what kind of objects are those and how can we find them in the database.

See link above https://github.com/OpenTechFund/opentech.fund/pull/862/files/90db685355556c800dcc5c7d7581bf83b49eff48#r248704254

4. "Action transition" - As we understood the end point requires just the action name. Is it correct?

You should be able to POST /submissions/<id>/<action_name> to perform that action of submission id.

Actions and how to transition them are defined as per the links in the document: 
Get valid actions - https://github.com/OpenTechFund/opentech.fund/blob/4e1a4fd5d465eaeef79f5b704f28b0b28b652b36/opentech/apply/funds/forms.py#L19
Perform transition - https://github.com/OpenTechFund/opentech.fund/blob/4e1a4fd5d465eaeef79f5b704f28b0b28b652b36/opentech/apply/funds/views.py#L87

5. How can we find the submission comments in database?

comments are based on the activity model https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/apply/activity/models.py#L92
See here for an example of fetching the comments.
https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/apply/activity/views.py#L40-L44

PR #870, opened 2019-01-21
Feature/api sprint 2

todd-dembrey:
Will need to be base winter sprint when everything else is merged.

Should really include some tests but this will unblock us for the next sprint.


todd-dembrey:
@tm-kn This is the branch which includes the new end points that are required for the sprint. I will merge it tomorrow morning when it passes the CI.

PR #891, opened 2019-01-22
Add notes list to the current submission view

tm-kn:


opentech/static_src/src/app/src/components/NotesPanelItem.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import moment from 'moment';
+
+export default class NotesPanelItem extends React.Component {
```
todd-dembrey: Surely this is just a `Note`?

opentech/static_src/src/app/src/components/NotesPanelItem.js
```diff
@@ -0,0 +1,24 @@
...
+    static propTypes = {
+        note: PropTypes.shape({
```
todd-dembrey: Maybe just accept the `user` `timestamp` `message` as props. Unsure about this one.

opentech/static_src/src/app/src/components/NotesPanel.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import LoadingPanel from '@components/LoadingPanel';
+
+export default class NotesPanel extends React.Component {
```
todd-dembrey: This shares a lot of functionality with the listing component. 

I think i would like to try and make the components less specific in that instance. 

So the NoteListing container would be:
```javascript

renderNote(id) {
    // assuming Note is connected to the store with a selector on ID
    // see https://github.com/reduxjs/reselect#accessing-react-props-in-selectors
    return <Note id=id>
}

error = <>
                <p><strong>Something went wrong!</strong>Sorry we couldn&rsquo;t load notes</p>
                <a onClick={handleRetry}>Refresh</a>.
            </>

<Listing 
    items={[<array of ids>]} // From `fetchNotesForSubmission`
    renderItem={renderNote}
    isError={isErrored}
    errorItem={error}
>

// ----------
<NoteListing submissionID=submissionID>
```

opentech/static_src/src/app/src/redux/actions/notes.js
```diff
@@ -0,0 +1,46 @@
...
+const startFetchingNotes = () => ({
+    type: START_FETCHING_NOTES,
+});
+
+const failFetchingNotes = message => ({
+    type: FAIL_FETCHING_NOTES,
+    message,
+});
+
+export const fetchNotesForSubmission = submissionID => {
```
todd-dembrey: We are starting to use this pattern quite a lot. Make an API request and handle `START`, `SUCCEED`, `FAIL`
I quite like what happens in the redux example. I might consider updating to this approach in the near future, I dont know if `redux-saga` or similar is required yet. What are your thoughts?

https://github.com/reduxjs/redux/blob/343cc97c194c69a1ac52e313efd6510e35a50703/examples/real-world/src/actions/index.js#L9-L15

https://github.com/reduxjs/redux/blob/343cc97c194c69a1ac52e313efd6510e35a50703/examples/real-world/src/middleware/api.js#L80

opentech/static_src/src/app/src/redux/reducers/notes.js
```diff
@@ -0,0 +1,68 @@
+import { combineReducers } from 'redux';
+
+import {
+    UPDATE_NOTE,
+    UPDATE_NOTES,
+    START_FETCHING_NOTES,
+    FAIL_FETCHING_NOTES,
+} from '@actions/notes';
+
+function notesFetching(state = false, action) {
```
todd-dembrey: We may want this on the note object or keyed by ID to the note.

I'm thinking of the case where we are fetching already and then the user changes the application they are looking at.

tm-kn: But they are submission specific, that's an issue. I guess they'd need to go to the submissions reducer.

opentech/static_src/src/app/src/components/GroupedListing.js
```diff
@@ -0,0 +1,124 @@
...
+                this.props.onItemSelection(firstGroup.items[0].id)
+            }
+        }
+    }
+
+    renderListItems() {
+        const { isLoading, error, items, onItemSelection, activeItem } = this.props;
+
+        if (isLoading) {
+            return (
```
todd-dembrey: Can we not delegate this to Listing such that:

```
<GroupedListing>
    <Listing onError={}>
</GroupedListing>
```

This way grouped listing is all about grouping?

opentech/static_src/src/app/src/redux/reducers/notes.js
todd-dembrey: As mentioned, I might need to look at this area when i am polling for new notes, so lets leave as is for now.

PR #896, opened 2019-01-23
Display errors and password rules on password reset form

frjo:
The password reset form was not displaying any errors making it hard for users to understand why the form did not submit.

<strike>When a user access a valid activation url their account are directly set to active and they are logged in. They are then redirected to a form where they should set a password.

For some reason many users simply skip this step and they end up with an active account with no password. Accounts with no password can not use the password reset form.

By using the "set_unusable_password()" to set a password in the first step users can at least use the password reset form if they skip creating a password for their account upon activation.

I have now set an unusable password on all active accounts with empty password as well.</strike>

~~~~
from django.contrib.auth import get_user_model
User = get_user_model()
users = User.objects.filter(password='', is_active=True).all()
for user in users:
	user.set_unusable_password()
	user.save()
~~~~

Since set_unusable_password does not allow the use of the password reset feature after all I have set a long random passord on the affected accounts

~~~~
from django.utils.crypto import get_random_string
from django.contrib.auth import get_user_model
User = get_user_model()
users = User.objects.filter(password__startswith='!', is_active=True).all()
for user in users:
        new_pass = get_random_string(length=32)
	user.set_password(new_pass)
	user.save()
~~~~

Old:

password_reset_old

New:

password_reset_new

### DIFF
--- a/opentech/apply/users/templates/users/password_reset/confirm.html
+++ b/opentech/apply/users/templates/users/password_reset/confirm.html
@@ -4,10 +4,31 @@
 {% block page_title %}Set your new password{% endblock %}
 {% block content %}
 <div class="wrapper wrapper--small wrapper--bottom-space">
-    <form class="form form--with-p-tags" method="post" novalidate>
+    <form class="form" method="post" novalidate>
+        {% if form.non_field_errors %}
+            <ul class="errorlist">
+                {% for error in form.non_field_errors %}
+                    <li>{{ error }}</li>
+                {% endfor %}
+            </ul>
+        {% endif %}
+
+        {% if form.errors %}
+            <ul class="errorlist">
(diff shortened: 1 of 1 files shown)


opentech/apply/users/views.py
```diff
@@ -71,6 +71,7 @@ def get(self, request, *args, **kwargs):
 
         if self.valid(user, kwargs.get('token')):
             user.is_active = True
+            user.set_unusable_password()
```
todd-dembrey: The alternative is to not activate on this redirect view and only set them as active when they have set a password on `create_password` view.

frjo: set_unusable_password() does not allow password reset to work, I could have sworn it worked when I first tested it but I must have done something wrong.

PR #899, opened 2019-01-23
WIP: Add notes submission

tm-kn:


todd-dembrey:
FYI: changed base for ease of review.

opentech/static_src/src/app/src/redux/reducers/notes.js
```diff
@@ -43,8 +45,44 @@ function note(state, action) {
...
+        case UPDATE_NOTE:
+        case FAIL_CREATING_NOTE_FOR_SUBMISSION:
+            return state.filter(v => v !== action.submissionID);
+        default:
+            return state
+    }
+}
+
+
+function notesFailedCreating(state = {}, action) {
```
todd-dembrey: I'm not sure how i feel about these two stores. They look like they work. But are they the best pattern. Not sure...

tm-kn: Should we add those to a particular submission object rather than having a separate store?

todd-dembrey: To be honest it is probably OK. What i'm trying to work out is the case:
> If we wanted to prevent changing the current submissions because we are performing an action on it - where would we be looking for the data? 

Perhaps it is ok where it is.

PR #902, opened 2019-01-24
Feature/845 jump to listing

chris-lawton:
Based off https://github.com/OpenTechFund/opentech.fund/tree/feature/872-notes-list which in turn is based off https://github.com/OpenTechFund/opentech.fund/tree/feature/875-submit-notes

- adds jump to listing functionality
- moves `<ListingDropdown>` into `<GroupedListing>`

### DIFF
--- a/opentech/apply/funds/templates/funds/base_submissions_table.html
+++ b/opentech/apply/funds/templates/funds/base_submissions_table.html
@@ -16,7 +16,6 @@
 
 {% block extra_js %}
     {{ filter.form.media.js }}
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/iamdustan-smoothscroll/0.4.0/smoothscroll.min.js"></script>
     <script src="{% static 'js/apply/all-submissions-table.js' %}"></script>
     <script src="https://cdn.jsdelivr.net/npm/symbol-es6@0.1.2/symbol-es6.min.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/url-search-params/1.1.0/url-search-params.js"></script>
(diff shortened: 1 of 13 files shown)


opentech/apply/funds/templates/funds/base_submissions_table.html
```diff
@@ -17,6 +17,7 @@
 
 {% block extra_js %}
     {{ filter.form.media.js }}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/iamdustan-smoothscroll/0.4.0/smoothscroll.min.js"></script>
```
todd-dembrey: Is this required still? and do we want it as a lib in package if including?

opentech/static_src/src/app/src/components/ListingDropdown/index.js
```diff
@@ -0,0 +1,41 @@
...
+import PropTypes from 'prop-types';
+
+export default class ListingDropdown extends React.Component {
+    static propTypes = {
+        listRef: PropTypes.object,
+        groups: PropTypes.array,
+    }
+
+    handleChange(e) {
+        const groupHeaderPosition = document.getElementById(e.target.value).offsetTop - 75;
```
todd-dembrey: this 75 is a bit magic...  i'm not sure i'm a fan. There must be a better way?

opentech/apply/funds/templates/funds/base_submissions_table.html
chris-lawton: nope, will remove

chris-lawton: still required to polyfill `element.scrollTo` in ms browsers, but will move it to be added as a package

opentech/static_src/src/app/src/components/ListingDropdown/index.js
chris-lawton: yup, agreed. I'll grab the height of the dropdown container then pass as a prop to `<ListingDropdown>` to use as an offset

PR #907, opened 2019-01-25
Updading social_auth_app_django to 3.x to handle new Google API for o…

frjo:
…auth2. The Google+ API is going away.

@todd-dembrey  I have set up new Authentication apps on Google as well but we need to switch to them at the same time as we update the social_auth_app_django package.

We should test this on test first to make sure I got all the parts right. How and when do we do this? The Google+ API is going away  by March 7, 2019 *but* will "intermittently fail as early as January 28, 2019" according to Google.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,7 +23,7 @@ responses==0.9.0
 
 flake8
 
-social_auth_app_django==2.1.0
+social_auth_app_django==3.1.0
 django-tables2==1.21.1
 django-filter==1.1.0
 django_select2==6.0.1


todd-dembrey:
I'm a bit confused as we are using the oauth 2 backend which hasn't changed between releases
https://github.com/OpenTechFund/opentech.fund/blob/60cfa85c21b1f1ad21021d14803d095f422d9ace/opentech/settings/base.py#L298

However i think what you have done here will not have an impact on the current behaviour.

**V2.0**
https://github.com/python-social-auth/social-core/blob/59987704f29b39d92a42ee2007c899acca706611/social_core/backends/google.py#L67-L70
```
    AUTHORIZATION_URL = 'https://accounts.google.com/o/oauth2/auth'
    ACCESS_TOKEN_URL = 'https://accounts.google.com/o/oauth2/token'
    ACCESS_TOKEN_METHOD = 'POST'
    REVOKE_TOKEN_URL = 'https://accounts.google.com/o/oauth2/revoke'
```

**V3.0**
https://github.com/python-social-auth/social-core/blob/5b7cf69c1a74bc836be7af5005608217e28968c1/social_core/backends/google.py#L74-L77
```
    AUTHORIZATION_URL = 'https://accounts.google.com/o/oauth2/auth'
    ACCESS_TOKEN_URL = 'https://accounts.google.com/o/oauth2/token'
    ACCESS_TOKEN_METHOD = 'POST'
    REVOKE_TOKEN_URL = 'https://accounts.google.com/o/oauth2/revoke'
```

PR #913, opened 2019-01-28
rounds fetching and ordering

todd-dembrey:
@tm-kn main reviewer. Would appreciate going over the fetching logic especially about loading and error states.

@chris-lawton can you confirm that you are happy with any of the drop down changes, functionally nothing should have changes

### DIFF
--- a/opentech/static_src/src/app/src/containers/ByStatusListing.js
+++ b/opentech/static_src/src/app/src/containers/ByStatusListing.js
@@ -5,6 +5,7 @@ import { connect } from 'react-redux'
 import GroupedListing from '@components/GroupedListing';
 import {
     loadCurrentRound,
+    loadCurrentRoundSubmissions,
     setCurrentSubmission,
 } from '@actions/submissions';
 import {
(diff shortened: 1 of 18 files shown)


opentech/static_src/src/app/src/containers/ByStatusListing.js
```diff
@@ -46,28 +49,34 @@ class ByStatusListing extends React.Component {
         }
     }
 
+    prepareOrder(round) {
+        if ( !round ) { return []}
+        const slugify = value => value.toLowerCase().replace(/\s/g, '-')
```
tm-kn: Maybe we could make it reusable and put it in utils?

opentech/static_src/src/app/src/redux/actions/submissions.js
```diff
@@ -33,16 +39,67 @@ export const setCurrentSubmission = id => ({
...
+export const fetchRound = roundID => {
+    return async function(dispatch) {
+        dispatch(startLoadingRound(roundID));
+        try {
+            const response = await api.fetchRound(roundID);
+            const json = await response.json();
+            if (response.ok) {
+                dispatch(updateRound(roundID, json));
+            } else {
+                dispatch(failLoadingRound(json.meta.error));
```
tm-kn: Won't it be `json.detail` if there's an error? Feels like we could write a generic function to handle errors instead.

PR #927, opened 2019-01-29
Feature/860 batch review

emullaney:
For https://github.com/OpenTechFund/opentech.fund/issues/860


opentech/apply/funds/views.py
```diff
@@ -68,22 +76,54 @@ def get_queryset(self):
...
+        Add any reviewers that were selected,  only if they are not
+        currently saved to that submission.
+        Send out a message of updates.
+        """
+        for submission_id in request.POST["submission_ids"].split(","):
+            submission = ApplicationSubmission.objects.get(id=submission_id)
+
+            old_reviewers = set(submission.reviewers.all())
+            reviewers = User.objects.filter(id__in=request.POST.getlist('staff_reviewers'))
+            for reviewer in reviewers:
```
todd-dembrey: `submission.reviews.add(*reviewers)`

opentech/apply/funds/views.py
```diff
@@ -68,22 +76,54 @@ def get_queryset(self):
...
+        Loop through all submissions selected on the page,
+        Add any reviewers that were selected,  only if they are not
+        currently saved to that submission.
+        Send out a message of updates.
+        """
+        for submission_id in request.POST["submission_ids"].split(","):
+            submission = ApplicationSubmission.objects.get(id=submission_id)
+
+            old_reviewers = set(submission.reviewers.all())
+            reviewers = User.objects.filter(id__in=request.POST.getlist('staff_reviewers'))
```
todd-dembrey: We probably want to be doing this outside the loop. so we only make the request for users once.

opentech/apply/funds/utils.py
```diff
@@ -0,0 +1,42 @@
+from opentech.apply.activity.messaging import messenger, MESSAGES
+from opentech.apply.users.models import User
+
+
+def can_alter_reviewers(instance, user):
```
todd-dembrey: For external only: `can_alter_external_reviewers` 

I should have got it right the first time :roll_eyes:

opentech/apply/funds/utils.py
```diff
@@ -0,0 +1,42 @@
...
+    return instance.stage.has_external_review and user == instance.lead
+
+
+def set_reviewers_fields(instance, user, fields):
```
todd-dembrey: I don't like that this is doing three things.

1. get the submitted reviewers
2. excluding people who have already submitted
3. hiding fields that we dont need

I'm also not too keen for this to become divorced from the form as it naturally belongs there. 

maybe more like the following on the form class: (maybe `get_excluded_reviewers`) isnt needed...
```python
assigned_reviewers = instance.reviewers.all()
excluded_reviewers = get_excluded_reviewers(instance)
self.fields['staff_reviewers'] = populate_field(self.fields['staff_reviewers'], assigned_reviewers, excluded_reviewers)
if can_alter_reviewers(instance, user):
    self.fields['reviewer_reviewers'] = populate_field(self.fields['reviewer_reviewers'], assigned_reviewers, excluded_reviewers)
else:
    self.fields.pop('reviewer_reviewers')
```

opentech/apply/funds/utils.py
```diff
@@ -0,0 +1,42 @@
...
+        review_field = fields['reviewer_reviewers']
+        review_field.queryset = review_field.queryset.exclude(id__in=submitted_reviewers)
+        review_field.initial = reviewers
+    else:
+        fields.pop('reviewer_reviewers')
+
+    return submitted_reviewers, fields
+
+
+def save_reviewers_message(old_reviewers, new_reviewers, request, submission):
```
todd-dembrey: You will need to be careful with this. I don't think it will be generic across all the two approached, single and batch.

We'll want this to handle `MESSAGES.BATCH_REVIEWERS_UPDATED` with slightly different messaging.

opentech/apply/funds/utils.py
emullaney: I moved it out of that form when I thought I would need to re-use it, and then I kept it out. I agree it should be moved back.

opentech/apply/funds/utils.py
emullaney: Yes, I was realizing that we'd need `MESSAGES.BATCH_REVIEWERS_UPDATED` as I was looking into the slack bulk message yesterday. I agree we might not even need a re-usable method here, we might want to move messaging to each individual form (batch and individual).

opentech/apply/activity/messaging.py
```diff
@@ -195,6 +239,12 @@ def send_message(self, message, user, submission, **kwargs):
         else:
             related_object = None
 
+        try:
+            # If this was a batch action we want to pull out the submission
+            submission = submissions[0]
+        except TypeError:
+            pass
+
```
emullaney: This makes sense after I looked at it for a minute but I think wrapping it in `if submissions` is clearer than a `try/except`
Not a deal breaker.

opentech/apply/activity/messaging.py
```diff
@@ -317,13 +385,17 @@ class EmailAdapter(AdapterBase):
...
-            subject = 'Application ready to review: {submission.title}'.format(submission=submission)
-        else:
-            subject = submission.page.specific.subject or 'Your application to Open Technology Fund: {submission.title}'.format(submission=submission)
+    def get_subject(self, message_type, submission):
+        if submission:
+            if message_type == MESSAGES.READY_FOR_REVIEW:
+                subject = 'Application ready to review: {submission.title}'.format(submission=submission)
+            else:
+                subject = submission.page.specific.subject or 'Your application to Open Technology Fund: {submission.title}'.format(submission=submission)
+            return subject
```
emullaney: I like this change.

opentech/apply/activity/messaging.py
```diff
@@ -365,24 +437,75 @@ def send_message(self, message, submission, subject, recipient, **kwargs):
             return 'Error: ' + str(e)
 
 
+class DjangoMessagesAdapter(AdapterBase):
```
emullaney: Good idea!

opentech/apply/activity/messaging.py
```diff
@@ -365,24 +437,75 @@ def send_message(self, message, submission, subject, recipient, **kwargs):
...
+        message = ['Reviewers updated.']
+        if added:
+            message.append('Added:')
+            message.append()
+
+        if removed:
+            message.append('Removed:')
+            message.append(', '.join([str(user) for user in removed]) + '.')
+
+        return ' '.join(message)
```
emullaney: Is this method used somewhere?

emullaney:
I'm hitting a few bugs on this latest version, but overall everything looks really great.

1. The select-all checkbox is somehow selecting each id twice. See my screenshots.
2. Now that DelegateableView is no longer inheriting from DetailView, the submission detail page is broken.
> AttributeError at /apply/submissions/3593/
type object 'AdminSubmissionDetailView' has no attribute 'as_view'

opentech/apply/funds/templates/funds/includes/batch_update_reviewer_form.html
```diff
@@ -0,0 +1,5 @@
+<div class="modal" id="batch-update-reviewers">
+    <h4>Update Reviewers</h4>
```
emullaney: I know that this matches the original requirements, however, we are really only adding reviewers here, not updating them (no reviewer gets removed by this update). I wonder if we need some more text here to explain what will happen or a different title. Just a thought.

opentech/apply/activity/messaging.py
todd-dembrey: no, i just copy pasted it and forgot to remove it...

emullaney:
All the updates look good on the back end side of things, this looks good to go to me.

PR #930, opened 2019-01-30
Feature/876 add note fe

chris-lawton:
dependant on `feature/873-notes-listing`

- styles add note form (react rte)


helenb:
CSS all looks good.

PR #932, opened 2019-01-30
Submission by status detail view

tm-kn:


opentech/static_src/src/app/src/redux/reducers/submissions.js
```diff
@@ -95,9 +99,47 @@ function currentSubmission(state = null, action) {
 }
 
 
+function submissionsByStatuses(state = {}, action) {
+    const key = () => action.statuses.join(',');
+    switch (action.type) {
+        case UPDATE_SUBMISSIONS_BY_STATUSES:
+            return {
+                ...state || null,
```
todd-dembrey: why is `|| null` necessary here?

opentech/static_src/src/app/src/redux/reducers/submissions.js
```diff
@@ -95,9 +99,47 @@ function currentSubmission(state = null, action) {
...
+                [key()]: {
```
todd-dembrey: I think i would rather key this by the actual status of the object and then we can pull multiple statuses out of the store with the selectors.

This would mean separating the error & loading logic out of here which I think is fine.

PR #936, opened 2019-01-30
Reset editor after submit

tm-kn:


opentech/static_src/src/app/src/containers/AddNoteForm.js
```diff
@@ -36,8 +36,7 @@ class AddNoteForm extends React.Component {
             message,
             visibility: 'internal',
         });
-
-        if (action === true) {
+        if (action.error === undefined) {
```
todd-dembrey: I think this is relying too heavily on the actual return values of the API. Should we be returning a promise from the API call here, which we are more likely to keep constant: https://github.com/OpenTechFund/opentech.fund/blob/581b6fc2f6ee625563beea51c5bad0c2a6ede86f/opentech/static_src/src/app/src/redux/middleware/api.js#L70-L79

That way we can then have a nicer interface here (`.then`)

PR #941, opened 2019-01-31
adds mock status block

chris-lawton:
- Adds mock data and styles for the status block

pasted_image_at_2019-01-31__4_45_pm

### DIFF
--- a/opentech/static_src/src/sass/apply/components/_status-block.scss
+++ b/opentech/static_src/src/sass/apply/components/_status-block.scss
@@ -1,0 +1,74 @@
+.status-block {
+    $root: &;
+    background-color: $color--white;
+    border: 1px solid $color--light-mid-grey;
+
+    @include media-query(tablet-landscape) {
+        display: flex;
+    }
+
+    &__item {
+        padding: 20px;
+        border-bottom: 1px solid $color--light-mid-grey;
+        background-color: $color--white;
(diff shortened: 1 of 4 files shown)


opentech/static_src/src/sass/apply/components/_status-block.scss
```diff
@@ -0,0 +1,74 @@
...
+    &__info {
+        font-size: 14px;
+        color: $color--dark-blue;
+    }
+
+    &__title {
+        font-weight: $weight--semibold;
+
+        @include media-query(tablet-landscape) {
+            height: 55px;
```
helenb: Guessing this ensures an even height for titles as if they were on 2 lines. Do you know for sure that all titles will wrap for no more than 2 lines? Could use min-height instead - might be more accessible if text size gets bumped up.

opentech/static_src/src/sass/apply/components/_status-block.scss
```diff
@@ -0,0 +1,74 @@
...
+
+    &__link {
+        text-transform: uppercase;
+        font-weight: $weight--semibold;
+        transition: opacity $quick-transition;
+
+        @include media-query(tablet-landscape) {
+            opacity: 0;
+            pointer-events: none;
+            height: 55px;
```
helenb: same comment as for title

PR #950, opened 2019-02-01
styles error, debug and info notification bars

chris-lawton:
- styles notifications 

screen shot 2019-02-01 at 12 40 49


opentech/templates/includes/message_item.html
```diff
@@ -1,8 +1,9 @@
 <li class="messages__text{% if tag %} messages__text--{{ tag }}{% endif %}{% if close %} js-message{% endif %}">
-    {{ message }}
-    {% if close %}
-    <a href="#" class="messages__close js-close-message">
```
helenb: add a title tag

opentech/templates/includes/message_item.html
```diff
@@ -1,8 +1,9 @@
...
-        <svg class="icon icon--close"><use xlink:href="#cross"></use></svg>
-    </a>
-    {% endif %}
+    <div class="messages__inner">
+        <svg class="messages__icon"><use xlink:href="#exclamation-point"></use></svg>
```
helenb: this svg can have aria-hidden="true" as it isn't necessary for understanding the text. According to the css tricks accessible svg articles it should also have a title tag, despite being aria-hidden.

PR #956, opened 2019-02-04
Bugfix/reduce bundle size

todd-dembrey:
I was a bit concerned about the bloat of the build size at the end of the sprint, so i've made a couple of tweaks to the process in order to bring it down.

Changes
* Remove moment.js - replace for native date object as timestamp is timezone aware from django
  * First attempt followed this process: https://github.com/jmblog/how-to-optimize-momentjs-with-webpack
  * Alternatives if we need more functionality https://github.com/you-dont-need/You-Dont-Need-Momentjs
* Remove Scss sourcemaps from production build

To note, CF is compressing the files at the edge giving ~9x compression which is why we never really noticed it. It is also why i haven't added in compression (gzip) as a plugin.

There is now the `gulp app:analyze` command which produces the following charts.

Bfefore: 2MB per build
image

After: 800KB per build
image

### DIFF
--- a/opentech/static_src/src/app/src/containers/Note.js
+++ b/opentech/static_src/src/app/src/containers/Note.js
@@ -17,7 +17,7 @@ class Note extends React.Component {
     render() {
         const { note } = this.props;
 
-        const date = new Date(note.timestamp).toLocaleDateString('en-gb', {day: 'numeric', month: 'short', year:'numeric'})
+        const date = new Date(note.timestamp).toLocaleDateString('en-gb', {day: 'numeric', month: 'short', year:'numeric', timezone:'GMT'})
 
         return <NoteListingItem
                 user={note.user}
--- a/opentech/static_src/src/app/src/containers/Note.js
+++ b/opentech/static_src/src/app/src/containers/Note.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
-import moment from 'moment';
 
 import { getNoteOfID } from '@selectors/notes';
 import NoteListingItem from '@components/NoteListingItem';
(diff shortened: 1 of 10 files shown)


opentech/static_src/src/app/webpack.base.config.js
```diff
@@ -1,83 +1,91 @@
...
+                    },
+                },
+                {
+                    test: /\.scss$/,
+                    use: [{
+                        loader: 'style-loader'
+                    }, {
+                        loader: 'css-loader',
+                        options: {
+                            sourceMap: !isProduction
```
tm-kn: I don't think webpack adds any source maps by default in the production mode. Did it?

opentech/static_src/src/app/src/containers/Note.js
```diff
@@ -18,10 +17,12 @@ class Note extends React.Component {
     render() {
         const { note } = this.props;
 
+        const date = new Date(note.timestamp).toLocaleDateString('en-gb', {day: 'numeric', month: 'short', year:'numeric'})
```
tm-kn: So we are going to have to do it for any date separately? We probably should create a reusable function. Also, shouldn't we specify `timeZone: 'GMT'`?

opentech/static_src/src/app/webpack.base.config.js
```diff
@@ -1,83 +1,91 @@
...
-                }, {
-                    loader: 'css-loader',
-                    options: {
-                        sourceMap: true
-                    }
-                }, {
-                    loader: 'sass-loader',
+                {
+                    test: /\.js$/,
+                    exclude: /node_modules/,
```
tm-kn: nice. ;)

opentech/static_src/src/app/webpack.base.config.js
todd-dembrey: it did because it is the css-loader not webpack adding the sourcemaps. Compare the *.scss sizes here:
image

image

opentech/static_src/src/app/src/containers/Note.js
todd-dembrey: So the datetime string coming from django is timezone aware. But i think you are correct the brower might localise it. Lets make sure

tm-kn: Yes, browser should use the time zone set in the OS if the date object is time zone-aware.

Issue #947, opened 2019-01-31
Improvements to standalone forms


danblah:
There are a number of issues with the current standalone forms implementation. I'm strongly recommending pulling it from live until we've got these. This is a priority issue that is required for us to cease using the old submission application's standalone forms.

**Required functionality for us to leave the old app:**
- [x] They have no styling. Have them take on the same styling as the fund/lab forms (considering this a bug)
- [ ] Add the ability to restrict form access to specified roles in the form editor. We need forms that are only available to the team, applicants, reviewers (and later for approved applicants from specific funds).
- [x] Require a recaptcha (or some other anti-spam) field for forms that are publically available. 
- [ ] File attachment field
- [x] A basic interface to review submissions (maybe this exists, can't tell cause currently erroring on test submissions)
- [ ] Field to define team lead
- [ ] Field to define slack channel to send notifications
- [ ] Notification to site-wide slack channel on all submissions
- [ ] Field to define webhook endpoint to send submissions to
- [x] Hide none wagtail forms from /admin/forms/
- [ ] Ability to preview completed form before submitting

frjo:
Related PR is https://github.com/OpenTechFund/opentech.fund/pull/948

frjo:
Since this is behind Cloudflare I do not think a captcha is needed. And if there are spam submissions we can do better than annoying captchas.

What is the "Team lead" field for. Just to direct notification?

The "webhook" field should end up in a separate issue I think, as an enhancement. I assume the system should not only ping the url put send some payload as well?

frjo:
@Parbhat Can you take on the following tasks?

1. Add a file upload field to the available fields when you build a form.
2. Make it possible to restrict access to a form by role/group.

frjo:
@danblah Wagtails forms on live is being blocked by Cloudflare cache, we have a page rule that tells Cloudflare to cache everything on www.opentech.fund. (By default Cloudflare only cache static assets like css, js and images, not HTML. So the cache bypass rules for test/dev/live is not really needed I think).

To get a page rule to bypass cache for pages with a cookie you would need to upgrade to a Cloudflare biz account. Overkill, and I think Cloudflare gets enough money already.

Could we decide that all wagtail forms get "-form" as part of their slug? I have added a page rule now that bypass the cache for "https://www.opentech.fund/*-form*".

The https://www.opentech.fund/about/jobs/test-form/ form seems to work now. I have also deployed the template fixes so the form looks like other forms on the site.

[...]

PR #963, opened 2019-02-04
Forms section only display wagtail forms not apply

Parbhat:
Link #947 

Hide none wagtail forms from `/admin/forms/`


frjo:
Works well for me so will deploy in a moment.

PR #976, opened 2019-02-06
Upload document field in Wagtail form page

Parbhat:
Link #947 

Initial implementation of Upload document field in Wagtail form page.

@frjo Uploaded document is currently saved as Wagtail document which can appear in the documents menu item of admin sidebar. Wagtail documents are served using document [serve](https://github.com/wagtail/wagtail/blob/master/wagtail/documents/views/serve.py) view. We can use [before_serve_document](https://docs.wagtail.io/en/v2.4/reference/hooks.html#document-serving) hook to serve for a particular group only. Are we using wagtail documents in other areas of the site? We can also use the implementation used in the submission (apply) site.

>@Parbhat and @frjo, want to be sure these are private files, only accessible to defined roles (or team only if that's easier right now). - [comment](https://github.com/OpenTechFund/opentech.fund/issues/947#issuecomment-459754218)


Parbhat:
@frjo Thanks for the review. Will update PR.

Parbhat:
@frjo PR updated to store uploaded document in the private bucket. We will probably need an interface to review form submissions. As the default form submissions interface (In Wagtail admin sidebar), only displays text data. Currently storing the file name with every submission, so we can retrieve it later.

Issue #957, opened 2019-02-04
Review updates: reviewer role and add reviewer roles


gibbsmaya:
The purpose of this is to fill the need for Reviewers fulfilling different roles within the review process within OTF. We will provide a means to “tag” reviewers with the type of review that they are required to complete. This will have no behavioural effect, you can have any number of roles and you will have access to the full range of review methods. It is up to the User to understand what the role means.

This issue is to add a new Wagtail taxonomy for reviewer role and ability to assign OTF staff to role. 

**Acceptance criteria**
- [x] 'Reviewer role' taxonomy added to Wagtail admin within the 'Apply' menu
    * Fields for Role: Name & icon  
- [x] "Add reviewer" form updated to display multiple lists for adding reviewers, one for each of the defined roles within the taxonomy.
- [x] The “Staff Reviewers” dropdown will be repeated for each Role as `“<Role> Reviewers”`
- [x] The "Reviewers" dropdown should remain unaffected (External reviewers only includes AC reviewers)
- [x] Users can only be selected for one Role. Assigning users to multiple roles will display validation error "Cannot be assigned to multiple roles"
- [x] Selection for each role should be single select rather than multiselect, i.e. you can only select one person per role
- [x] Only OTF staff can be selected for a Role.
- [x] Updates to messaging (activity feed, slack notification) to include the role of the assigned reviewer `"<Reviewer> added as <Role>"`

Design: https://projects.invisionapp.com/share/KZPIO1NENQJ#/screens/343687953

**Dev Notes**
Affected form and field:
https://github.com/OpenTechFund/opentech.fund/blob/e89491bfe9fe002cc79d9a5a4a6f09031c69c3ad/opentech/apply/funds/forms.py#L52-L57

**QA criteria**
Dev:

- [x] checked feature meets acceptance criteria/conforms exactly to the specification.
- [x] provided good unit test coverage (if this is non-trivial behaviour).
- [x] checked all tests for the project pass with this feature enabled.
- [x] checked code conforms to the project coding standards.
- [x] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end user of the website/app (Can I get to it? Is it useable? Can I break it? Does it work in an end-to-end context?)
- [x] checked that this feature works on the server/s I am deploying it to.

QA:

 * [ ] tested this feature as a front end user and it meets the acceptance criteria/conforms to the specification and design.
 * [ ] checked that the feature works on the server/s deployed to

[...]

PR #977, opened 2019-02-06
WIP Feature/957 Reviewer Roles

emullaney:
References #957  **Reviewer Role and Add Reviewer Role**
- 'Reviewer role' taxonomy added to Wagtail admin within the 'Apply' menu
-- Fields for Role: Name & icon
-- Also included `order` field which is strictly used for placement of the fields on the front-end
-  "Add reviewer" form updated to display multiple lists for adding reviewers, one for each of the defined roles within the texonomy.
-  The “Staff Reviewers” dropdown will be repeated for each Role as “<Role> Reviewers”
-  Users can only be selected for one Role. Assigning users to multiple roles will display validation error "Cannot be assigned to multiple roles"
-  Selection for each role should be single select rather than multiselect, i.e. you can only select one person per role
-  Only OTF staff can be selected for a Role.
-  Updates to messaging (activity feed, slack notification) to include the role of the assigned reviewer "<Reviewer> added as <Role>"

**TODO: Fix broken tests now that reviewer functionality has changed.**


opentech/apply/funds/views.py
```diff
@@ -269,20 +277,34 @@ class UpdateReviewersView(DelegatedViewMixin, UpdateView):
...
+                Q(reviewer_role=role)).delete()
 
         messenger(
             MESSAGES.REVIEWERS_UPDATED,
             request=self.request,
             user=self.request.user,
             submission=self.kwargs['submission'],
-            added=added,
-            removed=removed,
+            added_messages_list=added_messages_list,
```
todd-dembrey: I think i would still like to keep the removed part as well.

opentech/apply/funds/models/submissions.py
```diff
@@ -671,3 +671,19 @@ def get_absolute_url(self):
             'to': self.id,
             'from': previous_revision.id,
         })
+
+
+class ApplicationSubmissionReviewer(models.Model):
```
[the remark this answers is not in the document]
todd-dembrey: We should probably be considering the `unique_together` meta option for `('submission', 'role')`

opentech/apply/funds/views.py
```diff
@@ -269,20 +277,34 @@ class UpdateReviewersView(DelegatedViewMixin, UpdateView):
...
-        added = new_reviewers - old_reviewers
-        removed = old_reviewers - new_reviewers
+        if len(form.cleaned_data.values()) != len(set(form.cleaned_data.values())):  # If any of the users match
+            error_message = _('Users cannot be assigned to multiple roles.')
+            messages.error(self.request, mark_safe(error_message + form.errors.as_ul()))
+            form.add_error(None, error_message)
+            return self.form_invalid(form)
+        added_messages_list = []
+        # Loop through cleaned_data and save reviewers by role type to submission
+        for key, user in form.cleaned_data.items():
```
todd-dembrey: I think this is doing a lot of work and queries for what it is.

If we make the changes to allow the field to include the role object (or the cleaned data) then we do not need to do that again.

`update_or_create` is probably your friend here. That way we don't have to worry about the clean up at the end.

```
object, created = objects.update_or_create(role=role, submission=form.instance, defaults={'reviewer': user})
```

emullaney:
@todd-dembrey I made an abundance of changes per your review... you had so many great ideas.

I'm unable to work on this any longer today, so tests remain broken. But I can look at them in the morning.

opentech/apply/funds/models/reviewer_role.py
```diff
@@ -0,0 +1,20 @@
+from django.db import models
+
+
+class ReviewerRole(models.Model):
+    name = models.CharField(max_length=128)
+    icon = models.ForeignKey(
```
todd-dembrey: Can you add a panel configuration here so that we can get an image chooser panel for the imges

https://docs.wagtail.io/en/latest/reference/contrib/modeladmin/create_edit_delete_views.html#changing-which-fields-appear-in-createview-editview

emullaney: Yes, I have done this on the tweaks branch. I didn't realize you could use the wagtail panels on taxonomy items, but of course you can.

emullaney:
Closing this in favor of https://github.com/OpenTechFund/opentech.fund/pull/994

PR #994, opened 2019-02-11
Feature/957 tweaks

todd-dembrey:
@emullaney I've tidied up the code around the creation of the role reviewers and how the messaging is handled.

I think the biggest problem was that the various elements were trying to do things that they shouldn't have done.

Forms should be handling the saving of data
Views should be orchestrating the form and messaging
Widgets should be handling how they are rendered

This was a ll a bit muddled in the current implementation. I've split them out into their relevant parts.


opentech/apply/funds/templates/funds/includes/delegated_form_base.html
```diff
@@ -2,23 +2,7 @@
...
-            {% endif %}
-
-
-            {{ field.label_tag }} {{ field }}
-
-            {% if field.help_text %}
-                {{ field.help_text|safe }}
-            {% endif %}
-        {% endfor %}
+        {{ form }}
```
emullaney: Now that we have moved the icon out to the widget template, I don't think we need to load wagtailimages_tags.

opentech/apply/funds/forms.py
```diff
@@ -69,28 +67,42 @@ def __init__(self, *args, **kwargs):
...
+        assigned_roles = {
+            assigned.role: assigned.reviewer
+            for assigned in self.instance.assigned.filter(
+                role__isnull=False
+            )
+        }
+
         for role in ReviewerRole.objects.all().order_by('order'):
-            field_name = 'reviewer_' + str(role.pk)
+            field_name = 'reviewer_role_' + str(role)
```
emullaney: The reason I made this role.pk and not role.name is because the name could have spaces or any sort of other characters that don't work for form fields names. I think it's easiest to use pk, but I guess you could also do something fancy in the __str__ method on the model.

opentech/apply/funds/models/submissions.py
```diff
@@ -695,3 +697,19 @@ class AssignedReviewers(models.Model):
 
     class Meta:
         unique_together = ('submission', 'role')
+
+    def __str__(self):
+        return f'{self.reviewer} as {self.role}'
```
emullaney: This is a really great idea.

opentech/apply/funds/views.py
```diff
@@ -280,20 +279,23 @@ def get_form_kwargs(self):
         kwargs['request'] = self.request
         return kwargs
 
+    def form_invalid(self, form):
+        messages.error(
+            self.request,
+            mark_safe(
+                "There are errors with the Reviewers Form:" + form.errors.as_ul()
```
emullaney: I think you might want
`"There are errors with the Reviewers Form:" + form.errors.get('__all__').as_ul()`

Because the error in the message is adding in __all__, see this screenshot:
image

opentech/apply/activity/messaging.py
```diff
@@ -309,15 +328,17 @@ def batch_recipients(self, message_type, submissions, **kwargs):
...
+            message.append('Added:')
+            message.extend(reviewers_message(added))
 
         if removed:
-            message += 'Reviewers removed: ' + ', '.join([str(reviewer.reviewer) + ' as ' + str(reviewer.role) for reviewer in removed]) + '. '
+            message.append('Removed:')
+            message.extend(reviewers_message(removed))
 
-        return message
+        return '. '.join(message)
```
emullaney: If we look at `message` here in the breakpoint, it equals:
['Erin Super Django has updated the reviewers on <http://apply.localhost:8000/apply/submissions/3595/|Review Test 7>', 'Added:', 'Fiona Krakenbürger as Primary.', 'Tara Tarakiyee as Secondary.', 'Removed:', 'Adam Lynn as Primary.', 'Wei Fan as Secondary.']

This code is adding too many periods. We really only need one at the end of line 331 above, then we can join with a space.

opentech/apply/funds/views.py
todd-dembrey: @chris-lawton FYI whats the best way to style this?

we are returning a `<ul>` which sits along side the `<p>` and is causing an issue as the flex isnt assigned to it.

@emullaney If there are other errors it would be nice to render them nicely as well not just for the `__all__` case

opentech/apply/funds/forms.py
todd-dembrey: I can see your point. I was just trying to make it a bit more meaningful. 

I guess the options are [slugify](https://docs.djangoproject.com/en/2.1/ref/utils/#django.utils.text.slugify) it or use the id.

opentech/apply/activity/messaging.py
todd-dembrey: :+1: good spot.

opentech/apply/funds/views.py
chris-lawton: Can we wrap the `<p>` and the `<ul>` in a `<div>` to keep the flex alignment? So:
```
<div class="messages__inner">
        <svg class="messages__icon"><use xlink:href="#exclamation-point"></use></svg>
        <div>
            <ul>...</ul>    
            <p class="messages__copy">{{ message }}</p>
        </div>
        {% if close %}
            <a href="#" class="messages__button js-close-message">Ok</a>
        {% endif %}
    </div>
```

image

opentech/apply/funds/models/submissions.py
```diff
@@ -677,6 +677,11 @@ def get_absolute_url(self):
         })
 
 
+class AssignedReviewersQuerySet(models.QuerySet):
+    def role_reviewers_by_submission(self, submission):
```
todd-dembrey: This should probably be `with_roles` and then we can do `self.instance.assigned.with_roles().values('reviewers')` `assigned` will already be limited to the submission

opentech/apply/funds/forms.py
```diff
@@ -68,7 +68,10 @@ def __init__(self, *args, **kwargs):
         self.request = kwargs.pop('request', None)
         super().__init__(*args, **kwargs)
 
+        # All submitted reviews by non-role reviewers
         self.submitted_reviewers = User.objects.filter(id__in=self.instance.reviews.values('author'))
+        self.submitted_reviewers = self.submitted_reviewers.exclude(
+            id__in=AssignedReviewers.objects.role_reviewers_by_submission(self.instance).values('reviewer'))
```
todd-dembrey: There are two cases:
1. I leave a review and i don't want to be added again
2. I get removed from a role, but i've already submitted a review - i dont want to be removed.

Not sure if the above handles 2?

opentech/apply/funds/forms.py
```diff
@@ -165,13 +168,31 @@ def save(self, *args, **kwargs):
...
         )
 
+        # Update or create role reviewers
         for role, reviewer in assigned_roles.items():
             if reviewer:
-                AssignedReviewers.objects.update_or_create(
+                # Edge case: handle any role reviewers that reviewed and now have been removed from role assignment,
+                # We want to have a AssignedReviewer record for that orphaned reviewer
+                # because they have left a review but no longer have a role
+                existing = AssignedReviewers.objects.filter(
```
todd-dembrey: I think this is now too complex for what it needs to do, I think by excluding the roles from `submitted_users` before we create the new objects we are causing a headache. 

What we need to do is:
* Update role reviewers 
* Update non-reviewers 
   * remove people not on form
   * add people from form
* Add in anyone who has already reviewed

So the logic is:

* Users who need non role assignment `non_role` = set([submitted reviewers, assigned staff - no role, external reviewers])
* Update new assigned reviewers (as original) 
* Delete users without a role if they don't appear in `non_role` 
* Create new users without roles from `non_roles` if they do not appear in `assigned(role=None)`

The logic to handle adding users once they have reviewed is part of the post-save method in the `reviews/models.py`

opentech/apply/funds/tests/test_views.py
```diff
@@ -278,6 +280,34 @@ def test_staff_cant_remove_reviewers_proposal(self):
 
         self.assertCountEqual(submission.reviewers.all(), self.reviewers)
 
+    def test_lead_can_change_role_reviewer_and_review_remains(self):
+        submission = ApplicationSubmissionFactory(lead=self.user)
+
+        self.post_form(submission, reviewer_roles=[self.staff[0]])
+        self.assertCountEqual(submission.reviewers.all(), [self.staff[0]])
```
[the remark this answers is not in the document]
emullaney: ok, that makes sense

opentech/apply/funds/tests/test_views.py
```diff
@@ -278,6 +280,34 @@ def test_staff_cant_remove_reviewers_proposal(self):
...
+        self.post_form(submission, reviewer_roles=[self.staff[1]])
+
+        # Make sure that the ex-role-reviewer (self.staff[0]) still has a assignedreviewer record
+        self.assertEqual(submission.reviewers.all().count(), 2)
+        self.assertIn(self.staff[0], submission.reviewers.all())
+
+    def test_lead_adds_two_roles_one_reviewer_and_a_new_review_comes_in(self):
+        submission = ApplicationSubmissionFactory(lead=self.user, status='external_review', workflow_stages=2)
+
+        self.post_form(submission, reviewer_roles=[self.staff[0], self.staff[1]], reviewers=[self.reviewers[0]])
```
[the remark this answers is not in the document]
emullaney: Sure

opentech/apply/funds/tests/test_views.py
```diff
@@ -278,6 +280,34 @@ def test_staff_cant_remove_reviewers_proposal(self):
...
+        self.assertEqual(submission.reviewers.all().count(), 3)
+
+        # Add a review from a new reviewer who isn't assigned
+        ReviewFactory(submission=submission, author=self.reviewers[1])
+
+        # Now there should be 4 reviewers assigned, because an outsider reviewed
+        self.assertEqual(submission.reviewers.all().count(), 4)
+
```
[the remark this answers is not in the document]
emullaney: Are you asking that I simplify this test? Create a new test? I'm not sure what you are asking here.

opentech/apply/funds/forms.py
```diff
@@ -165,13 +168,31 @@ def save(self, *args, **kwargs):
             if reviewer not in instance.reviewers.filter(assignedreviewers__role=None)
```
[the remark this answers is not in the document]
emullaney: agree

opentech/apply/funds/tests/test_views.py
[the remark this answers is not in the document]
todd-dembrey: :man_facepalming: I missed a word. "I think the  test_role_reviewer_exists_after_submitting_review is *missing*"

emullaney:
@todd-dembrey please have another look.

PR #998, opened 2019-02-11
adds updated message style to public site

chris-lawton:
- adds updated message styling to the public site


helenb:
I'm fairly sure I saw these changes before on another pull request - but anyway, they all look good to me.

PR #1008, opened 2019-02-12
Fix/1005 mobile submit note button

chris-lawton:


todd-dembrey:
@frjo We're happy for this to go to master as well.

PR #1015, opened 2019-02-13
Feature/988 dashboard polish

chris-lawton:
- Allows for filter and search partial to accept a heading and not render batch actions markup
- Moves cell spacing from selected checkbox to title so that spacing is consistent for tables that have no checkbox

### DIFF
--- a/opentech/apply/funds/templates/funds/includes/table_filter_and_search.html
+++ b/opentech/apply/funds/templates/funds/includes/table_filter_and_search.html
@@ -2,12 +2,20 @@
     <div class="actions-bar">
         {# Left #}
         <div class="actions-bar__inner actions-bar__inner--left">
-            <p class="actions-bar__total"><span class="js-total-actions">0</span> Selected</p>
-            <form action="" class="js-batch-update-status">
-                <button class="button button--action button--change-status" type="submit">Change status</button>
-            </form>
+            {% if heading %}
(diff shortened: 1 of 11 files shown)


opentech/apply/funds/templates/funds/includes/table_filter_and_search.html
```diff
@@ -2,12 +2,20 @@
...
 
-            <button data-fancybox data-src="#batch-update-reviewers" class="button button--action button--reviewers js-batch-update-reviewers" type="submit">Reviewers</button>
+            {% if use_batch_actions %}
+                <div class="actions-bar__inner actions-bar__inner--batch-actions">
+                    <p class="actions-bar__total"><span class="js-total-actions">0</span> Selected</p>
+                    <form action="" class="js-batch-update-status">
+                        <button class="button button--action button--change-status" type="submit">Change status</button>
+                    </form>
+
+                    <button data-fancybox data-src="#batch-update-reviewers" class="button button--action button--reviewers js-batch-update-reviewers" type="submit">Reviewers</button>
```
helenb: should this button not also be inside the form element? what is it submitting?

chris-lawton: it just pops open a modal, where the actual `<form>` is, however with `type="submit"` I can see why that would be confusing. I'll update the type for clarity.

Issue #867, opened 2019-01-18
Status overview block: frontend


gibbsmaya:
Designed as the entry to Submissions by Status and provides an overview of the status for all submissions being processed. The statuses will be grouped into similar status to make the visualisation useful across multiple workflow types. 

**Acceptance criteria**

- [x] Status overview block styled as per design - not we have changed the design from the one below as we felt a "timeline" style was not relevant for these statuses since they are not always sequential - it is just really a list of statuses
- [x] 'View' link appears on hover to link through to submission by status page
- [x] On mobile, display statuses + #of applications as list

Design: https://projects.invisionapp.com/share/KZPIO1NENQJ#/screens/335805634
Specifically
![screen shot 2019-01-18 at 15 54 35](https://user-images.githubusercontent.com/19690895/51397586-6ef2c300-1b39-11e9-9cd5-e0c02eca689a.png)

**QA criteria**
Dev:

- [x] checked that this feature works across all agreed supported browsers and devices.
- [x] checked that this code meets the agreed accessibility level.
- [x] checked that this code meets the agreed browser performance levels.
- [x] checked that this code conforms to the project coding standards.
- [x] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end user.
- [x] checked that the feature works on the server/s I am deploying it to.

QA:

 * [x] tested this feature as a front end user and it meets the acceptance criteria/conforms to the specification and design.
 * [x] checked that the feature works on the server/s deployed to

chris-lawton:
@helenb - can you review this one please? - https://github.com/OpenTechFund/opentech.fund/pull/941

(It requires BE work hence the mock data)

helenb:
@chris-lawton - have reviewed so back with you.

chris-lawton:
@gibbsmaya - screens of the hardcoded status block:

![screen shot 2019-02-01 at 10 16 16](https://user-images.githubusercontent.com/31627284/52116976-79698e00-260a-11e9-9ee9-45c58e0ae3b0.png)
![screen shot 2019-02-01 at 10 16 26](https://user-images.githubusercontent.com/31627284/52116978-79698e00-260a-11e9-866c-ef6490a2e2bb.png)

gibbsmaya:
Nice one. Can you just confirm that the view link appears on hover for each status?

chris-lawton:
Apologies, thought it was in the screen!

[...]

Issue #866, opened 2019-01-18
Status overview block: backend


gibbsmaya:
Designed as the entry to Submissions by Status and provides an overview of the status for all submissions being processed. The statuses will be grouped into similar status to make the visualisation useful across multiple workflow types. 

**Acceptance criteria**

- [x] Status overview block added to Submission overview page
- [x] Statuses should be as follows:
  - [x] Received (all applications in Screening or Proposal received status)
  - [x] Internal Review (all applications in Internal review status)
  - [x] Ready for Discussion (all applications in Ready for discussion status) 
  - [x] More information requested (all applications in More information requested status)
  - [x] Invited for proposal (all applications in Invited for proposal status)
  - [x] Advisory Council Review (all applications in Advisory council review status)
  - [x] Ready for Determination (all applications in Ready for determination, Ready for preliminary determination and Ready for final determination statuses)
  - [x] Accepted (all applications in Accepted status)
  - [x] Dismissed (all applications in Dismissed status)
- [x] Each status displays a count of submissions for the status, counting only “active” rounds and labs 
- [x] Each status inks to the corresponding Submission by Status page.

Design: https://projects.invisionapp.com/share/KZPIO1NENQJ#/screens/335805634
Specifically
![screen shot 2019-01-18 at 15 54 35](https://user-images.githubusercontent.com/19690895/51397586-6ef2c300-1b39-11e9-9cd5-e0c02eca689a.png)

**QA criteria**
Dev:

- [x] checked feature meets acceptance criteria/conforms exactly to the specification.
- [x] provided good unit test coverage (if this is non-trivial behaviour).
- [x] checked all tests for the project pass with this feature enabled.
- [x] checked code conforms to the project coding standards.
- [x] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end user of the website/app (Can I get to it? Is it useable? Can I break it? Does it work in an end-to-end context?)
- [x] checked that this feature works on the server/s I am deploying it to.

QA:

 * [x] tested this feature as a front end user and it meets the acceptance criteria/conforms to the specification and design.
 * [x] checked that the feature works on the server/s deployed to

todd-dembrey:
@gibbsmaya This is now deployed on dev and with Erin for code review. Chris will need to tweak some styling.

Small variation on the AC - We split determined into accepted and dismissed last sprint for the by status view, have continued that here. 
[...]

PR #1017, opened 2019-02-14
GH-866: wire up the status block to the data

todd-dembrey:
@chris-lawton as discussed, small issue with heights of text. Do you want to tweak the styles here?
image

@emullaney FYI the boxy style above is correct and was approved in #867 it wont match what is shown in #866


chris-lawton:
@todd-dembrey - fix for the height issue - https://github.com/OpenTechFund/opentech.fund/pull/1019

todd-dembrey:
@emullaney I did wonder that, I didn't think there was any harm as we got to a table which has the "There is nothing here" message anyway.

PR #1019, opened 2019-02-14
cater for more than two lines in the status block

chris-lawton:


todd-dembrey:
Would it be possible to improve the alignment of the counts between boxes?

image

chris-lawton:
Have actioned: 

image

PR #1020, opened 2019-02-14
Submission by round status detail views review information

tm-kn:


opentech/apply/funds/serializers.py
```diff
@@ -19,10 +20,23 @@ def to_representation(self, instance):
...
+        reviews = instance.reviews.all()
         return {
-            'count': instance.reviews.count(),
-            'score': instance.reviews.score(),
-            'recommendation': instance.reviews.recommendation(),
+            'count': len(reviews),
+            'score': reviews.score(),
+            'recommendation': reviews.recommendation(),
+            'reviews': [
+                OrderedDict([
```
todd-dembrey: a) dict is ordered by default now as of py3.6
b) Why do we need ordering?

opentech/apply/funds/serializers.py
```diff
@@ -19,10 +20,23 @@ def to_representation(self, instance):
 
 class ReviewSummarySerializer(serializers.Field):
     def to_representation(self, instance):
+        reviews = instance.reviews.all()
```
todd-dembrey: do we want a `select_related` on the author?

opentech/static_src/src/app/src/containers/ReviewInformation.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+
+import ReviewsBlock from '@components/ReviewsBlock'
+import { getSubmissionOfID } from '@selectors/submissions'
+
+const ReviewInformation = ({ submission }) => {
+    return <ReviewsBlock submission={submission} />
```
todd-dembrey: Should be passing in the review objects here i think. The `ReviewBlock` doesn't really care about the submission, just the reviews.

opentech/static_src/src/app/src/components/ReviewsBlock/index.js
```diff
@@ -0,0 +1,51 @@
...
+        <p>
+            <a href={reviewUrl}>
+                {author} - {score} - {recommendation.display}
+            </a>
+        </p>
+    )
+}
+
+Review.propTypes = {
+    review: PropTypes.shape({
```
todd-dembrey: Can we not just unpack the review data into the prop directly without the intermediary review level?

opentech/static_src/src/app/src/components/ReviewsBlock/index.js
```diff
@@ -0,0 +1,51 @@
...
+        author: PropTypes.string.isRequired,
+        score: PropTypes.number.isRequired,
+        recommendation: PropTypes.shape({
+            display: PropTypes.string.isRequired,
+        }).isRequired,
+        reviewUrl: PropTypes.string.isRequired,
+    }),
+}
+
+const ReviewsBlock = ({ submission }) => {
```
todd-dembrey: as per comment on the container - should accept "reviews" and handle them

opentech/apply/funds/serializers.py
tm-kn: > a) dict is ordered by default now as of py3.6
It is a cpython implementation detail rather than the language feature as far as I know.

>  b) Why do we need ordering?
Not really, just seems like a practice of using ordered dictionaries in the APIs.

opentech/static_src/src/app/src/components/EmptyPanel/index.js
```diff
@@ -27,6 +27,10 @@ export default class EmptyPanel extends React.Component {
                 {column === 'applications' &&
                     <p>No results found.</p>
                 }
+
+                {column === 'reviews' &&
+                    <p>No reviews found</p>
```
todd-dembrey: This looks wrong... We should be using a new component for the `ReviewEmptyPanel` not passing a value in. Appreciate you are just copying what was done before. Look at it if you have time, but its ok for now,.

PR #1022, opened 2019-02-14
Feature/991 direct link to submission

chris-lawton:


todd-dembrey:
@chris-lawton I've tweaked the flow to get into a correct state when the submission we have set is not in the current active round.

opentech/static_src/src/app/src/redux/actions/submissions.js
```diff
@@ -85,7 +108,13 @@ export const loadCurrentRoundSubmissions = () => (dispatch, getState) => {
         return null
     }
 
-    return dispatch(fetchSubmissionsByRound(getCurrentRoundID(state)))
+    return dispatch(fetchSubmissionsByRound(getCurrentRoundID(state))).then(() => {
+        const state = getState()
+        const ids = getCurrentRoundSubmissionIDs(state)
+        if (!ids.includes(getCurrentSubmissionID(state))) {
+            return dispatch(setCurrentSubmission(null))
```
tm-kn: I don't understand rationale for changing the current submission when fetching. Is that when user sets non-existing ID? Can we add a comment?

opentech/static_src/src/app/src/redux/actions/submissions.js
```diff
@@ -49,10 +50,31 @@ export const setCurrentSubmissionRound = id => ({
...
+        return dispatch(setCurrentSubmission(activeId));
+    }
+    return null;
+}
+
+
+export const setCurrentSubmission = id => (dispatch, getState) => {
+    const submissionID = getCurrentSubmissionID(getState())
+
+    if (id && submissionID !== id) {
```
tm-kn: `id` could be `0` I guess and would fail this test. Should we check for `undefined` or `null` instead?

opentech/static_src/src/app/src/redux/actions/submissions.js
todd-dembrey: This is the case where you try and look up a submission that isnt in the list anymore. i.e. its has changed status. 

I was considering adding a message at this point "The submission you are trying to view is now available in this status/round anymore"

opentech/static_src/src/app/src/redux/actions/submissions.js
todd-dembrey: You are correct.

PR #1023, opened 2019-02-18
WIP: Feature/978 actions fe

chris-lawton:
- Updates modal styling
- Adds disabled batch actions styles
- Adds list reveal block to determination form
- Adds modal to determination form

Remaining BE work: 
- Hook up disabled batch actions button
- Populate list reveal block on determination form
- Populate modal on determination form
- Hook up modal to submit determination form

PR #1027, opened 2019-02-19
Feature/858 batch status

todd-dembrey:
@emullaney Need to add a few tests, but would appreciate a review

@chris-lawton Could you look at the js code as it will impact you on #1023


opentech/apply/utils/views.py
```diff
@@ -83,6 +83,9 @@ def post(self, request, *args, **kwargs):
 
 
 class DelegateableListView(DelegatableBase):
+    def get_form_args(self):
+        return None, self.request.user
+
```
emullaney: Nice.

opentech/apply/utils/views.py
```diff
@@ -71,8 +72,11 @@ def post(self, request, *args, **kwargs):
 
 
 class DelegateableView(DelegatableBase):
-    def get_form_args(self):
-        return self.object, self.request.user
+    def get_form_kwargs(self):
+        return {
+            'user': self.request.user,
+            'instance': self.object,
+        }
```
emullaney: Nice

PR #1029, opened 2019-02-19
Add missing notifications

tm-kn:


tm-kn:
@todd-dembrey I am having issues distinguishing status updates made by other users and inside the React component. Could you help me come up with an idea to sort that out please?

PR #1031, opened 2019-02-20
Fixup the behaviour with status

todd-dembrey:
@tm-kn this makes the statuses be more like the round - we set a current set of statuses and then refer to that elsewhere.

It removes the passing around of the statuses within the components.

I've also split out the selectors into multiple files to be more explicit and reduce confusion.

### DIFF
--- a/opentech/static_src/src/app/src/containers/GroupByRoundDetailView.js
+++ b/opentech/static_src/src/app/src/containers/GroupByRoundDetailView.js
@@ -9,15 +9,17 @@ import {
     getRoundsErrored,
 } from '@selectors/rounds';
 import {
-    getByGivenStatusesLoading,
-    getByGivenStatusesError,
-    getCurrentRoundSubmissions,
+    getCurrentStatusesSubmissions,
     getCurrentSubmissionID,
 } from '@selectors/submissions';
+import {
+    getByStatusesLoading,
+    getByStatusesError,
+} from '@selectors/statuses';
+
 
 class GroupByRoundDetailView extends React.Component {
     static propTypes = {
-        submissionStatuses: PropTypes.arrayOf(PropTypes.string),
         submissions: PropTypes.arrayOf(PropTypes.object),
         submissionID: PropTypes.number,
         isLoading: PropTypes.bool,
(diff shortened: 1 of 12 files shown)


opentech/static_src/src/app/src/containers/GroupByRoundDetailView.js
```diff
@@ -45,13 +46,11 @@ class GroupByRoundDetailView extends React.Component {
 }
 
 const mapStateToProps = (state, ownProps) => ({
-    isErrored: getRoundsErrored(state),
-    errorMessage: getByGivenStatusesError(ownProps.submissionStatuses)(state),
```
todd-dembrey: @chris-lawton This was returning a boolean, so moved to the `isErrored` prop.

opentech/static_src/src/app/src/containers/GroupByRoundDetailView.js
```diff
@@ -45,13 +46,11 @@ class GroupByRoundDetailView extends React.Component {
...
+    isErrored: getRoundsErrored(state) || getByStatusesError(state),
     isLoading: (
-        getByGivenStatusesLoading(ownProps.submissionStatuses)(state) ||
-        getRoundsFetching(state)
+        getByStatusesLoading(state) || getRoundsFetching(state)
     ),
-    submissions: getCurrentRoundSubmissions(state),
+    submissions: getCurrentStatusesSubmissions(state),
```
todd-dembrey: @tm-kn This is the only functional change. I think this change is correct... It seemed wrong that we were fetching round submissions when we are interested in the status submissions. (I guess because they are a similar thing here it was working before..)

opentech/static_src/src/app/src/containers/GroupByRoundDetailView.js
tm-kn: https://github.com/OpenTechFund/opentech.fund/commit/e7ef88d7cddbaf63ccf926801d9200dabb705b2a#diff-97f00faee6202c788f86bcc7b8eefe56R49

You will need to rebase. I've added a quick hack.

Issue #979, opened 2019-02-07
Review updates: front end


gibbsmaya:
**Acceptance criteria**

- [x] Review block on submission detail page styled as per design:
  - [x] if reviewer assigned but no reviews or agreement/disagreement yet, display as per design (include icon) but greyed out as per current styles 
![image](https://user-images.githubusercontent.com/6338661/52417859-b212d700-2ae4-11e9-94c7-88f4952000a1.png)
  - [x] agreement/disagreement shown as 'Agree' or 'Disagree' associated with the parent review (coloured in green for agree, red for disagree)
  - [x] Traffic light to remain as per current design.
- [x] Expanded submission on Submissions listing: agreement/disagreement shown in Review outcomes section as 'Agree' or 'Disagree' associated with the parent review. Styles based on review block on submission detail page
- [x] Review detail view (e.g. https://dev-apply.opentech.fund/apply/submissions/11/reviews/4/): will show agreement/disagreement from anyone who has agreed or disagreed with the review in the header
- [x] View all reviews view (e.g. https://dev-apply.opentech.fund/apply/submissions/6/reviews/) will show agreement/disagreement from anyone who has agreed or disagreed with the review in a new row above score in the table
- [x] Error notifications styled as per standard notification styles. Try adding the same person as a primary and secondary review and see error notification that appears.

Design: 
![image](https://user-images.githubusercontent.com/6338661/52417713-65c79700-2ae4-11e9-808c-7041001d77ed.png)

**QA criteria**
Dev:

- [x] checked that this feature works across all agreed supported browsers and devices.
- [x] checked that this code meets the agreed accessibility level.
- [x] checked that this code meets the agreed browser performance levels.
- [x] checked that this code conforms to the project coding standards.
- [x] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end user.
- [x] checked that the feature works on the server/s I am deploying it to.

QA:

 * [x] tested this feature as a front end user and it meets the acceptance criteria/conforms to the specification and design.
 * [x] checked that the feature works on the server/s deployed to

chris-lawton:
Blocked until BE work is complete and can be put on dev to test

gibbsmaya:
One FE tweak required which I noticed as I tested #959:

Reviewer block on submission page: don't show the line underneath the reviewers if there are no AC reviewers (e.g. https://dev-apply.opentech.fund/apply/submissions/5/)

gibbsmaya:
[...]

Issue #962, opened 2019-02-04
Review updates: reviewer consensus: display


gibbsmaya:
This issue covers the display of agreeing/disagreeing with a review.

**Acceptance criteria**
 - [x] Review block on submission detail page: agreement/disagreement shown as 'Agree' or 'Disagree' associated with the parent review (as per screenshot in design section below). 
- [x] Expanded submission on Submissions listing: agreement/disagreement shown in Review outcomes section as 'Agree' or 'Disagree' associated with the parent review. (Image shows location only, no need to match design)
![image](https://user-images.githubusercontent.com/6338661/52418115-47ae6680-2ae5-11e9-8e2c-68dd9d0d7a3b.png)
- [x] Review detail view (e.g. https://dev-apply.opentech.fund/apply/submissions/11/reviews/4/): will show agreement/disagreement from anyone who has agreed or disagreed with the review. A list of:  `<user> <role icon> <agree|disagree>s` will appear in the header to the right of the title.
- [x] View all reviews view (e.g. https://dev-apply.opentech.fund/apply/submissions/6/reviews/) will show agreement/disagreement from anyone who has agreed or disagreed with the review. This will be an extra row above the score with a list as per Review Detail page.

Design:
![image](https://user-images.githubusercontent.com/6338661/52417713-65c79700-2ae4-11e9-808c-7041001d77ed.png)

**Dev Notes**
Table for reviews is constructed here:
https://github.com/OpenTechFund/opentech.fund/blob/7b49da51dcc08407abf3d9a9a94b29fadb849a2b/opentech/apply/review/views.py#L136

**QA criteria**
Dev:

- [x] checked feature meets acceptance criteria/conforms exactly to the specification.
- [x] provided good unit test coverage (if this is non-trivial behaviour).
- [x] checked all tests for the project pass with this feature enabled.
- [x] checked code conforms to the project coding standards.
- [x] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end user of the website/app (Can I get to it? Is it useable? Can I break it? Does it work in an end-to-end context?)
- [x] checked that this feature works on the server/s I am deploying it to.

QA:

 * [x] tested this feature as a front end user and it meets the acceptance criteria/conforms to the specification and design.
 * [x] checked that the feature works on the server/s deployed to

todd-dembrey:
@gibbsmaya We have varied the acceptance criteria here. Your opinion will remain visible regardless of having submitting a review.

[...]

PR #1034, opened 2019-02-20
Feature/962 Display Agree/Disagree user opinion

emullaney:
References #962 Reviewer consensus (agree/disagree) display
 - Review block on submission detail page: agreement/disagreement shown as 'Agree' or 'Disagree' associated with the parent review 
-- Display all agree/disagree opinions
-- Assigned reviewers who have not submitted a review but HAVE submitted an *opinion* won't appear greyed out in the list (we aren't waiting for their review because they have submitted an opinion)

- Expanded submission on Submissions listing: agreement/disagreement shown in Review outcomes section as 'Agree' or 'Disagree' associated with the parent review. (Image shows location only, no need to match design)
image

- Review detail view shows agreement/disagreement from anyone who has agreed or disagreed with the review. A list of:  `<user> <role icon> <agree|disagree>s` will appear in the header to the right of the title.
-- The code displays agree/disagree data below the title, but this can be adjusted on the FE ticket #979 
-- Created a new include template to get this data `review_opinions_list.html` and a new model method on Review Opinion, `get_author_role`

- View all reviews view shows agreement/disagreement from anyone who has agreed or disagreed with the review. This will be an extra row above the score with a list as per Review Detail page.
-- I reused the include template `review_opinions_list.html` in order to display this data on the all reviews table below each score. I'm curious if you're okay with this approach @todd-dembrey 

- Tests


opentech/apply/review/tests/test_views.py
```diff
@@ -206,3 +207,65 @@ def test_review_detail_recommendation(self):
...
+
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+        cls.submission = ApplicationSubmissionFactory(status='draft_proposal', workflow_stages=2)
+
+    def get_kwargs(self, instance):
+        return {'pk': instance.id, 'submission_pk': instance.submission.id}
+
+    def test_can_see_opinion_buttons_on_others_review(self):
```
todd-dembrey: Test that others cant see the buttons and test they cant submit.

opentech/apply/review/views.py
```diff
@@ -30,32 +34,48 @@ def get_context_data(self, **kwargs):
         for review in reviews:
             reviews_dict[review.author.pk] = review
 
+        # Get all the authors of opinions, these authors should not show up in the 'xxx_not_reviewed' lists
+        opinion_authors = set(User.objects.filter(
```
todd-dembrey: `distinct()` rather than set

opentech/apply/review/views.py
```diff
@@ -30,32 +34,48 @@ def get_context_data(self, **kwargs):
...
+            pk__in=ReviewOpinion.objects.filter(review__submission=self.object).values('author__pk')))
```
todd-dembrey: `pk__in=self.object.reviews.opinions.values('author')` maybe?

Would involve an extra method on the review queryset.

opentech/apply/review/views.py
```diff
@@ -30,32 +34,48 @@ def get_context_data(self, **kwargs):
...
                 if review:
                     key = 'staff_reviewed'
-                else:
+                elif review not in opinion_authors:
                     key = 'staff_not_reviewed'
             else:
                 if review:
                     key = 'external_reviewed'
-                else:
+                elif review not in opinion_authors:
```
todd-dembrey: This shouldn't happen. only staff can leave reviews.

opentech/apply/review/models.py
```diff
@@ -168,3 +168,19 @@ def update_submission_reviewers_list(sender, **kwargs):
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.PROTECT,
+    )
+    opinion = models.IntegerField(choices=OPINION_CHOICES)
+
+    class Meta:
+        unique_together = ('author', 'review')
+
+    def get_author_role(self):
+        role = self.review.submission.assigned.with_roles().filter(reviewer=self.author).first()
```
todd-dembrey: this is `assignment` not `role`.

opentech/apply/review/views.py
```diff
@@ -30,32 +34,48 @@ def get_context_data(self, **kwargs):
...
+                review_info_dict = {
+                    'reviewer': reviewer,
+                    'review': review,
+                    'role': role,
+                }
+                opinions_list = []
+                if review:
+                    # Loop through all opinions and include the current author's role for this submission
+                    for opinion in review.opinions.all():
+                        opinions_list.append({
```
todd-dembrey: i dont think you need to prefetch this in the view. you have access to all of this information off the review object within the templates.

You can even `selected_related("author")` to reduce queries.

opentech/apply/review/views.py
```diff
@@ -181,6 +270,12 @@ def get_context_data(self, **kwargs):
 
         for i, review in enumerate(self.object_list):
             review_data['title']['answers'].append('<a href="{}">{}</a>'.format(review.get_absolute_url(), review.author))
+            if review.opinions:
```
todd-dembrey: do you need the if statement? Perhaps select related the authors on opinions again.

if there are no opinions could the template return nothing/minimal markup? 

You can use an `empty` template tag to handle this nicely.

opentech/apply/review/views.py
emullaney: good idea

opentech/apply/review/views.py
emullaney: true, I'll switch this to an `else`

opentech/apply/review/models.py
emullaney: I think you're just talking about renaming this method? I'll do that.

opentech/apply/review/views.py
emullaney: you are correct, excellent point

Issue #781, opened 2018-12-27
Approved Proposals Should be Editable [High/Bug] 


danblah:
-=Sensitive=-
No

-=Reporter=-
Tara

-=Ideal deadline=-
12/31/2018

-=Who's affected=-
OTF team, Applicants

-=Description=-
Proposals in "Approved" state can not be edited by staff or applicants. We need to edit them for post-approval fixes. 

-=Objective=-

-=Additiional info=-

-=Timestamp=-
2018-12-27T16:14:08.495Z

frjo:
@danblah Is this only for proposals or also for the other applications? I guess yes.

Do staff need to edit concept notes that have been invited for proposal? I guess not.

I have made an PR according to my guesses https://github.com/OpenTechFund/opentech.fund/pull/975.

PR #1042, opened 2019-02-21
Allow applicants to edit approved submissions.

frjo:
Followup fix for #781

### DIFF
--- a/opentech/apply/funds/views.py
+++ b/opentech/apply/funds/views.py
@@ -494,14 +494,17 @@ class ApplicantSubmissionEditView(BaseSubmissionEditView):
             )
 
         action = set(self.request.POST.keys()) & set(self.transitions.keys())
-        transition = self.transitions[action.pop()]
-
-        self.object.perform_transition(
-            transition.target,
-            self.request.user,
-            request=self.request,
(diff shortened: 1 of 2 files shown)


opentech/apply/funds/views.py
```diff
@@ -493,15 +493,18 @@ def form_valid(self, form):
...
-        action = set(self.request.POST.keys()) & set(self.transitions.keys())
-        transition = self.transitions[action.pop()]
-
-        self.object.perform_transition(
-            transition.target,
-            self.request.user,
-            request=self.request,
-            notify=not (revision or submitting_proposal),  # Use the other notification
-        )
+        try:
```
frjo: Added this to avoid an error when applicants saved submission in approved state. There are no next state to move to then.

opentech/apply/funds/views.py
```diff
@@ -493,15 +493,18 @@ def form_valid(self, form):
...
+            action = set(self.request.POST.keys()) & set(self.transitions.keys())
```
todd-dembrey: Keep as little as possible inside the try except. this first line should be outside.

PR #1044, opened 2019-02-21
CI test failure fix (testing to see if this fixes the issue)

emullaney:
I was thinking that maybe faker was creating words for role names that weren't playing well with the form field names, so I added slugify to see if it fixed the test that are sometimes failing on these tests. These tests all use the `reviewer_roles` field when they post the form, so I'm thinking that might have been the issue. No other tests use that field, and these are the 3 tests that keep occasionally failing.


todd-dembrey:
This was merged in another ticket - closing

PR #1058, opened 2019-02-22
Feature/958 batch add reviewer with roles

emullaney:
References https://github.com/opentechfund/opentech.fund/issues/958
- Batch add reviewer action updated to display multiple lists for adding reviewers, one for each of the defined roles (as per individual actions).

- Batch reviewer activity text changed to be `"<reviewer> assigned as <role>"`  per role.  
  * Slack
  * Activity feed


emullaney:
FYI @todd-dembrey I think I might have missed the form validation check for reviewers matching (you can't assign the same person to different reviewer roles). I just realized that this morning and wanted to drop a note here as a reminder.

PR #1094, opened 2019-03-07
Fix Validation error when creating copy of round page

### LINKED ISSUE #1018: Validation error for date when copy round
When user copy a round the date fields should be cleared to avoid a validation error. This can be done using the before_copy_hook in wagtail (tips from Todd).

https://sentry.io/otf/otf-webapp/issues/887293023/

https://docs.wagtail.io/en/v2.4/reference/hooks.html#before-copy-page

Parbhat:
Closes #1018


opentech/apply/funds/models/applications.py
```diff
@@ -144,7 +147,7 @@ class RoundBase(WorkflowStreamForm, SubmittableStreamForm):  # type: ignore
         limit_choices_to=LIMIT_TO_REVIEWERS,
         blank=True,
     )
-    start_date = models.DateField(default=date.today)
+    start_date = models.DateField(null=True, blank=True, default=date.today)
```
Parbhat: We have to make start_date `null=True` and `blank=True` so we can create a copy of round pages programmatically. Validation is shifted to [RoundBasePageAdminForm](https://github.com/OpenTechFund/opentech.fund/pull/1094/files#diff-a02bd3da8a1ce030a48a03e118ffc50bR46). The create copy will have None value for start_date and end_date. An info message will be shown to admin user after creating a copy.

```
Please select the date in the copied page. Newly copied pages have NONE value for the start and end date.
```

PR #1103, opened 2019-03-11
Allow both e-mail and full name to be searched for users in wagtail admin.

### LINKED ISSUE #470: Add full name to user search index in Wagtail admin

frjo:
Closes #470

### DIFF
--- a/opentech/apply/users/admin_views.py
+++ b/opentech/apply/users/admin_views.py
@@ -22,6 +22,10 @@ delete_user_perm = "{0}.delete_{1}".format(AUTH_USER_APP_LABEL, AUTH_USER_MODEL_
 @any_permission_required(add_user_perm, change_user_perm, delete_user_perm)
 @vary_on_headers('X-Requested-With')
 def index(request):
+    """
+    Override wagtail's users index view to filter by full_name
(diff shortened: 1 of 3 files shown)


frjo:
@Parbhat In the current site you can only search the e-mail field for users in wagtail admin "/admin/users/?q=something". But at least it's support partial match in any part of the e-mail address.

When I try to add full name in this PR I get it to work but without real partial match. IT seems to only match from the start of the string.

Tips how to solve this or better you take it over?

Parbhat:
@frjo Updated PR to search by full name of the user. This area of Wagtail admin does not use Wagtail search index. It performs [filtering](https://github.com/wagtail/wagtail/blob/af69cb4a544a1b9be1339546be62ff54b389730e/wagtail/users/views/users.py#L61) on the users by fields like `username`, `email`, `first_name` and `last_name`.  As we have our custom field named `full_name`, it does not perform the search. We have also turned [off](https://github.com/OpenTechFund/opentech.fund/blob/14e7b89ceac3f1679a83d71e9075bf129b3da8d3/opentech/apply/users/models.py#L79) fields like `username`, `first_name` and `last_name` as we are not using these fields.

Created a new index view which will overtake the handling of URL `/admin/users/` so we can search by full name. There are no hooks present to add custom filter fields.

frjo:
Work perfectly, thanks.

PR #1113, opened 2019-03-14
Allow only specific files in Uploads

### LINKED ISSUE #1099: Only allow specified file types in file uploads on submissions

Parbhat:
Closes #1099 

Frontend and Backend checks for files being uploaded. Only allow files with specific extensions.

### DIFF
--- a/opentech/apply/stream_forms/blocks.py
+++ b/opentech/apply/stream_forms/blocks.py
@@ -318,10 +318,16 @@ class UploadableMediaBlock(OptionalFormFieldBlock):
 
 class ImageFieldBlock(UploadableMediaBlock):
     field_class = forms.ImageField
+    widget = ClearableFileInput
 
     class Meta:
         label = _('Image field')
         icon = 'image'
+
+    def get_field_kwargs(self, struct_value):
+        kwargs = super().get_field_kwargs(struct_value)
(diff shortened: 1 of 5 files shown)


frjo:
@Parbhat Nicely done!

How complicated would it be to add the HTML accept attribut to the image stream field as well. That file already only accept image file types server side (default Django thing I believe).

With the test fixed this is RTBC.

Parbhat:
@frjo Similar code can be used to add `accept` attr to Image field as well. What image extensions do we need to support? In the backend, Django uses Pillow library to check the uploaded file is a valid image.

frjo:
@Parbhat I think we can hardcode it to simply "png gif jpg jpeg".

This will be only client side so easy to circumvent if someone really wants. But the field will only access image formats as specified by Pillow server side.

If a user manage to upload a tiff it doesn't really matter. The HTML accept attribut is there to help guide the user.

Parbhat:
@frjo Great. Adding the `accept` attr for Images as well.

Parbhat:
@frjo Fixed tests and add accept attr for images with value `image/*`. Django 2.1 includes `accept` attr for images so we can remove the code when we are on [2.1](https://docs.djangoproject.com/en/2.1/releases/2.1/#forms).

PR #1123, opened 2019-03-18
Apply: Group fields in application forms 

### LINKED ISSUE #1064: CFP: Make it possible to group forms
Make it possible to group forms and set them as collapsed by default. A trigger field can make them uncollapsed.

This could be done by adding a feature to group fields in application forms or by having multiple application forms for a fund with the new "Request with community review" workflow being stitched together.

Parbhat:
Closes #1064 

@frjo Initial draft PR which adds group number to form fields. Streamfield allows arranging fields in the admin area. PR adds a new block called `group` to form fields StreamBlock. It is a boolean block which will not render on the page but will act as a Flag to group forms. Please review the approach and we can improve it further if this looks okay.

Also, if we use one form and toggle some form fields on the page we have to consider required/non-required fields. If the other fields (hidden) are not meant for the user filling the form but they are required then the user will not be able to submit the form.


frjo:
@Parbhat  I think this looks good. A generic function that can be used for all kind of things, including the feature we are buying it for now.

Good catch on the required fields issue. I think we for now only tell OTF how it works, there is only a few people there who build these forms. If it becomes a problem we can deal with it later on.

opentech/apply/funds/templates/funds/application_base.html
```diff
@@ -32,9 +32,13 @@ <h3>{% blocktrans %}Sorry this {{ page|verbose_name }} is not accepting applicat
 
             {% for field in form %}
               {% if field.field %}
-                {% include "forms/includes/field.html" %}
+                <div class="field-group-{{ field.field.group }}">
```
frjo: Would it be possible to add this class to the wrapping div in field.html? So we avoid to many levels of divs.

opentech/apply/stream_forms/blocks.py
```diff
@@ -384,6 +384,10 @@ class FormFieldsBlock(StreamBlock):
     image = ImageFieldBlock(group=_('Fields'))
     file = FileFieldBlock(group=_('Fields'))
     multi_file = MultiFileFieldBlock(group=_('Fields'))
+    group = BooleanBlock(
+        group=_('Custom'), label=_('Group fields'), template='stream_forms/render_group_field.html',
+        icon='group', help_text='This field will not render on the page. Only helps to group fields.'
+    )
```
frjo: This needs to be checked I discovered, maybe have it be checked by default in that case.

frjo:
@Parbhat The toggle field will be a question on the submission form that applicants fill in. Depending how they answer the corrosponding group will revel or not.

opentech/apply/funds/templates/funds/application_base.html
Parbhat: Yeah, PR updated.

opentech/apply/stream_forms/blocks.py
Parbhat: Created a new block instead of using the boolean block for toggle effect.

Parbhat:
@frjo Thanks for the review. Updated PR to use custom block GroupToggleBlock instead of boolean block. Changed PR status from draft to open.

opentech/apply/stream_forms/blocks.py
```diff
@@ -186,6 +186,38 @@ def get_field_kwargs(self, struct_value):
         return kwargs
 
 
+class GroupToggleBlock(FormFieldBlock):
+    required = BooleanBlock(label=_('Required'), default=True)
+    choices = ListBlock(
+        CharBlock(label=_('Choice')),
+        help_text='Please create only two choices for toggle. Additional choices will be ignored.'
```
frjo: Please create only two choices for toggle. First choice will revel the group and the second hide it. Additional choices will be ignored.

Parbhat: Currently, we can not use `max_num` option in ListBlock. There is one open PR in [Wagtail](https://github.com/wagtail/wagtail/pull/3914). Therefore handle this case by ignoring choices after 2. And if fewer choices are provided, default to Yes and No.

frjo:
Just tested this and it works really well. Added a comment about improving the help text.

opentech/apply/stream_forms/blocks.py
Parbhat: Just realised comment is for help text. Updating PR.

Parbhat:
@frjo Thanks for the quick review. Updated help text for group toggle choices.

PR #1136, opened 2019-03-26
Open call index page and prefill data from submission with open call status

### LINKED ISSUE #1065: CFP: Add a "Build it" button to Submissions with new workflow and the "Open call" state.
Clicking the button will create an duplicate of the submission but with current author and date. All form groups will be uncollapsed.

### LINKED ISSUE #1066: CFP: Add a public view to display submission in "Open call" state
Related: 
https://github.com/OpenTechFund/opentech.fund/issues/1065

Parbhat:
Closes #1066 
Closes #1065

@frjo What data from submission should be displayed on Open Call index page for a submission? Currently listing submission title only. And when user click on built it button (#1065)

>Clicking the button will create an duplicate of the submission but with current author and date. All form groups will be uncollapsed.

Should we only prefill the form data with clicked submission or create the submission object in the DB first and then give the user form for an update? As the application form is for anonymous users, it seems that we can only prefill the data.


frjo:
@Parbhat Prefilling the form data sounds like the correct approach. Except for name and e-mail.

The title maybe should get some prefix/suffix. "Copy of [title] (please edit)" or something like that?

The field group toggle field should be profiled with the first option so the groups is displayed.

Regarding what to show, hmm, very good question. Need to have a discussion with OTF about that.

frjo:
@Parbhat I have talked with OTF now.

The first part of the submission will only have about three textfields and one text area. And there will likely not be more than five (5) open call submission at a time.

I suggest we make it simple and display the complete submission on the index page. If it proves to big we can later on use css/js to have a "fold out" function that displays full text.

The name should show after all but not the e-mail address.

CSS and any changes to the display of the content I can take on when your are done.

Parbhat:
@frjo Sounds good. I will update PR.

frjo:
@Parbhat I have rebased the branch from master and added some styling. IT all seems to work really well.

The last thing is to get the form to prepopulate. Any ideas how to best do that?

Parbhat:
@frjo Thanks for updating the styles. Pushed commit for the initial approach to prefill the first group form fields when user click Built it button. It is not complete. It would be great if you can check if it is the right direction. And then can complete it.

opentech/apply/funds/models/applications.py
```diff
@@ -280,9 +281,59 @@ def clean(self):
...
+        form_parameters = {}
+        initial_values = {}
+
+        if request:
+            copy_open_submission = request.GET.get('open_call_submission')
+            if copy_open_submission:
+                submission_class = self.get_submission_class()
+                try:
+                    submission = submission_class.objects.get(id=copy_open_submission)
+                    if self.get_parent().id == submission.page.id:
```
frjo: This checks that the open_call_submission is the the same fund but we also need to make sure that  open_call_submission is in state "open call".

If not there is a potential information leak.

frjo:
Looks good, made on inline comment about further checks needed.

Add in title and set the group toggle to open and it's done I believe.

opentech/apply/funds/models/applications.py
Parbhat: Thanks for noticing. Checking open call was on my list but got skipped. PR updated.

Parbhat:
@frjo PR is ready for final review.

PR #1144, opened 2019-04-01
Allow reviewers and partners to access their own submissions as applicants

### LINKED ISSUE #1143: Partners can't access their own submissions
A user that has the role "Partner" can't access their own submissions.

frjo:
Fixes #1143


frjo:
@Parbhat The goal is to direct partners and reviewers to the "ApplicantSubmissionDetailView" for their own submissions.

A single user account can be both partner and applicant and the system now assumes "ones a partner always a partner" and they can't access their own submissions.

This PR seems to solve it but can you take a look and see if it looks ok?

Better or nicer way to solve it?

frjo:
@Parbhat Thanks! I went with your solution, clean and simple.

PR #1145, opened 2019-04-02
Delete Submission permission, view and link

### LINKED ISSUE #728: Create a permission and an interface for deleting submissions, giving that permission to team and admins, and notification on action to channel

Parbhat:
Groups can be given delete submission permission from Wagtail admin.

Closes #728


opentech/apply/funds/views.py
```diff
@@ -746,3 +746,20 @@ class RoundListView(SingleTableMixin, FilterView):
...
+
+    def delete(self, request, *args, **kwargs):
+        submission = self.get_object()
+        messenger(
+            MESSAGES.DELETE_SUBMISSION,
+            user=request.user,
+            request=request,
+            submission=submission,
+        )
+        response = super().delete(request, *args, **kwargs)
```
Parbhat: Ideally, the message should be sent after deletion of submission. But the messaging system requires Event model object which has a non-null foreign key to submission. We have to make some changes in the messaging system to make [event](https://github.com/OpenTechFund/opentech.fund/blob/a281060d595cc5e20c445a214689008814aef504/opentech/apply/activity/messaging.py#L710) optional. Suggestions?

Parbhat:
@frjo Thanks for the view and adding missing migration.

opentech/apply/funds/views.py
Parbhat: yeah, it should not be an issue.

PR #1149, opened 2019-04-04
Notify Second Reviewer when First Review is finished.

### LINKED ISSUE #1125: Notify Second Reviewer when First Review is finished [High/Improvement]
-=Sensitive=-
No

-=Reporter=-
Tara

-=Ideal deadline=-

-=Who's affected=-
OTF team

-=Description=-
On slack, the notification of a review only notifies the round lead of the review. It should also tag the second reviewer on the CN/Proposal.

-=Objective=-
Second reviewers are notified when they are able to put in their review. 

-=Additiional info=-
no, thanks!

-=Timestamp=-
2019-03-21T01:42:00.852Z

frjo:
Fixes #1125

### DIFF
--- a/opentech/apply/activity/messaging.py
+++ b/opentech/apply/activity/messaging.py
@@ -128,7 +128,7 @@ class AdapterBase:
             self.process_send(message_type, recipients, events, request, user, submissions=submissions, submission=None, related=related, **kwargs)
 
     def process(self, message_type, event, request, user, submission, related=None, **kwargs):
-        recipients = self.recipients(message_type, submission=submission, **kwargs)
+        recipients = self.recipients(message_type, submission=submission, related=related, **kwargs)
         self.process_send(message_type, recipients, [event], request, user, submission, related=related, **kwargs)
 
     def process_send(self, message_type, recipients, events, request, user, submission, submissions=list(), related=None, **kwargs):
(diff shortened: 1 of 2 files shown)


opentech/apply/activity/messaging.py
```diff
@@ -159,10 +159,10 @@ def process_send(self, message_type, recipients, events, request, user, submissi
 
             if not settings.SEND_MESSAGES:
                 if recipient:
-                    message = '{} [to: {}]: {}'.format(self.adapter_type, recipient, message)
+                    debug_message = '{} [to: {}]: {}'.format(self.adapter_type, recipient, message)
```
frjo: Avoid messing up the message variable if there is more than one recipient.

opentech/apply/activity/messaging.py
```diff
@@ -128,7 +128,7 @@ def process_batch(self, message_type, events, request, user, submissions, relate
             self.process_send(message_type, recipients, events, request, user, submissions=submissions, submission=None, related=related, **kwargs)
 
     def process(self, message_type, event, request, user, submission, related=None, **kwargs):
-        recipients = self.recipients(message_type, submission=submission, **kwargs)
+        recipients = self.recipients(message_type, submission=submission, related=related, **kwargs)
```
frjo: Added the "related" argument. Will other instances of "recipients" functions complain if it's not present?

Parbhat: There will be no issue as other `recipients` functions will have `related` available in `kwargs` if not explicitly defined.

opentech/apply/activity/messaging.py
Parbhat: 👍

opentech/apply/activity/messaging.py
```diff
@@ -388,8 +388,15 @@ def extra_kwargs(self, message_type, **kwargs):
             'links': links,
         }
 
-    def recipients(self, message_type, submission, **kwargs):
-        return [self.slack_id(submission.lead)]
+    def recipients(self, message_type, submission, related, **kwargs):
+        recipients = [self.slack_id(submission.lead)]
+
+        # Notify second reviewer when first reviewer is done.
+        if message_type == MESSAGES.NEW_REVIEW:
```
Parbhat:
```suggestion
        if message_type == MESSAGES.NEW_REVIEW and related:
```

We can have check for related to avoid AttributeError if related is None.

Issue #985, opened 2019-02-07
Review updates: reviewer consensus: updates to table display


todd-dembrey:
For review summary indicator (traffic lights) wherever they appear: Agreeing with the review will be considered as having left a review of the same outcome. Disagreeing with a review will be considered as having left a review of the opposite outcome, always resulting in an Amber circle

**Acceptance criteria**
- [x] Agreements are reflected in the count of completed reviews in the review column of submission tables
- [x] Disagreements are not reflected in the count of completed reviews in the review column of submission tables
- [x] Disagreements will change a red or green traffic light to amber.

**Dev Note**
Both Disagreement and Agreement will be saved - `for_table` queryset method updated so that `review_submitted_count` includes agreements, and does not include disagreement.
https://github.com/OpenTechFund/opentech.fund/blob/7b49da51dcc08407abf3d9a9a94b29fadb849a2b/opentech/apply/funds/models/submissions.py#L122

`review_recommendation` impacts the traffic light. 
https://github.com/OpenTechFund/opentech.fund/blob/7b49da51dcc08407abf3d9a9a94b29fadb849a2b/opentech/apply/funds/models/submissions.py#L126

**QA criteria**
Dev:

- [x] checked feature meets acceptance criteria/conforms exactly to the specification.
- [x] provided good unit test coverage (if this is non-trivial behaviour).
- [x] checked all tests for the project pass with this feature enabled.
- [x] checked code conforms to the project coding standards.
- [x] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end user of the website/app (Can I get to it? Is it useable? Can I break it? Does it work in an end-to-end context?)
- [x] checked that this feature works on the server/s I am deploying it to.

QA:

 * [x] tested this feature as a front end user and it meets the acceptance criteria/conforms to the specification and design.
 * [x] checked that the feature works on the server/s deployed to

frjo:
@todd-dembrey Could making sure the list of reviewers being sorted in the same manner in all places be part of this as well?

todd-dembrey:
I've done the bulk of the work fro this ticket and will pick up on it again next friday. 

The only blocking issue is de-duplicating the authors from opinion and reviews (in progress in latest commit)

todd-dembrey:
@frjo I've created a new issue for your comment about the ordering of the reviewers.

todd-dembrey:
@fourthletter @danblah This is deployed to the test instance for review.

[...]

PR #1151, opened 2019-04-05
Feature/985 update table for opinions

todd-dembrey:
This should resolve the issues around the correct number of reviews being included in the tables and also reflects the outcome of opinions on the recommendation. 

See #985 for original issue


frjo:
@todd-dembrey Tried to test this but got a "subquery must return only one column" error on dashboard as well as submission list view.

frjo:
Rebased the branch.

fourthletter:
I've tested this issue. Thanks!

frjo:
@todd-dembrey I'm having an issue with the "/apply/submissions/all/" view and a clone of the live db. Returns a proxy error (Apache/Gunicorn). Postgres CPU usage also spikes.

Switching to master branch and all is well.

It works on test however and @Parbhat can not replicate it on his local dev env.

todd-dembrey:
@frjo I'll have a look at this.

todd-dembrey:
@frjo I would appreciate a review on this. There are some serious data migrations going on and would appreciate a second opinion. 

I'm not 100% happy with all the `author.reviewer` references going on at the moment. Will perhaps take a break and see if i can tidy that up.

The database queries are much quicker now so i think the performance issue is closed.

todd-dembrey:
This is awaiting an internal review from Dan. Then we will also need to go through QA at our end before we push to live.

opentech/apply/review/views.py
```diff
@@ -240,7 +224,11 @@ def test_func(self):
 
     def form_valid(self, form):
         self.review = self.get_object()
-        form.instance.author = self.request.user
+        author, _ = AssignedReviewers.objects.get_or_create_staff(
```
zerolab: Should be get_or_create_for_user() as this can be submitted by more than staff

opentech/apply/review/views.py
```diff
@@ -27,58 +26,38 @@
...
+
+        ordering = [models.When(type__name=review_type, then=models.Value(i)) for i, review_type in enumerate(review_order)]
+        assigned_reviewers = self.object.assigned.annotate(
+            type_order=models.Case(
+                *ordering,
+                output_field=models.IntegerField(),
+            )
+        ).order_by(
+            'role__order',
+            'review',
```
zerolab: Note on `nulls_last` discussed offline

Issue #1081, opened 2019-03-06
Investigate memory usage, often close to the max 512 MB


frjo:
Can we get Django to use less or do we need to upgrade to 2x dynos that has 1 GB memory.

todd-dembrey:
You can use the scout reporting to help identify some areas for improvements.

https://scoutapp.com/apps/28725?section=memory&p=memory&s=response_time&insights=more

Parbhat:
@todd-dembrey @frjo Can I get access to scout app OTF project to review possible performance issues?

frjo:
@Parbhat Invite sent.

I have upgraded to 2x dynos on Heroku for the moment to avoid problems. Would be nice if we could revert back to 1x.

Parbhat:
@frjo Thanks, got access.

Parbhat:
@frjo One of the reasons for the recent increase in database queries and increased response time is changes made in [table.html ](https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/apply/funds/templates/funds/tables/table.html). The changes made in the template increased the number of DB queries significantly. As this template is used for many views to create a table it reduces the overall performance. Given it a quick shot to reduce the effects but it will take more time. Also, are we moving some pages containing tables like `/apply/submissions/all/` to React frontend? If yes, we check the performance again after the migration is made.

frjo:
@Parbhat There is only a very specific use of react and it's for OTF staff only. Do not think that will be expanded much.

frjo:
This is no longer an issue after performance improvements in other issues.

PR #1154, opened 2019-04-08
Improve perf by not firing db query when review author and review opinion author is accessed

Parbhat:
Link #1081 

Currently, a query is fired for each row in the submissions list table when [review author](https://github.com/OpenTechFund/opentech.fund/blob/2c8c3de93b4b9581a6aaaf99867d8cb0f9ad5f2b/opentech/apply/funds/templates/funds/tables/table.html#L43) and review [opinion author](https://github.com/OpenTechFund/opentech.fund/blob/2c8c3de93b4b9581a6aaaf99867d8cb0f9ad5f2b/opentech/apply/funds/templates/funds/tables/table.html#L52) is accessed in the template. Uses [Prefetch](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#prefetch-objects) objects.

Also, are we using Opinion on the Review feature in the apply site? The code is present but not sure it is being used. If not we can update the modified code.

### DIFF
--- a/opentech/apply/funds/models/submissions.py
+++ b/opentech/apply/funds/models/submissions.py
@@ -6,7 +6,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.postgres.fields import JSONField
 from django.core.exceptions import PermissionDenied
 from django.db import models
-from django.db.models import Count, IntegerField, OuterRef, Subquery, Sum, Q
+from django.db.models import Count, IntegerField, OuterRef, Subquery, Sum, Q, Prefetch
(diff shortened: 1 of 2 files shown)


frjo:
A significant performance improvement for the dashboard. A busy reviewers dashboard went from 134 queries to 71 and CPU time from 5,7s to 3s.

frjo:
A reviewer can give opinion on another reviewers review (Agree/Disagree). This is not displayed on the dashboard but an opinion should count as "has been reviewed by".

Parbhat:
Great

PR #1157, opened 2019-04-10
Set correct Cache-Control max-age and s-maxage headers on all pages

frjo:
I have added one commit to allow CACHE_CONTROL_MAX_AGE and CACHE_CONTROL_S_MAXAGE environment variables to be set, they default to 0.

CACHE_CONTROL_MAX_AGE => max-age
CACHE_CONTROL_S_MAXAGE => s-maxage (Cloudflare will use this if present, otherwise standard max-age)

* Set never cache on all pages on the Apply site except the front page. They seem to have no cache_control header at all currently.

### DIFF
--- a/opentech/settings/base.py
+++ b/opentech/settings/base.py
@@ -212,11 +212,16 @@ else:
         }
     }
 
+# Set max-age header. See urls.py
+try:
+    CACHE_CONTROL_MAX_AGE = int(env.get('CACHE_CONTROL_MAX_AGE', 0))
+except ValueError:
+    pass
 
 # Set s-max-age header that is used by reverse proxy/front end cache. See
 # urls.py
 try:
-    CACHE_CONTROL_S_MAXAGE = int(env.get('CACHE_CONTROL_S_MAXAGE', 600))
+    CACHE_CONTROL_S_MAXAGE = int(env.get('CACHE_CONTROL_S_MAXAGE', 0))
 except ValueError:
     pass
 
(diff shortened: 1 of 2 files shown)


frjo:
@Parbhat Suggestions for the best implementation?

Parbhat:
@frjo There is some issue with [urlpatterns](https://github.com/OpenTechFund/opentech.fund/blob/2c8c3de93b4b9581a6aaaf99867d8cb0f9ad5f2b/opentech/urls.py#L32) list that will get `cache-control` header with `max-age/s-maxage`. Some URLs in the list should not be cached like 

- Admin URLs and users public [URLs](https://github.com/OpenTechFund/opentech.fund/blob/2c8c3de93b4b9581a6aaaf99867d8cb0f9ad5f2b/opentech/apply/users/urls.py#L11) should not be cached as they contain CSRF tokens etc.

- Wagtail URLs are cached in a different way. Decorating `path('', include(wagtail_urls)),` will not work. As in Wagtail, model's serve method should be decorated. Like we applied on public form pages. We can decorate [BasePage](https://github.com/OpenTechFund/opentech.fund/blob/2c8c3de93b4b9581a6aaaf99867d8cb0f9ad5f2b/opentech/public/utils/models.py#L245) serve method to add cache-control header on public wagtail pages.

- Search [URLs](https://github.com/OpenTechFund/opentech.fund/blob/master/opentech/urls.py#L39) should not be cached.

It seems that the `urlpatterns` list get modified in the meantime the feature to add cache-control is disabled or it may not be activated at all. The `urlpatterns` list should be divided into which URLs can be cached and which should not.

For adding a header to `urlpatterns` we can create a dictionary from settings by excluding None values. And apply like `cache_control(**cache_control_kwargs)`

frjo:
@Parbhat Excellent, please take this on after you have done the Django cache work.

Parbhat:
Thanks, will work on it.

Issue #949, opened 2019-02-01
Utilising Django cache to improve page load times, especially for tables listing many submissions


frjo:
@Parbhat Could you spend a few hours investigating if/how to best use the Django cache system for pages, views and queries?

Would e.g. the page cache work for the public pages? Could we cache queries/views for the tables that show many submissions (/apply/submissions/, /apply/submissions/all/, /apply/rounds/[round_id]/).

The Django site is quite a lot slower to load than the old Drupal site and caching is the main difference I believe.

PR #1158, opened 2019-04-11
Cache submission overview and list pages

Parbhat:
Link #949 

Initial cache for submission overview (`/apply/submissions/`) and list pages (`/apply/submissions/all/`). The cache will be specific to the user as add Vary by Cookie header. And pages will be cached for 30 mins. This is the initial trial and can extend further timing and more pages.

The first request to the page will be normal and subsequent requests will be served from cache. For testing locally, the backend should be changed from Dummy cache in [local.py](https://github.com/OpenTechFund/opentech.fund/blob/2c8c3de93b4b9581a6aaaf99867d8cb0f9ad5f2b/opentech/settings/local.py.example#L3). Dummy cache doesn’t actually cache – it just implements the cache interface without doing anything.

### DIFF
--- a/opentech/apply/funds/views.py
+++ b/opentech/apply/funds/views.py
@@ -10,6 +10,8 @@ from django.urls import reverse_lazy
 from django.utils.decorators import method_decorator
 from django.utils.text import mark_safe
 from django.utils.translation import ugettext_lazy as _
+from django.views.decorators.cache import cache_page
+from django.views.decorators.vary import vary_on_cookie
 from django.views.generic import DetailView, FormView, ListView, UpdateView, DeleteView
 
 from django_filters.views import FilterView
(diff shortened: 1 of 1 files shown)


frjo:
@Parbhat How is cache invalidation handled? As soon as a submission or any related content to it is updated the cache needs to be updated.

Parbhat:
@frjo The current cache is specific to a user. So, a user 1 will have a separate cache from user 2. We can burst a user cache when submission is updated for a submission that can be accessed by the user. Also, the cache is dependent on user cookie. Signing out and login will also burst the cache. Caching based on user helps to cache table data permitted to user and show correct username in menu.

frjo:
Yes, doing it by user is necessary I believe with the permissions we have in place.

But when staff is working with the system they will expect everything to be up to date. Status, number of reviews etc. needs to be updated directly if changed.

Can we tag caches with content id:s? That way we could clear all caches by id when thet content is updated.

frjo:
Maybe we can start by implementing cache for the public pages and the apply front page? They are easier to handle I believe.

Parbhat:
@frjo Currently we are using [per-view cache](https://docs.djangoproject.com/en/2.1/topics/cache/#the-per-view-cache). It caches the complete page. We can also use [low-level cache API ](https://docs.djangoproject.com/en/2.1/topics/cache/#the-low-level-cache-api) to control the bits that should be cached. And on some signal like submission update delete it from the cache.

For public pages, this package looks good - [wagtail-cache](https://github.com/coderedcorp/wagtail-cache). I will check if it fits our use case.

Parbhat:
@frjo Should we cache public pages server side or just send the cache-control response header with max-age, s-maxage.

frjo:
Also cache them, not everyone running this system will have a CDN in front. 

PR #1159, opened 2019-04-15
Light colours for submission diff.

frjo:


fourthletter:
We tested: https://test-apply.opentech.fund/ and believe it's ready to deploy. Thanks!

PR #1160, opened 2019-04-15
 Cache public site pages and Apply homepage

Parbhat:
Cache for the public site and apply homepage can be invalidated from `settings > cache` in Wagtail admin. The cache will also be invalidated automatically when a wagtail page is edited/created. The cache will not work for logged in users. And WAGTAIL_CACHE should be set `True` in local.py settings while testing locally.

Also, remove the unused code which adds cache_control in urlpatterns.

Link issue #949 and PR #1158


opentech/settings/base.py
```diff
@@ -198,6 +198,23 @@
...
+    CACHE_CONTROL_MAX_AGE = 3600
+
+# Set s-max-age header that is used by reverse proxy/front end cache.
+try:
+    CACHE_CONTROL_S_MAXAGE = int(env.get('CACHE_CONTROL_S_MAXAGE', 3600))
+except ValueError:
+    CACHE_CONTROL_S_MAXAGE = 3600
+
+# Set wagtail cache timeout (automatic cache refresh).
+WAGTAIL_CACHE_TIMEOUT = CACHE_CONTROL_S_MAXAGE
```
Parbhat: It should be set to CACHE_CONTROL_MAX_AGE.

opentech/apply/funds/templates/funds/tables/table.html
```diff
@@ -1,7 +1,8 @@
 {% extends 'django_tables2/table.html' %}
-{% load django_tables2 table_tags review_tags %}
+{% load adv_cache django_tables2 table_tags review_tags %}
 
 {% block table.tbody.row %}
+    {% cache 3600 submission_table_row row.record.pk row.record.status row.record.last_update %}
```
frjo: What do you think of this solution?

This will cache each table row that contains a single submission. Need to test to get the correct cache tag but this is a good start I believe.

Would like to use ADV_CACHE_TIMEOUT here instead of hardcoded "3600". Can you help with that? Tried using context but did not get it to work.

opentech/settings/base.py
```diff
@@ -197,28 +199,66 @@
...
+    CACHE_CONTROL_MAX_AGE = 3600
+
+# Set s-max-age header that is used by reverse proxy/front end cache.
+try:
+    CACHE_CONTROL_S_MAXAGE = int(env.get('CACHE_CONTROL_S_MAXAGE', 3600))
+except ValueError:
+    CACHE_CONTROL_S_MAXAGE = 3600
+
+# Set wagtail cache timeout (automatic cache refresh).
+WAGTAIL_CACHE_TIMEOUT = CACHE_CONTROL_MAX_AGE
```
frjo: Fixed this per your comment.

frjo:
@Parbhat I removed the Advanced cache template tag parts. Should be in another pr and when I started testing the performance impact was small.

Parbhat:
@frjo The solution for caching table using cache template tag looks good. There are some UI issues in some pages like dashboard start showing checkboxes at the start of the row. We can also use the normal [cache](https://docs.djangoproject.com/en/2.1/topics/cache/#template-fragment-caching) template tag instead of advanced `django-adv-cache-tag`. While testing noticed that while switching between Django cache tag and django adv cache tag, we need to clear the cache.

Issue #702, opened 2018-11-15
Ensure all form fields have sensible default values


danblah:

tmsndrs:
UX tasks
- [ ] Review and make recommendations

fourthletter:
Related to #1162  #714 #552

Issue #552, opened 2018-09-14
Allow reviews to be edited by staff and author


frjo:

danblah:
Assuming this will require enabling revisions for the content type. If so, we'd need to clearly indicate that an edit/new revision has been created when the viewer is reviewing a review.

danblah:
I want to bump this up as something to get moving on.

fourthletter:
Related to #1162 #714 #702

fourthletter:
@frjo AC reviewer was not able to edit their own review https://apply.opentech.fund/apply/submissions/4027/reviews/5276/

frjo:
@danblah Is it a requirement that the reviews have revisions?

danblah:
No, but please make a follow up ticket for this functionality.

frjo:
Revisions moved to #1311

fourthletter:
@frjo Test worked. Thanks!

frjo:
@fourthletter Found my bug and fixed it, so now it should work.

fourthletter:
@frjo Test was successful. Thanks!

PR #1162, opened 2019-04-17
Set sensible defaults for reviewer score fields.

### LINKED ISSUE #714: Set sensible defaults for reviewer score fields. Default is set to zero which could lead to mis-scoring in user error

frjo:
Fixes #714


fourthletter:
@frjo I've tested this issue. 1) NA is the default display, 2) A review receiving NA or default in each section of a review becomes a "no" or red spotlight score. 

Example NA:
image

Example 99:
image

frjo:
@fourthletter That behaviour is unchanged. Theis PR only changes the default answer, not how they are calculated. We can fix the calculation if needed in a separate PR.

fourthletter:
Relates to #702 #552

PR #1180, opened 2019-04-29
Adding news feed for all news and for each news type.

### LINKED ISSUE #1172: Create an RSS feed for all updates/news and sub-feeds for each term in the newstype taxonomy

frjo:
Fixes #1172


frjo:
@Parbhat Best way to cache the news feeds?

frjo:
All the OTF specific strings should really be settings when I think about it.

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
+from django.contrib.syndication.views import Feed
+
+from opentech.public.news.models import NewsPage, NewsType
+
+
+class NewsFeed(Feed):
+    link = "https://www.opentech.fund/news/"
```
frjo: Can we get this programatically? If not I should make it a setting.

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+    title = "OTF – News, updates, and announcements"
+    description = "News, updates, and announcements from The Open Technology Fund"
```
frjo: Will make these two in to settings.

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+    link = "https://www.opentech.fund/news/"
+
+    def get_object(self, request, type):
+        return NewsType.objects.get(id=type)
+
+    def title(self, type):
+        return f"OTF – News of type {type}"
+
+    def description(self, type):
+        return f"News, updates, and announcements of type {type} from The Open Technology Fund"
```
frjo: Best way to make this in to a setting?

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+    description = "News, updates, and announcements from The Open Technology Fund"
+
+    def items(self):
+        return NewsPage.objects.live().order_by('-first_published_at')[:20]
+
+    def item_title(self, item):
+        return item.title
+
+    def item_description(self, item):
+        return item.body
```
Parbhat: NewsPage's body field is a streamfield for content. Using `introduction` field will be good here as description is for item synopsis.

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+        return f"News, updates, and announcements of type {type} from The Open Technology Fund"
+
+    def items(self, type):
+        return NewsPage.objects.live().filter(news_types__news_type=type).order_by('-first_published_at')[:20]
+
+    def item_title(self, item):
+        return item.title
+
+    def item_description(self, item):
+        return item.body
```
Parbhat: Similar to https://github.com/OpenTechFund/opentech.fund/pull/1180/files#r279664274

opentech/public/news/views.py
Parbhat: We can get `NewsIndex` page in method `def link(self, obj):` and remove link attribute.

opentech/public/news/views.py
Parbhat: We can create the Feed Settings in the wagtail admin using [Site settings](https://docs.wagtail.io/en/v2.5/reference/contrib/settings.html) feature.

opentech/public/news/views.py
Parbhat: We can create a setting with string value "News, updates, and announcements of type {type} from The Open Technology Fund". And then get the setting here and apply string format on it like `<setting-str>.format(type=type)`

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def item_pubdate(self, item):
+        return item.display_date
+
+
+class NewsTypesFeed(Feed):
+    link = "https://www.opentech.fund/news/"
```
Parbhat: We need link value like `https://www.opentech.fund/news/?news_type=1` here, therefore can use a function `def link(self, obj)`.

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def get_object(self, request, type):
+        return NewsType.objects.get(id=type)
+
+    def title(self, type):
+        return f"OTF – News of type {type}"
```
Parbhat: This works but can create confusion. The title method receives the actual object and not the id. Therefore can change it to `def title(self, obj)`.

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def description(self, type):
```
Parbhat: Similar to https://github.com/OpenTechFund/opentech.fund/pull/1180/files#r279681502

opentech/public/urls.py
```diff
@@ -1,10 +1,13 @@
...
 
+from .news import views as news_views
 from .search import views as search_views
 from .mailchimp import urls as newsletter_urls
 
 
 urlpatterns = [
     path('search/', search_views.search, name='search'),
+    path('news/feed/', news_views.NewsFeed(), name='news_feed'),
+    path('news/<int:type>/feed/', news_views.NewsTypesFeed(), name='news_types_feed'),
```
Parbhat: We can use `news_type` here and other related places as `type` to make it more explicit and also type is a python built-in method.

opentech/public/news/views.py
```diff
@@ -0,0 +1,46 @@
...
+        return f"News, updates, and announcements of type {type} from The Open Technology Fund"
+
+    def items(self, type):
+        return NewsPage.objects.live().filter(news_types__news_type=type).order_by('-first_published_at')[:20]
```
Parbhat: We are using `display_date` property for `pubDate`. Therefore can use [Coalesce](https://docs.djangoproject.com/en/2.2/ref/models/database-functions/#coalesce) to have the same behaviour.

 ```python
NewsPage.objects.live().public().filter(news_types__news_type=type).annotate(
            date=Coalesce('publication_date', 'first_published_at')).order_by('-date')[:20]
```

**Comment by Parbhat** [comment]
To implement caching, we can override [\_\_call__](https://github.com/django/django/blob/6866c91b638de5368c18713fa851bfe56253ea55/django/contrib/syndication/views.py#L34) method and use django's low-level [cache API](https://docs.djangoproject.com/en/2.1/topics/cache/#the-low-level-cache-api).

File: opentech/public/news/views.py (hunk 4 above)
Comment (frjo) [reply opentech/public/news/views.py]: I see your point but introduction field is not used a lot in practice and I want a full content rss feed. Tested the feed in my feed reader and it seems to work as it should with body.

**Comment by frjo** [comment]
@Parbhat Thanks for all the tips, learning a lot about Django and Wagtail with your guides!

The tests are now complaining that "wagtailcore_site" does not exist. But should they not be set by "opentech/public/home/migrations/0002_create_homepage.py"?

**Comment by Parbhat** [comment]
@frjo Feed class from Django works a little different. Included a commit to fix tests and override the `__call__` method so we can use wagtail settings. The reason tests were failing: The code executes before any migrations were run by tests. Also, locally I was not able to run the migration to create the table for NewsFeedSettings. As NewsFeedSettings was accessed first in NewsFeed class attribute. In URLs file, we create an instance of the class NewsFeed and then the instance is called as a function (made callable using `__call__`).

**Comment by Parbhat** [comment]
For using full content in the description element:

Read from some sources and it looks fine to use description element to hold full news page content.

> An item's description element holds character data that contains the item's full content or a summary of its contents, a decision entirely at the discretion of the publisher. - [rssboard](http://www.rssboard.org/rss-profile#element-channel-item-description)

There is also [content:encoded](http://www.rssboard.org/rss-profile#namespace-elements-content-encoded) element whose precise purpose is to hold full content. But one recommendation mentioned in rssboard

>Publishers who don't want to employ item summaries in their feeds SHOULD use the description element for an item's full content rather than content:encoded because it has the widest support.

>Publishers who employ summaries SHOULD store the summary in the description and the full content in content:encoded, ordering description first within the item. On items with no summary, the full content SHOULD be stored in the description.

So, it seems good to use the description element for full content if we do not want a summary. Also, adding content:encoded needs a bit of work and an example is [here](https://github.com/chrisdev/django-wagtail-feeds/blob/df377afa7fc44316d1149183201461de2d8a99ca/wagtail_feeds/feeds.py#L58). Contributed to the package some years back.

### Need to modify Wagtail body content:

The full content in the description element is good for text. But images sources are relative. The HTML generated is good for rendering on our site but we have to modify it for use in feeds. Another issue is embed block, we need to remove `style="padding-bottom: 56.25%;"`. Looks we are not using images, embeds in news pages. Mentioned it here so we implement it now or later when the requirement arises.

File: opentech/public/news/views.py (hunk 2 above)
Comment (frjo) [reply opentech/public/news/views.py]: Need feature and easy to implement.

File: opentech/public/news/views.py (hunk 10 above)
Comment (frjo) [reply opentech/public/news/views.py]: Learned a new thing, thanks!

**Comment by frjo** [comment]
@Parbhat Added a cache implementation.

File: opentech/public/news/templates/news/news_index.html [hunk 11]
```diff
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% load wagtailcore_tags wagtailimages_tags static markdown_tags bleach_tags %}
+{% block feedlinks %}<link rel="alternate" type="application/rss+xml" title="RSS" href="{{ request.site.root_url }}{{ request.path }}feed/">{% endblock %}
```
Parbhat: @frjo We can use [url](https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#url) template tag to get URL. This way we can avoid hard-coded values.

frjo: A lot nicer now, thanks.

PR #1186, opened 2019-05-02
Add is archive option to terms in categories

### LINKED ISSUE #1183: Add a "archived" flag to all categories and allow it to be set via wagtail admin
Terms flagged with "archive" should:

* Show on content where they where added before being archived.
* Not be available to set on new content.
* Admins can toggle this flag on/off in wagtail admin edit view for each term.

frjo:
Fixes #1183


frjo:
@Parbhat This pr now adds a "Archived" option to terms.

My plan was to use this data to hide archived terms when users create new content but show them on both view and edit if the content already have the term selected.

If an existing content have an archived term, a user edit it, deselect the archived term and saves the archived term becomes hidden.

Tips how to make that happen or suggestion for a nicer solution?

Parbhat:
@frjo The approach sounds good. While editing we can also indicate which terms are archived and not allow the user to save without deselecting the archived terms.

opentech/apply/categories/blocks.py
```diff
@@ -58,7 +59,7 @@ def get_field_kwargs(self, struct_value):
         kwargs = super().get_field_kwargs(struct_value)
         category = struct_value['category']
         kwargs = self.use_defaults_from_category(kwargs, category)
-        choices = category.options.values_list('id', 'value')
+        choices = category.options.filter(Q(is_archived=False) | Q(id__in=[])).values_list('id', 'value')
```
frjo: @Parbhat Possible to get the current values in the form at this stage? Want to filter on "id__in" so to show only archived terms if they are set on the submission.

Parbhat: Not exactly here. This piece of code only runs while generating form fields for the form. It looks okay for new form submission but on submission edit page it will not display archived answers. The existing archived options values should display correctly in submission data until the user hit save on edit page. On the edit page, we can also add the archived choices if is in submitted data by updating [initial](https://github.com/OpenTechFund/opentech.fund/blob/c9fb73f3d79391ad2dcdb4ce780ccda336bce588/opentech/apply/funds/views.py#L654) values in BaseSubmissionEditView.

frjo: Thanks, now I know what I need to do.

PR #1188, opened 2019-05-03
Reuse parent header image if page has none.

### LINKED ISSUE #501: Can all the pages below an index page default to the index page's header (ie, people pages use the people page header)?

frjo:
Fixes #501

### DIFF
--- a/opentech/templates/base.html
+++ b/opentech/templates/base.html
@@ -70,8 +70,8 @@
         {% block header %}
             {% if page.header_image %}
                 {% image page.header_image fill-2560x410 as header_image %}
-            {% elif page.get_parent.header_image %}
-                {% image page.get_parent.header_image fill-2560x410 as header_image %}
+            {% elif page.get_parent.specific.header_image %}
+                {% image page.get_parent.specific.header_image fill-2560x410 as header_image %}
             {% endif %}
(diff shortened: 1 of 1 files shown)


opentech/templates/base.html
```diff
@@ -68,8 +68,12 @@
         {% include 'includes/messages.html' %}
 
         {% block header %}
-            {% image page.header_image fill-2560x410 as header_image %}
-            <header class="header header--standard {% if page.header_image %}header--has-bg-image{% endif %} {% block header_modifier %}{% endblock %}" {% if page.header_image %}style="background-image:url('{{ header_image.url }}')"{% endif %}>
+            {% if page.header_image %}
+                {% image page.header_image fill-2560x410 as header_image %}
+            {% elif page.get_parent.header_image %}
```
frjo: @Parbhat page.get_parent.header_image is always empty so I'm clearly doing it wrong? Tips?

Parbhat: @frjo The code is correct, we just need to add one wagtail related thing here. Using [.specific ](https://docs.wagtail.io/en/v2.5/reference/pages/model_reference.html#wagtail.core.models.Page.specific)we get page in its most specific subclassed form.

```django
{% elif page.get_parent.specific.header_image %}
    {% image page.get_parent.specific.header_image fill-2560x410 as header_image %}
{% endif %}
```

PR #1195, opened 2019-05-07
put people who don't have an order set after those who do

chris-lawton:
This was occurring because an `order` hadn't been set in the backend under Apply > Reviewer Roles > Secondary so the order value was `null`. Anyone with this value is now moved after those who have a value. Suggest we also make this field mandatory in the backend.


todd-dembrey:
I'm happy with the fix. I've also updated the live site to have a value as well. 

I've asked @frjo about how we want to handle these changes so will get back to you about merging this.

PR #1196, opened 2019-05-07
Make the comment editable through the API

todd-dembrey:
This makes the activities revision-able and adds the endpoints to work with this through the API.


zerolab:
LGTM

opentech/apply/funds/api_views.py
```diff
@@ -188,3 +191,38 @@ def perform_create(self, serializer):
...
+
+        comment_to_edit.previous = comment_to_update
+        comment_to_edit.pk = None
+        comment_to_edit.edited = timezone.now()
+
+        serializer = self.get_serializer(comment_to_edit, data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        if serializer.validated_data['message'] != comment_to_update.message:
+            self.perform_create(serializer)
```
zerolab: Without testing this locally, I am a uncertain whether this will add a messenger entry as in `CommentListCreate`

zerolab: This definitely doesn't.
We spoke offline -- was not specced. Could be beneficial, but may get noisy with small spelling edits (for example)

Issue #1130, opened 2019-03-25
batch action determination jumps to a different page


fourthletter:
After completing a batch action determination within a specific round, the page jumps to all submissions and the queryset filter no longer applies. 

After completing a batch action determination with a specific round, the page should stay within the round.

gibbsmaya:
**Acceptance criteria**

- [ ] Following batch submission, user returned to listing page they originated from including any applied filters/searches.
- [ ] QA criteria satisfied

abigail-hampson:
Looks good, this can be deployed to test

fourthletter:
Features are functional during test. Thanks! 💥

PR #1197, opened 2019-05-07
GH-1130: Redirect back to origin on batch determine

todd-dembrey:
Appends the `next` query param to help redirect the user back on a successful determination.

### DIFF
--- a/opentech/apply/determinations/views.py
+++ b/opentech/apply/determinations/views.py
@@ -1,3 +1,5 @@
+from urllib import parse
+
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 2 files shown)


opentech/apply/determinations/views.py
```diff
@@ -162,13 +165,17 @@ def should_redirect(cls, request, submissions, actions):
...
+                "&submissions=" + ','.join([str(submission.id) for submission in submissions]) +
+                "&next=" + parse.quote_plus(request.get_full_path()),
             )
         elif set(actions) != non_determine_states:
             raise ValueError('Inconsistent states provided - please talk to an admin')
 
     def get_success_url(self):
-        return reverse_lazy('apply:submissions:list')
+        try:
+            return self.request.GET['next']
```
zerolab: Do we need to worry about enforcing this to a `apply:submissions:list` base?

zerolab: other than that, :shipit:

PR #1203, opened 2019-05-09
When new partners are added to a submission, notify the applicant and…

### LINKED ISSUE #1085: Notify the Partner and Applicant when a new Partner has been added to an application
For applicants, send the name of added partners to the applicant
For partners, say they've been added a partner for an application with the name of the submission with and link

frjo:
… the added partners.

Fixes #1085


opentech/apply/funds/views.py
```diff
@@ -427,6 +427,16 @@ def form_valid(self, form):
             added=added,
             removed=removed,
         )
+
+        messenger(
+            MESSAGES.PARTNERS_UPDATED_PARTNER,
+            request=self.request,
+            user=self.request.user,
+            submission=self.kwargs['submission'],
+            added=added,
```
Parbhat: We can create a check above (if condition) if new partners are added then only call the messenger. SImilar for PARTNERS_UPDATED.

PR #1210, opened 2019-05-13
Feature/1194 comment editing

chris-lawton:


opentech/apply/funds/templates/funds/includes/activity-feed.html
```diff
@@ -15,7 +15,7 @@ <h4 class="activity-feed__heading">Activity Feed <svg class="icon icon--speech-b
 
     <div class="wrapper wrapper--medium wrapper--activity-feed js-tabs-content">
         <div class="tabs__content tabs__content--current" id="tab-1">
-            {% include "activity/include/comment_list.html" with submission_title=True %}
+            {% include "activity/include/comment_list.html" with submission_title=True editable=False %}
```
todd-dembrey: `editable=False` not necessarily required as will be falsey without this. But happy to include

opentech/static_src/src/javascript/apply/edit-comment.js
```diff
@@ -0,0 +1,131 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+                'X-CSRFToken': $.cookie('csrftoken')
+            },
+            body: JSON.stringify({
+                message: editedComment
+            })
+        }).then(response => {
+            if (response.status === 404) {
+                $(this).closest(editBlock).append('<p class="wrapper--error js-comment-error"><b>Update unsuccessful</b>. This comment has been edited elsewhere - please refresh to get the latest updates.</p>');
```
todd-dembrey: Do we want to note that they will lose unsaved changes by doing this?

opentech/static_src/src/javascript/apply/edit-comment.js
```diff
@@ -0,0 +1,131 @@
...
+                $(this).attr('disabled', true);
+                return;
+            }
+            return response.json();
```
todd-dembrey: Do we need to handle 500 or catch generally? network issues etc?

opentech/static_src/src/javascript/apply/edit-comment.js
chris-lawton: Good point, have updated.

opentech/static_src/src/javascript/apply/edit-comment.js
chris-lawton: Definitely. I've caught all other non-404 errors.

PR #1213, opened 2019-05-13
Use local jquery and normalize files. Run no-js to js class toggle as…

frjo:
… early as possible. Add empty tinymce skin.min.css.map to avoid missing file error on local dev.

Local versions of jQuery and normalise makes it easier to run on local dev and not everyone will run this system via Cloudflare.

For OTF Cloudflare will cache the files so no performance impact.

Also run the no-js to js class toggle as early as possible with inline js since this effects styling and we want to avoid and flicker effects.


frjo:
Yes, that makes more sense. Moved them now and updated lint settings to ignore them.

PR #1219, opened 2019-05-14
Submission private media shareable URL with permissions checks

### LINKED ISSUE #637: Within submissions, change the uploaded files link behavior to include a shareable link that upon clicking checks user permissions and then delivers file
Right now, it links to a non-sharable link, unique to the user that can't be shared through external mediums to users with permissions to view the file, so others get a 404 unless they go to the submission directly and then click the file attachment themselves.

### LINKED ISSUE #1108: Message "The contents of this application have been changed by someone else" constantly appears in the detailed view of a submission
The message (image below) appears constantly within the detailed view. Here is an example link to the detailed view: https://apply.opentech.fund/apply/rounds/796/?submission=3690

This message (image below) has been permanently visible in different submissions for several team members. Either 1) this message is correct and should appear or 2) this is a bug.

image

Thanks, Di

### LINKED ISSUE #1168: Route all s3 private file requests via Django, checking access privs
Currently the OTF WebApp generates download links for s3 private files with a token that is valid for 10 minuts. This does not work well. Many times it takes more than 10 minuts to read a submission so when the user gets to the bottom and tries to download the file the link is not valid anymore. It's also not possible to share the link since it times out in 10 minuts.

In Drupal we used the built in Private file storage. It route all file access via Drupal so the request can be checked for access privileges.

We need something similar for the OTF WebApp.

The goal is to have links to submission uploads (private files on s3) that checks that the currently logged in user has the right to view the submission. If yes server them the file, if not redirect to the login form.

* These links would never timeout.
* They would allow e.g. staff members to share links between them.

Related issues: #1108, #1166 and #637

Parbhat:
Closes #1168
Closes #637  
Closes #1108

The new URL for submission media can be accessed by users who have access to the submission detail view. The shareable URL will redirect to the actual file URL.

### DIFF
--- a/opentech/apply/funds/urls.py
+++ b/opentech/apply/funds/urls.py
@@ -1,4 +1,4 @@
-from django.urls import include, path
+from django.urls import include, path, re_path
 
 from .views import (
     RevisionCompareView,
@@ -12,6 +12,7 @@ from .views import (
     SubmissionOverviewView,
     SubmissionSealedView,
     SubmissionDeleteView,
+    SubmissionPrivateMediaRedirectView,
 )
 from .api_views import (
     CommentEdit,
@@ -36,6 +37,7 @@ app_name = 'funds'
 submission_urls = ([
     path('', SubmissionOverviewView.as_view(), name="overview"),
     path('all/', SubmissionListView.as_view(), name="list"),
+    re_path('documents/(.*)$', SubmissionPrivateMediaRedirectView.as_view(), name='private_media_redirect'),
     path('<int:pk>/', include([
         path('', SubmissionDetailView.as_view(), name="detail"),
         path('edit/', SubmissionEditView.as_view(), name="edit"),
--- a/opentech/apply/funds/urls.py
+++ b/opentech/apply/funds/urls.py
@@ -1,4 +1,4 @@
-from django.urls import include, path, re_path
+from django.urls import include, path
 
 from .views import (
     RevisionCompareView,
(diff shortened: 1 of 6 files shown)


opentech/apply/funds/views.py
```diff
@@ -812,3 +814,24 @@ def delete(self, request, *args, **kwargs):
...
+        except (IndexError, ValueError):
+            raise Http404
+
+        try:
+            ApplicationSubmission.objects.get(id=submission_id)
+        except ApplicationSubmission.DoesNotExist:
+            raise Http404
+
+        # If user can access submission detail view then show the media
+        can_access_detail_view = SubmissionDetailView.as_view()(self.request, pk=submission_id)
```
todd-dembrey: this is doing a lot of work to process the entire request through the view. This would also allow post requests to be processed against the detail view through the endpoint which is a bit weird.

opentech/apply/funds/views.py
```diff
@@ -812,3 +814,24 @@ def delete(self, request, *args, **kwargs):
...
+
+        submission_storage = get_storage_class(getattr(settings, 'PRIVATE_FILE_STORAGE', None))()
+        return submission_storage.url(file_name) if can_access_detail_view.status_code == 200 else None
```
todd-dembrey: shouldn't we be 403 if the user cannot access the file?

opentech/apply/stream_forms/files.py
```diff
@@ -16,11 +17,12 @@ def default(self, o):
 
 
 class StreamFieldFile(File):
-    def __init__(self, *args, filename=None, storage=default_storage, **kwargs):
+    def __init__(self, *args, filename=None, storage=default_storage, is_submission=False, **kwargs):
```
todd-dembrey: I'm not entirely happy about this. The stream_form files are meant to be agnostic to the implementation.

There is already a built in method to vary the URL as required. Provide a storage instance instead of the default and that implementation will handle varying the URL.

`submission_storage` from above.

opentech/apply/funds/urls.py
```diff
@@ -35,6 +36,7 @@
 submission_urls = ([
     path('', SubmissionOverviewView.as_view(), name="overview"),
     path('all/', SubmissionListView.as_view(), name="list"),
+    re_path('documents/(.*)$', SubmissionPrivateMediaRedirectView.as_view(), name='private_media_redirect'),
```
todd-dembrey: Can you not have a better path matching here rather than having to handling the 404 etc in the view itself?

opentech/apply/funds/views.py
Parbhat: Makes sense. The initial thought was not to copy all permission checks. PR updated. Thanks.

Parbhat:
@todd-dembrey Thanks for the review. PR updated for requested changes.

frjo:
Lower the S3 token timeout after deploy.

PR #1230, opened 2019-05-21
Allow staff users to see/set meta categories to submissions

### LINKED ISSUE #972: Create a  interface on submission page that allows a team member to see and set meta categories
Ideally, w/o leaving or having to reload the page. Labels set on 1st-stage applications should carry over to associates 2nd-state applications.

Parbhat:
Closes #972


Parbhat:
@frjo Do we need to send notifications to slack etc. when a meta-category is added to a submission?

frjo:
I do not think a Slack notice is needed.

Parbhat:
PR updated to exclude Root from the list and new meta categories section in the sidebar.

frjo:
@Parbhat 

1. Would it be possible to only allow users to set terms that have no children? Show the tree of terms as now but only show a checkbox for terms with no children.
2. How about showing each main branch (Car and Animal e.g.) in separate select fields? Save them the same way as now, just present the widget divided per main branch.

Further down OTF want this and all other meta data to be saved with an AJAX call so the page doesn't need to be reloaded on each save. This would make it quicker to set many meta data items in quick succession.

Parbhat:
@frjo 1 and 2 not look general cases, will probably need to custom rendering the form. I will check the possible options.

Parbhat:
@frjo PR updated for 1st point of the [comment](https://github.com/OpenTechFund/opentech.fund/pull/1230#issuecomment-502051890). User can only select meta categories with no children. Meta categories with children checkbox option will be disabled.

Parbhat:
@danblah @frjo Updated meta categories selection interface for submission to

Screen Shot 2019-07-15 at 8 43 48 PM

Meta categories are grouped by parent categories. Not the one requested above but looks cleaner than the earlier interface. Feedback?

frjo:
@Parbhat I like it! A clear improvement I think.

PR #1232, opened 2019-05-22
Implement 2fa with django-two-factor-auth.

### LINKED ISSUE #331: Add MFA/2FA options for all user accounts, including support for authenticator app and/or yubikey/hardware keys first

frjo:
Fixes #331


frjo:
@Parbhat Any idea why test_nonstaff_cant_post_opinion_to_review is failing?

Looked at a number of 2FA solution but django-two-factor-auth seemed best. How you tried any other solution that I should test?

The current implementation seems to do the basic of allowing users to set up 2FA and login.

Parbhat:
@frjo Looks strange why the test is failing. I am looking into it and will post if found anything.

`django-two-factor-auth` seems like the best package for implementing 2FA. We also need to enable 2FA in Wagtail admin login view. The package patched the Django admin which redirect admin URL to 2FA login URL. We can use a similar strategy for Wagtail admin as well.

frjo:
@Parbhat Found a solution to redirect wagtail admin login that seems to work. Looks ok to you?

opentech/apply/users/urls.py
```diff
@@ -11,9 +13,8 @@
 public_urlpatterns = [
     path(
         'login/',
-        auth_views.LoginView.as_view(
-            template_name='users/login.html',
-            redirect_authenticated_user=True
```
frjo: The two_factor LoginView does not support "redirect_authenticated_user" setting. Any idea how to get the functionality back?

opentech/urls.py
```diff
@@ -14,8 +16,13 @@
 from opentech.apply.users.urls import public_urlpatterns as user_urls
 
 
+def redirect_to_my_auth(request):
+    return redirect_to_login(reverse('wagtailadmin_home'), login_url='users_public:login')
+
+
 urlpatterns = [
     path('django-admin/', admin.site.urls),
+    path('admin/login/', redirect_to_my_auth, name='wagtailadmin_login'),
```
Parbhat: We can directly call the LoginView here instead of creating a new view. Also, using reverse('wagtailadmin_home') user will be redirected to Wagtail home page after login. But we should redirect the user to the requested page. Using view directly will help the request for `/admin/pages/1/` to `/admin/login/?next=/admin/pages/1/` instead of `/login/?next=/admin/`.

```python
path('admin/login/', LoginView.as_view(template_name='users/login.html'), name='wagtailadmin_login'),
```

opentech/apply/users/urls.py
Parbhat: We need to extend the Django 2fa [LoginView](https://github.com/Bouke/django-two-factor-auth/blob/87b644ce5bbd9f65959bd619b1782868f14888d3/two_factor/views/core.py#L51) for [redirecting](https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/contrib/auth/views.py#L53) authenticated users to Login Redirect URL say dashboard. 

Or we can also change template code so `My OTF` button detects the user is logged-in and take user to dashboard instead of the login page.

opentech/apply/users/templates/users/login.html
```diff
@@ -1,15 +1,65 @@
...
+                <button name="wizard_goto_step" type="submit" value="backup"
+                                class="btn btn-default btn-block">{% trans "Use Backup Token" %}</button>
+            </p>
+        {% endif %}
+
+        {% include "two_factor/_wizard_actions.html" %}
+      {% endif %}
     </form>
 
     <div class="wrapper wrapper--inner-space-large">
```
Parbhat: Login in OTF email appears in Token page as well while login. We may want to include it only in initial form as a choice between username and password. Also, need to test how it will work with new 2fa feature.

opentech/apply/users/urls.py
frjo: Implemented this now but it feels like I have duplicated a lot of code, better way do do it?

opentech/urls.py
frjo: A lot neater, fixed now.

opentech/apply/users/templates/users/login.html
frjo: Good catch, added a check to fix this now.

opentech/apply/users/urls.py
Parbhat: Looks good, some code duplication is required. We also need to add other method decorators to dispatch method as they are not carried over from parent 2fa login view.

```python
    @method_decorator(sensitive_post_parameters())
    @method_decorator(csrf_protect)
    @method_decorator(never_cache)
    def dispatch(self, request, *args, **kwargs):
    ....
    ....
```

Parbhat:
Login URL is currently working from the apply site and not from the public. Gives error 'users' is not a registered namespace. Although we are only using apply site login URL in the live site, it seems better to fix the public site's login URL. The reason for the error is this line 

https://github.com/OpenTechFund/opentech.fund/blob/7c5abf91832c83e1254193d65ba3847c64a1d2c8/opentech/apply/users/templates/users/login.html#L35

We have two URL patterns in users app, [public_urlpatterns](https://github.com/OpenTechFund/opentech.fund/blob/7c5abf91832c83e1254193d65ba3847c64a1d2c8/opentech/apply/users/urls.py#L11) and apply site [urlpatterns](https://github.com/OpenTechFund/opentech.fund/blob/7c5abf91832c83e1254193d65ba3847c64a1d2c8/opentech/apply/users/urls.py#L26). As reset password [users:password_reset](https://github.com/OpenTechFund/opentech.fund/blob/7c5abf91832c83e1254193d65ba3847c64a1d2c8/opentech/apply/users/urls.py#L46) is part of apply site [urlpatterns](https://github.com/OpenTechFund/opentech.fund/blob/7c5abf91832c83e1254193d65ba3847c64a1d2c8/opentech/apply/urls.py#L16), it is available only in apply site. We have a middleware called [apply_url_conf_middleware](https://github.com/OpenTechFund/opentech.fund/blob/7c5abf91832c83e1254193d65ba3847c64a1d2c8/opentech/apply/middleware.py#L4) which activate apply site urlpatterns when the request is coming from an apply site. The forgot password link can be public, we can move this and related URLs to [public_urlpatterns](https://github.com/OpenTechFund/opentech.fund/blob/7c5abf91832c83e1254193d65ba3847c64a1d2c8/opentech/apply/users/urls.py#L11) to fix the issue.

frjo:
Will not a line like this raise an error then?

https://github.com/OpenTechFund/opentech.fund/blob/f3d9c8f7a8c2b3dd29eec9d01cda4deeb27fd7f0/opentech/apply/users/urls.py#L59

I will add a separate issue for this to be fixed.

Parbhat:
Yeah, it will raise an error. Therefore, we need to move all related URLs as well. Creating a separate issue will be good.

Parbhat:
@frjo In case this comment not get skipped, posting it here https://github.com/OpenTechFund/opentech.fund/pull/1232#discussion_r287061168. It is about adding more method decorators to the LoginView's dispatch method.

opentech/apply/users/urls.py
frjo: Fixed now, thanks for the ping about it.

PR #1242, opened 2019-05-28
Serve private media instead of S3 redirect for a submission

Parbhat:
Discussed here https://github.com/OpenTechFund/opentech.fund/issues/1168#issuecomment-495614598


frjo:
@Parbhat Easy to get this to redirect back to the file when the user is not logged in? Now the user gets sent to the dashboard since "next" is not set.

The current implementation you built where users get redirected to S3 handels this already.

Parbhat:
@frjo The issue is only on the test site likely due to Cloudflare login. It is working fine on dev-apply site.

PR #1261, opened 2019-06-03
Set AssigneReviewers correctly on new submissions. Show review button.

frjo:


frjo:
Should add some test to make sure this does not happen again. 

Parbhat:
Discussed in Slack.

PR #1310, opened 2019-07-14
Allow author and groups with permission to edit reviews.

### LINKED ISSUE #552: Allow reviews to be edited by staff and author

frjo:
Fixes #552


frjo:
@Parbhat Yes, that would be the best. With all the moving parts I was hesitant to start messing with working functionality.

As it is I'm unsure how good the ReviewEditView class is. It seems to work but it might have parts that is not really needed or could be done better.

Parbhat:
@frjo `ReviewEditView` looks good and have all the required parts.

PR #1314, opened 2019-07-16
Update to sentry-python with Django integration.

frjo:


frjo:
@Parbhat sentry.io recommend that projects use the new sentry-python instead of raven.

The part I'm a bit unsure abut is the logging setup in base.py. The setup is straight forward in https://docs.sentry.io/platforms/python/django/ and I think that part is correct.

Logging is according to https://docs.sentry.io/platforms/python/logging/ automatically set up. But the current code seems to customise the output a bit.

frjo:
@Parbhat Thanks, makes perfect sense to move it to the production settings. Will deploy it on test for testing.

PR #1315, opened 2019-07-16
Updating to Django 2.1.x and Wagtail 2.5.x.

### LINKED ISSUE #1184: Update to Django 2.1.x and Wagtail 2.5.x
Django 2.2.0 is the latest release but all the packages we use does not yet support it so we first go to Django 2.1.x.

frjo:
Fixes #1184

### DIFF
--- a/opentech/apply/funds/admin_views.py
+++ b/opentech/apply/funds/admin_views.py
@@ -3,7 +3,7 @@ from django.shortcuts import redirect
 from django.utils.translation import ugettext as _
 
 from wagtail.admin import messages
-from wagtail.admin.forms import CopyForm
+from wagtail.admin.forms.pages import CopyForm
 from wagtail.admin.views.pages import get_valid_next_url_from_request
 from wagtail.core import hooks
 from wagtail.core.models import Page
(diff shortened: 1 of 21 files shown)


frjo:
@Parbhat I have managed to get the site working more or less but the tests generates a number of errors that I could use some help with. You find them in the CI reports above.

* TypeError: render() got an unexpected keyword argument 'renderer'
* TypeError: unhashable type: 'AssignedReviewers'
* AttributeError: 'str' object has no attribute 'step'

frjo:
@Parbhat Excellent work! Going 2.0 > 2.1 > 2.2 is a good idea.

frjo:
@Parbhat  The group names has disappeared from the wagtail user edit form. Reverting to Django 2.0.13 fixes the problem so. Wagtail version does not affect it.

I checked the wagtail user template and view overrides but found only one change and that has to do with pagination (updated that in a commit to the pr).

So it seems it's the data coming in that has changed?

Skärmavbild 2019-07-19 kl  10 55 27
(The label and the checkbox should be inline as well but that is a separate issue.)
Skärmavbild 2019-07-19 kl  10 55 37

Parbhat:
@frjo The issue is due to [this](https://docs.djangoproject.com/en/2.2/releases/2.1/#moved-wrap-label-widget-template-context-variable) backwards incompatible change in Django 2.1. The way the issue appears in group selection, it was not clear and took some hours to find the actual issue. Fixed it in the recent commit and removed wrap_label context in CategoriesWidget as it is fixed in Django 2.1.

The paginate function in wagtail.utils.pagination has been deprecated in [Wagtail 2.5](http://docs.wagtail.io/en/v2.5.1/releases/2.5.html#changes-to-admin-pagination-helpers).

frjo:
@todd-dembrey I have seen no issues with these updates. And it seems to work well on test as well.

What do you think about pushing this out this week?

My thought is that it can be good to let it run on live a bit before all your new features gets deployed.

PR #1316, opened 2019-07-18
Fix iSort

ghickman:
This applies `isort` to the codebase and checks imports are still ordered correctly in CI.


frjo:
Nice to get this cleaned up!

PR #1352, opened 2019-07-29
add show more block

chris-lawton:
- Adds 'show more' block styles and example markup

### DIFF
--- a/opentech/static_src/src/javascript/apply/toggle-proposal-info.js
+++ b/opentech/static_src/src/javascript/apply/toggle-proposal-info.js
@@ -1,0 +1,24 @@
+(function ($) {
+
+    'use strict';
+
+    function toggleProposalInfo() {
+        $('.js-toggle-propsoal-info').click(function (e) {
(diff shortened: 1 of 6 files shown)


opentech/static_src/src/javascript/apply/toggle-proposal-info.js
```diff
@@ -0,0 +1,24 @@
+(function ($) {
+
+    'use strict';
+
+    function toggleProposalInfo() {
+        $('.js-toggle-propsoal-info').click(function (e) {
+            e.preventDefault();
+            const activeClass = 'is-open';
+
+            if (this.innerHTML === 'Show more') {
```
thisoldbear: @chris-lawton _Could_ be neater like this?... but up to you, naturally (might not even work 😅 🤷‍♂) 

```
const innerHtml = this.innerHTML;
const showMore = 'Show more';
innerHtml = (innerHtml === showMore) ? 'Hide' : showMore;
```

Issue #1325, opened 2019-07-25
Application Detail Page: Create project from accepted submission


abigail-hampson:
- [x] From an accepted submission I can create a new project 
- [x] If a project exists I am unable to create another project for that submission 
- [x] Slack message is sent when a project is created 
- [x] QA is complete

abigail-hampson:
Looks good to me, can go onto test if QA is all complete

abigail-hampson:
Hello, 

This is now ready for testing against the acceptance criteria above, you can test this by going to an accepted submission and using the create a project button on the right. 
Let me know if you have any questions. 

Thanks,
Abigail

fourthletter:
Acceptance criteria were visible and functional during test. Thanks! 🔥

What does the last bit "QA is complete" mean?

todd-dembrey:
@fourthletter That is our internal QA checks before it gets to you (we've written tests, reviewed it, tested it on dev etc)

saoun:
Have an error on my end, but not sure what's going on. My workflow:
- I go to a submission
- Create a project
- Page is now updated with the project linked to that submission, on the right handside of the page
- I click on the project link, and I get an server error

The server error comes up for some, and not for others. Is that expected behavior at this point of testing?

fourthletter:
@saoun Torchbox proposed a fix to the issue you're referring to here #1327 and here #1403

saoun:
@abigail-hampson is this done? can we close?

saoun:
@abigail-hampson what's the status on this ?

todd-dembrey:
The work referenced by this ticket has been merged and deployed.

The issue you mentioned is likely the issue fixed in #1540 due to its inconsistent nature.

Please open a new bug report if this is still an issue.

PR #1353, opened 2019-07-29
Create Project from Submission

ghickman:
This adds the ability to create a Project from a submission which doesn't already have a Project attached.  When a Project is created Slack is notified.

Ref #1325


opentech/apply/projects/forms.py
```diff
@@ -0,0 +1,11 @@
+from django import forms
+
+
+class CreateProjectForm(forms.Form):
+    submission = forms.CharField(widget=forms.HiddenInput())
```
todd-dembrey: Should this be ModelChoice field so that we can skip all the checks in the view?
```
forms.ModelChoiceField(
    queryset=Submissions.objects.filter(project__isnull=True),
    widget=forms.HiddenInput(),
)
```

opentech/apply/funds/views.py
```diff
@@ -345,6 +347,35 @@ def form_valid(self, form):
...
+
+        messenger(
+            MESSAGES.CREATED_PROJECT,
+            request=self.request,
+            user=self.request.user,
+            submission=submission,
+            project=project,
+        )
+
+        return redirect('apply:submissions:detail', submission.id)
```
todd-dembrey: The following would work as well i think. 
```
return redirect(self.get_success_url())
```

opentech/apply/funds/views.py
```diff
@@ -345,6 +347,35 @@ def form_valid(self, form):
...
+    model = Project
+
+    def form_valid(self, form):
+        submission_id = form.cleaned_data['submission']
+        try:
+            submission = ApplicationSubmission.objects.get(pk=submission_id)
+        except ApplicationSubmission.DoesNotExist:
+            return self.form_invalid(form)
+
+        project, created = Project.create_from_submission(submission)
```
todd-dembrey: Should this be moved onto the form save method? then we do not need to customise this view as much.

opentech/apply/funds/views.py
```diff
@@ -345,6 +347,35 @@ def form_valid(self, form):
...
+class CreateProjectView(DelegatedViewMixin, CreateView):
+    context_name = 'project_form'
+    form_class = CreateProjectForm
+    model = Project
+
+    def form_valid(self, form):
+        submission_id = form.cleaned_data['submission']
+        try:
+            submission = ApplicationSubmission.objects.get(pk=submission_id)
+        except ApplicationSubmission.DoesNotExist:
```
todd-dembrey: See comment on `projects/forms.py` I think we can rely on the form  to validate without throwing invaid ourselves

opentech/apply/funds/views.py
ghickman: Fixed in 75b0a597

opentech/apply/projects/forms.py
ghickman: Fixed in 379d6b5d

opentech/apply/funds/views.py
ghickman: Fixed in 6c7dd74b

opentech/apply/funds/views.py
ghickman: Fixed in 379d6b5d

Issue #1327, opened 2019-07-25
Project detail view: Create project page 


abigail-hampson:
- [x] On the application submission there is a link to this page within a new 'project' side bar block at the top of the list 
- [x] Side bar block is only visible when there is a project associated with the submission 
- [x] Create skeleton page with key information (information copied from application submission, dummy page layout content) 
- [x] On successful creation I am redirected to the new project page
- [x] QA complete

abigail-hampson:
Looks good to me

abigail-hampson:
Hello, 

This is now ready to be tested against the acceptance criteria above. If you view a submission page with a project page you should be able to view the links in the side bar, and the skeleton project page is created. 

Thanks, 
Abigail

fourthletter:
The last two acceptance criteria failed during test: 1) Create skeleton page with key information and 2) On successful creation I am redirected to the new project page

1) Create skeleton page with key information
Test worked for this test submission: https://test-apply.opentech.fund/apply/submissions/5/
Test failed for these 2 test submission: https://test-apply.opentech.fund/apply/submissions/26/ AND https://test-apply.opentech.fund/apply/projects/5/

2) On successful creation I am redirected to the new project page
After creating a new project, I was not redirected to the new project page for any of the new projects I created. 

Thanks,
Di

todd-dembrey:
@fourthletter Thanks DI. Turns out there was an issue with projects of labs. I've made a fix here: #1403 and will redeploy soon

abigail-hampson:
Looks good, works with labs for me

fourthletter:
It brings me no joy to inform you that two (2) projects are created after a proposal has been 'accepted.'

Example1 from IFF: https://test-apply.opentech.fund/apply/submissions/79/
Example2 from Lab: https://test-apply.opentech.fund/apply/submissions/84/

Image: 
<img width="366" alt="Screen Shot 2019-09-04 at 6 56 53 AM" src="https://user-images.githubusercontent.com/20019656/64249416-90929600-cee1-11e9-9bbe-9d0302260e5c.png">

saoun:
@todd-dembrey is this done? can we close?

saoun:
@abigail-hampson updates? can we close?

peteandrew:
I've looked in to this and I wasn't able to replicate the reported bug. In a local environment, when a submission is accepted, only one project is created. In the test environment I couldn't find any submissions with multiple projects. I think this is ready to be closed.

PR #1354, opened 2019-07-29
Add Project Page Skeleton

ghickman:
This adds the Project page skeleton, redirects to this page when a Project is created from a submission, and adds the Project block to Submission detail pages.

Ref #1327


opentech/apply/funds/views.py
```diff
@@ -373,7 +373,7 @@ def form_valid(self, form):
             project=project,
         )
 
-        return redirect('apply:submissions:detail', submission.id)
+        return redirect('apply:projects:detail', project.id)
```
todd-dembrey: Could update the `get_success_url` method, but not too fussed. Depends on the outcome of any changes to #1353

opentech/apply/projects/views.py
```diff
@@ -0,0 +1,7 @@
+from django.views.generic import DetailView
+
+from .models import Project
+
+
+class ProjectDetailView(DetailView):
```
todd-dembrey: We'll need to `login_required` this view at some point. Can you make this:
`@method_decorator(staff_required, name='dispatch')`
and we'll handle the user stuff at a later date?

opentech/apply/funds/views.py
ghickman: Fixed in 6c7dd74b, which adds `Project.get_absolute_url()` so that now redirects to the Project detail page.

opentech/apply/projects/views.py
ghickman: Fixed in 9ee19678

Issue #1326, opened 2019-07-25
Application Detail Page: Copy relevant submission fields from accepted submission


abigail-hampson:
- [x] On creation of project the following fields are copied from the submission:
Title 
Email  
Address 
Value 
Full Name 
- [x] QA complete

abigail-hampson:
Looks good

abigail-hampson:
Hello, 

This is now ready for testing against the acceptance criteria, if you have created a project the above fields should be automatically populated if they were present on the submission.

Thanks, 
Abigail

fourthletter:
Acceptance criteria were visible and functional during test. Thanks!🔥

PR #1358, opened 2019-07-30
Copy Submission fields on Project Creation

ghickman:
This adds the relevant fields from a Submission when creating a Project.

Ref #1326


opentech/apply/projects/models.py
```diff
@@ -26,6 +31,10 @@ def create_from_submission(cls, submission):
         project = Project.objects.create(
             submission=submission,
             name=submission.title,
+            contact_email=submission.user.email,
+            contact_legal_name=submission.user.full_name,
+            contact_address=submission.form_data.get('address', ''),
+            value=submission.form_data.get('value', ''),
```
todd-dembrey: so... there is some magic here: https://github.com/OpenTechFund/opentech.fund/blob/767de79cc46550939055576e7c554054e1d3eabc/opentech/apply/funds/models/submissions.py#L336-L358
Meaning we can do `submission.address` and `submission.value`. 

It is essentially doing what you are doing, but hiding the underlying implementation
https://github.com/OpenTechFund/opentech.fund/blob/767de79cc46550939055576e7c554054e1d3eabc/opentech/apply/funds/models/submissions.py#L719-L720

opentech/apply/projects/models.py
```diff
@@ -6,6 +6,11 @@ class Project(models.Model):
 
     name = models.TextField()
 
+    contact_legal_name = models.TextField(default='')
+    contact_email = models.TextField(default='')
+    contact_address = models.TextField(default='')
```
todd-dembrey: Address field doesn't have a database representation, only the form field. I think this will be fine.
https://github.com/OpenTechFund/opentech.fund/blob/master/addressfield/fields.py

opentech/apply/projects/models.py
```diff
@@ -6,6 +6,11 @@ class Project(models.Model):
...
+    value = models.TextField(default='')
```
todd-dembrey: Value is an integer in the submission stage, but we'll probably want to change that. We should go decimal here because money.

opentech/apply/projects/models.py
ghickman: When those fields aren't in the `form_data` column we get a `KeyError`, so rather than wrap them in try/excepts I accessed the field directly.  Is there a better way to handle that error path?

opentech/apply/projects/models.py
ghickman: That's a Form Field which outputs JSON, should I switch to a JSON column instead?

opentech/apply/projects/models.py
ghickman: Fixed and rebased to avoid extra migrations.

opentech/apply/projects/models.py
todd-dembrey: They are required on the submission so you can to rely on them here if you want to. I would rather it failed a bit nosily than silently in the previous use case, but It doesn't make much difference in this instance I don't think. (it would have already complained trying to progress the application)

I just wanted to highlight the implementation for you incase you missed it.

opentech/apply/projects/models.py
todd-dembrey: JSON does serialise to and from textfield iirc, so ok for our use case?

I don't think we need the power of a JSON column but it wouldn't harm apart from portability (which we've not really focused on).

opentech/apply/projects/templates/application_projects/project_detail.html
```diff
@@ -0,0 +1,154 @@
...
+                    </div>
+
+                    <div>
+                        <h5>Project Proposed end date</h5>
+                        <p>25 January 2020</p>
+                    </div>
+
+                    <div>
+                        <h5>Legal name</h5>
+                        <p>The Company</p>
```
todd-dembrey: Can you hook this up to the model please.

opentech/apply/projects/models.py
ghickman: Ah, I misunderstood your initial comment. I'll leave it as a textfield then!

opentech/apply/projects/templates/application_projects/project_detail.html
ghickman: Fixed in b6a88dfb

Issue #1343, opened 2019-07-25
Application Detail Page: Create project when accepted determination sent


abigail-hampson:
- [ ] When an accepted determination is sent a project is created 
- [x] QA is complete

ghickman:
Should we redirect to the Project detail page as part of this?

abigail-hampson:
Looks good to me!

abigail-hampson:
Is this on test?

abigail-hampson:
Hello, this is ready to be tested on dev against the acceptance criteria. This can be tested by setting an accepted determination and the project page being created from this

fourthletter:
Acceptance criteria were visible and functional during test. Test was smooth as silk. Thanks!

PR #1360, opened 2019-07-31
Create Project on Accepted Determination

ghickman:
This creates a Project when an accepted Determination is applied to a submission.

I took the opportunity to refactor the `.form_valid()` method to add a fast failure condition to make it clearer that draft submissions are ignored.

Ref #1343

### DIFF
--- a/opentech/apply/determinations/tests/test_views.py
+++ b/opentech/apply/determinations/tests/test_views.py
@@ -126,6 +126,143 @@ class DeterminationFormTestCase(BaseViewTestCase):
         self.assertEqual(submission_original.status, 'invited_to_proposal')
         self.assertEqual(submission_next.status, 'draft_proposal')
 
+    def test_first_stage_accepted_determination_does_not_create_project(self):
+        submission = ApplicationSubmissionFactory(
+            status='concept_review_discussion',
+            workflow_stages=2,
+            lead=self.user,
+        )
+
+        self.post_page(submission, {
+            'data': 'value',
+            'outcome': ACCEPTED,
+            'message': 'You are invited to submit a proposal',
+        }, 'form')
+
+        # Cant use refresh from DB with FSM
+        submission_original = self.refresh(submission)
+        submission_next = submission_original.next
+
+        self.assertIsNotNone(submission_next)
+
+        # Confirm a Project was not created for either submission since it's
+        # not in the final stage of its workflow.
+        self.assertFalse(hasattr(submission_original, 'project'))
+        self.assertFalse(hasattr(submission_next, 'project'))
+
+    def test_first_stage_rejected_determination_does_not_create_project(self):
+        submission = ApplicationSubmissionFactory(
+            status='concept_review_discussion',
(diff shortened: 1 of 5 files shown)


opentech/apply/determinations/views.py
```diff
@@ -240,27 +241,37 @@ def get_success_url(self):
...
 
-            self.submission.perform_transition(transition, self.request.user, request=self.request, notify=False)
+        self.submission.perform_transition(
+            transition,
+            self.request.user,
+            request=self.request,
+            notify=False,
+        )
+
+        if self.object.outcome == ACCEPTED:
```
todd-dembrey: I've realised I missed an acceptance criteria. This should only happen when Accepted and there isnt another stage in the application workflow.

I don't know if we have a method to check that yet, you might want to do some digging. But pseudo code:
```python
 submission.workflow.stages.index(submission.stage) == len(submission.workflow.stages)
```
/musing
Maybe something to add to the workflow class

ghickman: Fixed in 8baa461f

ghickman:
@todd-dembrey – added some tests to check determinations create a Project under the correct criteria and added the check you requested.

opentech/apply/determinations/tests/test_views.py
```diff
@@ -126,6 +126,44 @@ def test_can_progress_stage_via_determination(self):
...
+
+        # Cant use refresh from DB with FSM
+        submission_original = self.refresh(submission)
+        submission_next = submission_original.next
+
+        self.assertTrue(hasattr(submission_next, 'project'))
+
+    def test_non_accepted_determination_does_not_create_project(self):
+        submission = ApplicationSubmissionFactory(
+            status='concept_review_discussion',
```
todd-dembrey: this should be `proposal_review_discussion` the `concept_review_discussion` would fall foul of the `is_final_stage` check, which i cant see being tested anywhere?

ghickman: What's the difference between these two statuses?

todd-dembrey: stage 1: `concept_review_discussion` -> `is_final_stage` = False
stage 2: `proposal_review_discussion`-> `is_final_stage` = True

opentech/apply/determinations/views.py
```diff
@@ -260,7 +258,23 @@ def form_valid(self, form):
...
 
-            self.submission.perform_transition(transition, self.request.user, request=self.request, notify=False)
+            self.submission.perform_transition(
+                transition,
+                self.request.user,
+                request=self.request,
+                notify=False,
+            )
+
+            if self.object.outcome == ACCEPTED and self.submission.in_final_stage:
```
todd-dembrey: We should check if its in the final stage before peforming the transition.

opentech/apply/determinations/tests/test_views.py
ghickman: Fixed and rebased.

Issue #1328, opened 2019-07-25
Project Detail View: Assign Lead 


abigail-hampson:
**NB: updated Program Manager to Lead**

- [x] Lead can be assigned 
- [x] Lead is limited to OTF staff
- [x] Reusing the modal component 
- [x] Slack message is sent to OTF staff member to tell them they have been assigned to the project 
- [x] Projects can be reassigned 
- [x] Lead can't be unassigned 
- [x] Can only have one Lead at a time 
- [x] Lead is listed below the title in the same style as an application submission lead 
- [x] QA is complete

ghickman:
Should a Lead be assigned when we create a Project?

abigail-hampson:
I think it shouldn't have to be, it can be or it can be added later. Todd do you agree?

abigail-hampson:
Still says program manager on the main project page other than that looks good

abigail-hampson:
This is still showing Program Manager for me - https://dev-apply.opentech.fund/apply/projects/2/

abigail-hampson:
Looks good, ready to go to test

abigail-hampson:
This can now be tested against the acceptance criteria above, this can be tested by assigning the project lead on a project page

frjo:
@abigail-hampson When I Try to set lead I get an 500 error.

fourthletter:
All acceptance criteria were visible and functional during test. The test was as smooth as nutella. Thanks!

frjo:
@fourthletter You did not the an error when assigning a lead?

fourthletter:
@todd-dembrey @frjo 

I found the application that receives errors: https://test-apply.opentech.fund/apply/projects/4/
[ ] Go to submission
[ ] Click on 'Lead'
[ ] See error (image below)

![image](https://user-images.githubusercontent.com/20019656/64734330-6aa65c00-d4b4-11e9-853b-22acda82bfcb.png)

danblah:
Tested, works but I'm not sure about the "Lead can't be unassigned" test. 🤷‍♂
Otherwise, all good.

PR #1362, opened 2019-08-01
Assign Project Lead

ghickman:
This adds a `lead` field to the Project model and lets an admin change the Lead via a modal on the detail view.

Ref #1328


opentech/apply/activity/messaging.py
```diff
@@ -368,6 +369,7 @@ class SlackAdapter(AdapterBase):
         MESSAGES.DELETE_SUBMISSION: '{user} has deleted {submission.title}',
         MESSAGES.DELETE_REVIEW: '{user} has deleted {review.author} review for <{link}|{submission.title}>.',
         MESSAGES.CREATED_PROJECT: '{user} has created a Project: <{link}|{project.name}>.',
+        MESSAGES.UPDATE_PROJECT_LEAD: 'The lead of <{link}|{project.name}> has been updated from {old_lead} to {project.lead} by {user}',
```
todd-dembrey: Probably worth differentiating being submission and project in the notifications text for clarity.

`The lead of project <{link}|{project.name}>`?

opentech/apply/projects/views.py
```diff
@@ -1,11 +1,66 @@
...
+class ReviewerProjectDetailView(DetailView):
+    model = Project
+
+
+class ProjectDetailView(ViewDispatcher):
+    admin_view = AdminProjectDetailView
+    applicant_view = ApplicantProjectDetailView
+    community_view = CommunityProjectDetailView
+    partner_view = PartnerProjectDetailView
+    reviewer_view = ReviewerProjectDetailView
```
todd-dembrey: I dont think these make sense as users that should be accessing the page.

At the moment it is just the admin view and applicant. Lets keep it that way for now.
```
    community_view = CommunityProjectDetailView
    partner_view = PartnerProjectDetailView
    reviewer_view = ReviewerProjectDetailView
```

opentech/apply/projects/templates/application_projects/project_admin_detail.html
```diff
@@ -0,0 +1,60 @@
+{% extends "application_projects/project_detail.html" %}
```
todd-dembrey: I just noticed the name, `application_projects` not sure that the `application` part is truly necessary.

opentech/apply/projects/forms.py
```diff
@@ -20,3 +22,20 @@ def __init__(self, instance=None, user=None, *args, **kwargs):
...
+        model = Project
+
+    def __init__(self, user=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        lead_field = self.fields['lead']
+        lead_field.label = f'Update lead from {self.instance.lead} to'
+
+        qwargs = Q(groups__name=STAFF_GROUP_NAME) | Q(is_superuser=True)
+        lead_field.queryset = (lead_field.queryset.exclude(pk=self.instance.lead_id)
```
todd-dembrey: The underlying DB field is Nullable - do you need to tweak the settings on the form field to prevent empty response or am i over thinking it?

opentech/apply/projects/templates/application_projects/project_admin_detail.html
ghickman: The CBVs use this app name when they build their `template_name` value.  We could override it but I'm not sure what it would gain us?

opentech/apply/activity/messaging.py
ghickman: Fixed and rebased.

opentech/apply/projects/views.py

opentech/apply/projects/forms.py
ghickman: `lead_field.queryset` is a QS of User so this should be fine.  If `self.instance.lead_id` is `None` then we're just excluding User's with `pk=None`.

Issue #1330, opened 2019-07-25
PAF Page: Make PAF Page 


abigail-hampson:
Contracting fields on project 

-  [ ] If an approval form has not been started, there is a link to create an approval form from the project page 
- [ ] If an application form exists there is the ability to edit this 
- [ ] Application form opens in a new page 
- [ ] Application form has the following fields: 
project title 
personal/organisational name
email
phone 
address
project value 
proposed start 
proposed end 

 - [ ] From the application form page there is the ability to save and return to the project page 
- [ ] Project page information updated based on content of approval form 
- [x] QA is complete

abigail-hampson:
Looks good to me, ready to go to test

abigail-hampson:
Hello, this can now be tested on test. You can test this by going onto a project and creating and editing and approval form

fourthletter:
Acceptance criteria were visible and functional during test. Test was as smooth as butter. Thanks!

PR #1365, opened 2019-08-02
Add PAF Page

ghickman:
This adds the Project Application Form edit page, updates the Project with changes from it, implements the show more toggle, and toggles the checkmark in the Support Docs.

Ref #1330


opentech/apply/projects/views.py
```diff
@@ -49,3 +50,20 @@ class ApplicantProjectDetailView(DetailView):
...
     applicant_view = ApplicantProjectDetailView
+
+
+@method_decorator(staff_required, name='dispatch')
+class ProjectEditView(UpdateView):
+    form_class = ProjectEditForm
+    model = Project
+
+    def form_valid(self, form):
+        self.object = form.save(commit=False)
```
todd-dembrey: I like to do the following to stay out the way of inheritance.
```python      
if form.has_changed():
    form.instance.user_has_updated_details = True
return super().form_valid(form)
````
I honestly dont know which is the "correct" approach, but personally dislike re-writing the `form_valid` method this way. 

Alternatively should this be on the form save method?

File: opentech/apply/projects/templates/application_projects/project_detail.html [hunk 2]
```diff
@@ -101,10 +101,19 @@ <h5>Email</h5>
                         -
                         {% endif %}
                     </div>
+                </div>
+
+                <a class="link--reveal-proposal js-toggle-propsoal-info" href="#">Show more</a>
```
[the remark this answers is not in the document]
ghickman: Fixed in 63bd1de8

opentech/apply/projects/views.py
```diff
@@ -49,3 +50,20 @@ class ApplicantProjectDetailView(DetailView):
...
+
+        if form.has_changed():
+            self.object.user_has_updated_details = True
+
+        print(self.object.user_has_updated_details)
```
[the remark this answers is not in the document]

opentech/apply/projects/views.py

opentech/apply/projects/forms.py
```diff
@@ -24,6 +25,35 @@ def save(self, *args, **kwargs):
...
+    class Meta:
+        fields = [
+            'title',
+            'contact_legal_name',
+            'contact_email',
+            'contact_address',
+            'contact_phone',
+            'value',
+            'proposed_start',
+            'proposed_end',
```
todd-dembrey: Would it be worth some logic to confirm end > start?

opentech/apply/projects/models.py
```diff
@@ -11,10 +12,16 @@ class Project(models.Model):
...
     title = models.TextField()
 
-    contact_legal_name = models.TextField(default='')
-    contact_email = models.TextField(default='')
-    contact_address = models.TextField(default='')
+    contact_legal_name = models.TextField(_('Person or Organisation name'), default='')
+    contact_email = models.TextField(_('Email'), default='')
+    contact_address = models.TextField(_('Address'), default='')
+    contact_phone = models.TextField(_('Phone'), default='')
     value = models.DecimalField(default=0, max_digits=10, decimal_places=2)
```
todd-dembrey: Do we want a min number validator on this? Should have been on the model definition PR but was thinking about it when reviewing the form.

`validators=[MinValueValidator(Decimal('0.01'))`

https://docs.djangoproject.com/en/dev/ref/validators/#minvaluevalidator

ghickman: Fixed in dc518330

opentech/apply/projects/forms.py
ghickman: Fixed in ad8154a

PR #1366, opened 2019-08-05
Feature/gh1329 link project to activity

todd-dembrey:
* Convert the foreign key (`submission`) to  submission model to a generic foreign key (`source`) on:
    * Activity
    * Event 
* Update reference to `submission/s` to `source/s` in the messaging template where appropriate
* Update the query reference for projects and submissions.

There will be a part two where the project activities are integrated


opentech/apply/funds/models/submissions.py
```diff
@@ -396,6 +398,12 @@ class ApplicationSubmission(
         related_name='submissions',
         blank=True,
     )
+    activities = GenericRelation(
+        Activity,
```
ghickman: If you wrap this in a string it gives the reader a little more insight into which Activity model you're talking about since you need the app name if it's not in the same app.  Slight improvement: it also removes the requirement to order model definitions based on dependent models using them.

I won't gate the review on it though, your choice.

opentech/apply/projects/models.py
```diff
@@ -14,6 +17,12 @@ class Project(models.Model):
     contact_address = models.TextField(default='')
     value = models.DecimalField(default=0, max_digits=10, decimal_places=2)
 
+    activities = GenericRelation(
+        Activity,
```
ghickman: Same thing here.

opentech/apply/activity/migrations/0029_migrate_old_submission_relation.py
```diff
@@ -6,12 +6,13 @@
 
 def submission_to_source(apps, schema_editor):
     Activity = apps.get_model('activity', 'Activity')
-    ContentType = apps.get_model('contenttypes', 'ContentType')
-    content_type = ContentType.objects.get(model='applicationsubmission', app_label='funds')
-    Activity.objects.update(
-        source_object_id=F('submission_id'),
-        source_content_type=content_type,
-    )
+    if Activity.objects.exists():
```
ghickman: Consider flipping this to a negative to fail fast:
```
if not Activity.objects.exists():
    return
```

It lowers the cognitive burden on the reader when considering the outcome of the conditional (a reader only has to look at one line vs 5).  You can rely on Activity objects existing after that check.

Alternatively drop it entirely, it only saves two database calls, one of which will a near no-op when there are no Activity rows.

opentech/apply/activity/migrations/0032_migrate_submission_to_generic_event.py
```diff
@@ -6,17 +6,18 @@
 
 def submission_to_source(apps, schema_editor):
     Event = apps.get_model('activity', 'Event')
-    ContentType = apps.get_model('contenttypes', 'ContentType')
-    content_type = ContentType.objects.get(model='applicationsubmission', app_label='funds')
-    Event.objects.update(
-        object_id=F('submission_id'),
-        content_type=content_type,
-    )
+    if Event.objects.exists():
```
ghickman: Same thing here.

PR #1367, opened 2019-08-06
Feature/gh1329 add projects activity

todd-dembrey:
* tie in the new activities
* add tests for the new activities


opentech/apply/activity/messaging.py
```diff
@@ -526,9 +529,10 @@ def batch_notify_reviewers(self, sources, links, **kwargs):
         )
 
     def slack_id(self, user):
-        if user.slack:
-            return f'<{user.slack}>'
-        return ''
+        if not user or not user.slack:
```
ghickman: Combine negative checks like so: `if not (user or user.slack):`.  It's much easier for a reader to consider `A or B` then apply the negative to it.

ghickman: This is [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) if you want to read up the reasoning behind it!

Now I think about this more I think this could be more readable as two checks returning the same thing but make the first one `if user is not None` so it's more explicit what you're confirming.

PR #1403, opened 2019-08-12
Fix sentry:OTF-WEBAPP-TEST-2G - projects of labs not rendering

todd-dembrey:
Fixes a bug where the when there is no round on the submission (e.g. labs) the page fails to render. 

Also tests were not fully rendering the page, so rewrote to use the full request in order to get a working render. (could have done this on the class, but moved to be inline with other tests.

### DIFF
--- a/opentech/apply/projects/tests/test_views.py
+++ b/opentech/apply/projects/tests/test_views.py
@@ -1,8 +1,6 @@
 from io import BytesIO
 
-from django.core.exceptions import PermissionDenied
-from django.test import RequestFactory, TestCase
-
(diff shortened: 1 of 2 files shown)


opentech/apply/projects/tests/test_views.py
```diff
@@ -1,8 +1,8 @@
 from io import BytesIO
 
-from django.core.exceptions import PermissionDenied
-from django.test import RequestFactory, TestCase
-
+from opentech.apply.funds.tests.factories import (
```
ghickman: Did you mean to wrap this single line import?

PR #1406, opened 2019-08-13
Add Debug Toolbar to Dev

ghickman:
This sets up development dependencies so we're not installing testing libraries into production environments and installs the Django Debug Toolbar in development, configuring it so it's always running but with its panels disabled.  I've disabled the panels to cope with the prohibitive performance penalty some of the panels bring, but they can be toggled on in the UI easily.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 -r requirements.txt
 
+django-debug-toolbar==2.0
 factory_boy==2.9.2
 Faker==1.0.8  # Pinning to avoid API changes to pydecimal in 0.x - could go higher in future
 flake8==3.7.9
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,9 @@
+-r requirements.txt
+
+factory_boy==2.9.2
+Faker==1.0.8  # Pinning to avoid API changes to pydecimal in 0.x - could go higher in future
+flake8==3.7.9
+responses==0.10.6
+stellar==0.4.5
+wagtail-factories==1.1.0
+Werkzeug==0.14.1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
 
       - restore_cache:
           keys:
-            - v2-python-{{ .Branch }}-{{ checksum "requirements.txt" }}
+            - v2-python-{{ .Branch }}-{{ checksum "requirements-dev.txt" }}
             - v2-python-{{ .Branch }}-
             - v2-python-
 
@@ -46,7 +46,7 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             pip install coverage codecov
-            pip install -r requirements.txt
+            pip install -r requirements-dev.txt
 
       - run:
           name: install node dependencies
@@ -57,7 +57,7 @@ jobs:
       - save_cache:
           paths:
             - ./venv
-          key: v2-python-{{ .Branch }}-{{ checksum "requirements.txt" }}
+          key: v2-python-{{ .Branch }}-{{ checksum "requirements-dev.txt" }}
 
       - save_cache:
           paths:
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install:
   - pip install codecov
 
   # Install project dependencies
-  - pip install -r requirements.txt
+  - pip install -r requirements-dev.txt
 
   # Install node
   - nvm install 10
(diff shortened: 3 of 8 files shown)


opentech/settings/base.py
```diff
@@ -674,3 +674,35 @@
...
+# Debug Toolbar
+# https://django-debug-toolbar.readthedocs.io/en/latest/index.html
+if DEBUG:
+    INSTALLED_APPS.append('debug_toolbar')
+    MIDDLEWARE = [
+        'debug_toolbar.middleware.DebugToolbarMiddleware',
+        *MIDDLEWARE,
+    ]
+
+INTERNAL_IPS = ['127.0.0.1', '10.0.2.2']
```
frjo: Should this not be moved in to "if DEBUG:"?

And setting an arbitrary privat ip as 10.0.2.2 on production feels wrong. It's specifically for the VirtualBox environment I believe.

opentech/settings/base.py
```diff
@@ -674,3 +674,35 @@
...
     PROJECTS_ENABLED = True
+
+
+# Debug Toolbar
+# https://django-debug-toolbar.readthedocs.io/en/latest/index.html
+if DEBUG:
+    INSTALLED_APPS.append('debug_toolbar')
+    MIDDLEWARE = [
+        'debug_toolbar.middleware.DebugToolbarMiddleware',
+        *MIDDLEWARE,
```
frjo: I notice you changed the pattern here. Just a preference or does it have benefits? Still learning Python so curious.

opentech/settings/base.py
ghickman: That's a [private address](https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses) but you're right I'll put it under `DEBUG` just in case.

opentech/settings/base.py
ghickman: It avoids mutating the `MIDDLEWARE` setting.  Mutation is typically considered non-ideal because it often has side effects.

I've just noticed I didn't used it for `INSTALLED_APPS`, I'll update that line to have the same pattern.

ghickman:
Updated as per comments and rebased into existing changes.

opentech/settings/base.py
```diff
@@ -674,3 +674,38 @@
 PROJECTS_ENABLED = False
 if env.get('PROJECTS_ENABLED', 'false').lower().strip() == 'true':
     PROJECTS_ENABLED = True
+
+
+# Debug Toolbar
+# https://django-debug-toolbar.readthedocs.io/en/latest/index.html
+if DEBUG:
```
todd-dembrey: Isn't `DEBUG` always `False` at this point because of the definition at the top of the file, or am i missing something? Where as in `settings/dev.py` it is always `True`

Wouldn't it make sense to keep it int he dev settings and then there is no need to the debug check at all?

Also worth noting the vagrant box sets the default settings like so: 
https://github.com/OpenTechFund/opentech.fund/blob/3d2e02a16e38894282ebf2f5f79eab405a61c784/vagrant/provision.sh#L49-L52
Which is the preferred entrypoint for dev.

opentech/settings/base.py
```diff
@@ -674,3 +674,38 @@
...
+    INSTALLED_APPS = [
+        *INSTALLED_APPS,
+        'debug_toolbar',
+    ]
+    MIDDLEWARE = [
+        'debug_toolbar.middleware.DebugToolbarMiddleware',
+        *MIDDLEWARE,
```
todd-dembrey: :+1:

opentech/settings/base.py
```diff
@@ -674,3 +674,38 @@
...
+    ]
+    INTERNAL_IPS = ['127.0.0.1', '10.0.2.2']
```
todd-dembrey: Not strictly sure this is necessary anymore as debug toolbar would handle all that internal ips does.

requirements-dev.txt
```diff
@@ -0,0 +1,9 @@
+-r requirements.txt
```
todd-dembrey: :+1:

opentech/settings/base.py
ghickman: 🤦‍♂️

Too used to single settings files!

opentech/settings/base.py
ghickman: `INTERNAL_IPS` is a [debug toolbar setting](https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#configuring-internal-ips) 🙂

todd-dembrey: Ah sweet, did not know it was used there as well! Assumed it was from django core.

Can we add a comment about `10.0.2.2` being for the vagrant box then.

ghickman: Yea, it's a poorly named setting!

That's a good idea, if there's a more in-depth docs page I'll link to that too.

PR #1416, opened 2019-08-15
Sort the transitions by the order they are listed.

### LINKED ISSUE #1407: Allow ordering of statuses in the update status form on submissions
Now the statuses are sorted be the machine name (key) of the status, e.g "post_review_discussion". This looks very arbitrary to a human.

The overall goal is to have the first option be the most common one and the rest to be in some logical order.

Would it be possible to get the status in the same order they are set in "transitions" for each status? Then we could manually order them there to have them in the most social order, I think they already are in decent shape for that.

https://github.com/OpenTechFund/opentech.fund/blob/3d2e02a16e38894282ebf2f5f79eab405a61c784/opentech/apply/funds/workflow.py#L194-L200

The default now is almost never the right one now so if staff is a bit too quick the wrong status is often set.

frjo:
Fixes #1407

### DIFF
--- a/opentech/apply/funds/forms.py
+++ b/opentech/apply/funds/forms.py
@@ -27,6 +27,9 @@ class ProgressSubmissionForm(forms.ModelForm):
         self.user = kwargs.pop('user')
         super().__init__(*args, **kwargs)
         choices = list(self.instance.get_actions_for_user(self.user))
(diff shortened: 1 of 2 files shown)


frjo:
@Parbhat Your tips on #1407 got me going and I came up with this. It seems to work but is it a good solution?

PR #1417, opened 2019-08-15
Bugfix/ensure uploaded files are private

todd-dembrey:
This reworks the changes to the private media sharing to be more similar to standard files.

I'm not 100% about the override of the url field on the stream field file, might look at moving to a reverse in locations where the file is used, but it seems to work well for now.

Also updates the project documents to use the private storage and a view to serve them.


opentech/settings/test.py
```diff
@@ -8,6 +8,8 @@
 
 SECRET_KEY = 'NOT A SECRET'
 
+PROJECTS_ENABLED = True
```
ghickman: You shouldn't need this, the Travis and Circle configs both add it as an environment variable.

PR #1421, opened 2019-08-19
Feature/performance audit

todd-dembrey:
Pages with activities and submissions had slowed down (~1000 queries and we were hitting the PSQL page buffer size :boom: )

Turns out i was a little overzealous with my refactor of the deserialisation method - comment added to prevent future me doing the same thing again.

Also made a few more tweaks to generally improve the performance on the listing pages:
* Tweaked prefetching to fetch the correct things
* Tweaked logic order to avoid DB lookup unless nessecary
* Removed unused template include that seemed to be firing of a DB request
* Cache the URL for the reviewer icons as this is a very repetitive lookup that can be cached.

Now down to ~80 queries for a standard page with listings and some n+1 exist, but generally much more stable and manageable.

### DIFF
--- a/opentech/apply/utils/image.py
+++ b/opentech/apply/utils/image.py
@@ -1,12 +1,22 @@
+from django.core.cache import cache
 from django.urls import reverse
 from django.utils.html import format_html
 
 
+def image_url_cache_key(image_id, spec):
+    return f'image_url_cache_{image_id}_{spec}'
+
+
 def generate_image_url(image, filter_spec):
+    cache_key = image_url_cache_key(image.id, filter_spec)
+    url = cache.get(cache_key)
+    if url:
+        return url
     from wagtail.images.views.serve import generate_signature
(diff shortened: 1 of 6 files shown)


opentech/apply/utils/image.py
```diff
@@ -1,12 +1,22 @@
...
 from django.urls import reverse
 from django.utils.html import format_html
 
 
+def image_url_cache_key(image_id, spec):
+    return f'image_url_cache_{image_id}_{spec}'
+
+
 def generate_image_url(image, filter_spec):
+    cache_key = image_url_cache_key(image.id, filter_spec)
```
ghickman: Since this is a single line function pull the string it builds to this line (and ditch the function).  You'll improve readability by putting the magic string near where it's used.

todd-dembrey: I was expecting to use it twice then didn't. You are correct.

PR #1437, opened 2019-08-20
Select Document from Submission

ghickman:
This adds the "choose file" option to Supporting Documents to pick files from the related Submission.

### DIFF
--- a/opentech/apply/projects/forms.py
+++ b/opentech/apply/projects/forms.py
@@ -1,4 +1,6 @@
 from django import forms
+from django.core.files.base import ContentFile
+from django.db import transaction
 from django.db.models import Q
 
 from addressfield.fields import AddressField
@@ -6,10 +8,12 @@ from opentech.apply.funds.models import ApplicationSubmission
 from opentech.apply.stream_forms.fields import MultiFileField
 from opentech.apply.users.groups import STAFF_GROUP_NAME
 
+from .files import get_files
 from .models import (
     COMMITTED,
     Approval,
     Contract,
+    DocumentCategory,
     PacketFile,
     PaymentReceipt,
     PaymentRequest,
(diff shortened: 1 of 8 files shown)


opentech/apply/projects/views.py
```diff
@@ -104,6 +129,38 @@ def percentage(total, value):
...
+    def get_file_paths(self, project):
+        """
+        Get MEDIA_ROOT-relative paths for all files linked to a Project's Submission
+        """
+        # Get field instances as an iterable
+        fields = project.submission.form_data.values()
+
+        # Make sure that list is flat.  We specify the base_type here so
+        # collapse doesn't try to iterate over the StreamFieldFiles as well.
+        fields = collapse(fields, base_type=StreamFieldFile)
```
todd-dembrey: This relies far too much on the internals of the form data rather than relying on the more explicit methods:

```
submission = project.submission 
file_fields = submission.file_field_ids
files = [submission.data(field) for field in file_fields]
```

Perhaps `submission.files` should be a new prop on the mixin to do this...

opentech/apply/projects/views.py
```diff
@@ -104,6 +129,38 @@ def percentage(total, value):
...
+
+        # Filter the list down to just StreamFieldFile instances
+        file_fields = (f for f in fields if isinstance(f, StreamFieldFile))
+
+        for field in file_fields:
+            # return the on-disk file path
+            yield field.name
```
todd-dembrey: I feel we should be returning the file descriptor which is an acceptable public API to be working with for files.

opentech/apply/projects/views.py
```diff
@@ -230,6 +287,45 @@ def form_valid(self, form):
...
+        return super().dispatch(request, *args, **kwargs)
+
+    def form_invalid(self, form):
+        for error in form.errors:
+            messages.error(self.request, error)
+
+        return redirect(self.project)
+
+    def form_valid(self, form):
+        form.save()
```
todd-dembrey:
```python
response =  super().form_valid(form)
messenger(...)
return response
```

opentech/apply/projects/forms.py
```diff
@@ -148,6 +154,48 @@ def save(self, commit=True):
...
+
+    @transaction.atomic()
+    def save(self, *args, **kwargs):
+        category = self.cleaned_data['category']
+        files = self.cleaned_data['files']
+
+        for path in files:
+            title = self.to_filename(path)
+
+            absolute_path = os.path.join(settings.BASE_DIR, settings.MEDIA_ROOT, path)
```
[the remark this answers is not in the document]
ghickman: Fixed and rebased.

opentech/apply/projects/views.py
ghickman: Done and pulled the method out into a utility function since we need to use it in the form's `.save()` method too.

opentech/apply/projects/views.py

opentech/apply/projects/views.py

requirements.txt
```diff
@@ -40,6 +40,7 @@ django~=2.0.0
 gunicorn==19.9.0
 mailchimp3==3.0.7
 mistune==0.8.4
+more-itertools==7.2.0
```
todd-dembrey: Still needed?

ghickman: Yea, I'm using `collapse` in the tests to avoid duplicating the logic in get_files.

Issue #1379, opened 2019-08-07
Project detail page in progress state


abigail-hampson:
- [ ] Template updating to reflect the design (contracts subheading within supporting docs sidebar) 
- [ ] PAF block moved to side bar 
- [ ] RFP block added 
- [ ] Applicant can only see proposal and contracts

fourthletter:
✔Template updating to reflect the design (contracts subheading within supporting docs sidebar)
✔PAF block moved to side bar
✔RFP block added

Applicant can only see proposal and contracts, but can NOT open. Applicant clicks on proposal or PAF and gets 403 Forbidden.

![Screenshot 2019-09-11 at 2 03 50 PM](https://user-images.githubusercontent.com/20019656/64725439-96b8e180-d4a2-11e9-9364-74ba4bdb3ecc.png)

PR #1438, opened 2019-08-20
Project In Progress State

ghickman:
This tweaks so Project page bits so we conform to the expectations of #1379.  I've also used it to capture some other little bugs I found along the way.

### DIFF
--- a/opentech/apply/projects/templates/application_projects/project_detail.html
+++ b/opentech/apply/projects/templates/application_projects/project_detail.html
@@ -132,7 +132,9 @@
                 </div>
                 {% endif %}
 
+                {% if not object.is_in_progress %}
                 {% include "application_projects/includes/supporting_documents.html" %}
+                {% endif %}
             </article>
 
             <aside class="sidebar">
--- a/opentech/apply/projects/templates/application_projects/project_detail.html
+++ b/opentech/apply/projects/templates/application_projects/project_detail.html
@@ -171,7 +171,7 @@
 
                     <a data-fancybox
                        data-src="#request-payment"
-                       class="button button--primary button--full-width"
+                       class="button button--primary button--bottom-space button--full-width"
                        href="#">
                         Add payment request
                     </a>
--- a/opentech/apply/projects/templates/application_projects/project_detail.html
+++ b/opentech/apply/projects/templates/application_projects/project_detail.html
@@ -183,6 +183,21 @@
 
                 </div>
 
+                {% if object.is_in_progress %}
+                <div class="sidebar__inner">
+                    <h5>Supporting Documents</h5>
+
+                    <p><a href="{{ object.submission.get_absolute_url }}">Proposal</a></p>
+
+                    {% if request.user.is_apply_staff %}
+                    <p><a href="#">Approval form</a></p>
+                    <p><a href="#">Supporting documents</a></p>
+                    {% endif %}
+
+                    <p><a href="{{ object.contracts.first.file.url }}">Draft contract</a></p>
+                </div>
+                {% endif %}
+
                 {% if request.user.is_apply_staff and approvals %}
                 {% block project_approvals %}{% endblock %}
                 {% endif %}
--- a/opentech/apply/projects/templates/application_projects/project_detail.html
+++ b/opentech/apply/projects/templates/application_projects/project_detail.html
@@ -193,8 +193,6 @@
                     <p><a href="#">Approval form</a></p>
                     <p><a href="#">Supporting documents</a></p>
                     {% endif %}
-
-                    <p><a href="{{ object.contracts.first.file.url }}">Draft contract</a></p>
                 </div>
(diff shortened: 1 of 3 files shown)


opentech/apply/projects/templates/application_projects/project_detail.html
```diff
@@ -166,11 +166,26 @@ <h4 class="modal__header-bar">Request Payment</h4>
...
+                    <p><a href="{{ object.contracts.first.file.url }}">Draft contract</a></p>
+                </div>
+                {% endif %}
+
                 {% if request.user.is_apply_staff and approvals %}
                 {% block project_approvals %}{% endblock %}
                 {% endif %}
 
-                {% if contracts %}
+                {% if not object.is_in_progress and contracts %}
```
todd-dembrey: We should be showing the contract when they exist, they are relevant at all times.

opentech/apply/projects/templatetags/contract_tools.py
```diff
@@ -10,6 +10,9 @@ def user_can_upload_contract(project, user):
     if user.is_apply_staff:
         return project.status != COMMITTED
 
+    if project.status != CONTRACTING:
```
todd-dembrey: Why did you add this? If there is a contract to sign they should be able to upload, which the following logic covers.

opentech/apply/projects/templates/application_projects/project_detail.html
```diff
@@ -166,11 +166,26 @@ <h4 class="modal__header-bar">Request Payment</h4>
...
+                    <h5>Supporting Documents</h5>
+
+                    <p><a href="{{ object.submission.get_absolute_url }}">Proposal</a></p>
+
+                    {% if request.user.is_apply_staff %}
+                    <p><a href="#">Approval form</a></p>
+                    <p><a href="#">Supporting documents</a></p>
+                    {% endif %}
+
+                    <p><a href="{{ object.contracts.first.file.url }}">Draft contract</a></p>
```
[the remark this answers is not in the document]
ghickman: > Template updating to reflect the design (contracts subheading within supporting docs sidebar)

As per the spec (Page Variations -> In Progress -> Design (opens Page 10)) the designs show "Draft contract" in the sidebar.

Could we do a better job at finding a draft contract? In hindsight, yes, but this is here because the ticket and design say it should be, if either of those are incorrect then we should fix those.

opentech/apply/projects/templatetags/contract_tools.py
ghickman: Poor assumption on my part, we've gated access to other things after moving out of a relevant state and I thought that applied here too since the spec didn't call it out one way or another.

Removed and updated tests to reflect the correct flow.

opentech/apply/projects/templates/application_projects/project_detail.html
ghickman: Ah, I misunderstood this line in the ticket:

> Applicant can only see proposal and contracts

Fixed and rebased.

ghickman:
> You also miss off removing the supporting documents block from the detail page when in progress.

Thanks, I must have removed the check from the model at some point!  Fixed and rebased.

opentech/apply/projects/templates/application_projects/project_detail.html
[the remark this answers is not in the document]
todd-dembrey: lets remove it - we've promoted contracts to its own block and are showing that at all times so the designs are no longer relevant.

todd-dembrey:
`{% include "application_projects/includes/supporting_documents.html" %}`
still showing here with no surrounding logic:
https://github.com/OpenTechFund/opentech.fund/blob/4048ec371c8c9c4a297da8aa7619078faae3d8a4/opentech/apply/projects/templates/application_projects/project_detail.html#L118

opentech/apply/projects/templates/application_projects/project_detail.html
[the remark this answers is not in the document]
ghickman: Fixed in 09d602c3

todd-dembrey:
> `{% include "application_projects/includes/supporting_documents.html" %}`
> still showing here with no surrounding logic:
> https://github.com/OpenTechFund/opentech.fund/blob/4048ec371c8c9c4a297da8aa7619078faae3d8a4/opentech/apply/projects/templates/application_projects/project_detail.html#L118

^^ still unaddressed.

ghickman:
> > `{% include "application_projects/includes/supporting_documents.html" %}`
> > still showing here with no surrounding logic:
> > https://github.com/OpenTechFund/opentech.fund/blob/4048ec371c8c9c4a297da8aa7619078faae3d8a4/opentech/apply/projects/templates/application_projects/project_detail.html#L118
> 
> ^^ still unaddressed.

Oops, missed that one, thanks.  Fixed in acd2b0d9

PR #1446, opened 2019-08-21
Fix File and Checkbox Elements on Modal Forms

ghickman:
This applies the field template include used for the Application Form so we get the same styling for file uploads and checkboxes.


chris-lawton:
Thanks @ghickman! 😄

PR #1452, opened 2019-08-22
Change Payment Request Status & Delete

ghickman:
This adds two controls to the Payment Requests table: change status and delete.

Change status is implemented with the standard modal popup form but posts to another view (like contracts) since we need to instantiate each form with the current PaymentRequest instance so we can tailor the options.

The Delete button is a form posting to another view which checks access.


opentech/apply/activity/messaging.py
```diff
@@ -59,6 +59,8 @@ def reviewers_message(reviewers):
     MESSAGES.CREATED_PROJECT: 'submission',
     MESSAGES.UPDATE_PROJECT_LEAD: 'old_lead',
     MESSAGES.APPROVE_CONTRACT: 'contract',
+    MESSAGES.UPDATE_PAYMENT_REQUEST_STATUS: 'payment_request',
+    MESSAGES.DELETE_PAYMENT_REQUEST: 'payment_request',
```
todd-dembrey: These should have different keys

opentech/apply/projects/views.py
```diff
@@ -169,6 +231,38 @@ def form_valid(self, form):
...
+class DeletePaymentRequestView(DeleteView):
+    model = PaymentRequest
+    pk_url_kwarg = 'payment_request_id'
+
+    def dispatch(self, request, *args, **kwargs):
+        self.project = get_object_or_404(Project, pk=self.kwargs['pk'])
+
+        is_admin = request.user.is_apply_staff
+        is_owner = request.user == self.project.user
+        if not (is_owner or is_admin):
```
todd-dembrey: This doesn't match with the logic in the templatetag - only owners can delete

ghickman: Pulled the templatetag logic into the model and called that here and in the templatetag so there's one source of truth for that logic.  Updated tests to reflect that too.

Issue #1441, opened 2019-08-21
Checkbox styling on receipt delete 


abigail-hampson:
- [ ] Removing receipts UI is easier to use for users

chris-lawton:
Possibly fixed by https://github.com/OpenTechFund/opentech.fund/pull/1446

fourthletter:
Acceptance criteria visible and functional during test. Thanks! 💥

PR #1458, opened 2019-08-23
Edit Payment Requests

ghickman:
This adds part of the edit Payment Requests flow.  I held off tackling the removal of existing files until #1441 has been handled.


opentech/apply/activity/templates/messages/email/payment_request_updated.html
```diff
@@ -0,0 +1,8 @@
+{% extends "messages/email/applicant_base.html" %}
+
+{% block content %}
+An OTF staff member has edited a Payment Request
```
todd-dembrey:
```
An OTF staff member has updated your payment request for {{ payment_request.project }} for period {{ payment_request.date_from }} to {{ payment_request.date_to }}. It is now {{ payment_request.get_status_display }}.
```

or something like that

opentech/apply/activity/tests/test_messaging.py
```diff
@@ -642,3 +650,51 @@ def test_slack_lead_change(self):
...
+        data = json.loads(responses.calls[0].request.body)
+        self.assertIn(str(applicant), data['message'])
+        self.assertIn(str(project), data['message'])
+
+    @override_settings(
+        SLACK_DESTINATION_URL=target_url,
+        SLACK_DESTINATION_ROOM=target_room,
+    )
+    @responses.activate
+    def test_slack_staff_update_payment_request(self):
```
todd-dembrey: Staff should be updated via slack when this happens. It was absent from the requirements added now

It is correct that applicants dont get an email when they submit though.

opentech/apply/projects/forms.py
```diff
@@ -107,6 +108,26 @@ def __init__(self, user=None, *args, **kwargs):
...
+        fields = ['invoice', 'value', 'date_from', 'date_to', 'receipt_list', 'comment']
+        model = PaymentRequest
+
+    def __init__(self, user=None, instance=None, *args, **kwargs):
+        super().__init__(*args, instance=instance, **kwargs)
+
+        self.instance = instance
+
+        self.fields['receipt_list'].choices = [
+            (r.pk, os.path.basename(r.file.url))
```
todd-dembrey: Should be working with the `file.name` because storage messes with the basename in the url?

ghickman: I'm getting the same output for `file.name` and `file.url` after passing through `os.path.basename`.  I looked into `SubmissionStreamFileField` which has a `filename` but that's the same output for me locally too so doesn't gain us anything.

opentech/apply/activity/templates/messages/email/payment_request_updated.html
ghickman: Fixed and rebased.

opentech/apply/activity/tests/test_messaging.py
ghickman: Fixed and rebased.

PR #1472, opened 2019-08-29
Payment Request Feedback

ghickman:
This lets a Staff member leave a comment and set the value paid on a PaymentRequest when changing the status.

It handles tracking that new paid value, using it instead of the requested amount in the various places we track values or aggregated values.


opentech/apply/projects/views.py
```diff
@@ -120,12 +138,28 @@ def percentage(total, value):
...
 
+        def get_values(qs):
+            """
+            Get definitive values for each PaymentRequest
+
+            This finds the definitive value for each PaymentRequest, picking
+            `paid_value` when it has a value or `requested_value` when it does
+            not.  These annotated values are then returned as an iterable.
+            """
+            return qs.annotate(
```
todd-dembrey: You should just be able to use Coalesce: `Coalesce(F('paid_value'), F('requested_value'))`

could also pull the sum in if you fancy...
```
qs.aggregate(
    total=Sum(Coalesce(F('paid_value'), F('requested_value')))
)['total']
```

Kind of tempted to move this into the queryset. (which you may be doing as part of the sprint 3 work on tables)

opentech/apply/projects/forms.py
```diff
@@ -54,32 +71,40 @@ def clean(self):
...
-        if instance.status == SUBMITTED:
-            wanted = [CHANGES_REQUESTED, UNDER_REVIEW, PAID, DECLINED]
-            status_field.choices = filter_request_choices(wanted)
-            return
+        statuses_lut = {
+            CHANGES_REQUESTED: filter_request_choices([DECLINED]),
+            SUBMITTED: filter_request_choices([CHANGES_REQUESTED, UNDER_REVIEW, PAID, DECLINED]),
+            UNDER_REVIEW: filter_request_choices([PAID]),
+        }
+        status_field.choices = statuses_lut.get(instance.status, status_field.choices)
```
todd-dembrey: Whats the scenario where instance.status is not in `statuses_lut`. I feel returning all choices in that situation might be less safe than returning no choices or a list of "safe" choices.

opentech/apply/projects/forms.py
```diff
@@ -54,32 +71,40 @@ def clean(self):
...
         self.instance = instance
+        self.initial['paid_value'] = instance.requested_value
 
         status_field = self.fields['status']
 
-        if instance.status == SUBMITTED:
-            wanted = [CHANGES_REQUESTED, UNDER_REVIEW, PAID, DECLINED]
-            status_field.choices = filter_request_choices(wanted)
-            return
+        statuses_lut = {
```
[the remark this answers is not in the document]
ghickman: "look-up table"

opentech/apply/projects/forms.py
ghickman: When `instance.status` is either `PAID` or `DECLINED`.  There are checks elsewhere to stop the user reaching that state but I could return an empty list here too?

opentech/apply/projects/views.py

I kept the logic in the inner function for now, we can pull it out later if there are more places which need it.

opentech/apply/projects/forms.py
todd-dembrey: Please,

opentech/apply/projects/forms.py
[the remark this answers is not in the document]
todd-dembrey: Can we get something a bit more meaningful? `status_transitions`?

opentech/apply/projects/forms.py
ghickman: Fixed and rebased.

opentech/apply/projects/forms.py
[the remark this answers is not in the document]
ghickman: Fixed and rebased.

PR #1473, opened 2019-08-30
Use settings for organisation name etc. in templates.

frjo:
This is a first step in making the app more generic.

### DIFF
--- a/opentech/apply/funds/templates/funds/application_base_landing.html
+++ b/opentech/apply/funds/templates/funds/application_base_landing.html
@@ -2,11 +2,11 @@
 {% block header_modifier %}header--light-bg{% endblock %}
 {% block content %}
 <div class="wrapper wrapper--small">
-    <h3>Thank you for your submission to the Open Technology Fund.</h3>
+    <h3>Thank you for your submission to the {{ ORG_LONG_NAME }}.</h3>
     <div class="rich-text">
         <p>An e-mail with more information has been sent to the address you entered.</p>
         <p>If you do not receive an e-mail within 15 minutes please check your
-spam folder and contact help@opentech.fund for further assistance.</p>
+spam folder and contact {{ ORG_EMAIL }} for further assistance.</p>
     </div>
 </div>
 {% endblock %}
--- a/opentech/public/utils/context_processors.py
+++ b/opentech/public/utils/context_processors.py
@@ -1,6 +1,7 @@
+from django.conf import settings
+
 from opentech.apply.home.models import ApplyHomePage
 from opentech.public.home.models import HomePage
-
 from opentech.public.mailchimp.forms import NewsletterForm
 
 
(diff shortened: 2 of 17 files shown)


opentech/apply/determinations/forms.py
```diff
@@ -204,21 +204,20 @@ class BaseConceptDeterminationForm(forms.Form):
 
     principles = RichTextField(
         label='Goals and principles',
-        help_text='Does the project contribute and/or have relevance to OTF goals and principles?'
+        help_text='Does the project contribute and/or have relevance to funds goals and principles?'
```
frjo: @danblah I have made the help text here more generic. Will this work?

opentech/apply/funds/templates/funds/application_base_landing.html
```diff
@@ -2,11 +2,11 @@
 {% block header_modifier %}header--light-bg{% endblock %}
 {% block content %}
 <div class="wrapper wrapper--small">
-    <h3>Thank you for your submission to the Open Technology Fund.</h3>
+    <h3>Thank you for your submission to the {{ ORG_LONG_NAME }}.</h3>
     <div class="rich-text">
         <p>An e-mail with more information has been sent to the address you entered.</p>
         <p>If you do not receive an e-mail within 15 minutes please check your
-spam folder and contact help@opentech.fund for further assistance.</p>
+spam folder and contact {{ ORG_EMAIL }} for further assistance.</p>
```
frjo: @danblah In most places a "hello" address is used but here a "help" address is used. Can we simplify this to a single "ORG_EMAIL" or do we need to have a "ORG_SUPPORT_EMAIL" as well?

opentech/public/utils/context_processors.py
```diff
@@ -1,12 +1,16 @@
...
 
 def global_vars(request):
     return {
         'APPLY_SITE': ApplyHomePage.objects.first().get_site(),
         'PUBLIC_SITE': HomePage.objects.first().get_site(),
-        'newsletter_form': NewsletterForm()
+        'newsletter_form': NewsletterForm(),
+        'ORG_LONG_NAME': settings.ORG_LONG_NAME,
+        'ORG_SHORT_NAME': settings.ORG_SHORT_NAME,
+        'ORG_EMAIL': settings.ORG_EMAIL,
```
frjo: @Parbhat Is this a decent way to make settings available in all templates?

opentech/settings/base.py
```diff
@@ -23,6 +23,14 @@
     ALLOWED_HOSTS = env['ALLOWED_HOSTS'].split(',')
 
 
+# Organisation name and e-mail address, used in e-mail templates etc.
+
+ORG_LONG_NAME = env.get('ORG_LONG_NAME', 'Acme Corporation')
+ORG_SHORT_NAME = env.get('ORG_SHORT_NAME', 'ACME')
+ORG_EMAIL = env.get('ORG_EMAIL', 'info@example.org')
+ORG_GUIDE_URL = env.get('ORG_GUIDE_URL', 'https://guide.example.org/')
```
frjo: @danblah Should we use real OTF data as defaults or generic like I have here?

opentech/public/utils/context_processors.py
Parbhat: @frjo Wagtail has a [Site settings](https://docs.wagtail.io/en/v2.6.1/reference/contrib/settings.html) app specifically for this purpose. It will allow admins to change values from Wagtail admin. This method is also good as an organisation name, email etc. not change very often.

opentech/apply/funds/templates/funds/application_base_landing.html
danblah: yes

Issue #1429, opened 2019-08-19
Project Overview page


todd-dembrey:
`/projects/`
- [ ] Accessible by staff only
- [ ] Includes Project Status summary block (see submissions page for example)
    - Link to  all projects prefiltered on status
![image](https://user-images.githubusercontent.com/6338661/63760847-fff00080-c8b7-11e9-9038-700a1ef71dd1.png)

- [ ] Includes Projects summary table - as per Project list view
    - limit 10
    - link "all projects" to project list view
- [ ] Includes PR summary table - as per PR list view
    - limit 10
    - link "all payment requests" to PR list view 
- [ ] Add `Projects` to the main nav linking to this page

fourthletter:
Acceptance criteria were visible and functional during test. Thanks! 💥

PR #1477, opened 2019-09-02
Project List View

ghickman:
This adds the Project List View and configures the dashboard view to redirect there when a user sorts the Project table on that page.  It adds a mixin to handle the redirection on a more generic basis since there are now two list views we redirect to and we know there's another one coming.

Note: the mixin included here will get used in the PR for #1429 to handle the same situation there.


opentech/apply/dashboard/views.py
```diff
@@ -25,42 +25,41 @@
...
-
-    def get(self, request, *args, **kwargs):
-        # redirect to submissions list when we use the filter to search for something
-        if len(request.GET):
-            query_str = '?'
-            for key, value in request.GET.items():
-                query_str += key + '=' + value + '&'
-            return HttpResponseRedirect(reverse_lazy('funds:submissions:list') + query_str)
+class AdminDashboardView(RedirectToListViewMixin, TemplateView):
+    destination_lut = {
```
todd-dembrey: not a fan of `lut` usage - can we be a bit more descriptive about this variable.

opentech/apply/dashboard/views.py
```diff
@@ -25,42 +25,41 @@
...
 
-class AdminDashboardView(TemplateView):
-
-    def get(self, request, *args, **kwargs):
-        # redirect to submissions list when we use the filter to search for something
-        if len(request.GET):
-            query_str = '?'
-            for key, value in request.GET.items():
-                query_str += key + '=' + value + '&'
-            return HttpResponseRedirect(reverse_lazy('funds:submissions:list') + query_str)
```
todd-dembrey: You've broken the search and filter aspects of the table with the current implementation.

TBH, i think this whole view is a bit of a mess (historic reasons)  and could do with some tidying up. 

associating filtersets with tables and prefixes i think would help this out. rather than just dumping everything into the context

opentech/apply/utils/views.py
```diff
@@ -169,3 +173,66 @@ def post(self, request, *args, **kwargs):
...
+
+    def redirect_to_list_view(self, query_dict, destination_lut):
+        """
+        Attempt to redirect to a list view when filtering a table
+
+        Each table is created with a prefix which is used when sorting keys are
+        generated.  We use those keys here to decide where the user should be
+        redirected to.
+        """
+        first_key, sep, _ = list(query_dict.keys())[0].rpartition('--')
```
todd-dembrey: Are we sure that the first query param is the one we are interested in?

opentech/apply/utils/views.py
```diff
@@ -169,3 +173,66 @@ def post(self, request, *args, **kwargs):
...
+        prefix = f'{first_key}{sep}'
+
+        if prefix not in destination_lut.keys():
+            # the prefix we found wasn't in the expected list so exit now
+            return
+
+        path = destination_lut[prefix]
+
+        query_args = dict(self.strip_prefix(query_dict, prefix))
```
todd-dembrey: do we need to drop querystrings that do not have the prefix?

opentech/apply/dashboard/views.py
ghickman: I missed those on the submissions table, should they redirect to the submissions page?  Or should we remove them from the dashboard and push users to use the Submissions List page instead?

opentech/apply/utils/views.py
ghickman: It depends on the outcome of [this discussion](https://github.com/OpenTechFund/opentech.fund/pull/1477#discussion_r320667734).

If we move to only handling prefixed query args (which I mistakenly thought was the case) then yes since only one will be active from the UI since clicking any column header redirects away.  However if we keep Submission filtering and search on the page then either we prefix their query args (which seems the logical choice to me) or we change this to search for the first prefixed arg.

opentech/apply/utils/views.py
ghickman: Again I think this depends wholly on the outcome of [this discussion](https://github.com/OpenTechFund/opentech.fund/pull/1477#discussion_r320667734).

opentech/apply/dashboard/views.py
ghickman: What would you like it to be?

opentech/apply/projects/views.py
```diff
@@ -697,3 +701,16 @@ def dispatch(self, request, *args, **kwargs):
...
+@method_decorator(staff_required, name='dispatch')
+class ProjectListView(SingleTableMixin, FilterView):
+    filterset_class = ProjectListFilter
+    model = Project
+    table_class = ProjectsListTable
+    template_name = 'application_projects/project_list.html'
+
+    def get_queryset(self):
+        return (Project.objects.annotate(amount_paid=Sum('payment_requests__paid_value'))
+                               .annotate(last_payment_request=Max('payment_requests__date_to'))
```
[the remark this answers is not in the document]

ghickman:
Updated based on discussions to:
* tidy up the context construction so we group tables and related items (filters, etc) into namespace dicts
* construct redirection URLs and pass those into filter/search forms to avoid handling redirection in the view

PR #1493, opened 2019-09-04
Command to fix so all stream fields have unique ids.

frjo:


frjo:
@todd-dembrey Does this look correct and safe to run?

The problem is the migration seed I built that is also fixed in this PR. I copied and pasted to quickly and ended up with the same id for a number of "text_markup" fields. A bit of luck it was not fields that hold review data.

PR #1494, opened 2019-09-04
Unauthenticated Submission and Project Views

ghickman:
This adds unauthenticated views for both Submissions and Projects, reusing the simplified templates for both.  To log via the messaging framework this adds a sentinel user so we don't need to rewrite how Event.user is used around the site.


opentech/apply/users/migrations/0014_add_sentinel_user.py
```diff
@@ -0,0 +1,38 @@
...
+
+SENTINEL_EMAIL = 'sentinel'
+
+
+def add_sentinel_user(apps, schema_editor):
+    User = apps.get_model('users', 'User')
+    User.objects.create(
+        email=SENTINEL_EMAIL,
+        full_name='sentinel',
+        slack='sentinel',
```
todd-dembrey: is slack required?

opentech/apply/users/migrations/0014_add_sentinel_user.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+SENTINEL_EMAIL = 'sentinel'
+
+
+def add_sentinel_user(apps, schema_editor):
+    User = apps.get_model('users', 'User')
+    User.objects.create(
+        email=SENTINEL_EMAIL,
+        full_name='sentinel',
```
todd-dembrey: I was expecting the sentinal to be specific to the use case. e.g. "Compliance", or similar so that when we log the activity we have a sensible public facing name for the activity log.

opentech/apply/users/migrations/0014_add_sentinel_user.py
ghickman: It's not nullable so I could set it to a blank string, but this seemed more obvious if someone found it exposed somewhere.  Which do you prefer, `'sentinel'` or `''`?

opentech/apply/users/migrations/0014_add_sentinel_user.py
ghickman: Fixed and rebased.

PR #1496, opened 2019-09-06
Notify Finance of Payment Request Approval

ghickman:
This adds an unauthenticated (and logged) view for a single PaymentRequest and sends an email to the configured Finance email when a PaymentRequest is approved.

Note: we haven't defined the approval of a PaymentRequest before so I opted for `{CHANGES_REQUESTED, UNDER_REVIEW}`.


fourthletter:
Do we will need this feature?

PR #1530, opened 2019-09-19
Add a help link field to all stream fields.

### LINKED ISSUE #1490: Add help link field in addition to help text on application form stream fields
In to this field it should be possible to enter an internal or external link and has it be rendered as a link on the form.

The link text should be "More information" and the link tag should open in a new window/tab.

frjo:
Fixes #1490


frjo:
@Parbhat What I'm I missing here to get the help_link value to the template?

frjo:
@Parbhat Thank you, that works perfectly. I have added some styling now so it looks better as well.

PR #1564, opened 2019-09-25
Added Heroku production settings file, and removed heroku lines from generic production settings file.

### LINKED ISSUE #1561: Production settings cause static files to be stored in /staticfiles/, not /static/
A deployment issue:

Using DJANGO_SETTINGS_MODULE=opentech.settings.dev creates static files in the expected directory defined by base.py: /static/, however, when using DJANGO_SETTINGS_MODULE=opentech.settings.production, `python manage.py collectstatic` creates static files in a directory called /staticfiles/.

maxpearl:
This creates a new settings file, called production_heroku.py, and removes the lines regarding heroku from the generic production file. This fixes #1561 - using generic production settings creates static files in the correct place.


opentech/settings/production_heroku.py
```diff
@@ -0,0 +1,36 @@
...
+
+from sentry_sdk.integrations.django import DjangoIntegration
+
+from .base import *  # noqa
+
+# Disable debug mode
+DEBUG = False
+
+# Configuration from environment variables
+# Alternatively, you can set these in a local.py file on the server
```
frjo: This is not true anymore, the local file is only loaded in `dev.py`.

What do you think of adding it to the new `production.py`?

https://github.com/OpenTechFund/opentech.fund/blob/7266d2d7e712172f7ddaa33a3972972a90622f85/opentech/settings/dev.py#L32-L35

maxpearl: Great idea. Makes sense to me.

frjo:
See https://github.com/OpenTechFund/opentech.fund/pull/1580

PR #1580, opened 2019-10-07
Make the production settings work on more enviroments than Heroku.

### LINKED ISSUE #1561: Production settings cause static files to be stored in /staticfiles/, not /static/

frjo:
Fixes #1561

Alternativ to https://github.com/OpenTechFund/opentech.fund/pull/1564

### DIFF
--- a/opentech/settings/production.py
+++ b/opentech/settings/production.py
@@ -1,19 +1,18 @@
 import os
 
-import django_heroku
-import sentry_sdk
-
-from sentry_sdk.integrations.django import DjangoIntegration
-
 from .base import *  # noqa
 
 # Disable debug mode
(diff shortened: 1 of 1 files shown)


frjo:
@maxpearl What do you think of this. Not tested yet, but I think it should work.

maxpearl:
OK, that makes a lot of sense to me to leave it in one file, and just do a check for environment.

PR #1592, opened 2019-10-09
Fix: New file versions created on status update, adding partner forms

### LINKED ISSUE #1572: New file versions created on status update, adding parter etc.
The PrivateMediaStorage class has "file_overwrite" set too "false". Each time a new version of a file is uploaded the system then adds a suffix to the filename to version them. This makes each revision have the correct version of a file.

When saving the submission form itself the system only increments the file name when needed, when a new version of a file is actually uploaded.

The problem is that when updating a submissions status, adding partners etc. the system treat the files as changed and saves a new copy with an added suffix to the filename on S3.

This adds up over a submissions lifetime so the filenames becomes very long (+1600 chars I have seen). At a point the system can no longer handle them. 

There is also a problem downloading files with very long filenames since OS:es can have limits.

Parbhat:
Fixes #1572

We have various Update forms on submission detail page like a status update, partners, lead forms etc. Application submission model has a custom [save](https://github.com/OpenTechFund/opentech.fund/blob/9a740785a3f6375cc8ec4dd95dcce90de747033f/opentech/apply/funds/models/submissions.py#L601) method. And a portion of the code [saves](https://github.com/OpenTechFund/opentech.fund/blob/9a740785a3f6375cc8ec4dd95dcce90de747033f/opentech/apply/funds/models/submissions.py#L595) a new version of a file every time save method is invoked. It results in long filenames as file overwrite setting is False. The update forms on detail page do not need to run the custom code in the save method. Therefore adding a flag to skip the custom operation for these forms. This old PR #558 also seems related. For the future, we should refactor the file save behaviour in save call.

### DIFF
--- a/opentech/apply/funds/forms.py
+++ b/opentech/apply/funds/forms.py
@@ -17,7 +17,39 @@ from .widgets import Select2MultiCheckboxesWidget, MetaCategorySelect2Widget
 from .workflow import get_action_mapping
 
 
-class ProgressSubmissionForm(forms.ModelForm):
+class ApplicationSubmissionModelForm(forms.ModelForm):
+    """
+    Application Submission model's save method performs several operations
+    which are not required in forms which update fields like status, partners etc.
+    It also has a side effect of creating a new file uploads every time with long filenames (#1572).
+    """
+
+    def save(self, commit=True):
+        """
+        Save this form's self.instance object if commit=True. Otherwise, add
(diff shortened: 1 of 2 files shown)


frjo:
Running this on test now and it works perfectly, nice work! Will deploy this tomorrow.

The problem with adding up hashes is also solved by this. I think the wrong filename was sent to S3 on "Update status" etc saves.

Issue #1567, opened 2019-09-30
Additional deployment options for production use


danblah:
* Scalingo: Hoping there isn't much required, see https://doc.scalingo.com/platform/getting-started/heroku-compatibility.
* Docker image on https://hub.docker.com/
* Digital Ocean 1-click build Droplet: https://github.com/digitalocean/marketplace-partners

Documentation for each and code changes/additions to make it as easy as possible.

PR #1599, opened 2019-10-17
Docker setup for local development

frjo:
Related: #1567


frjo:
@maxpearl How does this looks to you?

I have tried to read up on Docker and this code seems to work at least.

I ended up using the "circleci/python:3.7-buster-node" image to avoid having to install anything, it has both python and node installed and working for this app. Had a lot of problem install python on node images and vide versa. Always more things that needed to be installed it seemed.

maxpearl:
This looks good to me - since it works - nothing stands out as problematic.

Issue #1373, opened 2019-08-07
Download submission and PAF information as PDF 


abigail-hampson:
- [ ] Download button available for the PAF (all information that is not docs in the simplified view) 
- [ ] Download button available for the simplified submission view

todd-dembrey:
This was merged and deployed as part of 1.30.3
#1677
https://github.com/OpenTechFund/hypha/releases/tag/v1.30.3

PR #1627, opened 2019-10-28
Feature/gh 1373 reportlab download

todd-dembrey:
First part of #1373 which adds basic pdf exporting to the simplified submission page. There will be a follow up PR for the work to export the PAF

- Add download to simplified report page
- Basic pdf export with light touch formatting in line with page styles
- Improvements to formatting on submissions and simplified submissions so that the user entered rich text doesn't overpower the questions for `h` tags.


opentech/apply/utils/pdfs.py
```diff
@@ -0,0 +1,274 @@
...
+    'NotoSans',
+    normal='NotoSans',
+    bold='NotoSansBold',
+    italic='NotoSansItalic',
+    boldItalic='NotoSansBoldItalic'
+)
+
+DARK_GREY = Color(0.0154, 0.0154, 0, 0.7451)
+
+PAGE_WIDTH, PAGE_HEIGHT = pagesizes.legal
```
frjo: Would be good to have this as a setting.

todd-dembrey: Yeah, it probably should be. That's a bit more work as currently i'm using the global values throughout the methods. Can we add it as a backlog ticket?

PR #1636, opened 2019-11-01
Fix the creation of files under the submission's revision path 

Parbhat:
Not related to the issue fixed. But we may need a strategy to only [save](https://github.com/OpenTechFund/opentech.fund/blob/2ffd404d628ad4f255ac3892cef306980f6e4a4d/opentech/apply/funds/models/mixins.py#L75) files if the existing files changes. Currently, a new file version is created even if the file is not changed in an edit form. Long file names have problems like #1572.

Fixes https://github.com/OpenTechFund/issues.otf/issues/13

### DIFF
--- a/opentech/apply/funds/files.py
+++ b/opentech/apply/funds/files.py
@@ -14,7 +14,13 @@ def generate_submission_file_path(submission_id, field_id, file_name):
 
 class SubmissionStreamFieldFile(StreamFieldFile):
     def generate_filename(self):
-        return generate_submission_file_path(self.instance.pk, self.field.id, self.name)
+        from opentech.apply.funds.models.submissions import ApplicationRevision
+        submission_id = self.instance.pk
+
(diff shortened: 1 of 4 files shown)


todd-dembrey:
Can you add a test for this @Parbhat, this is the second time there has been an issue around this and i would like to avoid that in future :)

Parbhat:
@todd-dembrey Sure, I will add tests.

frjo:
Tested and it works well, no more revision folders. Nice work!

A test would indeed be good to have here.

Confirmed that your fix for not saving new versions of the files on each status etc change still works as well. This makes the file names less of an issue but it still should get fixed. Will add a separate issue about that.

frjo:
When the test is in place and working I will deploy this.

Parbhat:
@frjo PR updated with test.

PR #1642, opened 2019-11-04
Allow staff to see 16 reviewers in the reviewers field, before only f…

### LINKED ISSUE #1639: Make reviewers field on fund/rounds taller so more text lines are visible
There are a lot of reviewers and staff commonly pick a selection of them for a fund/round.

With only 4 lines showing it's a bit cumbersome.

frjo:
…ive was visible at a time.

Fixes #1639

### DIFF
--- a/opentech/apply/funds/models/applications.py
+++ b/opentech/apply/funds/models/applications.py
@@ -1,5 +1,6 @@
 from datetime import date
 
+from django import forms
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import models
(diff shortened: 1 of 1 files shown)


opentech/apply/funds/models/applications.py
```diff
@@ -108,7 +109,7 @@ def serve(self, request):
 
     content_panels = WorkflowStreamForm.content_panels + [
         FieldPanel('approval_form'),
-        FieldPanel('reviewers'),
+        FieldPanel('reviewers', widget=forms.SelectMultiple(attrs={'size': '16'})),
```
frjo: @Parbhat A better way to do this?

Parbhat: @frjo This is the best way as mentioned in [docs](http://docs.wagtail.io/en/v2.6.3/reference/pages/panels.html#wagtail.admin.edit_handlers.FieldPanel.widget).

Issue #1623, opened 2019-10-28
Receive reminders when reports are due


todd-dembrey:
- [ ] Default period 1 week before due
- [ ] reminder sent on day of report due if not already submitted
- [ ] Email reminder about report sent to applicant when within time period

abigail-hampson:
This issue is with Frederik for testing. Once this has been approved we can deploy this as it is quite difficult to get this set up for testing by end users on the test site

frjo:
@abigail-hampson  With two small fixes from @todd-dembrey this now works well when I test it.

PR #1660, opened 2019-11-05
Feature/gh 1623 notify users of report

todd-dembrey:
@frjo probably worth you having a quick look at this one.

Will need to run the management with a daily cron job which will trigger the sending of an email to the user when the reach the threshold.

I've also add in the `notified` attribute in case we get errors and need to run this twice etc.


frjo:
Related #1623

PR #1662, opened 2019-11-06
Feature/gh 1646 change frequency

chris-lawton:
Adds reporting calculator which is a breakdown of:
- Next report due period (next report date - today)
- Report period start date (last report end date + 1)
- Report period end date (next report date)
- Frequency (no. of months/weeks)
- Project end date

### DIFF
--- a/opentech/apply/funds/templates/funds/includes/delegated_form_base.html
+++ b/opentech/apply/funds/templates/funds/includes/delegated_form_base.html
@@ -22,7 +22,7 @@
 
     {% if report_frequency_form %}
         <div class="form__info-box">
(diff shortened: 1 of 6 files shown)


opentech/apply/funds/templates/funds/includes/delegated_form_base.html
```diff
@@ -20,6 +20,24 @@
         {% endif %}
     {% endfor %}
 
+    {% if report_frequency_form %}
```
todd-dembrey: I didn't want to do this, adding a highly specific implementation into the base template will get messy long term and totally disassociates it from the only place it gets used.

Is there no way to move this out to somewhere near here: https://github.com/OpenTechFund/opentech.fund/pull/1662/files#diff-407ee86547708479420457906b7a732aR19

chris-lawton: Moved to the top of the modal which goes against the design, but don't think there's a neat way to achieve what the design asks for. Adding it all via JS seems overkill.

todd-dembrey: Thanks Chris, i like that - lets review it on dev

Issue #1643, opened 2019-11-04
Report Page FE 


abigail-hampson:
Includes:

- [x] Alert for reporting period at the top of the form 
- [x] Modal for saving as draft and confirming submission 
- [x] General FE tidy of form page

abigail-hampson:
Just missing confirmation modals

chris-lawton:
PR for confirmation modals: https://github.com/OpenTechFund/opentech.fund/pull/1678

abigail-hampson:
Looks good, ready to go to test

fourthletter:
All acceptance criteria are visible. Thanks! :boom:

PR #1664, opened 2019-11-06
feature/gh-1643-report-page

chris-lawton:
Style report page and update wording to match design

### DIFF
--- a/opentech/apply/projects/forms.py
+++ b/opentech/apply/projects/forms.py
@@ -381,10 +381,10 @@ class UpdateProjectLeadForm(forms.ModelForm):
 
 class ReportEditForm(forms.ModelForm):
     public_content = RichTextField(
-        help_text="This section will be shared with the wider community."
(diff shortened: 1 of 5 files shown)


opentech/apply/projects/forms.py
```diff
@@ -381,10 +381,10 @@ def __init__(self, user=None, *args, **kwargs):
 
 class ReportEditForm(forms.ModelForm):
     public_content = RichTextField(
-        help_text="This section will be shared with the wider community."
+        help_text="This section of the report will be shared with the broader OTF community."
     )
     private_content = RichTextField(
-        help_text="This section will be shared with staff members only."
+        help_text="This section of the report will be shared with OTF staff only."
```
todd-dembrey: we want to avoid using the term "OTF" as we need to consider other organisations. Can you remove it please.

chris-lawton: ah fair, was just matching the wording in the designs, will swap put

PR #1689, opened 2019-11-13
Admin: Create copy button in Application form

### LINKED ISSUE #606: Add a clone/copy function on the admin/funds/applicationform/ listing to create a new application based on a previous one

Parbhat:
Closes #606

### DIFF
--- a/opentech/apply/funds/admin_helpers.py
+++ b/opentech/apply/funds/admin_helpers.py
@@ -80,13 +80,13 @@ class ApplicationFormButtonHelper(ButtonHelper):
 
     def copy_form_button(self, pk, form_name, **kwargs):
         classnames = self.prepare_classnames(
-            start=self.edit_button_classnames + ['icon', 'icon-plus'],
+            start=self.edit_button_classnames,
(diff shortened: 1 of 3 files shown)


opentech/apply/funds/admin_helpers.py
```diff
@@ -69,3 +69,37 @@ def queryset(self, request, queryset):
...
+
+    def copy_form_button(self, pk, form_name, **kwargs):
+        classnames = self.prepare_classnames(
+            start=self.edit_button_classnames + ['icon', 'icon-plus'],
+            add=kwargs.get('classnames_add'),
+            exclude=kwargs.get('classnames_exclude')
+        )
+        return {
+            'classname': classnames,
+            'label': f'Create Copy',
```
frjo: The button on the rounds admin says only "Copy" and they do the same thing more or less so I think they should look the same. Can you add that to this pr? Pick the look you think is best.

Parbhat: Thanks for noticing. PR updated to have similar styling and text for copy button.

PR #1702, opened 2019-11-25
Add a open/closed filter to the round admin page in wagtail admin.

### LINKED ISSUE #1701: Add filter for active/inactive/all to rounds admin page in wagtail admin
Similar to how it looks on the application forms page.

frjo:
Fixes #1701

### DIFF
--- a/opentech/apply/funds/admin_helpers.py
+++ b/opentech/apply/funds/admin_helpers.py
@@ -1,10 +1,13 @@
+from urllib.parse import urlencode
+
 from django import forms
 from django.contrib import admin
 from django.urls import reverse
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext as _
 
 from wagtail.contrib.modeladmin.forms import ParentChooserForm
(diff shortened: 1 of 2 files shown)


opentech/apply/funds/admin_helpers.py
```diff
@@ -71,6 +71,25 @@ def queryset(self, request, queryset):
         return queryset
 
 
+class RoundStateListFilter(admin.SimpleListFilter):
+    title = 'state'
+    parameter_name = 'form-state'
+
+    def lookups(self, request, model_admin):
+        return (
+            ('open', _('Open')),
```
frjo: @Parbhat I would like to set "open" as the default for this page. The user should then be able to select all and closed as well. Best way to do that.

Parbhat: @frjo Included a commit for setting index URL with Open filter.

frjo: Excellent, thanks!

PR #1835, opened 2020-02-28
fix wiki links on readme file

vaibhavmule:
The current links do not work and that leads to 404 pages. This fix leads to correct pages and it will be beneficial for a newcomer to save time finding the deploy docs.


frjo:
@vaibhavmule Welcome to the Hypha team!

PR #1901, opened 2020-04-01
Allow staff to edit final determination except status

### LINKED ISSUE #635: Ability for team to edit/delete a determination msg with an option to not re-send the applicant notifications
This exists in the previous system. Use case, we need to go back and edit the final determination msg outcome due to error or some other reason but don't want the applicant to get another duplicate notification with the only thing that's changed is the status.

vaibhavmule:
Fixes #635

(This PR now includes commits from branch [enhancement/add_sent_notice_option_determinations](https://github.com/OpenTechFund/hypha/tree/enhancement/add_sent_notice_option_determinations))

I have gone through CreateOrUpdateView for Determination and checked if there is a possibility to use the existing `Update` view for Determination. 

The existing Update view is quite complex, 
- it has created in the mind that when you are updating the determination when in draft state
- has permission-based on if the user can create or edit determination

After going through this, decided to create a new update view for editing Determination.

**Following things need to be done in case you want to update status for Submission.**

There is no provision to revert back to `Ready for determination` stage.

I have looked at existing use cases found that there is a [restart_stage method on Submission](https://github.com/OpenTechFund/hypha/blob/ceb74015590ae24d1412478505faeda6951c080d/hypha/apply/funds/models/submissions.py#L462). 

After that, I have look at [FSMField Docs](https://github.com/viewflow/django-fsm#usage) and [blog](https://hashedin.com/blog/a-guide-to-managing-finite-state-machine-using-django-fsm/),  came with the following function to programmatically change the status of the submission. 

https://github.com/OpenTechFund/hypha/pull/1901/files#diff-dc3c11f62e7001fabe8c26c7ace3a61fR477
```python
@transition(
    status, source='*',
    target=RETURN_VALUE(
        'determination',
        'ext_determination',
        'proposal_determination',
        'concept_determination',
        'com_determination',
    ),
    permission=make_permission_check({UserPermissions.STAFF}),)
def revert_to_determination(self, **kwargs):
    """
    Revert back to determination
    """
    if self.stage == RequestExt:
        return 'ext_determination'
    elif self.stage == RequestCom:
        return 'com_determination'
    elif self.stage == Concept:
        return 'concept_determination'
    elif self.stage == Proposal:
        return 'proposal_determination'

    return 'determination'
```

So before updating the Determination above `revert_to_determination` is called and the following side effects are taken into consideration.
- if the determination was approved, then you change it to dismissed, `you may need to delete the project object created for submission`.
- if the determination was dismissed, then you change it to approved, `you must create a project`.

**For delete determination, we need to implement the above status change part so when we delete the determination, we can revert back status of a submission to` ready for determination.`**


hypha/apply/determinations/views.py
```diff
@@ -429,3 +431,82 @@ class DeterminationDetailView(ViewDispatcher):
     partner_view = PartnerDeterminationDetailView
     community_view = CommunityDeterminationDetailView
     applicant_view = ApplicantDeterminationDetailView
+
+
+@method_decorator(staff_required, name='dispatch')
+class DeterminationUpdateView(UpdateView):
```
frjo, line 437: Call it "DeterminationEditView" instead to sync up with the naming in reviews.

hypha/apply/funds/models/submissions.py
```diff
@@ -474,6 +478,31 @@ def restart_stage(self, **kwargs):
             return 'invited_to_proposal'
         return INITIAL_STATE
 
+    @transition(
```
frjo, line 481: The state of the submission should not change when the determination is edited. My understanding is that the edits are only for archival purposes.

hypha/apply/determinations/forms.py
```diff
@@ -403,3 +404,18 @@ def __init__(self, *args, submissions, initial={}, **kwargs):
...
 
         self.fields = self.apply_form_settings('proposal', self.fields)
+
+
+class UpdateDeterminationForm(forms.ModelForm):
+    message = RichTextField(label='Determination message')
+
+    class Meta:
+        model = Determination
+        fields = ['outcome', 'message']
```
frjo, line 414: Remove outcome and add send_notice.

hypha/apply/determinations/forms.py
```diff
@@ -403,3 +404,18 @@ def __init__(self, *args, submissions, initial={}, **kwargs):
         self.fields['outcome'].widget = forms.HiddenInput()
 
         self.fields = self.apply_form_settings('proposal', self.fields)
+
+
+class UpdateDeterminationForm(forms.ModelForm):
```
frjo, line 409: Rename to "DeterminationEditForm".

hypha/apply/determinations/templates/determinations/determination_detail.html
```diff
@@ -6,7 +6,8 @@
 {% block content %}
 <div class="admin-bar">
     <div class="admin-bar__inner">
-        <h2 class="heading heading--no-margin">Determination {% if determination.is_draft %}[DRAFT]{% endif %}</h2>
+        <h2 class="heading heading--no-margin">Determination {% if determination.is_draft %}[DRAFT]{% else %}<a class="button" href="{% url 'apply:submissions:determinations:update' submission_pk=determination.submission.id pk=determination.id %}">Edit</a>{% endif %}
```
frjo, line 9: See how the edit button look and is placed on reviews and copy that.

frjo:
@vaibhavmule I rebased this branch of enhancement/add_sent_notice_option_determinations that in turn was rebased of latest master.

Please to a hard reset to get your local branch updated. I think this should work.

~~~~
git fetch origin
git checkout feature/edit-delete-determination
git reset origin/feature/edit-delete-determination --hard
~~~~

Rebasing instead of merging is a bit of a hassel but you get a nicer commit history that is easier to read and understand.

hypha/apply/determinations/forms.py
frjo: They also want all the other text fields.

vaibhavmule: oh okay, Need to add more changes in that case, as form fields are different for 4 workflows.

hypha/apply/determinations/templates/determinations/determination_form.html
```diff
@@ -3,7 +3,7 @@
 {% block header %}
 <div class="admin-bar">
     <div class="admin-bar__inner">
-        <h2 class="heading heading--no-margin">{% if object %}Update Determination draft{% else %}Add Determination{% endif %}</h2>
+        <h2 class="heading heading--no-margin">{% if object %}Update Determination {% if object.is_draft %}draft{% endif %}{% else %}Add Determination{% endif %}</h2>
```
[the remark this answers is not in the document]
vaibhavmule: @frjo Yes, it is the same as the review.
The only difference is here it is `Add Determination` as compare for Review it is `Create Review`
Let me know what do you think of it.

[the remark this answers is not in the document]
vaibhavmule: Done.

hypha/apply/determinations/forms.py
vaibhavmule: I have used existing form with few modifications.

vaibhavmule:
@frjo I have made the requested changes, Please review and let me know.

Issue #1913, opened 2020-04-06
Documentation - Continuation of docs 


abigail-hampson:
- [x] Principles of development  - https://github.com/OpenTechFund/hypha/wiki/System-principles
  - Generic solution
    - Specific yet flexible
  - Security
- [ ] Implementation Details- how the principles are applied (WIP) https://github.com/OpenTechFund/hypha/wiki/Implementation
  - [ ] Infrastructure
    - [x] Media
  - [ ] App
    - [x] Split between apply and public site
- [ ] Specific Architectural details - things that are a little more complex technically
  - [ ] Application staging
  - [x] Storage of media- in implementation details and repo docs
  - [ ] Storage and versioning of application
- [ ] Extending the application
  - [ ] Application types
  - [ ] Workflows
- [ ] Development advice
  - [x] Writing and working with tests (WIP) https://github.com/OpenTechFund/hypha/wiki/Testing

PR to improve the documentation in the repo aligned with changes to wiki: https://github.com/OpenTechFund/hypha/pull/1920

PR #1920, opened 2020-04-09
Docs/updates in line with wiki

todd-dembrey:
Updates to repo in line with changes to wiki as part of #1913


frjo:
@Parbhat What do you think of these changes, can they go in?

Trying to cleanup the PR:s a bit.

Parbhat (commented):
@frjo The changes look good and will be a good addition to the docs. The branch is old so need to first update with the master.

PR #1922, opened 2020-04-09
Improve display of screen application primary action

### LINKED ISSUE #1903: Better Communicate Actions: Screen Applications
- [x] Appears in primary when application has not been screened for staff, text is 'Screen application', uses primary button styling 
- [x] Appears in secondary and is changeable at all times for staff, matches designs below 

 

- [x] It is never visible to applicants, reviewers or partners 

From #1851

peteandrew:
This PR makes the 'Screen application' action prominent when an application hasn't been screened. The 'Screen application' primary action is only displayed when an application's screening status hasn't been set. A 'Change' link has been added to the 'Screening Status' sidebar block to enable screening status to be changed after it has been set.

### DIFF
--- a/hypha/apply/funds/tests/test_views.py
+++ b/hypha/apply/funds/tests/test_views.py
@@ -251,6 +251,20 @@ class TestStaffSubmissionView(BaseSubmissionViewTestCase):
         assert_add_determination_not_displayed(submission, 'Add determination')
         assert_add_determination_not_displayed(submission, 'Complete draft determination')
 
+    def test_screen_application_primary_action_is_displayed(self):
+        # Submission not screened
+        response = self.get_page(self.submission)
+        buttons = BeautifulSoup(response.content, 'html5lib').find(class_='sidebar').find_all('a', text='Screen application')
+        self.assertEqual(len(buttons), 1)
+
+    def test_screen_application_primary_action_is_not_displayed(self):
+        # Submission screened
+        self.submission.screening_status = ScreeningStatusFactory()
+        self.submission.save()
+        response = self.get_page(self.submission)
+        buttons = BeautifulSoup(response.content, 'html5lib').find(class_='sidebar').find_all('a', text='Screen application')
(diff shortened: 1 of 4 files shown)


hypha/apply/funds/tests/test_views.py
```diff
@@ -207,6 +207,30 @@ def test_can_create_project(self):
         self.assertTrue(hasattr(submission, 'project'))
         self.assertEquals(submission.project.id, project.id)
 
+    def test_screen_application_primary_action(self):
```
todd-dembrey, line 210: This should be two tests - one with and one without the screening status.

hypha/apply/funds/tests/test_views.py
```diff
@@ -207,6 +207,30 @@ def test_can_create_project(self):
...
+        def assert_screen_application(content, action_displayed):
+            screen_application_button = '<a data-fancybox data-src="#screen-application" class="button button--bottom-space button--primary button--full-width is-not-disabled" href="#">Screen application</a>'
+
+            sidebar = content[content.find('<aside class="sidebar'):].replace('\\n', '')
```
todd-dembrey, line 214: We have beautifulsoup installed on this project - going to be easier to use it long term. Use that to extract the sidebar and get all anchor elements from within that. You can then confirm that it contains the correct number of "Screen Application" references. 

Copying in the entire HTML fragment feels like it's going to be a pain to maintain long term. 

maybe something like:
```python
buttons = BeautifulSoup(response.content, 'html5lib').find(class_='sidebar').find_all('a', text='Screen applications')
self.assertEqual(len(buttons), 1)
```

hypha/apply/funds/tests/test_views.py
peteandrew: ok, updated

hypha/apply/funds/tests/test_views.py
peteandrew: yes, that makes more sense. I've updated the tests with this

PR #1923, opened 2020-04-09
Enhancement/allow visibility edit on comments

### LINKED ISSUE #944: Allow the role team to change the visibility setting for submission comments

harumhelmy:
Fixes #944 

All users can now edit a comment's visibility as part of the edit comment pagedown form on the Django view of submission details. 

Edited: When users post comments/notes on submissions the React app, they can't set the visibility on those — is this ability something we want added?

### DIFF
--- a/hypha/apply/activity/templatetags/activity_tags.py
+++ b/hypha/apply/activity/templatetags/activity_tags.py
@@ -50,13 +50,14 @@ def display_for(activity, user):
 
     return message_data[ALL]
 
+
 @register.filter
 def visibility_options(activity, user):
     if user.is_apply_staff:
-            return 'applicant, team, reviewers, partners, all'
+        return 'applicant, team, reviewers, partners, all'
     if user.is_reviewer:
-            return 'reviewers, all'
+        return 'reviewers, all'
     if user.is_partner:
-            return 'partners, all'
+        return 'partners, all'
 
     return 'applicant, all'
--- a/hypha/apply/activity/templatetags/activity_tags.py
+++ b/hypha/apply/activity/templatetags/activity_tags.py
@@ -49,3 +49,14 @@ def display_for(activity, user):
         return message_data[TEAM]
 
     return message_data[ALL]
+
+@register.filter
+def visibility_options(activity, user):
+    if user.is_apply_staff:
+            return 'applicant, team, reviewers, partners, all'
+    if user.is_reviewer:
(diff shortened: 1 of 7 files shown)


frjo:
@harumhelmy Regarding the React app it sounds like you need to set API_BASE_URL.

```
$ export API_BASE_URL='http://apply.hypha.test/api'
$ gulp watch:app
```

harumhelmy:
@frjo Haha yeah I just remembered to do that yesterday! It's working now. And I realized there's no way to set visibility on notes when users post them from the React app anyway — is there a need for that? Let me know if that needs to be added.

hypha/apply/activity/templatetags/activity_tags.py
```diff
@@ -49,3 +49,15 @@ def display_for(activity, user):
         return message_data[TEAM]
 
     return message_data[ALL]
+
+
+@register.filter
+def visibility_options(activity, user):
```
frjo, line 55: Can we not reuse the "visibility_choices_for" function in "hypha/apply/activity/models.py"? It has this exact information and I like to avoid duplication.

harumhelmy: Thanks for the review @frjo! Yes, I totally wanted to avoid duplication. I tried using the visibility_choices_for class method in the template, but Django templates only work with class methods/functions that don't accept params. Making it into a filter is the best way I found to get this logic into the Django template, though I agree with you the duplication isn't at all deal. I'll timebox researching another way to do this, maybe about 30 mins. Tagging @rudietuesdays in case Rae has an idea how to do this!

frjo: @harumhelmy You can use "visibility_choices_for " directly in the filter like this.

hypha/static_src/src/javascript/apply/edit-comment.js
```diff
@@ -32,15 +35,36 @@
...
+                <div>Visible to:</div>
+            </div>
+        `;
+
+        const radioButtonsDiv = '<div id="edit-comment-visibility"></div>';
+        let key = '';
+        let label = '';
+        let radioButtons = '';
+
+        $.each(visibilityOptions, function (idx, value) {
```
frjo, line 48: @harumhelmy This seems to work but Javascript is not my strong suit so please take a look at it.

PR #1927, opened 2020-04-13
Improve display of Assign Reviewers action

### LINKED ISSUE #1904: Better Communicate Actions: Assign Reviewers
- [x] Assign reviewers button only shows if the submission does not have all reviewer roles assigned and is in internal review 
- [x] Assign reviewers button appears when it is in the external review stage 
- [x] Staff will always see 'update reviewers' in the reviews box, uses secondary button styling 

From #1851

peteandrew:
This PR makes the 'Assign Reviewers' action prominent when an application is in review stage. When in internal review, the action is displayed until reviewers are assigned to all roles. When in external review, the action is always displayed.

### DIFF
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -53,6 +53,7 @@ from ..workflow import (
     review_statuses,
 )
 from .mixins import AccessFormData
+from .reviewer_role import ReviewerRole
 from .utils import (
     COMMUNITY_REVIEWER_GROUP_NAME,
     LIMIT_TO_PARTNERS,
@@ -631,6 +632,10 @@ class ApplicationSubmission(
             self.live_revision = first_revision
             self.draft_revision = first_revision
             self.save()
+
+    @property
+    def has_all_reviewer_roles_assigned(self):
+        return self.assigned.with_roles().count() == ReviewerRole.objects.count()
 
     @property
     def community_review(self):
--- a/hypha/apply/funds/tests/test_views.py
+++ b/hypha/apply/funds/tests/test_views.py
@@ -328,6 +328,74 @@ class TestStaffSubmissionView(BaseSubmissionViewTestCase):
         assert_create_review_not_displayed(submission, 'Add a review')
         assert_create_review_not_displayed(submission, 'Complete draft review')
 
+    def test_can_see_assign_reviewers_primary_action(self):
+        def assert_assign_reviewers_displayed(submission):
+            response = self.get_page(submission)
+            buttons = BeautifulSoup(response.content, 'html5lib').find(class_='sidebar').find_all('a', class_='button--primary', text='Assign reviewers')
+            self.assertEqual(len(buttons), 1)
+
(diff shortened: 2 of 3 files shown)


hypha/apply/funds/models/submissions.py
```diff
@@ -634,6 +635,12 @@ def save(self, *args, update_fields=list(), skip_custom=False, **kwargs):
             self.draft_revision = first_revision
             self.save()
 
+    @property
+    def has_all_reviewer_roles_assigned(self):
+        all_role_ids = set(ReviewerRole.objects.values_list('id', flat=True))
+        assigned_role_ids = set(self.assigned.with_roles().values_list('role__id', flat=True))
```
todd-dembrey, line 641: `assignedReviewers` is unique together with the role so you should be able to just `count()` roles and assignees here.
 https://github.com/OpenTechFund/hypha/blob/6ea1d55451f5f1d2231605553c0d79833dd62fb0/hypha/apply/funds/models/submissions.py#L962-L963

hypha/apply/funds/templates/funds/includes/actions.html
```diff
@@ -14,6 +14,12 @@ <h5>Actions to take</h5>
 
 <a data-fancybox data-src="#screen-application" class="button button--bottom-space button--primary button--full-width {% if screening_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Screen application</a>
 
+{% if object.in_internal_review_phase or object.in_external_review_phase %}
+    {% if object.in_internal_review_phase and not object.has_all_reviewer_roles_assigned or object.in_external_review_phase %}
```
todd-dembrey, line 18: hmm, can we tidy this up a bit? i think the following might be clearer.
```
{% if object.in_internal_review_phase or object.in_external_review_phase %}
    {% if object.in_external_review_phase or not object.has_all_reviewer_roles_assigned %}
```

hypha/apply/funds/models/submissions.py
peteandrew: yeah, that's better. I've updated that

hypha/apply/funds/tests/test_views.py
```diff
@@ -207,6 +207,79 @@ def test_can_create_project(self):
         self.assertTrue(hasattr(submission, 'project'))
         self.assertEquals(submission.project.id, project.id)
 
+    def test_assign_reviewers_primary_action(self):
```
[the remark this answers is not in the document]
peteandrew: I've changed the button assertions to use BeautifulSoup (that definitely makes more sense than updating all of the html) and I've added tests for applicant and reviewer views. With the staff view, I've split the existing primary action test into can see and can't see. I'm not sure that there's going to be that much value in splitting the tests further than that.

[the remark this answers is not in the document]
todd-dembrey: I think you could split it down, i like to maintain separation so that we dont cross contaminate and can see which state do/dont work rather than bailing early from the assertions.

It works though, approving.

PR #1928, opened 2020-04-14
Make it so reviewers can flag submission.

### LINKED ISSUE #1698: Make user flags available to reviewers

frjo:
Fixes #1698

### DIFF
--- a/hypha/apply/dashboard/views.py
+++ b/hypha/apply/dashboard/views.py
@@ -45,7 +45,19 @@ class MySubmissionContextMixin:
         )
 
 
-class AdminDashboardView(TemplateView):
+class MyFlaggedMixin:
+    def my_flagged(self, submissions):
+        submissions = submissions.flagged_by(self.request.user).order_by('-submit_time')
+        row_attrs = dict({'data-flag-type': 'user'}, **SummarySubmissionsTable._meta.row_attrs)
+
+        limit = 5
+        return {
+            'table': SummarySubmissionsTable(submissions[:limit], prefix='my-flagged-', attrs={'class': 'all-submissions-table flagged-table'}, row_attrs=row_attrs),
+            'display_more': submissions.count() > limit,
+        }
+
+
+class AdminDashboardView(MyFlaggedMixin, TemplateView):
     template_name = 'dashboard/dashboard.html'
 
     def get_context_data(self, **kwargs):
@@ -138,18 +150,8 @@ class AdminDashboardView(TemplateView):
             'open': rounds.open()[:limit],
         }
 
-    def my_flagged(self, submissions):
-        submissions = submissions.flagged_by(self.request.user).order_by('-submit_time')
-        row_attrs = dict({'data-flag-type': 'user'}, **SummarySubmissionsTable._meta.row_attrs)
 
-        limit = 5
-        return {
-            'table': SummarySubmissionsTable(submissions[:limit], prefix='my-flagged-', attrs={'class': 'all-submissions-table flagged-table'}, row_attrs=row_attrs),
-            'display_more': submissions.count() > limit,
-        }
-
-
-class ReviewerDashboardView(MySubmissionContextMixin, TemplateView):
+class ReviewerDashboardView(MyFlaggedMixin, MySubmissionContextMixin, TemplateView):
     template_name = 'dashboard/reviewer_dashboard.html'
 
     def get(self, request, *args, **kwargs):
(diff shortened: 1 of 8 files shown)


hypha/apply/dashboard/views.py
```diff
@@ -163,6 +163,7 @@ def get(self, request, *args, **kwargs):
             'active_statuses_filter': active_statuses_filter,
             'my_submissions': my_submissions,
             'my_inactive_submissions': my_inactive_submissions,
+            'my_flagged': self.get_my_flagged(self.request, qs),
```
frjo, line 166: Is "qs" the correct collection off submissions to use here?

hypha/apply/dashboard/views.py
```diff
@@ -216,6 +217,16 @@ def get_context_data(self, **kwargs):
...
+    def get_my_flagged(self, request, qs):
+        qs = qs.flagged_by(request.user).order_by('-submit_time')
+        row_attrs = dict({'data-flag-type': 'user'}, **SummarySubmissionsTable._meta.row_attrs)
+
+        limit = 5
+        return {
+            'table': SummarySubmissionsTable(qs[:limit], prefix='my-flagged-', attrs={'class': 'all-submissions-table flagged-table'}, row_attrs=row_attrs),
+            'display_more': qs.count() > limit,
+        }
+
```
frjo, line 229: This feels wrong since the exact same function already exist for the admin dashboard.

hypha/apply/flags/views.py
```diff
@@ -26,3 +27,6 @@ def post(self, request, type, submission_pk):
             flag.delete()
 
         return JsonResponse({"result": created})
+
+    def test_func(self):
+        return self.request.user.is_apply_staff or self.request.user.is_reviewer
```
frjo, line 32: Would be nice to have this as a wagtail permission that you set per group/role.

frjo:
@Parbhat Could you take over this? See some comments I made in the PR.

Also the whole "views.py" for "dashboard" feels messy.

Can you clean it up? In my mind it looks like we could hava a base dashboard class with common function and then one for each role with only custom things. Maybe this is best done in a separate PR?

Now they seem to use different names etc. with make it harder to understand.

Parbhat:
@frjo Yeah will continue work on it. Looks like we have two types of flags currently. One for staff and one for the user itself. And currently, only staff can create flags. We want to give similar functionality to reviewers but they can only create flags for themselves. Is this correct?

frjo:
@Parbhat Yes exactly. This PR already adds the needed functionality. But as you see in my comments on my own solution, I'm unsure how good it is.

Also working on this I found the Dashboard views code non optimal.

Parbhat:
@frjo Thanks for confirming. I will review the current PR and look into dashboard views code for refactoring.

hypha/apply/dashboard/views.py
Parbhat: Yeah, but it could have been named better. The admin dashboard view uses [submissions](https://github.com/OpenTechFund/hypha/blob/d7181b723de077d311477b58763f7b5d4bdfa8ad/hypha/apply/dashboard/views.py#L30). Also, this context code should be in [get_context_data](https://github.com/OpenTechFund/hypha/blob/d7181b723de077d311477b58763f7b5d4bdfa8ad/hypha/apply/dashboard/views.py#L210) method instead of get. We will refactor it in dashboard view code refactoring.

hypha/apply/dashboard/views.py
Parbhat: A lot of other methods are duplicated as well like [get_my_submissions](https://github.com/OpenTechFund/hypha/blob/d7181b723de077d311477b58763f7b5d4bdfa8ad/hypha/apply/dashboard/views.py#L196). We will fix it dashboard views refactoring. But okay to duplicate for now.

hypha/apply/flags/views.py
Parbhat: The current code will give reviewers access to update Flags for Staff. As the same `FlagSubmissionCreateView` view is used for flags of type user and staff. Looks like we should create a separate view and create an abstract class for common code. Updating it. We can add permission for flags to certain groups like staff, reviewers from admin. But will also have to add table HTML code to all dashboard templates and flag create button in all submission detail pages. In case one group gets permission like a partner then creating a flag and table should appear in their dashboard.

Parbhat: [SubmissionUserFlaggedView](https://github.com/OpenTechFund/hypha/blob/d7181b723de077d311477b58763f7b5d4bdfa8ad/hypha/apply/funds/views.py#L341) also need updates, it is used when display more options appear.

frjo:
See https://github.com/OpenTechFund/hypha/issues/1936

PR #1932, opened 2020-04-15
Sort by active/inactive status in admin users listing

### LINKED ISSUE #550: Ability to sort/view/filter users by active/inactive and role

Parbhat:
Fixes #550 

Allow admin users to order users listing by status. We can not add order by roles as it gives
 duplicate results due to

>Here, there could potentially be multiple ordering data for each Event; each Event with multiple children will be returned multiple times into the new QuerySet that order_by() creates. In other words, using order_by() on the QuerySet could return more items than you were working on to begin with - which is probably neither expected nor useful.
>Thus, take care when using multi-valued field to order the results. If you can be sure that there will only be one ordering piece of data for each of the items you’re ordering, this approach should not present problems. If not, make sure the results are what you expect. - [docs](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#order-by)

### DIFF
--- a/hypha/apply/users/admin_views.py
+++ b/hypha/apply/users/admin_views.py
@@ -59,10 +59,7 @@ def index(request):
         form = SearchForm(placeholder=_("Search users"))
 
     if not is_searching:
-        users = User.objects.all()
-
-    if 'last_name' in model_fields and 'first_name' in model_fields:
-        users = users.order_by('last_name', 'first_name')
+        users = User.objects.all().order_by('-is_active', 'full_name')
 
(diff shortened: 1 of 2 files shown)


Parbhat:
@frjo Looks good. The default ordering currently used by users index view is from User model meta [ordering = ('full_name', 'email')](https://github.com/OpenTechFund/hypha/blob/d7181b723de077d311477b58763f7b5d4bdfa8ad/hypha/apply/users/models.py#L153). Default ordering by `full_name` make more sense than full_name and email both.

PR #1935, opened 2020-04-16
Moving the edit comment UI to inline in detailed view

### LINKED ISSUE #1528: Edit Note Section in Detailed View
**For Testing**: https://test-apply.opentech.fund/apply/rounds/824/?submission=81

**Describe the bug**
UI issue: in Note section in detailed view: When clicking on "Edit," it's very hard to tell that the place to edit is the submit box, as opposed to the comment box (where the comment already is). Everything greys out and it looks like the page freezes.

Is it possible to make the edit within that comment box itself where the comment already is, as opposed to sending the user back to the submit box? It would be way more intuitive.

**To Reproduce**
Go to 'detailed view'
Click on 'Notes'
Add a comment
Click on 'Edit'

**Expected behavior**
Would be better if the Editing UI was more intuitive. Right now, hard to tell that it sends you back to the lower 'submit' box rather than being able to edit directly in the comment itself. 

**Priority**
Low priority

**Affected roles**
Staff

harumhelmy:
Fixes #1528 

I made some edits to the React app/detailed view of submission notes, so that when a user is editing a comment/note, the rich text editing form appears inline instead of having the users scroll down to the initial comment submit box.

### DIFF
--- a/hypha/static_src/src/app/src/containers/DisplayPanel/index.js
+++ b/hypha/static_src/src/app/src/containers/DisplayPanel/index.js
@@ -15,7 +15,6 @@ import CurrentSubmissionDisplay from '@containers/CurrentSubmissionDisplay'
 import ReviewInformation from '@containers/ReviewInformation'
 import ScreeningOutcome from '@containers/ScreeningOutcome'
 import AddNoteForm from '@containers/AddNoteForm'
-import EditNoteForm from '@containers/EditNoteForm'
 import NoteListing from '@containers/NoteListing'
 import StatusActions from '@containers/StatusActions'
 import Tabber, {Tab} from '@components/Tabber'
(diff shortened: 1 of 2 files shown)


hypha/static_src/src/app/src/containers/DisplayPanel/index.js
```diff
@@ -69,10 +68,7 @@ const DisplayPanel = props => {
         </Tab>,
         <Tab button="Notes" key="note">
             <NoteListing submissionID={submissionID} />
-            {isEditing ? (
-                    <EditNoteForm submissionID={submissionID}/>
-
-                ) : (
+            {isEditing ? null : (
```
harumhelmy, line 71: this makes sure that the comment submit box disappears when a user is editing a comment

frjo:
Have moved this to test now for OTF to test and approve. Hopefully we can deploy later this week.

PR #1945, opened 2020-04-23
Add a submissions result page to get sum and average of values per ro…

### LINKED ISSUE #297: As a manager, I want to see the total number of concept notes and their total value within a defined time range, fund, etc

frjo:
…und/found/status etc.

Fixes #297

Opted for not calculating values for the initial view that shows all submissions. It makes the page slow to load and I assumed it was not an interesting number in any case.


frjo:
@Parbhat Can you take a look at this and see what improvements there are to be made?

PR #1953, opened 2020-04-28
Improve display of Add Review action

### LINKED ISSUE #1905: Better Communicate Actions: Add a review
From #1851

- [x] When an application is moved into "internal review" the button should display for all staff 
  - https://github.com/OpenTechFund/hypha/blob/master/hypha/apply/funds/workflow.py#L974
- [x] When an application is in "external review" the button should display for all external reviewers 
  - https://github.com/OpenTechFund/hypha/blob/master/hypha/apply/funds/workflow.py#L990
- [x] Button should not display when a user has added a review (including if the review was added in a different status) 
- [x] The button should not display once the application moves out of internal review 
- [x] If user has a draft review, button says 'complete draft review' 
- [x] If the user doesn't have a draft review, button says 'add a review' 
- [x] This is never visible to applicants or partners

peteandrew:
This PR makes the 'Review' action prominent when an application in in internal or external review phases. When a user hasn't created a review, 'Create review' is displayed as a primary action. When a user has created a draft review, 'Complete draft review' is displayed. When a user has completed a review, the button is not displayed.

### DIFF
--- a/hypha/apply/review/templates/review/includes/review_button.html
+++ b/hypha/apply/review/templates/review/includes/review_button.html
@@ -3,7 +3,7 @@
     {% if request.user|has_draft:submission or request.user|can_review:submission %}
         <a target="_blank" href="{% url 'apply:submissions:reviews:form' submission_pk=submission.id %}" class="button button--primary {{ class }}">
             {% if request.user|has_draft:submission %}
-                Update draft
+                {{ draft_text|default:"Update draft" }}
             {% elif request.user|can_review:submission %}
                 Create review
             {% endif %}
--- a/hypha/apply/review/templates/review/includes/review_button.html
+++ b/hypha/apply/review/templates/review/includes/review_button.html
@@ -5,7 +5,7 @@
             {% if request.user|has_draft:submission %}
                 {{ draft_text|default:"Update draft" }}
             {% elif request.user|can_review:submission %}
-                Create review
+                Add a review
             {% endif %}
         </a>
     {% endif %}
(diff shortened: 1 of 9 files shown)


hypha/apply/funds/templates/funds/includes/actions.html
```diff
@@ -14,6 +14,10 @@ <h5>Actions to take</h5>
 
 <a data-fancybox data-src="#screen-application" class="button button--bottom-space button--primary button--full-width {% if screening_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Screen application</a>
 
+{% if object.in_internal_review_phase or object.in_external_review_phase %}
+    {% include 'review/includes/review_button.html' with submission=object class="button--full-width button--bottom-space" extended_draft_text=True %}
```
todd-dembrey, line 18: Consider just passing in `draft_text="Complete draft review"` and default the text to the current value.

peteandrew: I've updated that, thanks

hypha/apply/review/templates/review/includes/review_button.html
```diff
@@ -3,7 +3,11 @@
     {% if request.user|has_draft:submission or request.user|can_review:submission %}
         <a target="_blank" href="{% url 'apply:submissions:reviews:form' submission_pk=submission.id %}" class="button button--primary {{ class }}">
             {% if request.user|has_draft:submission %}
-                Update draft
+                {% if draft_text %}
```
todd-dembrey, line 6: consider: `{{ draft_text|default:"Update draft" }}`
but currently ok.
https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#default

PR #1958, opened 2020-04-28
Improve display of 'Add determination' action

### LINKED ISSUE #1907: Better Communicate Actions: Add determination
From #1851 

- [x] When in 'ready for determination' there is an 'add determination' as a primary button 
- [x] This is only visible to staff 
- [x]  If there is a draft determination, button says 'Complete draft determination'

peteandrew:
This PR makes the 'Add determination' action prominent when an application is in the 'ready-for-determination' phase. When a user hasn't created a determination, 'Add determination' is displayed as a primary action. When a user has created a draft determination, 'Complete draft determination' is displayed.

### DIFF
--- a/hypha/apply/determinations/templates/determinations/includes/determination_button.html
+++ b/hypha/apply/determinations/templates/determinations/includes/determination_button.html
@@ -1,6 +1,14 @@
 {% load determination_tags workflow_tags %}
 {% if request.user|show_determination_button:submission %}
-    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width">
-        {% if submission.determination.is_draft %}Update draft{% else %}Add Determination{% endif %}
+    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width {{ class }}">
+        {% if submission.determinations.last.is_draft %}
+            {% if draft_text %}
+                {{ draft_text }}
+            {% else %}
+                Update draft
+            {% endif %}
+        {% else %}
+            Add determination
+        {% endif %}
     </a>
 {% endif %}
--- a/hypha/apply/funds/templates/funds/includes/actions.html
+++ b/hypha/apply/funds/templates/funds/includes/actions.html
@@ -14,6 +14,10 @@
 
 <a data-fancybox data-src="#screen-application" class="button button--bottom-space button--primary button--full-width {% if screening_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Screen application</a>
 
+{% if object.ready_for_determination %}
+    {% include 'determinations/includes/determination_button.html' with submission=object class="button--bottom-space" draft_text="Complete draft determination" %}
+{% endif %}
+
 <a data-fancybox data-src="#update-status" class="button button--primary button--full-width {% if progress_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Update status</a>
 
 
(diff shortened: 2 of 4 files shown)


hypha/apply/funds/templates/funds/includes/actions.html
```diff
@@ -14,6 +14,10 @@ <h5>Actions to take</h5>
 
 <a data-fancybox data-src="#screen-application" class="button button--bottom-space button--primary button--full-width {% if screening_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Screen application</a>
 
+{% if object.ready_for_determination %}
+    {% include 'determinations/includes/determination_button.html' with submission=object class="button--bottom-space" extended_draft_text=True %}
```
todd-dembrey, line 18: as per #1953 - consider vary at the include level

hypha/apply/determinations/templates/determinations/includes/determination_button.html
```diff
@@ -1,6 +1,14 @@
 {% load determination_tags workflow_tags %}
 {% if request.user|show_determination_button:submission %}
-    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width">
-        {% if submission.determination.is_draft %}Update draft{% else %}Add Determination{% endif %}
+    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width {{ class }}">
+        {% if request.user|has_draft_determination:submission %}
```
todd-dembrey, line 4: I'm not sure why you updated this. Determination should be a 1-1 with the submission - so the user is not relevant only that they can work on determinations for this submission.

hypha/apply/determinations/templatetags/determination_tags.py
```diff
@@ -12,3 +12,8 @@ def show_determination_button(user, submission):
         return can_edit_determination(user, submission.determinations.active(), submission)
     except ObjectDoesNotExist:
         return can_create_determination(user, submission)
+
+
+@register.filter
+def has_draft_determination(user, submission):
```
todd-dembrey, line 18: See comment above `show_determination_button` should be enough to identify ability to edit.

hypha/apply/determinations/templates/determinations/includes/determination_button.html
peteandrew: hmm, I was thrown a bit by the existing code around determinations. In determination_block.html there is a `{% for determination in object.determinations.all %}` that implies that there can be multiple determinations for a submission. In fact this only ever shows one determination because if a determination is added when a draft determination already exists, the draft determination is updated. Also this existing conditional doesn't work as `determination` doesn't exist on `submission` so the button always shows `Add Determination`

hypha/apply/determinations/templatetags/determination_tags.py
peteandrew: Removed this and updated determination_button.html to use `submission.determinations.last.is_draft` instead

hypha/apply/determinations/templates/determinations/includes/determination_button.html
```diff
@@ -1,6 +1,14 @@
 {% load determination_tags workflow_tags %}
 {% if request.user|show_determination_button:submission %}
-    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width">
-        {% if submission.determination.is_draft %}Update draft{% else %}Add Determination{% endif %}
+    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width {{ class }}">
+        {% if submission.determinations.last.is_draft %}
```
todd-dembrey, line 4: hmm, well i was incorrect then. good fix :+1:

PR #1965, opened 2020-05-01
Improve display of 'View determination' action

peteandrew:
This PR makes the 'View determination' action prominent when an application has been determined (accepted or rejected). The button is visible to any user that has access to the submission.

### DIFF
--- a/hypha/apply/funds/templates/funds/includes/admin_primary_actions.html
+++ b/hypha/apply/funds/templates/funds/includes/admin_primary_actions.html
@@ -28,6 +28,14 @@
     {% endif %}
 {% endif %}
 
+{% if object.is_finished %}
+    {% with determination=object.determinations.last %}
+    {% if determination %}
+        <a class="button button--primary button--full-width button--bottom-space" href="{% url 'apply:submissions:determinations:detail' submission_pk=object.id pk=determination.id %}">View determination</a>
+    {% endif %}
+    {% endwith %}
+{% endif %}
+
 <a data-fancybox data-src="#update-status" class="button button--primary button--full-width {% if progress_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Update status</a>
 
 
--- a/hypha/apply/funds/templates/funds/includes/generic_primary_actions.html
+++ b/hypha/apply/funds/templates/funds/includes/generic_primary_actions.html
@@ -7,5 +7,13 @@
         {% if object.in_internal_review_phase or object.in_external_review_phase %}
             {% include 'review/includes/review_button.html' with submission=object class="button--full-width button--bottom-space" draft_text="Complete draft review" %}
         {% endif %}
+
+        {% if object.is_finished %}
+            {% with determination=object.determinations.last %}
+            {% if determination %}
+                <a class="button button--primary button--full-width button--bottom-space" href="{% url 'apply:submissions:determinations:detail' submission_pk=object.id pk=determination.id %}">View determination</a>
+            {% endif %}
+            {% endwith %}
+        {% endif %}
     </div>
 {% endif %}
--- a/hypha/apply/funds/tests/test_views.py
+++ b/hypha/apply/funds/tests/test_views.py
@@ -396,6 +396,40 @@ class TestStaffSubmissionView(BaseSubmissionViewTestCase):
         AssignedReviewersFactory(submission=submission, reviewer=ReviewerFactory(), role=reviewer_role)
         assert_assign_reviewers_secondary_displayed(submission)
 
+    def test_can_see_view_determination_primary_action(self):
+        def assert_view_determination_displayed(submission):
+            response = self.get_page(submission)
+            buttons = BeautifulSoup(response.content, 'html5lib').find(class_='js-actions-sidebar').find_all('a', class_='button--primary', text='View determination')
+            self.assertEqual(len(buttons), 1)
+
+        # Phase: accepted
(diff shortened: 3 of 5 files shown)


hypha/apply/funds/tests/test_views.py
```diff
@@ -643,6 +755,7 @@ def test_can_screen_applications_in_final_status(self):
         we can still screen it because we are super user
         """
         submission = ApplicationSubmissionFactory(rejected=True)
+        DeterminationFactory(submission=submission, rejected=True, submitted=True)
```
todd-dembrey, line 758: Why was this needed here?

hypha/apply/determinations/tests/test_views.py
```diff
@@ -460,7 +461,7 @@ def get_kwargs(self, instance):
 
     def test_can_edit_determination(self):
         submission = ApplicationSubmissionFactory(status='rejected')
-        determination = DeterminationFactory(submission=submission)
+        determination = DeterminationFactory(submission=submission, author=self.user, accepted=True, submitted=True)
```
todd-dembrey, line 464: This seems wrong, a submitted determination is not editable i think. Also why did you need to change ownership of the determination, as previously mentioned that isnt important (as long as you are staff).

hypha/apply/determinations/tests/test_views.py
```diff
@@ -60,6 +60,7 @@ def test_can_access_form_if_lead(self):
 
     def test_cant_access_wrong_status(self):
         submission = ApplicationSubmissionFactory(status='rejected')
+        DeterminationFactory(submission=submission, author=self.user, rejected=True, submitted=True)
```
todd-dembrey, line 63: Hmm, i think 
https://github.com/OpenTechFund/hypha/blob/ceb74015590ae24d1412478505faeda6951c080d/hypha/apply/determinations/tests/factories.py#L33
should also pass `submission__rejected=True` so that a submission of the right shape is created.

hypha/apply/funds/templates/funds/includes/admin_primary_actions.html
```diff
@@ -14,6 +14,12 @@ <h5>Actions to take</h5>
 
 <a data-fancybox data-src="#screen-application" class="button button--bottom-space button--primary button--full-width {% if screening_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Screen application</a>
 
+{% if object.is_determined %}
+    {% with determination=object.determinations.final.first %}
```
todd-dembrey, line 18: we want the latest determination here i think so `last`?

hypha/apply/funds/templates/funds/includes/generic_primary_actions.html
```diff
@@ -0,0 +1,13 @@
+{% load primaryactions_tags %}
+
+{% if request.user|should_display_primary_actions_block:object %}
```
todd-dembrey, line 3: This will be fun to maintain & debug. Cant think of something better at this point in time.

hypha/apply/funds/templates/funds/includes/admin_primary_actions.html
```diff
@@ -14,6 +14,12 @@ <h5>Actions to take</h5>
 
 <a data-fancybox data-src="#screen-application" class="button button--bottom-space button--primary button--full-width {% if screening_form.should_show %}is-not-disabled{% else %}is-disabled{% endif %}" href="#">Screen application</a>
 
+{% if object.is_determined %}
```
todd-dembrey, line 17: Whats interesting here is that `is_determined` is coming from the submission state and not from the presence of an actual determination. Which is by design, the determination object is only extra data that doesn't affect the state of the object.

I think what needs to happen is that `object.is_determined` is renamed to something else to reflect the state is final or closed and then we want to be confirming that the determination exists as a separate task. So the submission says that its in a state to show determinations as a primary action and then we check the determinations to confirm they exists.

```
{% if object.is_finished %}
    {% with determination=object.determinations.final.last %}
    {% if determination %}
        ...
    {% endif %}
{% endif %}
```

peteandrew: Yeah, this makes more sense. The unecessary test changes were the result of my not understanding that an application being finished doesn't necessarily mean it will have a determination. I've made this change and reverted the previously changed tests

hypha/apply/funds/tests/test_views.py
peteandrew: Reverted this after making the change below

hypha/apply/determinations/tests/test_views.py
peteandrew: Reverted this after making the change below

hypha/apply/determinations/tests/test_views.py
peteandrew: Reverted this after making the change below

hypha/apply/funds/templates/funds/includes/admin_primary_actions.html
peteandrew: I don't think it would make much difference to use first or last as the view guarentees there's only one determination but I've changed this to be consistent with generic_primary_actions.html

hypha/apply/funds/templates/funds/includes/generic_primary_actions.html
peteandrew: Yeah, I know. Not pretty but I can't think of a better solution either

PR #1976, opened 2020-05-04
Display a warning to admin user in case a page model like Fund, Round can not be deleted due to ProtectedError exception

### LINKED ISSUE #1912: When attempting to delete a round with submissions, 500 error. Catch that and present the user a warning that they can't do that.

### LINKED ISSUE #1971: Delete button for Fund in wagtail admin should not show if there are related rounds
Similar solution to https://github.com/OpenTechFund/hypha/pull/1968 I believe.

Parbhat:
Fixes #1912 Fixes #1971

### DIFF
--- a/hypha/apply/utils/views.py
+++ b/hypha/apply/utils/views.py
@@ -1,13 +1,21 @@
+from django.db.models import ProtectedError
+
 from django.contrib.auth.decorators import login_required
 from django.forms.models import ModelForm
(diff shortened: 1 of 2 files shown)


Parbhat:
Yeah, not a good solution. But a smaller code which can be removed later when fix arrives in Wagtail core. Thanks for the review.

PR #1978, opened 2020-05-05
Do the set_status_after_reviewers_assigned for batch review update also.

### LINKED ISSUE #1977: Status changes not applied when batch update reviewers
The status change after two staff reviewers have been assigned to an application do not work when using the batch function.

frjo:
Fixes #1977

### DIFF
--- a/hypha/apply/funds/views.py
+++ b/hypha/apply/funds/views.py
@@ -90,6 +90,33 @@ from .workflow import (
 )
 
 
+class UpdateReviewersMixin:
+    def set_status_after_reviewers_assigned(self, submission):
+        # Check if two internal reviewers have been selected.
+        internal_reviewers_count = submission.assigned.with_roles().count()
+        if internal_reviewers_count > 1:
+            # Automatic workflow actions.
+            action = None
+            if submission.status == INITIAL_STATE:
+                # Automatically transition the application to "Internal review".
+                action = submission.workflow.stepped_phases[1][0].name
+            elif submission.status == 'proposal_discussion':
+                # Automatically transition the proposal to "Internal review".
+                action = 'proposal_internal_review'
+
+            # If action is set run perform_transition().
+            if action:
+                try:
+                    submission.perform_transition(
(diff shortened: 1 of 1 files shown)


hypha/apply/funds/views.py
```diff
@@ -521,29 +553,8 @@ def form_valid(self, form):
             removed=removed,
         )
 
-        # Check if two internal reviewers have been selected.
-        internal_reviewers_count = form.instance.assigned.with_roles().count()
```
frjo, line 525: @Parbhat If I understand correctly "form.instance" and "self.object", used a few lines down, is the exact same thing here?

hypha/apply/funds/views.py
```diff
@@ -90,6 +90,33 @@
 )
 
 
+class UpdateReviewerActions:
```
frjo, line 93: Is this the best way to do this? I want to avoid having the same logic in two places.

hypha/apply/funds/views.py
Parbhat: Yeah, it will be submission object.

hypha/apply/funds/views.py
```diff
@@ -90,6 +90,33 @@
...
+    def set_status_after_reviewers_assigned(self, submission):
+        # Check if two internal reviewers have been selected.
+        internal_reviewers_count = submission.assigned.with_roles().count()
+        if internal_reviewers_count > 1:
```
frjo, line 97: Reminder to self: Make use of has_all_reviewer_roles_assigned when https://github.com/OpenTechFund/hypha/pull/1927 is merged.

hypha/apply/funds/views.py
Parbhat: Looks good, we can name it a Mixin say UpdateReviewersMixin to convey it is to be used by other views. Also, later common methods can be included in it.

frjo: Renamed, thanks!

PR #1979, opened 2020-05-06
Allow Applicants to save application as a draft

### LINKED ISSUE #309: As an applicant, I'd like the ability create drafts of applications (and create an account, if required)

vaibhavmule:
Fixes #309 

Added Save Draft button on all fund pages.

Free_Fund

When the applicant clicks on the Save Draft button, submission is considered as a `draft`. 
Following things are taken into consideration while building this feature. 

- The applicant can see and edit draft submission made in the dashboard.

Dashboard

- The applicant can edit and submit a submission that becomes `in_discussion` aka INITIAL_STATUS
Editing_free4

- Staff cannot see draft submission made by the applicant also can't see in the submission list.

### DIFF
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -160,6 +160,10 @@ class ApplicationSubmissionQueryset(JSONOrderable):
             Sum('value'),
         )
 
+    def exclude_draft(self):
+        # Applications which have the current stage active (have not been progressed)
+        return self.exclude(status='draft')
+
     def with_latest_update(self):
         activities = self.model.activities.rel.model
         latest_activity = activities.objects.filter(submission=OuterRef('id')).select_related('user')
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -790,13 +790,21 @@ def log_status_update(sender, **kwargs):
     notify = kwargs['method_kwargs'].get('notify', True)
 
     if request and notify:
-        messenger(
-            MESSAGES.TRANSITION,
-            user=by,
-            request=request,
-            source=instance,
-            related=old_phase,
-        )
+        if kwargs['source'] == DRAFT:
+            messenger(
+                MESSAGES.NEW_SUBMISSION,
+                request=request,
+                user=by,
+                source=instance,
+            )
+        else:
+            messenger(
+                MESSAGES.TRANSITION,
+                user=by,
+                request=request,
+                source=instance,
+                related=old_phase,
+            )
 
         if instance.status in review_statuses:
             messenger(
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -45,7 +45,7 @@ from ..blocks import NAMED_BLOCKS, ApplicationCustomFormFieldsBlock
 from ..workflow import (
     COMMUNITY_REVIEW_PHASES,
     DETERMINATION_RESPONSE_PHASES,
-    DRAFT,
+    DRAFT_STATE,
     INITIAL_STATE,
     PHASES,
     PHASES_MAPPING,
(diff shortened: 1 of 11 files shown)


hypha/apply/funds/templates/funds/application_base.html
```diff
@@ -47,6 +47,7 @@ <h3>{% blocktrans %}Sorry this {{ page|verbose_name }} is not accepting applicat
                     {% endif %}
                 {% endif %}
             {% endfor %}
+            <input  class="link link--button-tertiary" type="submit" value="Save Draft" name="draft" />
```
frjo, line 50: I prefer to use "button" tag throughout unless there are good reason to use "input". I know "input" is in use in many places but my plan is to one day clean it up throughout.

Also all draft buttons should have "formnovalidate" added. mora about that in separate comment.

hypha/apply/funds/tests/test_models.py
```diff
@@ -214,6 +214,8 @@ def submit_form(self, page=None, email=None, name=None, user=AnonymousUser(), ig
                 data[field.id] = self.email if email is None else email
             if isinstance(field.block, FullNameBlock):
                 data[field.id] = self.name if name is None else name
+            if draft:
+                data['draft'] = "Save Draft"
```
frjo, line 218: Maybe make this possible to translate. (This is another area that really needs some work overall, a lot of text is not translatable in the Hypha app.)

hypha/apply/funds/templates/funds/application_base.html
vaibhavmule: Done.

hypha/apply/funds/tests/test_models.py
```diff
@@ -9,6 +9,7 @@
 from django.core.exceptions import ValidationError
 from django.test import TestCase, override_settings
 from django.urls import reverse
+from django.utils.translation import gettext
```
frjo, line 12: `from django.utils.translation import gettext_lazy as _` is used in other places.

vaibhavmule: Done.

Parbhat:
@frjo Not sure if I understood it correctly

>When I as an admin/staff try to save a draft application I can not see or view it. I can edit it do. This needs a fix I believe.

Do we want the draft submissions submitted by Admin/Staff to appear in Submissions table? Currently, we do not show Admin submissions in their dashboard like Applicant under "Your active submissions". The submissions table in /apply/submissions/ exclude draft submissions. I think we should only display finally submitted applications there. For staff, if required we can add Your active submissions in their dashboard to allow them to edit their submission.

hypha/apply/funds/tests/test_models.py
vaibhavmule: @frjo I have done but it is failing for linting as `./hypha/apply/funds/tests/test_models.py:417:13: F402 import '_' from line 12 shadowed by loop variable`

Parbhat:
@vaibhavmule Instead of showing the following message after successful draft submission, we should ask them to complete the application later after log-in.

>Thank you for your submission to the Acme Corporation.

So, in the case of draft submissions, we can show a message like

> Your application is saved as a draft. Please note that it is not submitted for review. You can complete your application by following the log-in details emailed to you.

OTF can change it later based on requirements.

This [template](https://github.com/OpenTechFund/hypha/blob/32679916c2f9e5e82322296b14966659b5d708b9/hypha/apply/funds/templates/funds/application_base_landing.html) will need updates.

hypha/apply/funds/models/submissions.py
```diff
@@ -770,13 +774,21 @@ def log_status_update(sender, **kwargs):
...
-            source=instance,
-            related=old_phase,
-        )
+        if kwargs['source'] == DRAFT_STATE:
+            messenger(
+                MESSAGES.NEW_SUBMISSION,
+                request=request,
+                user=by,
+                source=instance,
+            )
```
frjo, line 783: Why do we add the notification here but disable it below in "render_landing_page"? Want to understand the code.

Parbhat: Currently, in production, we notify OTF slack for a new submission, other messages using NEW_SUBMISSION. But now if the user saves the application as a draft then we do not send NEW_SUBMISSION message. Therefore disabled in "render_landing_page". But it will be sent later when the user edits the draft application and send for review.

frjo: Now I understand, thanks!

PR #1995, opened 2020-05-12
Give user login option before raising 403

Parbhat:
`UserPassesTestMixin` used without `login_required` method decorator, raise 403 Permission Denied without giving user option to login first. It seems better to give the login option before checking permissions.

### DIFF
--- a/hypha/apply/review/views.py
+++ b/hypha/apply/review/views.py
@@ -58,6 +58,7 @@ def get_fields_for_stage(submission):
         return forms[0].form.form_fields
 
 
+@method_decorator(login_required, name='dispatch')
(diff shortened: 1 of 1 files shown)


frjo:
@Parbhat Excellent, should it not also be added for ReviewOpinionFormView and ReviewDisplay?

Parbhat:
@frjo No, `ReviewOpinionFormView` and `ReviewDisplay` are not used directly. But are called in [ReviewDetailView](https://github.com/OpenTechFund/hypha/blob/cf47517f16710e2be12dcab50da3d7e0826319ac/hypha/apply/review/views.py#L327) which has a `login_required` decorator.

PR #1996, opened 2020-05-13
Add pattern to ValueBlock to only allow number input.

frjo:
Adds a "'pattern=[0-9]+" attribute to the value field so only numbers are allows as input.


frjo:
@Parbhat  You are of course correct, will close this. It's by design that the numbers type allow commas and periods in number like 100.000.00. Will open a new issue to dela with that. Thanks!

PR #1997, opened 2020-05-13
Use db functions to calculate submission results. Performs significantly better.

### LINKED ISSUE #1999: Clean db value data and use db function for result calculation
Using db functions for calculations all make this feature far more performant, at least around 5x.

For this to work the db needs to have clean values in the db. The linked PR has a script to clean old values.

An issue is that file the correct number fined will make sure only positive numbers are submitted from now on it does allow commas and periods as thousand separators etc.

How can we handle this?

frjo:
Fixes #1999

We also need to do a one off cleanup of existing values in the db.

```
from hypha.apply.funds.models import ApplicationSubmission
submissions = ApplicationSubmission.objects.all().order_by('pk')
for s in submissions:
	try:
		value = s.data('value')
	except KeyError:
		pass
	else:
		value = str(value)
		value = value.replace('USD', '')
		value = value.replace('$', '')
		value = value.replace('-', '')
		value = value.replace(',', '')
		try:
			value_as_float = float(value)
		except (TypeError, ValueError):
			value = '0'
	finally:
		if s.data('value') and s.data('value') != value:
			s.form_data['value'] = value
			s.save(skip_custom=True)
```

### DIFF
--- a/hypha/apply/funds/templates/funds/submissions_result.html
+++ b/hypha/apply/funds/templates/funds/submissions_result.html
@@ -26,9 +26,12 @@
     <h4 class="heading heading--normal">Filter submissions to calculate values</h4>
     {% include "funds/includes/table_filter_and_search.html" with filter_form=filter_form search_term=search_term use_search=True filter_action=filter_action use_batch_actions=False filter_classes="filters-open" %}
     <div class="grid">
-        <div><strong>Number of submission:</strong> {{ object_list.count }}</div>
+        <div><strong>Number of submission:</strong> {{ count_values }}{% if not count_values == object_list.count %} ({{ object_list.count }}){% endif %}</div>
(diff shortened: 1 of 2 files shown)


frjo:
@Parbhat What do you think of the command I pasted in above to cleanup the db? Looks safe to use?

Parbhat:
Yeah, looks good. Btw only older submissions may have values other than a number. As `ValueBlock` widget was changed to `NumberInput` from quite some time https://github.com/OpenTechFund/hypha/commit/fe4e4328d2a6f46c29f5a842cf341b02ad6a9c87

frjo:
The live db has now been cleaned. This is a one time thing to clean out old badly formatted values that was submitted before the value field was turned in to a number field.

PR #2007, opened 2020-05-20
Feature/staff assignments view

### LINKED ISSUE #1224: Staff assignment view to see staff current workload
- Account name or name of person: Di

We need to count how many reviews have been assigned to a specific person. Example: How many reviews Di has been assigned? What is the number of applications she is the primary reviewer and the number of applications she is the secondary reviewer?

Feature might be related to creating tags to filter or sort applications into sub-categories

We currently have filters for reviewers, but this filter does not distinguish primary from secondary. We require each person to sign up for about the same amount of primary and secondary to ensure equity and even distribution of applications. 

 #613 is a suggested feature to randomly assign reviewers. This new feature request might be similar, however, reviewers with expertise in a certain area are more likely to gravitate towards certain themes. A random assignment would not take this into consideration. 

**Priority**

**Affected roles**
- Staff

**Ideal deadline**
Q4

frjo:
Fixes #1224

### DIFF
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -20,8 +20,8 @@ from hypha.apply.utils.image import generate_image_tag
 from hypha.images.models import CustomImage
 
 from .models import ApplicationSubmission, Round, ScreeningStatus
-from .workflow import STATUSES, get_review_active_statuses
 from .widgets import Select2MultiCheckboxesWidget
+from .workflow import STATUSES, get_review_active_statuses
 
 User = get_user_model()
 
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -14,13 +14,13 @@ from django_tables2.utils import A
 from wagtail.core.models import Page
 
 from hypha.apply.categories.models import MetaTerm
-from hypha.apply.funds.models import ApplicationSubmission, Round, ScreeningStatus
-from hypha.apply.funds.workflow import STATUSES, get_review_active_statuses
 from hypha.apply.review.models import Review
 from hypha.apply.users.groups import STAFF_GROUP_NAME
 from hypha.apply.utils.image import generate_image_tag
 from hypha.images.models import CustomImage
 
+from .models import ApplicationSubmission, Round, ReviewerRole, ScreeningStatus
+from .workflow import STATUSES, get_review_active_statuses
 from .widgets import Select2MultiCheckboxesWidget
 
 User = get_user_model()
@@ -443,3 +443,18 @@ class ReviewerLeaderboardDetailTable(tables.Table):
         order_by = ('-created_at',)
         attrs = {'class': 'all-reviews-table'}
         empty_text = _('No reviews available')
+
+
+class StaffAssignmentsTable(tables.Table):
+    full_name = tables.LinkColumn('funds:submissions:list', orderable=True, verbose_name="Staff", attrs={'td': {'class': 'title'}})
+    reviewer_roles = ReviewerRole.objects.all().order_by('order')
+    role1 = tables.Column(verbose_name=reviewer_roles.first())
+    role2 = tables.Column(verbose_name=reviewer_roles.last())
+
+    class Meta:
+        model = User
+        fields = [
+            'full_name',
+        ]
+        attrs = {'class': 'all-reviews-table'}
+        empty_text = _('No staff available')
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -6,6 +6,7 @@ import django_tables2 as tables
 from django import forms
 from django.contrib.auth import get_user_model
 from django.db.models import F, Q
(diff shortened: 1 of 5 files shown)


hypha/apply/funds/views.py
```diff
@@ -1230,9 +1235,29 @@ class ReviewerLeaderboardDetail(SingleTableMixin, ListView):
...
+    template_name = 'funds/staff_assignments.html'
+
+    def get_queryset(self):
+        # Only list staff.
+        return self.model.objects.staff()
+
+    def get_table_data(self):
+        reviewer_roles = ReviewerRole.objects.all().order_by('order')
+        reviewer_role_id_1 = reviewer_roles.first().id
+        reviewer_role_id_2 = reviewer_roles.last().id
```
frjo, line 1259: @Parbhat I would like to dynamically create a column for each reviewer role. OTF happens to have two but the system allows you to create an arbitrary number of them. Any neat way to do this?

hypha/apply/funds/views.py
```diff
@@ -1230,9 +1235,29 @@ class ReviewerLeaderboardDetail(SingleTableMixin, ListView):
...
+        return super().get_table_data().annotate(
+            role1=Count('assignedreviewers', filter=Q(assignedreviewers__role_id=reviewer_role_id_1)),
+            role2=Count('assignedreviewers', filter=Q(assignedreviewers__role_id=reviewer_role_id_2)),
+        )
```
frjo, line 1263: @Parbhat I would like to filter this down to only include assigned reviewers on submissions that are in status "active_statuses". Any tips?

frjo: @Parbhat Here also I would like to accommodate an arbitrary number of reviewer roles if possible.

hypha/apply/funds/tables.py
```diff
@@ -443,3 +443,18 @@ class Meta:
         order_by = ('-created_at',)
         attrs = {'class': 'all-reviews-table'}
         empty_text = _('No reviews available')
+
+
+class StaffAssignmentsTable(tables.Table):
+    full_name = tables.LinkColumn('funds:submissions:list', orderable=True, verbose_name="Staff", attrs={'td': {'class': 'title'}})
```
frjo, line 449: @Parbhat I would like to add a query to this linked column so the view gets filtered on the reviewer.

`…/apply/submissions/all/?reviewers=123`

frjo: or easier to just create a separate view for this?

hypha/apply/funds/views.py
Parbhat: `django-tables2` Table has an [extra_columns](https://django-tables2.readthedocs.io/en/latest/pages/api-reference.html#table) parameter to dynamically add columns. As we use `SimpleTableMixin`, these can be included in [get_table_kwargs](https://github.com/jieter/django-tables2/blob/6f5b5b6b228c3fc3a2491b83fe2c20f0e744dd01/django_tables2/views.py#L126) method.

hypha/apply/funds/views.py
Parbhat: Similar to the role, we can add a [submission](https://github.com/OpenTechFund/hypha/blob/32679916c2f9e5e82322296b14966659b5d708b9/hypha/apply/funds/models/submissions.py#L967) query to check status is in the active statuses.

Parbhat: Looks like we do not need to annotate here as can use `extra_columns` parameter. More details in another comment.

hypha/apply/funds/tables.py
```diff
@@ -443,3 +443,18 @@ class Meta:
...
+    reviewer_roles = ReviewerRole.objects.all().order_by('order')
+    role1 = tables.Column(verbose_name=reviewer_roles.first())
+    role2 = tables.Column(verbose_name=reviewer_roles.last())
```
Parbhat, line 452: This code will run only when the code loads but will not update when data changes in db. We may not need this by using `extra_columns`. More details in another comment.

hypha/apply/funds/tables.py
Parbhat: For filtering, we can use `django-filters`'s FilterView like https://github.com/OpenTechFund/hypha/blob/32679916c2f9e5e82322296b14966659b5d708b9/hypha/apply/funds/views.py#L1009

frjo:
@Parbhat Could your team take this on? The issue is about filtering on primary and secondary reviewer but I want to test this staff assignments table and see if that does not solve the need OTF staff have in a better way.

They want to get a picture of the workload put on staff. So the table should list all staff and what current submissions they are assigned to and it what reviewer role.

Parbhat:
@frjo Yeah, will pick it.

Parbhat:
@frjo PR updated with some updates. Can you check if it meets requirements?

frjo:
@Parbhat Looks perfect, neat and easy to understand as all your code.

PR #2018, opened 2020-05-27
Always use gettext_lazy.

frjo:
I noticed that `ugettext_lazy` is deprecated in Django 3 and only needed if the code support Python 2 from what I understand. Use `gettext_lazy` instead.

I also replaced all calls to `ugettext` and `gettext` with `gettext_lazy`.

### DIFF
--- a/hypha/apply/categories/blocks.py
+++ b/hypha/apply/categories/blocks.py
@@ -1,6 +1,6 @@
 from django import forms
 from django.utils.functional import SimpleLazyObject, cached_property
-from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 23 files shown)


frjo:
@Parbhat Did I get it right?

frjo:
@Parbhat Updated all views to use `gettext`. How about table code?

Parbhat:
@frjo Tables file can use the lazy version. `admin_views.py` [file](https://github.com/OpenTechFund/hypha/blob/fed7650b98cbd79e18c63b713c328621346d80f6/hypha/apply/funds/admin_views.py#L4) can also use `gettext`.

PR #2023, opened 2020-06-01
Upgrade to Wagtail 2.9

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/2011

In order to upgrade to Wagtail 2.9:

- We need to use `Site.find_for_request(request)` in place of `request.site` in Python files and tag `{% wagtail_site %}` in templates.

- And then remove `wagtail.core.middleware.SiteMiddleware` from the settings.


sks444:
Thanks @frjo.

PR #2033, opened 2020-06-08
Add multi inputs single line text field and render url/e-mail as links 

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/1954 Fixes https://github.com/OpenTechFund/hypha/issues/2046

This is an MVP with fixed number of inputs(that can be taken from the user for a single field) defined by the admin.

Screenshot 2020-06-08 at 1 39 37 AM
Screenshot 2020-06-08 at 1 35 27 AM
Screenshot 2020-06-08 at 1 36 17 AM

Next we will have to have button to give user an option to add that many inputs dynamically.


Parbhat:
@sks444 If there is no answer by the applicant on multi char inputs field, then the detail view shows **''** to staff. The system currently shows `-` in case a field has no answer. So, we can follow the same pattern here as well.

https://github.com/OpenTechFund/hypha/blob/a0f9dcaf889e92f535a76dd082e5e816a3bd0af3/hypha/apply/funds/models/mixins.py#L232

sks444:
@Parbhat Added a check to not include empty answers from the list: `data = ', '.join(filter(None, answers))`. Now if there is no other answers it will just show the first answer without the whitespace and the commas.

Parbhat:
@sks444 There is an API endpoint for a submission detail. We need to update the Submission detail API endpoint (`/api/v1/submissions/<id>/`) on apply site for multi-input data. Currently, it displays None. We need to make changes in [serialize](https://github.com/OpenTechFund/hypha/blob/a0f9dcaf889e92f535a76dd082e5e816a3bd0af3/hypha/apply/funds/models/mixins.py#L216) method similar to changes made in `render_answer`.

https://github.com/OpenTechFund/hypha/blob/a0f9dcaf889e92f535a76dd082e5e816a3bd0af3/hypha/apply/api/v1/serializers.py#L143

Parbhat:
@frjo This PR is ready for review.

Parbhat:
Thanks for the review. For

>Can we render e-mail and URL as clickable links with a target blank and rel="noopener noreferrer"?

It will be a better experience for staff if the links and emails are clickable. Btw the normal [block](https://github.com/OpenTechFund/hypha/blob/fa75d959e7f9e71f4bee7f1d71a3c2cff95c1366/hypha/apply/stream_forms/blocks.py#L134) for URL and Email also give the text output. So, maybe we handle it in a separate issue. Or handle in this PR.

>How mush work is it to add a "Add new item" button that revels another field for the user instead of showing all the empty fields at once?

It should not be much work. We can add some JS to allow the applicant to add inputs until the max number of inputs available for that field.

frjo:
Since the normal block also render plain text we can take it all in a separate issue. I will create an issue and assign to you.

Adding a "Add more" (or something like that) button that revels more fields up to the maximum would be nice to have in this PR. The application forms are big as they are so anything that can make them shorter and more manageable is good.

I think the JS only need to revel the fields, they can be in the DOM just as now, just hidden.

Parbhat:
Yeah, it would be better to have "Add new item" button as part of this PR. JS only solution would be good. @sks444

sks444:
@Parbhat Added JS with a button to reveal more fields up to the maximum number of inputs submitted by admin. Let me know your views.

Parbhat:
@frjo Yeah, we are working on these issues. I also had a chat with @sks444 about the empty space. But in my review, I see a smaller empty space equal to the height of hidden fields and not this large space. Some complexity will be there as we need to track which fields should be displayed. Also, there can be multiple multi-input fields on the form. Also, some CSS work is required to show button inline with the first input. There is another issue to resolve in edit form of submission which is

>But technically if user has submitted two inputs for multi inputs, we should be showing two inputs initially in the edit form with their data, but the fact that edit form also uses fields.html all the inputs will be hidden except one.

Parbhat:
@frjo Updated this PR for rendering URL/e-mail as links as it seems better to roll out in one go instead of separate PR #2060

frjo:
Deployed to test now, thanks.

PR #2039, opened 2020-06-09
Allow draft to be saved without filling reuqired fields

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/2012

fields from blocks inheriting [ApplicationMustIncludeFieldBlock](https://github.com/OpenTechFund/hypha/blob/a0f9dcaf889e92f535a76dd082e5e816a3bd0af3/hypha/apply/funds/blocks.py#L22) should required to create account for the user, even if it's a draft submission.

### DIFF
--- a/hypha/apply/stream_forms/models.py
+++ b/hypha/apply/stream_forms/models.py
@@ -9,6 +9,7 @@ from .blocks import (
     GroupToggleEndBlock,
     TextFieldBlock,
 )
+from hypha.apply.funds.blocks import EmailBlock, FullNameBlock, TitleBlock
 from .forms import BlockFieldWrapper, PageStreamBaseForm
 
 
@@ -42,7 +43,7 @@ class BaseStreamForm:
     def get_defined_fields(self):
         return self.form_fields
 
-    def get_form_fields(self):
+    def get_form_fields(self, draft=False):
         form_fields = OrderedDict()
         field_blocks = self.get_defined_fields()
         group_counter = 1
@@ -50,9 +51,10 @@ class BaseStreamForm:
         for struct_child in field_blocks:
             block = struct_child.block
             struct_value = struct_child.value
-
             if isinstance(block, FormFieldBlock):
                 field_from_block = block.get_field(struct_value)
+                if draft and not isinstance(block, (EmailBlock, FullNameBlock, TitleBlock)):
+                    field_from_block.required = False
                 field_from_block.help_link = struct_value.get('help_link')
                 field_from_block.group_number = group_counter if is_in_group else 1
                 if isinstance(block, GroupToggleBlock) and not is_in_group:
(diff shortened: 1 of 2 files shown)


hypha/apply/funds/templates/funds/applicationsubmission_form.html
```diff
@@ -24,7 +24,11 @@ <h2 class="heading heading--no-margin">Editing: {{ object.title }}</h2>
                 {% endif %}
             {% endfor %}
             {% for button_name, button_type, button_value in buttons %}
-                <input class="button button--submit button--top-space button--{{ button_type }}" type="submit" name="{{ button_name }}" value="{{ button_value }}" />
+                {% if button_value == 'Save Draft' %}
+                    <input class="button button--submit button--top-space button--{{ button_type }}" type="submit" name="{{ button_name }}" value="{{ button_value }}" formnovalidate/>
+                {% else %}
+                    <input class="button button--submit button--top-space button--{{ button_type }}" type="submit" name="{{ button_name }}" value="{{ button_value }}" />
+                {% endif %}
```
frjo, line 31: Can't we do a simple `{% if button_value == 'Save Draft' %} formnovalidate{% endif %}`, less code.

hypha/apply/stream_forms/models.py
```diff
@@ -74,8 +77,8 @@ def get_form_fields(self):
 
         return form_fields
 
-    def get_form_class(self):
-        return type('WagtailStreamForm', (self.submission_form_class,), self.get_form_fields())
+    def get_form_class(self, draft=False):
+        return type('WagtailStreamForm', (self.submission_form_class,), self.get_form_fields(draft))
```
Parbhat, line 81: @sks444 Ideally `stream_forms` app should be considered like a third party app and funds related changes should happen in the apply app. So, we should try to not make changes here. Instead, we can make changes in [SubmittableStreamForm](https://github.com/OpenTechFund/hypha/blob/a0f9dcaf889e92f535a76dd082e5e816a3bd0af3/hypha/apply/funds/models/utils.py#L58) which is inherited by [RoundBase](https://github.com/OpenTechFund/hypha/blob/a0f9dcaf889e92f535a76dd082e5e816a3bd0af3/hypha/apply/funds/models/applications.py#L145). By overriding the `get_form_class` method in it.

Parbhat: Looks like we will have to copy all the code in `get_form_fields` method if we override it in SubmittableStreamForm class. So, we can send the list of blocks `(EmailBlock, FullNameBlock, TitleBlock)` which should not have required False along with the draft value which calling `get_form_class(draft)` method. Instead of hard coding the block names here.

sks444: Yeah we might have to copy all the code which doesn't look good.

Idea of sending the list of blocks as tuple sounds good. But it feels a bit weird to send all the those block names to get_form_class, that condition will be dependent on the draft value of it anyway. I was thinking if there is way to only check for the block from which it is inherited `ApplicationMustIncludeFieldBlock` and not do required False for them.

Parbhat: Good idea. Checking `ApplicationMustIncludeFieldBlock` is better than option than checking individual blocks like `EmailBlock`.

frjo:
Just tested and it works as it should for me,  more to do or is it done?

Parbhat:
@frjo It is completed. Thanks for testing.

Parbhat:
Thank you.

Issue #2054, opened 2020-06-16
Setting for Salesforce Integration


maxpearl:
Suggestion: Add a setting under Wagtail admin: /admin/settings/utils/systemmessagessettings/2/ to activate salesforce integration. This would requires modifying the class SystemMessagesSettings(BaseSetting) in hypha/public/utils/models.py, giving something like this:

![Screenshot_2020-06-16_12-16-46](https://user-images.githubusercontent.com/144151/84818429-55260b80-afcb-11ea-89fe-afb5886cd40b.png)

maxpearl:
[Here's a potential pull request](https://github.com/maxpearl/hypha/commit/bac7cb9543845683a0f55715409527f08ca833ab) but I'd like to wait until we have a new branch for this work.

frjo:
@maxpearl I think I would prefer to add this as an environment variable. OR is it something an site admin will need to turn on/off from time to time?

The API keys will need to be environment variables in any case and a developer will need set it all up.

Maybe we could use the existens of an Salesforce API be the toggle itself? If the key(s) is preset the integration is active? Or if you see a usecase for a separate toggle we add that of course.

maxpearl:
I'm adding a combination of an environment variable and an additional settings file, as there are a lot of settings to include.

PR #2056, opened 2020-06-17
Feature/salesforce integration

maxpearl:
Addresses #2054

Adds an environment variable: SALESFORCE_INTEGRATION as well as a new optional settings file.


hypha/settings/dev.py
```diff
@@ -145,3 +145,11 @@
...
     'STATS_FILE': os.path.join(BASE_DIR, './hypha/static_compiled/app/webpack-stats.json'),
 })
+
+# Salesforce Integration
+if 'SALESFORCE_INTEGRATION' in env and env['SALESFORCE_INTEGRATION']:
+    try:
+        from .salesforce import *
+    except ImportError:
+        pass
+    DATABASES.update(SALESFORCE_DB)
```
frjo, line 155: Since this needs to be on all environments I thinks it better to put it in `base.py`. Then we can add it only once.

hypha/settings/salesforce.py.example
```diff
@@ -0,0 +1,14 @@
+SALESFORCE_DB = {
+    'ENGINE': 'salesforce.backend',
+    'CONSUMER_KEY': 'Consumer Key from Salesforce App',
+    'CONSUMER_SECRET': 'Consumer Secret from Salesforce APP',
+    'USER': 'user@email.com',
+    'PASSWORD': 'Password and security token concatenated',
+    'HOST': 'https://login.salesforce.com' # Or test or URL of salesforce org
+}
```
frjo, line 8: I think we should make all these environment variables. On hosting environments like Heroku you can't upload a separate configuration file but setting environment variables is easy and reasonable secure.

frjo: Having the option to use environment variables or a settings file might be the absolut best.

maxpearl:
I added a new commit which moves all of the variables to the environment. I added a sf.env file to docker-compose for sensitive data, and added that to .gitignore.

hypha/settings/base.py
```diff
@@ -201,12 +203,34 @@
...
+    DATABASE_ROUTERS = [
+        "salesforce.router.ModelRouter"
+    ]
+else:
+    DATABASES = {
+        'default': dj_database_url.config(
+            conn_max_age=600,
+            default=f"postgres:///{APP_NAME}"
+        )
+    }
```
frjo, line 234: I would prefer to put all the Salesforce settings in a separate section. Something like this.

~~~~
# Salesforce integration

if env.get('SALESFORCE_INTEGRATION', 'false').lower().strip() == 'true':
    DATABASES = {
        **DATABASES,
        'salesforce': {
            'ENGINE': 'salesforce.backend',
            'CONSUMER_KEY': env.get('SALESFORCE_CONSUMER_KEY', ''),
            'CONSUMER_SECRET': env.get('SALESFORCE_CONSUMER_SECRET', ''),
            'USER': env.get('SALESFORCE_USER', ''),
            'PASSWORD': env.get('SALESFORCE_PASSWORD', ''),
            'HOST': env.get('SALESFORCE_LOGIN_URL', '')
        }
    }

    SALESFORCE_QUERY_TIMEOUT = (30, 30)  # (connect timeout, data timeout)

    DATABASE_ROUTERS = [
        "salesforce.router.ModelRouter"
    ]
~~~~

.gitignore
```diff
@@ -12,6 +12,7 @@
 node_modules/
 /npm-debug.log
 /.idea/
+sf.env
```
frjo, line 15: Maybe name this "salesforce.env" just to be explicit.

frjo:
@maxpearl You have write access to this repo so you do not need to keep things in your own fork. Feel free to commit directly to the Hypha salesforce-integration branch.

.gitignore
maxpearl: Good idea.

maxpearl:
Thanks, but I'd rather use this workflow so I can get comments on my code before I add them to the main repo, if that's OK with you. But I'm happy to merge the pull requests after we decide the code is going to work.

PR #2059, opened 2020-06-24
Disable view label from summary card when there is no data to view

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/1884

Considering https://github.com/OpenTechFund/hypha/issues/1884 as reference, I think view label should also be disabled for the other cards when there is no data to view.


frjo:
@sks444 Having things work the same everywhere is good indeed. Can you do it in this PR or better to open a separate issue for it?

sks444:
@frjo I have also included the changes for other cards in this pr.

frjo:
Deployed to test, thanks!

PR #2060, opened 2020-06-24
Make output from textfields of type url/e-mail render as links

### LINKED ISSUE #2046: Make output from textfields of type url/e-mail render as links
This is the fields that can be added to application forms. The "Text fields (single line)", both normal and the new multiple input version, can be set to URL or E-mail type.

In these cases it would be nice if they render as clickable links.

For e-mail as standard mailto. And for URL a link with target blank and rel="noopener noreferrer"

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/2046

This pr is rebased on top of https://github.com/OpenTechFund/hypha/pull/2033 as it needed some modification on in the code of that branch:

We need to render multi-input answers individually in order to make each link/email clickable in itself. Otherwise data would come something like this:

```
<a class="link" href="https://twitter.com, https://google.com" target="_blank" rel="noopener noreferrer">https://twitter.com, https://google.com</a>
```
Screenshot 2020-06-25 at 1 53 07 AM

---

I have test answers rendered by each of the field with my current implementation and it works well.

Screenshot 2020-06-25 at 1 29 51 AM


hypha/apply/templates/forms/includes/field.html
```diff
@@ -25,7 +25,10 @@
     {% endif %}
 
     {% if field.field.help_link %}
+        {% if field.field.only_input %}
```
Parbhat, line 28: Seems like a leftover from old work as we do not use `only_input` now and separate `multi_input_field.html` is used.

Parbhat (approved):
Looks great. Can you add these changes in PR #2033 and close this PR? As it would be good to roll out in single go.

sks444:
Closing this pr, updated these changes in multi-input pr itself https://github.com/OpenTechFund/hypha/pull/2033

PR #2066, opened 2020-07-01
Allow determination forms to be staff created with stream fields

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/1562

**Steps to test this feature:**
### 1. Creating Determination forms in Wagtail

- Create determination form from wagtail. If your application workflow is going to be of two stage create two determination forms: Concept and Proposal.
   - While creating determination form if you want group your form fields under a certain title(similar to old determination form), then you can make use of `Section text/header` block to create titles.
- Once the forms are created they can be added to funds similar to review forms:

Screenshot 2020-07-01 at 4 19 41 PM

   - And they will be automatically copied while creating rounds for that funds, similar to other forms.

### 2. Adding determination to application submissions
- All determination functionality(i.e. Add, Edit, Save Draft, etc.) works in same way as the old determinations.

### 3. Handling old and active rounds:
- Existing rounds(i.e rounds which are closed or active rounds or the rounds which doesn't have wagtail created determination forms attached to it) will use old determination forms for Create, Edit and Detail view as expected.

### 4. Batch Determinations
- Batch determinations can also be created in same way, but note that we can not select two different submissions(I.e. One submission is using old determination forms and another is using new determination forms attached to it.) to create batch determinations.


hypha/apply/funds/models/submissions.py
```diff
@@ -536,6 +536,10 @@ def get_from_parent(self, attribute):
             # We are a lab submission
             return getattr(self.page.specific, attribute)
 
+    @property
+    def is_determination_form_attached(self):
+        return self.get_from_parent('determination_forms').count() > 0
```
Parbhat, line 541: Working smoothly, great work. Noticed an issue with Lab. A lab works differently from Funds. A Fund has Rounds. So, we are able to use new determination streamfield forms for new Rounds and current active/old rounds can use old forms because they do not have the determination streamfield forms set. And there is no way to edit round as well. So, we are safe here. But Labs do not have Rounds. So, In admin when a staff user attaches new determination streamfield form to a lab, old submissions for a lab also start showing new forms instead of old. So, we need a come up with an approach for labs.

Parbhat:
@frjo The PR is almost ready and needs updates related to labs as commented above. This implementation does not need data migrations because we are able to use current normal determination forms and stream field-based forms together. So, when we deploy it, current active rounds will continue to work like before. The staff can create new streamfield based determination forms and set it in the Fund page. So, new Rounds will start using new streamfield based forms. Old Rounds will continue to show data/forms like before. Basically we are handling which forms to use based on condition (streamfield based form is set or not). And use current normal determination forms or new streamfield based forms accordingly. Some queries

1. Determination forms are currently a required field while editing Fund Pages like review forms. Should we make them optional so if a funds/lab do not need streamfield based determination forms can continue to use old forms?

2. There are seed commands in codebase like [seed_concept_note.py](https://github.com/OpenTechFund/hypha/blob/0e0df9e730dd1a3bfdd3dc771068f46bab1f3560/hypha/apply/funds/management/commands/seed_concept_note.py#L34) and many others. Do we use these and should we update them for new streamfield based determination forms? @sks444 has made some changes in some of these commands but thought to confirm should we update these or left untouched?

frjo:
@Parbhat Sounds like an excellent approach! Will test this week or on Monday.

1. The old determinations forms should be considered depreciated so the determinations forms should be required fields just as you have done.
2. The sees and other migration scripts are only for past submissions from the old Drupal system so leave them untouched please. There are a small amount of data we should import from the old system that was generated in the short transition period but I have not taken time to do it yet.

Parbhat:
Yeah, and once all rounds/labs start to use new streamfield based forms. Then we can remove the old normal forms code and only use details views and no determination forms for older submissions.

Parbhat:
**Migrations note**: We will need to update the migration name (`funds/migrations/0075_add_determination_stream_field_forms.py`) based on which PR merged first (this or #2033). @sks444 Master branch has [migration](https://github.com/OpenTechFund/hypha/tree/master/hypha/apply/funds/migrations) 0075, so we can update the migration in this PR to 0076. Then based on which PR gets out first, we can update accordingly.

PR #2084, opened 2020-07-13
Update to latest django-two-factor-auth with followup up changes.

frjo:


hypha/apply/users/views.py
```diff
@@ -28,47 +21,6 @@
...
-        )
-        return redirect_to if url_is_safe else ''
-
-    def get_context_data(self, **kwargs):
-        return super().get_context_data(
-            next=self.request.GET.get('next', ''),
-            **kwargs,
-        )
-
-
```
frjo, line 71: @Parbhat It looks to me that this has been built in to django-two-factor-auth now so it can be removed?

Any other followup changes I have missed.

Parbhat: Yeah, we can now use LoginView from the package instead of the custom one implemented in #1232. As redirect URL is implemented in the package https://github.com/Bouke/django-two-factor-auth/commit/90df643fc2f3eda5551b308df9cc200c283002f1#diff-b4dd7622064f0338584dd3ca5426d009

Looks like we do not need more changes.

Parbhat (approved):
Looks good.

PR #2085, opened 2020-07-13
Add django-file-form to upload fields

### LINKED ISSUE #1440: Improve the file upload widget
- [x] An drag and drop area
- [x] Show a list of the uploaded files
- [x] Allow user to add files to the list
- [x] Allow user to remove files from the list

Something like this:

Skärmavbild 2019-08-21 kl  10 12 14

(from commercial https://innostudio.de/fileuploader/)

My first option would be to look for an open source javascript widget that works with the file upload stream field we have now.

frjo:
Fixes #1440


hypha/static_src/src/javascript/apply/file-uploads.js
```diff
@@ -1,7 +1,13 @@
-(function ($) {
+jQuery(function ($) {
```
frjo, line 1: Why change this? The original style is how all jQuery code looks on the site.

Happy to see it converted to vanilla JS but that is not important.

requirements.txt
```diff
@@ -41,3 +42,7 @@ tomd==0.1.3
 wagtail==2.9
 wagtail-cache==1.0.0
 whitenoise==5.1.0
+
+# django-file-form==2.1.3
+# Fork that supports storing temporary files on S3:
+-e git+https://github.com/balazs-endresz/django-file-form.git@custom-storage#egg=django_file_form
```
frjo, line 48: Good chance of getting the changes included in the main project you think?

hypha/static_src/src/javascript/apply/file-uploads.js
balazs-endresz: I changed it when testing jquery-file-upload and dropzone.js. I think was necessary because now we include this file with `Widget.Media`, which (probably) outputs it further up the page, and we need domcontentloaded. If this general approach works then we'll need to clean up this file anyway (and might as well use vanilla JS then).

requirements.txt
balazs-endresz: I think so. Especially since they've tried to do something like this before. Maybe the problem was just that they didn't use the proper django File/storage api (which is a bit weird). But I guess we'd have a better case to make if we tried it out on heroku (dev) first.

Btw, django-file-form itself has some tests, which did pass locally, except for the selenium based ones. (I suspect it's still not easy to set up selenium locally. And last time I tried to get travis to run a whole test suite locally I gave up after a day. I haven't tried that with circle-ci yet, which is used on django-file-form.) 

So, django-file-form probably needs some more tests and documentation changes. I've added only one more config option, so updating the docs shouldn't take too long. I guess it'd be helpful to add tests for a custom (local) storage backend too. Or see how they mock it for boto and do something similar.

If getting a new version released takes too long then in the meanwhile you can fork it under OTF and install it the same way as I did here. Not ideal but works fine in the short term.

frjo:
https://github.com/mbraak/django-file-form/pull/313#event-3565428960

hypha/settings/base.py
```diff
@@ -672,3 +684,19 @@
 PROJECTS_AUTO_CREATE = False
 if env.get('PROJECTS_AUTO_CREATE', 'false').lower().strip() == 'true':
     PROJECTS_AUTO_CREATE = True
+
+
+# django-file-form settings
+
+FILE_FORM_MUST_LOGIN = True
```
sks444, line 691: >Must the user be logged in to upload a file.

I think this should be set as False as users without account can also apply to our forms.

balazs-endresz: Sorry, I assumed it's logged in only.

Btw, I guess you might want to consider adding file size validation too. I've just tried a [large pdf](https://download.lenovo.com/pccbbs/mobiles_pdf/t480_hmm_en.pdf) and it returned 503 after about 20MB but there was no indication that it failed, the error appeared on the console only. (I'm on mobile data so this might have been a network error too.)

sks444: >Sorry, I assumed it's logged in only.

No problem, I was just checking.

>Btw, I guess you might want to consider adding file size...

Thanks, will look into that.

frjo:
@sks444 It seems you need to be logged in for the file upload to work. For submission forms we need to allow anonymous users as well.

hypha/apply/funds/views.py
```diff
@@ -911,7 +928,10 @@ def form_valid(self, form):
                     source=self.object,
                     related=revision,
                 )
-
+        
+        # Required for django-file-form: delete temporary files for the new files
+        # uploaded while edit.
+        form.delete_temporary_files()
```
balazs-endresz, line 934: Does this delete files from S3? It might be worth checking and clarifying because with the S3 backend the [tmp files on the local filesystem are deleted](https://github.com/mbraak/django-file-form/blob/master/django_file_form/views/tus.py#L50) as soon as they are uploaded to S3 (and those are still considered temporary until the form is submitted).

sks444: I hope not as this is being executed after form data has been processed.

Temp files has to be deleted after they are saved in db otherwise they come up again while processing edit forms. It's better that we are deleting temp files at django-file-form when setup with s3 backend. I am not locally setup with s3, maybe that is why I was facing this issue.

@frjo when you deploy this to test env can you please verify that the files are not being deleted from S3.

balazs-endresz: So, the UploadedFile instances you can see in the django-admin are temporary and are on S3 as well. 

And looking at the instructions again you're right to call `.delete_temporary_files()`. I think I just forgot to add it but it is there in the readme.

So, you could maybe just **check in the django admin if they are in fact removed after submit** when this is deployed (but they should be).

sks444:
@frjo this is ready for review. Have fixed all the backend issues that were there.

hypha/static_src/src/javascript/apply/file-uploads.js
```diff
@@ -1,57 +1,38 @@
...
+            var allowedExtensions = fileField.accept.split(', ');
+
+            // TODO: this currently raises an exception when trying to remove an invalid file.
+            $(fileField).next('.dff-files').on('drop', function (e) {
+                var items = e.originalEvent.dataTransfer.items;
+                var files = e.originalEvent.dataTransfer.files;
+                for (var i = 0; i < files.length; i++) {
+                    var extension = '.' + files[i].name.split('.').slice(-1)[0];
+                    if (!allowedExtensions.includes(extension)) {
+                        items.remove(i);
```
sks444, line 31: Only thing that is pending now is the client side validation of the files with invalid extension. This doesn't seem to work and gives error:
```
Failed to execute 'remove' on 'DataTransferItemList': The list is not writable.
```
Unfortunately, we can not remove items from drop box like this. Looked over the internet for a similar possible solution but couldn't find much. I have also opened an issue related to this at https://github.com/mbraak/django-file-form/issues/317, maybe we could show a small warning and let user remove files with invalid extension, if there is no way to remove those files automatically.

frjo: A warning sounds like an excellent solution to this problem. Friendlier then an error on submit.

balazs-endresz: Besides a warning simply `.preventDefault()` might work too. I guess it would just cancel all files that are being dropped.

sks444: `. preventDefault()` also doesn't seem to work here.

sks444: Added a warning on invalid file uploads for now. Maybe we can fix it at django-file-form later on to have those warning on specific files.

sks444:
@frjo this is ready for final reviews.

frjo:
@sks444 Can we add a help text that states the allowed file extensions (FILE_ACCEPT_ATTR_VALUE) for upload fields?

Something like:

Accepted file types are: .doc, .docx, .odp, .ods, .odt, .pdf, .ppt, .pptx, .rtf, .txt, .xls, .xlsx

sks444:
@frjo do we need to add the help_text via javascript? Or

We can also make use of our block help_text field to display this:
```
    def get_field_kwargs(self, struct_value):
        kwargs = super().get_field_kwargs(struct_value)
        kwargs['help_text'] = kwargs['help_text'] + f"Accepted file types are {settings.FILE_ACCEPT_ATTR_VALUE}"
        return kwargs
```

sks444:
@frjo added the help text; Seemed better to use the block way as it a permanent text on file upload fields, would have been good to use javascript if it is temporary and only needed on invalid file uploads.

frjo:
@balazs-endresz and @sks444 Really excellent work on this! Will deploy to test for OTF to try out now. This will make applicants really happy I think.

frjo:
@sks444  Just noticed that when running "migrate" a second time on this branch it complains about missing migrations. Do we need to add this?

```
# Generated by Django 2.2.14 on 2020-07-29 08:11

from django.db import migrations, models
import hypha.apply.projects.models
import hypha.storage_backends


class Migration(migrations.Migration):

    dependencies = [
        ('application_projects', '0034_multi_input_char_block'),
    ]

    operations = [
        migrations.AlterField(
            model_name='contract',
            name='file',
            field=models.FileField(storage=hypha.storage_backends.PrivateMediaStorage(), upload_to=hypha.apply.projects.models.contract_path),
        ),
        migrations.AlterField(
            model_name='packetfile',
            name='document',
            field=models.FileField(storage=hypha.storage_backends.PrivateMediaStorage(), upload_to=hypha.apply.projects.models.document_path),
        ),
        migrations.AlterField(
            model_name='paymentreceipt',
            name='file',
            field=models.FileField(storage=hypha.storage_backends.PrivateMediaStorage(), upload_to=hypha.apply.projects.models.receipt_path),
        ),
        migrations.AlterField(
            model_name='paymentrequest',
            name='invoice',
            field=models.FileField(storage=hypha.storage_backends.PrivateMediaStorage(), upload_to=hypha.apply.projects.models.invoice_path),
        ),
        migrations.AlterField(
            model_name='reportprivatefiles',
            name='document',
            field=models.FileField(storage=hypha.storage_backends.PrivateMediaStorage(), upload_to=hypha.apply.projects.models.report_path),
        ),
    ]
```

sks444:
@frjo, there was a bug that I noticed on the edit forms with single file upload. To reproduce for a new submission:

- Upload a file to single file uploader
- Go to edit, and then do submit. File submitted on single file uploader would be automatically deleted.

I have fixed this in my last commit and pushed it on branch https://github.com/OpenTechFund/django-file-form/tree/upload-widget-placeholder-file. I have also opened an [issue](url) related to this at django-file-form.

frjo:
@sks444 Deployed to test now, thanks!

PR #2086, opened 2020-07-13
Enhancement/1440 django file form

balazs-endresz:
Ticket: #1440

Rebased from master and added support for S3 temporary storage using a slightly modified django-file-form.

There are some hints in `local.py.exmaple` for setting up S3 locally if you want to try that first.


frjo:
@balazs-endresz I have sent you an invite to the Hypha repo now, sorry for not noticing this before!

Now you can commit directly to the 1440-django-file-form branch in this repo.

PR #2089, opened 2020-07-20
Implement score field without text box block

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/2053

This pr implements a new block `ScoreFieldWithoutTextBlock`, which can be used when we need to accept score on review forms without requiring to fill in an answer.


hypha/apply/review/options.py
```diff
@@ -9,6 +9,18 @@
...
+
+SCORE_CHOICES = (
+    (0, '0. Need more info'),
+    (1, '1. Poor'),
+    (2, '2. Not so good'),
+    (3, '3. Is o.k.'),
+    (4, '4. Good'),
+    (5, '5. Excellent'),
+    ('', 'n/a - choose not to answer'),
+)
```
frjo, line 21: Can we not reuse "RATE_CHOICES"?

sks444: We can but then we will have to manipulate `RATE_CHOICES` to use `''` instead of `NA`. And then I also thought maybe it's good idea to use a separate choices for a new block. So that maybe if we want to change choices for the other block we won't have to change for this. I have included a comment in the block on why `''` is needed:
```
    This block uses SCORE_CHOICES instead of RATE_CHOICES, only difference is to
    have empty string('') in place of NA for text value `n/a - choose not to answer`
    as it helps to render this value as default to the forms and also when this field is
    required it automatically handles validation on empty string.
```

Anyway, should I use `RATE_CHOICES` and manipulate it accordingly? Or is it okay.

frjo: Good points. Can we rebuild RATE_CHOICES to have the empty value for n/a?

If that is a lot of work we go with your solution.

PR #2097, opened 2020-07-30
Integrate drf-yasg for doc ui and document existing apis

sks444:
This pr integrates with https://github.com/axnsan12/drf-yasg which will provide a nice UI to interact with all the APIs that we have.

Most of the information which one needs to interact with the APIs are automatically generated by `drf-yasg` and additional information can also be added inform of a doc string to the views.

This UI is currently available at `/api/doc/` endpoint.

Screenshot 2020-07-30 at 7 39 29 PM

### DIFF
--- a/hypha/apply/api/urls.py
+++ b/hypha/apply/api/urls.py
@@ -1,10 +1,28 @@
 from django.urls import include, path
+from drf_yasg import openapi
+from drf_yasg.views import get_schema_view
+from rest_framework import permissions
 
 from .v1 import urls as v1_urls
 
 app_name = 'api'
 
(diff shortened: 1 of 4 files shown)


frjo:
This looks nice!

frjo:
What happens one day when we have "api/v2"? How will this documentation handle that?

sks444:
@frjo, thanks for noticing that, made the current doc path to `api/v1/doc`. So when we support v2, we would just have to define `api/v2/doc` including v2 url patterns.

frjo:
Merging this in to master, nice work!

PR #2099, opened 2020-08-03
Improve current apis to make use of drf viewset and routers

sks444:
This pr makes following changes:
- Instead of creating separate API views for each methods we can use a single [viewsets](https://www.django-rest-framework.org/api-guide/viewsets/) class and then instead of hardcoding url patterns in urls.py, we can use [routers](https://www.django-rest-framework.org/api-guide/routers/) to make use of automatically generated url patterns.
- By using [drf-nested-routers](https://github.com/alanjds/drf-nested-routers) we can define `submission_router` and then register it's child url-views under that router. This makes easy for us to define any number of nested urls for a submission without hardcoding the patterns in urls.py. 
- moves filters from views.py to a separate file named `filters.py`.


sks444:
@Parbhat @frjo, At this stage as there are not many apis, thought it would be a good idea to switch our existing apis to use new drf features which makes it easy for us to develop new apis in future.

sks444:
That looks good @frjo.

PR #2108, opened 2020-08-18
Add per page option to submission all list view filter.

### LINKED ISSUE #2075: On the application listings, on the same line as filters and search bar, present pagination options 25 (default), 50, and 100 per page and show how many of total applications is presented, eg "1-25 of 100".

frjo:
Fixes #2075

### DIFF
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -244,12 +244,18 @@ class StatusMultipleChoiceFilter(Select2MultipleChoiceFilter):
 
 
 class SubmissionFilter(filters.FilterSet):
+    PAGE_CHOICES = (
+        (25, '25'),
+        (50, '50'),
+        (100, '100'),
+    )
     round = Select2ModelMultipleChoiceFilter(queryset=get_used_rounds, label='Rounds')
     fund = Select2ModelMultipleChoiceFilter(field_name='page', queryset=get_used_funds, label='Funds')
     lead = Select2ModelMultipleChoiceFilter(queryset=get_round_leads, label='Leads')
     reviewers = Select2ModelMultipleChoiceFilter(queryset=get_reviewers, label='Reviewers')
     screening_status = Select2ModelMultipleChoiceFilter(queryset=get_screening_statuses, label='Screening')
     meta_terms = Select2ModelMultipleChoiceFilter(queryset=get_meta_terms, label='Terms')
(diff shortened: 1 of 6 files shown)


hypha/apply/funds/tables.py
```diff
@@ -244,12 +244,18 @@ def get_filter_predicate(self, v):
...
+        (50, '50'),
+        (100, '100'),
+    )
     round = Select2ModelMultipleChoiceFilter(queryset=get_used_rounds, label='Rounds')
     fund = Select2ModelMultipleChoiceFilter(field_name='page', queryset=get_used_funds, label='Funds')
     lead = Select2ModelMultipleChoiceFilter(queryset=get_round_leads, label='Leads')
     reviewers = Select2ModelMultipleChoiceFilter(queryset=get_reviewers, label='Reviewers')
     screening_status = Select2ModelMultipleChoiceFilter(queryset=get_screening_statuses, label='Screening')
     meta_terms = Select2ModelMultipleChoiceFilter(queryset=get_meta_terms, label='Terms')
+    per_page = filters.OrderingFilter(choices=PAGE_CHOICES, empty_label=None, label='Per page')
```
Parbhat, line 258: [OrderingFilter](https://django-filter.readthedocs.io/en/stable/ref/filters.html?highlight=empty_label#orderingfilter) can be replaced with [ChoiceFilter](https://django-filter.readthedocs.io/en/stable/ref/filters.html?highlight=empty_label#choicefilter) as we do not need to perform queryset ordering. But then we also need a [method](https://django-filter.readthedocs.io/en/stable/ref/filters.html?highlight=empty_label#method) parameter to handle it.

```python
per_page = filters.ChoiceFilter(choices=PAGE_CHOICES, empty_label=None, label='Per page', method='per_page_handler')
```

```python
    def per_page_handler(self, queryset, name, value):
        # Pagination is already implemented in view. We only need to add per_page query parameter.
        return queryset
```

We can use `empty_label` to give user some context for the filter. Currently, it displays 25, 50 and 100 in the dropdown and it may not be clear what it is for.

Parbhat (commented):
Works as expected. We may tweak the filter for its actual purpose.

hypha/apply/funds/tables.py
frjo: Excellent improvements. Implemented now.

PR #2116, opened 2020-08-27
Use reviewer settings to give submission access to reviewers

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/1182


sks444:
@frjo added two more settings variables:
- `use_settings`: which actually decides if we want to use the settings variable to filter out submissions or display them like it was used to.

- `assigned`: I actually figured it might not be a good idea to show submissions to a reviewer for which they are not assigned to, so there is also a setting for that.

frjo (approved):
Excellent work! Will put on test for OTF to check out.

PR #2126, opened 2020-09-14
Add investments table and partners page

sks444:
Fixes https://github.com/ResetNetwork/apply-app/issues/79

Steps to test this feature:

- Add Partner Index Page as a child of `About` page. (similar to Person Index Page)
- Add partners as child of partnerindex. 
- Go to `Investments` and create Investments for partners that you created.
- Visit the partners in frontend by going to `about` tab. You'll get a table of all partners with their investments made. 

---
TODOs:

- [ ]  From Dan's [comment](https://github.com/ResetNetwork/apply-app/issues/79#issuecomment-686569317): Option to include and require one or more Category Question, for example:
-- type: Can we use the Category Question "Organization Type"?
    - If I understand correctly this means that we will have a fixed number of fields defined in model and we will have a category question defined for them in the admin. So that we can add any number of options/choices to that.
    - OR, we want to have the functionality of adding any number of category questions(fields) to these pages/models?
       - But for fields to be viewed in the table and for searching/filtering they should be defined in the models and not be dynamically defined via admin.


hypha/public/partner/models.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+    description = models.TextField()
+    application = models.OneToOneField(
+        ApplicationSubmission,
+        on_delete=models.SET_NULL,
+        blank=True, null=True
+    )
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now_add=True)
```
Parbhat:
```suggestion
    updated_at = models.DateTimeField(auto_now=True)
```

Parbhat (commented):
Good first step. @frjo Can we deploy this branch to the testing environment, to get feedback from Dan? It will help to get more inputs about the requirements.

frjo:
Deploying this to test now.

frjo:
Redeployed latest changes to test.

sks444:
Thanks @frjo.

One thing that we need to discuss is the requirements of `Category Questions` and the feasibility of its implementations.

Other than that, we can add search/filter and other functionalities later on when the dataset grows.

danblah:
This is great. Feedback here: https://github.com/ResetNetwork/apply-app/issues/79#issuecomment-692903040

sks444:
@frjo, can we redeploy this to test.

[Here are the updates that has been made.](https://github.com/ResetNetwork/apply-app/issues/79#issuecomment-699820053)

frjo:
@sks444 Deployed to test now.

sks444:
@frjo, can we redeploy this to test.

It looks like few of the things failing [here](https://github.com/ResetNetwork/apply-app/issues/79), were because of the migration conflict.

frjo:
@sks444 Redeployed to test now.

sks444:
Thanks @frjo.

It looks like the data entry created from the old migration is still there in the cms. https://test-apply.opentech.fund/admin/pages/17/ **(Our Partners)**. Now I can not delete/edit/add as it will raise 500. 

This problem I have been noticing in wagtail from quite some time now. Pages are cached and remain there even if we unapply the corresponding migrations. So the process is to first delete the pages from wagtail and then unapply the migrations.

@frjo, can we do the following:

We need to delete current `Our Partners` entry from the admin.

- python manage.py migrate partner zero
- Checkout to some other branch, maybe `main`. In this case, that means deploying `main` to test or reverting the deployment of current branch. 
- Go ahead and delete `Our Partners` from cms.
- redeploy the current branch.

frjo:
@sks444 Done, hope I managed to fix it.

sks444:
Thank you so much for the effort @frjo.

I still see `Our Partners` at https://test-apply.opentech.fund/admin/pages/17/

Can you try once again to delete that when you're on other branch where partners code is not deployed. And then redeploy the partners branch.

Screenshot 2021-03-22 at 7 29 55 PM

This is cached page I believe. So will have to delete it from where the current branch code is not deployed. And then redeploy the current branch.

Let me know if that doesn't work out. I am also not 100% sure on this, I just do this fix when I face similar kind of problem locally sometime.

frjo:
@sks444 Now that page is gone as well.

danblah:
@frjo and @sks444, why don't I have the option to create an about child page on our production? Did I fail to fully test/confirm this?

PR #2128, opened 2020-09-16
Add optional consent checkbox to login form.

### LINKED ISSUE #2127: Require content checkbox on login
Possibility to configure text  and help text with link in wagtail admin.

frjo:
Fixes #2127


Parbhat (commented):
Looks good. One concern is a long length of help text. And clicking the checkbox every time while login may not be good UX. If we can allow the user to login and then give his/her consent then we can use some ideas from [django-cookie-law](https://github.com/TyMaszWeb/django-cookie-law) package. It saves the user consent in the [cookie](https://github.com/TyMaszWeb/django-cookie-law/blob/master/cookielaw/static/cookielaw/js/cookielaw.js) and render consent box [accordingly](https://github.com/TyMaszWeb/django-cookie-law/blob/36e12cd4133d93b60b1cf00e1dfd6d993b763e61/cookielaw/templatetags/cookielaw_tags.py#L27).

frjo:
Good input!

What do you think about about adding a consent value with a timestamps to the db? A setting would then allow staff to set how often users need to confirm the consent.

Parbhat:
Adding to the DB would be good also as the user will not have to give consent on every device (change in a browser session and cookies).

frjo:
The draw back is that you need to create a new login "step" for the consent. At start of login you do not know what user is logging in so can't retrieve the consent timestamp and value.

Maybe a cookie is easiest and best after all.

Parbhat:
We can also ask for consent and save it in DB after login if login is allowed without consent. Similar to cookie and we can save consent status in DB model. But Cookie method will be simpler in this case.

PR #2131, opened 2020-09-18
Cookie consent feature

### LINKED ISSUE #1950: Add cookie consent option to the core app
@scott-psycle and @stuartmessenger, this one is for y'all and @frjo

Reset.tech has a cookie consent pop-up. Let's bring that over to the main app as an option to enable and configure. This is needed for those who have privacy policies to comply with.

frjo:
Fixes #1950

This is an improvement of the Reset version.

Wagtail admin settings for text and to turn it on/off.

When on a banner can be inserted in to templates with `{% cookie_banner %}`.

Any analytic or other tracking scripts that only should run when user opts in are wrapped in `{% if COOKIES_ACCEPTED %}Code goes here{% endif %}`.

If the user have an adblocker installed the banner will in most cases never display. The site will work as it should but  `COOKIES_ACCEPTED` will be false as that is the default.

cookieconcent


frjo:
@Parbhat Should this app be inside the hypa dir or not in your opinion? The addressfield app is outside in the root, is this a good choice? How should one reason in these cases?

Then overall comment on the approach of the cookieconcent app.

cookieconsent/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
frjo, line 3: Delete these empty files or keep them?

Parbhat: Good to delete the empty files if not used.

cookieconsent/context_processors.py
```diff
@@ -0,0 +1,8 @@
+
+
+def cookies_accepted(request):
+    cookieconsent = False
+    if request.COOKIES.get('cookieconsent', 'decline') == 'accept':
+        cookieconsent = True
+
+    return {'COOKIES_ACCEPTED': cookieconsent}
```
Parbhat, line 8: COOKIES_ACCEPTED context does not seem to be used. or it will be used later as mentioned in the issue description.

cookieconsent/templates/includes/banner.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}
+
+{% if show_banner %}
+<div class="wrapper wrapper--cookieconsent cookieconsent no-js-hidden">
+    <div class="wrapper wrapper--small">
+        <h2 class="cookieconsent__title">{% trans title %}</h2>
+        <div class="cookieconsent__message">
+            {% trans message|safe %}
```
Parbhat, line 8: It would be good to use [|richtext](https://docs.wagtail.io/en/v2.8.1/topics/writing_templates.html#rich-text-filter) filter.

Parbhat (approved):
The approach and implementation look great.

Parbhat:
@frjo It depends on the pattern used by the project. Normally while creating apps with `python manage.py startapp` the app is created in the same directory as `manage.py` (root directory). But as the number of apps grows larger the root directory gets more cluttered. So, some projects place the apps inside the project directory like [hypha](https://github.com/OpenTechFund/hypha/tree/master/hypha). It will also help us to divide apps between [public](https://github.com/OpenTechFund/hypha/tree/master/hypha/public) and [apply](https://github.com/OpenTechFund/hypha/tree/master/hypha/apply) site. So, it seems better to place `cookieconsent` app in the hypha project directory like the [images](https://github.com/OpenTechFund/hypha/tree/master/hypha/images) app if cookie banner will show in both public and apply sites. So, we can follow one pattern and continue to use it everywhere. It seems the reason to place `addressfield` at the root is it may have been developed with a goal to make it a third party package or inspired from a package for custom requirements. So, to keep uniformity in import statements it may have been at root. There is no right or wrong method but just a convention used by projects.

PR #2133, opened 2020-09-21
Add Review form container

vimal1083:
This is to add review forms rendered using react. I have created a `FormContainer` component, It just takes the form fields configuration and renders the form and validate it. `FormContainer` can be reused in other parts of application as well

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -94,6 +94,7 @@ gulp.task('clean:css', function clean () {
 // Clean JavaScript files.
 gulp.task('clean:js', function clean () {
   return del([
+      options.theme.dest + 'app/*.*',
       options.theme.js_dest + '**/*.js',
       options.theme.js_dest + '**/*.map'
     ], {force: true});
--- a/hypha/static_src/src/app/src/SwitcherApp.js
+++ b/hypha/static_src/src/app/src/SwitcherApp.js
@@ -38,7 +38,13 @@ class SwitcherApp extends React.Component {
         if (success) {
             this.openDetail()
         }
-    }
+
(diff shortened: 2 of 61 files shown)


vimal1083:
rebuild

sks444 (approved):
Works really well. Great work @vimal1083.

@frjo this can be deployed to test for testing purposes.

This implements frontend of adding reviews in react interface. Along with that this also adds the functionalities for `Update Draft Reviews`, `Edit a review` and `Delete a review` to react interface.

frjo (commented):
Tested to save and edit reviews. Update review form and confirm that the updated version is used in React as well. It all seems to work perfectly, excellent work!

I get a message in each rich text field "This domain is not registered with Tiny Cloud. Please review your approved domains." 

1. Is the TineMCE editor depended on an external service? Cost? Can it be locally hosted instead?
2. Is the content of the field  transfered to this service in any way?

vimal1083:
It was dependent on tinyCloud, which is a free service and it din't send any data to the tinyCloud. But now I have removed the dependency. It uses the our self hosted library file which is already being used in Django pages. We no more depend on tinyCloud. Kindly pull the code and check you won't be getting that banner.

Parbhat:
One thing noticed while working on this feature: As we are creating more functionality in the React app like Review and determinations. We should also look into the future requirement of giving access to different user groups like reviewers. Currently, the react app can only be accessed by staff. Mentioning it so we can check with the OTF and Reset about this.

frjo:
@Parbhat The React app is only for staff for the forcible future. 

PR #2136, opened 2020-09-28
Updated Wagtail to 2.10.2

frjo:


frjo:
@Parbhat Anything else one should be aware of with this update?

Parbhat:
@frjo We will need to check the effect of [rich-text](https://docs.wagtail.io/en/latest/releases/2.10.html#div-class-rich-text-wrappers-removed-from-rich-text) wrapper class removal. I also asked @sks444 last week to look into the upgrade so he can add more details. Wagtail 2.10 comes with a lot of new features.

sks444 (commented):
@Parbhat @frjo, This looks good to me.

I have added `wagtail.contrib.legacy.richtext` which replicates the old behaviour for rich-text fields. Rest everything works well.

sks444:
@frjo, we figured there are not many places where we use `|richtext` filter which is responsible for creating the `<div class="rich-text"></div>` wrapper. So we could avoid the dependency on the legacy richtext app and manually wrap richtext values with `<div class="rich-text"></div>` at the places we have used the filter.

We need to add this wrapper class because our css for specific elements is dependent on the `.rich-text` css.

Reverting the commit which adds the legacy `richtext` to the settings.

frjo:
@sks444 I have added div wrappers where `|richtext` is used now.

sks444 (approved):
Looks good.

sks444:
@frjo, just noticed that upgrading to `wagtail==2.10.2` breaks the react interface.

frjo:
@sks444 Yes look like the API is having issues.

`ValueError at /api/v1/rounds/
Cannot assign "'Jane Doe'": "Round.lead" must be a "User" instance.`

sks444:
Yeah @frjo, debugging on why that is happening.

sks444:
@frjo, I have figured out the problem and have found a way around to fix the api.

### Why it is happening
[This code](https://github.com/wagtail/wagtail/commit/f43db1e69e67946b08207aeb39e2949f907bba2e#) was added in Wagtail 2.10 to fix an issue around annotate when using `specific()` which is causing our code to fail.

The error that we are getting is: `ValueError: Cannot assign "'Staff 2'": "Round.lead" must be a "User" instance` which is happening because of line `setattr(page, annotation, value)`. Where page is the page object(Round), annotation is `lead` , and `Staff 2` is the value that is being set. But given that `Round.lead`  should be user instance as it is related field in Round model, setting it as string(Staff 2) is causes the error.

This feels like an issue around Wagtail when using annotate with related fields and using `specific` Or there could be an issue with our [annotate](https://github.com/OpenTechFund/hypha/blob/29e9565bcc8ca695e62ccca9a2a1b77b624123c6/hypha/apply/funds/models/applications.py#L496) itself. But changing our queryset logic won't be a good idea as it is used at many places in our codebase.

### Our Fix
We were using `specific` [RoundsAndLabs.objects.specific()](https://github.com/OpenTechFund/hypha/blob/29e9565bcc8ca695e62ccca9a2a1b77b624123c6/hypha/apply/api/v1/views.py#L122) method for the list and detail view. Which is not needed and we can use `RoundsAndLabs.objects.all()` for the list view and override the get_object to get specific for each object.
```
    def get_object(self):
        obj = super(RoundViewSet, self).get_object()
        return obj.specific
```

I have pushed to code for the same which fixes the api and react interface. Let me know if everything also works for you.

frjo:
Deployed to test now.

frjo:
This seems to work as it should. Merging now.

@sks444 Excellent bug tracking and fixing!

PR #2139, opened 2020-09-30
Use official django-file-form release. Now includes our feature additions

frjo:


sks444 (commented):
@frjo, there are couple changes that we have to do:

- As the client side validation is now being done at [django-file-form](https://github.com/mbraak/django-file-form/issues/317). We can remove the [code which adds the warning](https://github.com/OpenTechFund/hypha/blob/29e9565bcc8ca695e62ccca9a2a1b77b624123c6/hypha/static_src/src/javascript/apply/file-uploads.js#L20).
Screenshot 2020-10-02 at 12 28 59 PM

- There is a [refactor change to check placeholder file](https://github.com/mbraak/django-file-form/pull/328/files). Earlier it was `f'placeholder-{field_name}'` and now it is `f'{field_name}-placeholder'` . So we would have to change [our code](https://github.com/OpenTechFund/hypha/blob/29e9565bcc8ca695e62ccca9a2a1b77b624123c6/hypha/apply/funds/views.py#L923) accordingly. `initial.pop('placeholder-' + field_id, False)` should be `initial.pop(field_id + '-placeholder', False)`.

frjo:
@sks444  Thanks for the information, have that fixed now. I get one error on tests that seems to have with checking an empty file field.

https://github.com/OpenTechFund/hypha/blob/29e9565bcc8ca695e62ccca9a2a1b77b624123c6/hypha/apply/projects/tests/test_forms.py#L140

~~~~
ERROR: test_receipt_not_required (hypha.apply.projects.tests.test_forms.TestCreatePaymentRequestForm)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/frjo/Sites/otf/hypha/apply/projects/tests/test_forms.py", line 144, in test_receipt_not_required
    self.assertTrue(form.is_valid(), msg=form.errors)
  File "/Users/frjo/Sites/otf/venv/hypha/lib/python3.8/site-packages/django/forms/forms.py", line 185, in is_valid
    return self.is_bound and not self.errors
  File "/Users/frjo/Sites/otf/venv/hypha/lib/python3.8/site-packages/django/forms/forms.py", line 180, in errors
    self.full_clean()
  File "/Users/frjo/Sites/otf/venv/hypha/lib/python3.8/site-packages/django_file_form/forms.py", line 41, in full_clean
    super(FileFormMixin, self).full_clean()
  File "/Users/frjo/Sites/otf/venv/hypha/lib/python3.8/site-packages/django/forms/forms.py", line 381, in full_clean
    self._clean_fields()
  File "/Users/frjo/Sites/otf/venv/hypha/lib/python3.8/site-packages/django/forms/forms.py", line 397, in _clean_fields
    value = field.clean(value, initial)
  File "/Users/frjo/Sites/otf/hypha/apply/stream_forms/fields.py", line 36, in clean
    old_files = [
TypeError: 'NoneType' object is not iterable
~~~~

Any ideas?

sks444:
@frjo, looking into it.

sks444:
@frjo, In the [clean method](https://github.com/OpenTechFund/hypha/blob/29e9565bcc8ca695e62ccca9a2a1b77b624123c6/hypha/apply/stream_forms/fields.py#L35) of MultiFileUpload we need to handle the `None` value like we have done In SingleFileUpload [clean](https://github.com/OpenTechFund/hypha/blob/29e9565bcc8ca695e62ccca9a2a1b77b624123c6/hypha/apply/stream_forms/fields.py#L70). But as this MultiFileUpload I think we should return an empty list instead of `None`.
```
if not value:
   return []
```
For this field earlier we were getting `[]` which was automatically handled in the loop. But seems like they changed something internally to have the same behaviour for both the fields.

sks444:
@frjo, before merging, we should also put this on testing env to check if this upgrade is working properly with our existing files.

frjo:
@sks444 Thanks, that seems to do the trick!

frjo:
@sks444 Deploying to test now.

sks444:
I tested this on test env and it worked well for me.

frjo:
For me as well so merging it now.

PR #2148, opened 2020-10-21
Change submission screening flow

### LINKED ISSUE #1948: Change how we screen applications, make it easier
A common need is for managers to quickly flag an application as within remit or within the scope of the open call). At its most basic, this is an "in" or an "out". Not every deployment will have a need for multiple different types of screening statuses. Here's my proposal.

* Add a required "In" or "Out" option for all screening status, "/admin/funds/screeningstatus/create/"
* Add a "Set to default" option for all screening status. 
* If no screening statuses exist, hide the option from application pages for those who don't use them.
* Replace the "screen application" button with a red down and greed up arrow (or thumbs up or down).
* On click of red icon, set status as default out status. If not default set, use the first out status created.
* On click of green icon, set status as default in status. If not default set, use the first in status created.
* On click and hold/longpress of either icon, an overlay comes on up top of only the "Actions to take" box, listing the in/out options according to the clicked icon. On click of an option sets the status. Clicking outside/away closes the overlay and does not set a status.

Any thoughts? Issues/concerns? Refinements?

cc @georgiamoon 

Assuming all good to proceed, would like to get @psycle folks cracking on this.

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/1948


frjo:
@sks444 For new installs we should add one generic default yes and one generic default no.

sks444:
> @sks444 For new installs we should add one generic default yes and one generic default no.
> 
> Accept/Dismissed is good generic terminology.

@frjo, you mean for a new org which starts using Hypha?

We could add this in the data migration. But that will create these statuses for reset and otf as well.

Okay..we can have a check to create these only if no default statuses exist. And there is always an option to change default..so I guess this would work out?

sks444:
@frjo, can we deploy this to test? Would be good to get some feedback.

We have implemented the new screening flow at both the places, at submission detail page and at the react page.

Some key points:

- make sure one or more screening statuses are created both for yes and no. (If no statuses exist two(Accept and Dismiss) screening statuses will be automatically created and set as default)
- set two as default one for yes and one for no, if defaults not set, first screening statuses create will be set as default.
- submission can be screened with thumbs by going to Django or react page.
- once submission is screened by thumbs, we can also set more yes or no statuses to the submission.
- in case we change defaults screening statuses, submission screened by old defaults will be replaced by new defaults.
- see https://github.com/OpenTechFund/hypha/issues/1948 and https://github.com/OpenTechFund/hypha/issues/2158 for more.

frjo:
@sks444 Will get this on test today.

frjo:
Will be deploying this tomorrow. Will also add some followup changes but they really like this and want it now.

Parbhat:
@sks444 Noticed 1 minor bug: Sorting by screening status on [all](https://test-apply.opentech.fund/apply/submissions/all/) submission table gives an [error](https://sentry.io/share/issue/c126cbd2d3f04418a6da8cc35ee913ca/). It can be produced by clicking Screening title in Table headers.

sks444:
Thanks for noticing that @Parbhat.

I have fixed the sorting in https://github.com/OpenTechFund/hypha/pull/2176

Also I noticed we are also showing the screening status in the `>arrow view` of the table. So in the column I am only showing the default screening status set to make it good look on the table and on detail we can view all the screening statuses set. Filters are also there, so I think it would be okay...? cc: @frjo 

Screenshot 2020-12-02 at 3 51 05 PM

PR #2153, opened 2020-10-31
Required attribute in field groups

### LINKED ISSUE #1294: Make it possible to have required fields in file groups
[Issuehunt badges](https://issuehunt.io/r/OpenTechFund/opentech.fund/issues/1294)

When a required field is in a field group make it unrequired when the field group is closed.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $120.00)

- [danblah danblah](https://issuehunt.io/u/danblah) ($20.00)
- [resetnetwork resetnetwork](https://issuehunt.io/u/resetnetwork) ($100.00)

### Submitted pull Requests
- [#2153 Required attribute in field groups](https://issuehunt.io/r/OpenTechFund/hypha/pull/2153)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/OpenTechFund/opentech.fund/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

bunysae:
Fixes #1294.

The server set's fields in field-groups to non required, if they are not displayed, and set's the attribute `data-hidden="false"` in the response, when an incomplete form is posted and the field-group is visible.
On the client-side, the `required`-attribute for the input fields in the field-group is removed, when an incomplete form is posted the user changes hides the field-group. This avoids a "component not focusable"-error in google chrome.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#1294: Make it possible to have required fields in file groups](https://issuehunt.io/repos/45578367/issues/1294)
---
</details>


sks444 (commented):
Thank you for the pr @bunysae.

Changes look good so far..

One suggestion if possible: as these changes is specific to application submission maybe we can try to override get_form_fields or get_form_class in [RoundBase](https://github.com/bunysae/hypha/blob/82a2a8bd4209f87bd2eeee6181232f19936c3749/hypha/apply/funds/models/applications.py#L145)..?

Here is an example of overriding [get_form_class](https://github.com/bunysae/hypha/blob/82a2a8bd4209f87bd2eeee6181232f19936c3749/hypha/apply/determinations/views.py#L328) for specific need for Determination.

sks444 (commented):
>When a required field is in a field group make it unrequired when the field group is closed.

@bunysae, As mentioned in the issue, I think we only want the required fields in group to be marked as not required when option 2 is selected(i.e. group is closed). If the user select the option 1(i.e. group is open). We want the fields to be required as they were.

These changes marks all the group field as not-required when the form is loaded, even when  option 1 is selected.

Screenshot 2020-11-06 at 6 38 27 PM
Screenshot 2020-11-06 at 6 39 10 PM

bunysae:
> One suggestion if possible: as these changes is specific to application submission maybe we can try to override get_form_fields or get_form_class in [RoundBase](https://github.com/bunysae/hypha/blob/82a2a8bd4209f87bd2eeee6181232f19936c3749/hypha/apply/funds/models/applications.py#L145)..?
It's done.

hypha/apply/funds/models/applications.py
```diff
@@ -343,9 +344,73 @@ def get_form_parameters(self, submission_id=None):
 
         return form_parameters
 
+    def get_form_fields(self, draft=False, posted_field_values={}):
```
sks444, line 347: I didn't mean to write all the get_form_fields code here, instead I was thinking if we could override it, like I have done [here](https://github.com/OpenTechFund/hypha/blob/e7c64e0223f4db1fcfd62fbc3ebe0f88fbc25584/hypha/apply/funds/models/applications.py#L346) for this [pr](https://github.com/OpenTechFund/hypha/pull/2162).

But anyway, I think for this issue it would be good to do it the way you were doing it earlier. As we would also need this in other places where we use group fields. E.g. to get this working for submission edit form, you would also need to send form_data in the [edit form](https://github.com/OpenTechFund/hypha/blob/f6a1934de94a4a349be62ea55706aa5f0ba463cc/hypha/apply/funds/views.py#L945). I think you can use `self.object.raw_data`.

So can you revert your changes to do this in [get_form_fields](https://github.com/OpenTechFund/hypha/blob/f6a1934de94a4a349be62ea55706aa5f0ba463cc/hypha/apply/stream_forms/models.py#L49) only.

hypha/apply/funds/models/applications.py
```diff
@@ -343,9 +344,73 @@ def get_form_parameters(self, submission_id=None):
...
+
+        return form_fields
+
+    def get_form_class(self, draft=False, posted_field_values={}):
+        return type('WagtailStreamForm', (self.submission_form_class,), self.get_form_fields(draft, posted_field_values))
+
     def get_form(self, *args, **kwargs):
         draft = kwargs.pop('draft', False)
-        form_class = self.get_form_class(draft)
+        from_class = None
```
sks444, line 409: Why do we need to assign form_class as None?

sks444 (changes requested):
Works well. Thank you!

Just a couple of more changes, I believe this should be the final round of review.

I think we should not interfere with the required attribute of the group field itself. E.g. In the second image `Grouping some field` should be required as it is set as required. Otherwise user can submit forms without even selecting a choice. 

Screenshot 2020-11-10 at 7 57 40 PM
Screenshot 2020-11-10 at 7 57 13 PM

hypha/apply/funds/models/applications.py
bunysae: Changes are reverted

hypha/apply/funds/models/applications.py
bunysae: Removed. Isn't necessary.

sks444:
@frjo, @Parbhat, group fields doesn't work as expected on the edit forms. Irrespective of the choice being selected all the grouped fields are shown.

So even if we merge this pr to make required group fields not required based on choices, it would not work correctly on the edit form until we have the group functionality working there.

Is that a known issue? If that is fixable maybe I could give it a try. Or other option is to not allow changing the choice of group field on edit forms.

Parbhat:
@sks444 IIRC grouping functionality was not implemented in edit forms. You can confirm it in master branch.

sks444:
@frjo, can we merge this on the branch `enhancement/field-groups-required-fields` ? And I can work on the remaining issues if there is any or will finalise the changes and create a pr to master.

frjo:
@sks444 Done

frjo:
@sks444 See https://github.com/OpenTechFund/hypha/pull/2174

sks444:
Thanks @frjo.

Thanks @bunysae for the contribution!

PR #2155, opened 2020-11-02
Create codeql-analysis.yml

frjo:


frjo:
@Parbhat @sks444 Do you think this is useful to have?

Parbhat:
Yeah, would be good to have.

PR #2156, opened 2020-11-02
Move the rating above text field for score block

sks444:
Towards https://github.com/OpenTechFund/hypha/issues/2149


frjo:
@Parbhat @sks444 Does this looks ready to go? I added some small fixes to the tests to make it pass.

sks444:
Not yet @frjo.

These changes just adds the support for new review forms.

But for old ones data comes in the form of `[text, score]`. So need to handle that as well. E.g. When viewing and editing old review forms.

This also needs to be done in the API/React-app as well.

I think we never had a final decision on this. Let me know if this needs to be implemented, will move forward.

frjo:
@sks444 Thanks, will ask OTF and Reset what do do with this.

PR #2162, opened 2020-11-05
Minimise application forms to only show required field

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/1951

**How to test**

- Create an application form. Mark some fields as required and some not-required.
- Create a fund selecting that application form.
Screenshot 2020-11-04 at 8 17 18 PM

- Mark `Minimize form` as true if you want minimize the form. This will show only the required fields on public-facing apply page.
-  After submission, upon editing, all fields are available for editing and completion.


frjo:
@sks444 This looks really neat! And done with a minimum of code as well!

sks444:
@frjo, yes, I believe, this feature could also be used for concept and proposal purpose. Although the submission won't be going through all the proposal stages that we go through for `concept and proposal` workflow..etc.

Also, currently we only allow editing for a submission by applicants in specific stages. I.e. When a `More Information Requested` determination has been made. Do we want to allow editing in other stages as well, or leave the editing permissions as it is now?

frjo:
Leave the permissions as is, from what I know of OTF and Reset workflows that will work well.

They will likely use the information in the required fields to determine if the application is at all interesting. If it's they will ask for more info and the application can fill in more fields.

sks444:
Great! Then I think this should suffice the requirements mentioned in https://github.com/OpenTechFund/hypha/issues/1951. @Parbhat, just to confirm, could also take a look at the changes once?

Parbhat:
I took a look a few days back but did not add review.

Parbhat (approved):
Looks good.

hypha/apply/funds/models/applications.py
```diff
@@ -343,6 +343,15 @@ def get_form_parameters(self, submission_id=None):
 
         return form_parameters
 
+    def get_form_fields(self, draft=False):
+        form_fields = super(RoundBase, self).get_form_fields(draft)
+        applicationform = self.fund.specific.forms.first()
+        if applicationform.minimize_form:
+            for field in form_fields.values():
+                if not field.required:
```
frjo, line 351: It seems not all fields have the "required" attribute. I get a "'BlockFieldWrapper' object has no attribute 'required'" error when I test this locally. Wrapping this in a try/except seems to make it work.

frjo (changes requested):
When trying to submit an application I get an TypeError regarding the address field, "string indices must be integers" in "/Users/frjo/Sites/otf/addressfield/fields.py in clean, line 42".

I know we have had issues regarding required and the address field. It marks some parts of the address as required based on the country selected and the postal regulations in that country. This makes the address field as a whole always required even when the overall address field is not marked as such.

hypha/apply/funds/models/applications.py
frjo: I hope this is not big problems, OTF and Reset would really like this feature I believe.

sks444: Oh! I have had fixed this problem in my local. Forget to push it. Sorry.

sks444: Fixed it.

sks444:
@frjo, I have pushed a fix for addressfield error. It works. But I am still not completely sure about this. We should also put this on test for testing purposes.

>This makes the address field as a whole always required even when the overall address field is not marked as such.

If the above fix doesn't workout then maybe we could just mark the address field as required when it is selected in the form.

frjo (approved):

frjo:
Works perfectly in my testing. Deploying to test now.

frjo:
After showing this for OTF and Reset they love the feature but it doesn't work well to use the required checkbox to decide what shows on the minimised form.

Parbhat:
@frjo Is by the required checkbox you mean the checkbox to minimum application form or not? And it is not clear which fields will be shown in the minimized form.

frjo:
@Parbhat @sks444 With this PR all fields marked as required will show on the minimised form. 

Reset and OTF needs to have some none required folds on the minimised form as well as some required fields that only show on the full form. 

We therefore need another method to pick the files that should show on the minimised form.

sks444:
I think we would need to add another checkbox field to field blocks `minimize` and if that is checked we should not show those fields to apply form.

PR #2171, opened 2020-11-19
Fix reset production errors on sentry 

sks444:
There are three commits:

1. Fixes error:
```
Determination.DoesNotExist /{var}
Unhandled Determination matching query does not exist.
```

2. Fixes the error while assigning reviewers.

3. Fixes the error while accessing `/search` on public site.

### DIFF
--- a/hypha/public/search/views.py
+++ b/hypha/public/search/views.py
@@ -7,12 +7,15 @@ from django.shortcuts import render
 from wagtail.core.models import Page, Site
 from wagtail.search.models import Query
 
-from hypha.public.home.models import HomePage
-
 
 def search(request):
     site = Site.find_for_request(request)
-    if site != HomePage.objects.first().get_site():
+    if (
+        not site.is_default_site and
+        'apply' in site.site_name.lower() and
+        'apply' in site.hostname and
+        'apply' in site.root_page.title.lower()
+    ):
         raise Http404
 
     search_query = request.GET.get('query', None)
--- a/hypha/apply/determinations/views.py
+++ b/hypha/apply/determinations/views.py
@@ -430,7 +430,7 @@ class AdminDeterminationDetailView(DetailView):
     model = Determination
 
     def get_object(self, queryset=None):
-        return self.model.objects.get(submission=self.submission, id=self.kwargs['pk'])
+        return get_object_or_404(self.model, submission=self.submission, id=self.kwargs['pk'])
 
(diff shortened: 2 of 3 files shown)


frjo (approved):
Seems to work well. Safe to deploy do you think?

Nice work!

hypha/public/search/views.py
```diff
@@ -7,12 +7,15 @@
...
 
 def search(request):
     site = Site.find_for_request(request)
-    if site != HomePage.objects.first().get_site():
+    if (
+        not site.is_default_site and
+        'apply' in site.site_name.lower() and
+        'apply' in site.hostname and
+        'apply' in site.root_page.title.lower()
+    ):
```
sks444, line 18: `/search` functionality is only available on the public site. And to make sure that, we were checking the first site created for [HomePage](https://github.com/OpenTechFund/hypha/blob/f6a1934de94a4a349be62ea55706aa5f0ba463cc/hypha/public/home/models.py#L63) and if it is not current site we return 404.

```python
if site != HomePage.objects.first().get_site():
    return Http404
```

But reset uses [ResetNetworkHomePage](https://github.com/ResetNetwork/apply-app/blob/b01eb31e6e895d9c678bf6ef47cce11464ce2295/hypha/reset_network/reset_network_home/models.py#L58) to create its public home page. So there were no pages in HomePage and when accessing  get_site() it was throwing:
```
'NoneType' object has no attribute 'get_site'
```
---
To fix that now I am checking if the current site is the default site(as generally public site is set as default site)and also it is not the apply site. So in that case I am assuming site must be the public site and search functionality should be there.

I can't find a way to get the public site used for home page.

@Parbhat @frjo let me know if this works out or if there is other way to get the public site used for homepage, so that we can verify and only show search for that.

sks444:
@frjo, I am still trying to figure out the error we are getting while saving/publishing the funds.

If that is not done by today, and other fixes are urgent you may go ahead and deploy these.

hypha/public/search/views.py
frjo: The public site should be the default but the extra checks does not hurt. Checking for "first()" happened to work for OTF but not good to depend on for all sites as we have seen. I think this is a good solution.

frjo:
I will merge this one now since they work and fix issues we are having. I also prefer to many smaller deploys.

Please open new PR when you figured out the fund issue Reset are having. I tried for two days before giving up and passed it on to you :-).

PR #2173, opened 2020-11-20
Make application, determination and review forms hashable

sks444:
This fixes the Type Error: unhashable type: 'ApplicationBaseForm', when editing and publishing funds.

### How to reproduce

This bug is not only specific to Reset. And can be reproduced with any funds. And also not only with attaching application form but it can also be reproduced for `Review` and `Determination` forms. Here are the steps:

- Select a fund for editing.
- As you see, it has a review form attached:
Screenshot 2020-11-20 at 3 30 09 PM

- Delete the `Review Form` attached:
Screenshot 2020-11-20 at 3 29 58 PM

- Attach a `Review Form`:
Screenshot 2020-11-20 at 3 29 50 PM

- Do publish.
- This will lead to the error `TypeError: unhashable type: 'ApplicationBaseReviewForm'`

### Why is it happening

All the forms listed here, overrides their [__eq__](https://github.com/OpenTechFund/hypha/blob/392b789d15b547eb3955879daa79e8752d2ca69f/hypha/apply/funds/models/forms.py#L46) methods. But doesn't overrides the `__hash__` method.

It has been mandatory to override __hash__ when you override `__eq__` of an object. Otherwise `__hash__` of that object will be set as `None`. Hence making the object unhashable. 

See [this](https://stackoverflow.com/a/61235311) stackoverflow answer for more.

### Fix

- Add `__hash__` method for all the forms overriding `__eq__`.
- There was one other problem in the `__eq__` method which is fixed now: if the objects are saved in db, pk should be checked to make sure if they are equal. Otherwise in the above flow that I described, form will not get added, as the deleted and the new form would be considered equal.

### DIFF
--- a/hypha/apply/funds/models/forms.py
+++ b/hypha/apply/funds/models/forms.py
@@ -45,13 +45,18 @@ class AbstractRelatedForm(Orderable):
 
     def __eq__(self, other):
         try:
-            return self.fields == other.fields and self.sort_order == other.sort_order
+            if self.fields == other.fields and self.sort_order == other.sort_order:
+                # If the objects are saved to db. pk should also be compared
+                if hasattr(other, 'pk') and hasattr(self, 'pk'):
+                    return self.pk == other.pk
+                return True
+            return False
         except AttributeError:
             return False
 
     def __hash__(self):
         fields = [field.id for field in self.fields]
-        return hash((tuple(fields), self.sort_order))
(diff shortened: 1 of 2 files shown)


frjo (approved):
Works!

frjo:
Deploying to test now, later on today to live.

Excellent work!

frjo:
Deploying to live now, both Reset and OTF.

PR #2194, opened 2021-01-08
Code review changes of all submissions entry point in react app

vimal1083:
Add code review changes of [PR 2184](https://github.com/OpenTechFund/hypha/pull/2184)

### DIFF
--- a/hypha/apply/api/v1/filters.py
+++ b/hypha/apply/api/v1/filters.py
@@ -1,4 +1,3 @@
-from django.contrib.auth import get_user_model
 from django.db.models import Q
 from django_filters import rest_framework as filters
 from wagtail.core.models import Page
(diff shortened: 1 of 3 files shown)


frjo:
Have deployed to test, seems to work fine.

hypha/apply/funds/tables.py
```diff
@@ -190,16 +190,16 @@ def get_used_funds(request):
...
     return User.objects.filter(submission_lead__isnull=False).distinct()
 
 
-def get_reviewers(request):
+def get_reviewers():
     """ All assigned reviewers, staff or admin """
     return User.objects.filter(Q(submissions_reviewer__isnull=False) | Q(groups__name=STAFF_GROUP_NAME) | Q(is_superuser=True)).distinct()
 
 
-def get_screening_statuses(request):
```
sks444, line 202: @vimal1083, do not edit these methods. Rather create new methods in the api directory, and use those for the api at both places(For filter class and filter options api).

The idea is to not be dependent django view methods for api views.

Also editing old django view methods would break existing changes.

sks444 (approved):
Looks good.

@frjo, can we redeploy this to test.

frjo:
Deploying to test now, takes only a few minuts normally.

PR #2196, opened 2021-01-14
Add wagtail setting for site logo link

frjo:
Also fixing so custom site logo used everywhere.


hypha/public/utils/templatetags/util_tags.py
```diff
@@ -31,3 +34,14 @@ def field_type(bound_field):
...
+
+# Get the site logo link
+@register.simple_tag()
+def site_logo_link(site):
+    if SystemMessagesSettings.for_site(site).site_logo_link:
+        logo_link = SystemMessagesSettings.for_site(site).site_logo_link
+    else:
+        home = Page.objects.filter(slug='home').first()
+        logo_link = home.url
+    return logo_link
```
frjo, line 47: @Parbhat This seems to be working but is it the best way to do it?

Using the slug "home" as default is a bit of a hack I feel but that is how it works now. Best not to disturb existing sites.

Another organisation, CIPESA, that are going to start using Hypha wanted the ability to set this link.

hypha/public/utils/models.py
```diff
@@ -264,6 +271,7 @@ class Meta:
         MultiFieldPanel([
             FieldPanel('site_logo_default'),
             FieldPanel('site_logo_mobile'),
```
Parbhat, line 273: Not related to the PR but these should be [ImageChooserPanel](https://docs.wagtail.io/en/stable/reference/pages/panels.html#module-wagtail.images.edit_handlers) for easily choosing images instead of a dropdown.

hypha/public/utils/templatetags/util_tags.py
Parbhat: Looks like the requirement is to have different logo URL for different sites. As SystemMessagesSettings can be different for sites like Public and Apply. `slug` value `home` is for the public site Homepage. So, we can also get URL from [HomePage](https://github.com/OpenTechFund/hypha/blob/41db2f7da612d9523ca30ced29f57d3d2105c06e/hypha/public/home/models.py#L63) first instance.

hypha/public/utils/models.py
frjo: A lot nicer, thanks!

hypha/public/utils/templatetags/util_tags.py
frjo: That looks nicer and will give the same result for current OTF and Reset sites.

PR #2197, opened 2021-01-18
Updated pip packages January 2021

frjo:


frjo:
The tests are failing, seems to be django_file_form that gives an error. IS it the packages or something in our code?

```
System check identified no issues (0 silenced).
Migrations for 'django_file_form':
  venv/lib/python3.7/site-packages/django_file_form/migrations/0007_auto_20210118_1009.py
    - Create model TemporaryUploadedFile
    - Delete model UploadedFile

```

sks444:
Looks like it is happening because of a missing migrations file in django-file-form. Opened an [issue](https://github.com/mbraak/django-file-form/issues/422) there.

sks444:
Other upgrades are looking good except the issue with django-file-form.

@frjo, @Parbhat what are your views on using https://github.com/marketplace/dependabot-preview?

Parbhat:
Dependabot opens a number of PRs each day even for minor upgrades. Seen in other repos there are a lot of unwanted notifications and emails.

frjo:
Dependabot was active once upon a time. Have turned it off/on now, lets see how that goes.

I also remember it being a bit noisy.

sks444:
@frjo, we can upgrade django-file-form to 3.1.3 https://github.com/mbraak/django-file-form/issues/422

frjo:
Thanks @sks444! Deploying to test now.

PR #2209, opened 2021-01-27
Add proposal form field to streamfield determination forms

### LINKED ISSUE #2204: Multiple proposal forms for Double stage funds do not show up on new stream form Determinations
See PR https://github.com/OpenTechFund/hypha/pull/1412.

When setting multiple proposal forms on a Fund they do not show up when creating an determination to "Invite for proposal".

With the old hardcoded determinations forms a select widget with the proposal forms from the Fund show up. Allowing the staff to pick different proposal forms for different applications.

OTF needs this soon so if it can be fixed this week would be excellent.

sks444:
Fixes https://github.com/OpenTechFund/hypha/issues/2204

### DIFF
--- a/hypha/apply/determinations/views.py
+++ b/hypha/apply/determinations/views.py
@@ -1,5 +1,6 @@
 from urllib import parse
 
+from django import forms
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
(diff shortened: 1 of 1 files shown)


frjo (approved):
Works perfectly! Deployed to test. Thanks for the quick fix!

PR #2210, opened 2021-01-27
Use category question options to filter submissions

### LINKED ISSUE #2206: Add category questions as filters on submission table view
Add checkbox "Add to submission filter" to each category.

frjo:
… of adding categories to submission table filters.

Fixes #2206


hypha/apply/funds/tables.py
```diff
@@ -204,6 +204,10 @@ def get_screening_statuses(request):
         id__in=ApplicationSubmission.objects.all().values('screening_statuses__id').distinct('screening_statuses__id'))
 
 
+def get_categories(request):
+    return Category.objects.filter(filter_on_dashboard=True)
```
frjo, line 208: This should return all categories and the options belong to each category. I do not think categories themselves should have a checkbox, only the options.

hypha/apply/funds/tables.py
```diff
@@ -262,8 +266,9 @@ class SubmissionFilter(filters.FilterSet):
     lead = Select2ModelMultipleChoiceFilter(queryset=get_round_leads, label='Leads')
     reviewers = Select2ModelMultipleChoiceFilter(queryset=get_reviewers, label='Reviewers')
     screening_statuses = Select2ModelMultipleChoiceFilter(queryset=get_screening_statuses, label='Screening', null_label='No Status')
+    form_data = Select2ModelMultipleChoiceFilter(queryset=get_categories, label='Categories')
```
frjo, line 269: This obviously do not work at all but I'm hopeful you can figure out how to do it.

hypha/apply/funds/templates/funds/includes/screening_form.html
```diff
@@ -1,6 +1,6 @@
 {% if screening_form.should_show %}
 <div class="modal" id="screen-application" data-yes-statuses="{{ submission.yes_screening_statuses }}" data-no-statuses="{{ submission.no_screening_statuses }}" data-default-yes="{{ default_yes }}" data-default-no="{{ default_no }}">
-    <h4 class="modal__header-bar">Update status</h4>
+    <h4 class="modal__header-bar">Screen application</h4>
```
frjo, line 3: Unrelated change but noticed the header was wrong so opted for adding a fix here.

hypha/apply/funds/utils.py
```diff
@@ -38,3 +42,33 @@ def get_default_screening_statues():
...
+
+    return: list of set of suitable category options
+    """
+    ApplicationSubmission = apps.get_model('funds', 'ApplicationSubmission')
+    submission_data = ApplicationSubmission.objects.values_list('form_data', 'form_fields')
+    used_category_options = []
+    for item in submission_data:
+        for field in item[1]:
+            if isinstance(field.block, CategoryQuestionBlock):
+                used_category_options.append(item[0].get(field.id, 0))
```
frjo, line 66: I believe it's this row that gives me the "int() argument must be a string, a bytes-like object or a number, not 'list'" error.

hypha/apply/funds/utils.py
```diff
@@ -38,3 +42,33 @@ def get_default_screening_statues():
...
+      - Use the list of correct option to filter options
+
+    return: list of set of suitable category options
+    """
+    ApplicationSubmission = apps.get_model('funds', 'ApplicationSubmission')
+    submission_data = ApplicationSubmission.objects.values_list('form_data', 'form_fields')
+    used_category_options = []
+    for item in submission_data:
+        for field in item[1]:
+            if isinstance(field.block, CategoryQuestionBlock):
```
frjo, line 65: OTF have currently around 6000 submissions, iterating through all the form data takes 9s on my machine. Line 63 3s and line 65 the other 6s.

sks444: Hm, then this doesn't seem like a good idea to get category questions filters options. _(and also tests not passing in the ci is another issue with this method)_

@frjo, there are couple of work arounds I can think of, basically the problem is to show limited number of options in the filter.(the correct options which are used by submissions, and without looping through all the submission we cant get those options)

- One is to show all the options for category which has filter_on_dashboard `True`, and in order to minimise that more we can also have a filter_on_dashboard check for each options. So along with checking filter_on_dashboard for a category we can also check that for its options and then list those as filter options)

- Second is to have a field used_by_submission to option and when the submission is getting saved for the first time, mark it automatically, but the problem is this won't work for existing submissions. (Note that if we go with this way I would prefer using a field `category_id` instead of `used_by_submission` and checking if its null or not, as category id is used in form_data to save correct options so it will be helpful.)

frjo: I think we can simply show all options where the category has filter_on_dashboard set too true. No need to have that checkbox on each option.

If staff happens to pick an option with no submission they get an empty table, no biggy.

sks444:
Have completed the filter in Django views.

@vimal1083, we added a new filter `Category Questions` to our submission table view.

I have added this in the [filter options api](https://github.com/OpenTechFund/hypha/blob/f448b8fd8cbf45c89275272611110a7cdc22e44c/hypha/apply/api/v1/views.py#L108) and in the [submission api as a new filters options](https://github.com/OpenTechFund/hypha/blob/f448b8fd8cbf45c89275272611110a7cdc22e44c/hypha/apply/api/v1/filters.py#L48). Could you also add this to react table view.

hypha/apply/funds/tables.py
```diff
@@ -257,13 +258,18 @@ class SubmissionFilter(filters.FilterSet):
...
         (100, '100'),
     )
+
     round = Select2ModelMultipleChoiceFilter(queryset=get_used_rounds, label='Rounds')
     fund = Select2ModelMultipleChoiceFilter(field_name='page', queryset=get_used_funds, label='Funds')
     lead = Select2ModelMultipleChoiceFilter(queryset=get_round_leads, label='Leads')
     reviewers = Select2ModelMultipleChoiceFilter(queryset=get_reviewers, label='Reviewers')
     screening_statuses = Select2ModelMultipleChoiceFilter(queryset=get_screening_statuses, label='Screening', null_label='No Status')
+    category_options = Select2MultipleChoiceFilter(
+        choices=[], label='Category Options',
```
frjo, line 268: Make label only "Category", the filters takes to mush space as it is.

frjo (commented):
When filtering with only Category I get a proxy error each time (local Apache/Gunicorn).

It also seem to bog down the system, it takes many seconds for the error to appear.

When I combine it with other filters it always return 0 results when there should be some hits.

hypha/apply/funds/tables.py
sks444: Renamed it.

sks444:
@frjo, the filters are working for me locally.

Although I don't use Apache/Gunicorn setup locally. Can we deploy this to test, so that I can also check what is the issue.

frjo:
Deploying to test now, takes a few minuts.

sks444:
Thanks @frjo,

So I didn't see any error interacting with the category filter.

Although there is a bug, "when a category question has multiple correct selected options", the filters are not working properly.

Will push a fix for that.

frjo:
Yes, on test it seem to work just fine. Will need to investigate what the problem is in my local setup.

frjo:
Test has far fewer submissions, just 126. Will test that.

sks444:
Fixed the bug that I mentioned above.

@frjo, this should now work properly with any filter options.

Have tried to improve the performance a bit, still that is where I am a bit concerned. (just to mention, this will only happen when we use the category filter options so we are somewhat saved).

Let me know how it works for you on a larger set of submissions.

frjo:
@sks444 I can confirm that the filter works when the query set contains few submissions. But with thousands it breaks down.

sks444:
@frjo, what actually is happening with larger set of submissions? Can you post the error details/traceback if any?

PR #2217, opened 2021-02-01
Restore old submissions.html and create a submissions_react.html for …

### LINKED ISSUE #2216: User and staff flagged view are not working, template assumes React app present
The table flashes and then disappears, the consul show a lot of React errors.

frjo:
…React extended views to use.

Fixes #2216


vimal1083:
This fix looks good, but it solves the problem only for admin, in reviewer we will be having the same issue. so I have fixed it in react code https://github.com/OpenTechFund/hypha/pull/2218

frjo:
@vimal1083 Thanks, neat and clean fix in #2218 so closing this PR.

PR #2269, opened 2021-03-03
Remove the jump menu from the react app, no longer needed with filters in place

### LINKED ISSUE #2263: Remove jump menu from React app now that we have filters
Also see about removing the row that the jump manu is in, it will be mostly empty now and is a waste of space.

frjo:
Fixes #2263


frjo:
@vimal1083 @Parbhat Does this look right. React is not my area of expertise.

vimal1083 (approved):

PR #2282, opened 2021-03-09
Make duration not required.

### LINKED ISSUE #2274: Stop making "project duration" field required by Wagtail system

frjo:
Fixes #2274


Parbhat (approved):
Looks good

PR #2283, opened 2021-03-12
Scroll to top while switching submissions

### LINKED ISSUE #2233: In React app, scroll to top when switching submission
Related #2203

vimal1083:
Fixes #2233


frjo:
This is working on Chromium based browsers only, OTF and Reset  only use that so Ok for now.

Will open a new issue to get it working on Safari and Firefox, I think it's really a deeper issue with the React UI.

frjo (commented):
Approved

PR #2286, opened 2021-03-16
Add keyboard shortcuts in React 

### LINKED ISSUE #2246: Add keyboard shortcuts in the React app
Related #2203

Most wanted is to move to the next/previous submission. But nice to have on the most used buttons and the thumbs.

vimal1083:
Fixes #2246

### DIFF
--- a/hypha/static_src/src/app/src/components/GroupedListing/index.js
+++ b/hypha/static_src/src/app/src/components/GroupedListing/index.js
@@ -41,6 +41,11 @@ export default class GroupedListing extends React.Component {
 
     componentDidMount() {
         this.orderItems();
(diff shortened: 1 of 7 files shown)


frjo (changes requested):
@vimal1083  Works perfectly in my testing.

Can we add shortcuts for thumbs up/down as well? Test with using left and right arrows for that or if you have better options please use that.

Then we can deploy and let people play with it and come up with more shortcuts they would like to see.

vimal1083:
@frjo Have used ~ctrl~+u and ~ctrl~+d as shortcuts for thumbs up & down respectively. Please let me know if you need any changes

frjo (approved):

PR #2290, opened 2021-03-17
Add setting for extra text on application landing page.

### LINKED ISSUE #2288: Modify post-submission landing page until 30th of April.
**Describe why this work is needed.**

As a result of the redesign of the IFF Concept Note application form, we need a way to invite applicants to give us feedback on the new design so that we can evaluate it.

**For a time-fixed period**, when the applicant submits their application, they are brought to the page: https://apply.opentech.fund/internet-freedom-fund-concept-note/.

I want this page to be modified to include the following text:

-----

## We welcome your feedback on the applciation process.

We have redesigned the Internet Freedom Fund application form, and we welcome your feedback. 

We invite you to complete [this short survey](https://saneuxdesign.survey.fm/internet-freedom-fund-application-redesign-2021).

(This research is completely anonymous and is managed by our third-party partner, Simply Secure.)

-----

This text and link needs to be removed on the 30th of April, to the original page.

We thought about putting a question in the form asking if applicants would agree to OTF sharing their details with the research team, but this was disgarded as it was not seen as a positive approach.

The *best* way to deal with this would be to have post-submission landing pages that fund managers could modify as needed.

**Design wireframes**
Here is how it's expected to look like. If there's any issues, please tag me here.

image

**Additional context**

**NB:** This change will need to be reverted on the 30 of April. 

**Priority**

**Affected roles**

- Applicants
- General public

**Ideal deadline**

As soon as possible. Please!

frjo:
Fixes #2288


hypha/apply/funds/models/applications.py
```diff
@@ -364,6 +366,9 @@ def serve(self, request, *args, **kwargs):
 
                 if form.is_valid():
                     form_submission = self.process_form_submission(form, draft=draft)
+                    site = Site.find_for_request(request)
+                    application_settings = ApplicationSettings.for_site(site=site)
+                    extra_text = application_settings.extra_text_round
```
frjo, line 371: @Parbhat I want this extra text variable to be available in the template, see below. How does one make that happen in the best way?

hypha/apply/funds/templates/funds/application_base_landing.html
```diff
@@ -16,6 +16,8 @@ <h3>Thank you for your submission to the {{ ORG_LONG_NAME }}.</h3>
         <p>If you do not receive an e-mail within 15 minutes please check your
 spam folder and contact {{ ORG_EMAIL|urlize }} for further assistance.</p>
         {% with email_context=page.specific %}<p>{{ email_context.confirmation_text_extra|urlize }}</p>{% endwith %}
+
+        {{ extra_text | safe }}
```
Parbhat, line 20: It should be [richtext](https://docs.wagtail.io/en/stable/topics/writing_templates.html#rich-text-filter) filter instead of `safe`.

hypha/apply/funds/models/applications.py
Parbhat: Settings can be accessed in templates directly with settings context variable. More in [docs](https://docs.wagtail.io/en/stable/reference/contrib/settings.html#using-in-django-templates) and [example](https://github.com/OpenTechFund/hypha/blob/174a0317e20416fd94f0d731f673e2bb64999c34/hypha/apply/utils/templates/apply/404.html#L10).

Parbhat:
Is the requirement to show extra text on all submissions or only IFF? Currently, it will show on all submissions page after successful submission.

hypha/apply/funds/models/applications.py
frjo: Thanks! This I should have remembered.

frjo:
@Parbhat I think it is ok to show for a  round based applications, I have separate text for round and lab applications.

Any idea about the test error?

Parbhat:
@frjo It is related to the recent Wagtail 2.11 release. We need to add [run_before](https://docs.wagtail.io/en/stable/releases/2.11.html#run-before-declaration-needed-in-initial-homepage-migration) in homepage migrations [1](https://github.com/OpenTechFund/hypha/blob/main/hypha/public/home/migrations/0002_create_homepage.py) and [2](https://github.com/OpenTechFund/hypha/blob/main/hypha/apply/home/migrations/0002_add_apply_homepage.py). More in upgrade [considerations](https://docs.wagtail.io/en/stable/releases/2.11.html#run-before-declaration-needed-in-initial-homepage-migration) from release.

PR #2310, opened 2021-04-02
Make react/node optional

### LINKED ISSUE #2250: Make React/Node optional
Now you get an error if Node is not running when loading Django views that also have a React version. 

Can we make it so it loads without errors and works as normal sans React when Node is not running?

vimal1083:
Fixes #2250

### DIFF
--- a/hypha/apply/funds/templates/funds/grouped_application_list.html
+++ b/hypha/apply/funds/templates/funds/grouped_application_list.html
@@ -1,6 +1,9 @@
 {% extends "base-apply.html" %}
 {% load static %}
-{% load render_bundle from webpack_loader %}
+
+{% if can_render_bundle %}
+    {% load render_bundle from webpack_loader %}
+{% endif %}
 
 {% block extra_css %}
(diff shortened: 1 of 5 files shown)


Parbhat (changes requested):
@vimal1083 Can we look into some other solution for this? Checking for file existence can be slow and also makes an assumption about the directory path. Issue #2250 is looking for an old way where the Django table view loads even if the local node was not running. There was an error when clicking on 3 Panel-view as that need webpack running. As per the issue, now the Django table view also starts giving errors when webpack is not running. It would be good to have same old behaviour. Looks like something we changed in our React integration created this side effect.

frjo (changes requested):
Agree with @Parbhat. I would not mind a simple setting allowing a sysadmin to disable the React environment.

`REACT_ENVIROMENT_DISABLE = True` (or a better name maybe)

If not set or False then we run the React/node stuff as normal.

Parbhat:
Closed in #2384

PR #2311, opened 2021-04-04
Add wrapper around docker-compose

jvasile:
Adjust docker config to enable/disable salesforce integration based on whether we actually have any settings in salesforce.env

The docker-compose.yaml file references `salesforce.env` but that file doesn't exist.  Docker complains if it is missing.  We can just use an empty one, but we can't check it in because `salesforce.env` is in .gitignore, presumably because it has local settings out in the wild somewhere.  So now the wrapper makes that file if we need it to.

Similarly, `Dockerfile.dev` has a setting for salesforce integration.  The wrapper takes care of twiddling that setting for us, based on whether `salesforce.env` has content.

The wrapper lets us accommodate either enabled or disabled salesforce integration.  It need not be run every time.  You can still run `docker-compose` so long as you're confident the settings are already correct.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,7 @@
+This directory has a wrapper that runs docker-compose for you.  Pass
+it the same arguments that you would to docker-compose.
+
+If you are want salesforce integration, put your `salesforce.env` file
+in this directory.  If you don't want salesforce integration, delete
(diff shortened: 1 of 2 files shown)


frjo:
See https://github.com/HyphaApp/hypha/pull/2316, we can just remove the Salesforce stuff for now.

PR #2312, opened 2021-04-04
Set Salesforce integration default to 'off'

jvasile:
It makes no sense to default salesforce integration to enabled.  Anybody who wants it enabled has to add a salesforce.env file, as we don't ship one.  If we disable salesforce by default, the user at least gets a working hypha docker config.


frjo:
See https://github.com/HyphaApp/hypha/pull/2316, we can just remove the Salesforce stuff for now.

PR #2322, opened 2021-04-07
Prevent scroll to top in react when app updates automatically

### LINKED ISSUE #2318: Submissions scrolls to top every x seconds
Related to https://github.com/HyphaApp/hypha/issues/2233

The react app reloads automatically every x seconds. When that happens the app is scrolled to top.

vimal1083:
Fixes #2318


frjo (approved):
Thanks for quick fix!

Issue #1943, opened 2020-04-21
Proposed changes to reminder function


danblah:
Here are some refinements needed for the reminder function on applications.

* TIme, make the hh:mm:ss not required. If not set, default to 00:00:00 UTC.

Replace the actions drop-down with the following fields:
* Title: A required field. A plain-text field, max char of 60 chars.
* Assign: A required field. Options (multi-select): Applicant, Partners, Lead, Reviewers, Team.
* Description: Not-required field. A plain-text box.

Changes to notifications:

TBD

Later tasks:
* Ability to click a reminder in the reminders box and view/edit full reminder
* In the admin panel, a place to define default reminder templates that auto-populate the above fields
* A small/light buttons on the reminder interface to choose templates
* Allowing rich-text in the description field

danblah:
@georgiamoon and @bumbleblue for ui/ux  thoughts.

Parbhat:
@frjo Is this ready to work on? Or need UI/UX and requirements updates.

bumbleblue:
I don't have data from the research on the reminder function, but I can second @danblah's suggestions. I would prioritize having default reminder templates (at all) over adding the admin interface for creating new templates.

Is the description field for finding the reminder ("full proposal late - first reminder") or the actual text in the email/notification ("Dear ..., We have not yet received you full proposal ..."). I haven't seen it in action yet!

saoun:
@ei8fdb @bumbleblue are any of you working on this?

fourthletter:
Please remove the **clock** from the reminder function. We only need the date. Thank you.

Program Managers are sending reminders to applicants regarding upcoming application deadlines and potentially setting reminders for other colleagues.

<img width="579" alt="Screen Shot 2021-04-23 at 1 49 42 PM" src="https://user-images.githubusercontent.com/20019656/115866992-c54abc80-a43a-11eb-9d46-e2fd99057bc8.png">

ei8fdb:
We're closing this ticket as it is being handled in #2247

PR #2334, opened 2021-04-16
Add Reminders in React

### LINKED ISSUE #2247: Make it possible to set reminders in the React app
Make it possible to set reminders for the PM and applicant to action some task.

The reminder is sent by email, default at a time the system chooses.

Related #2203

vimal1083:
Fixes #2247 

https://user-images.githubusercontent.com/8245654/114980981-e8112a00-9eaa-11eb-8b77-c1b12c946391.mp4

If reminder created in Django, it will appear as "untitled reminder" in react

https://user-images.githubusercontent.com/8245654/114981056-024b0800-9eab-11eb-9dbf-63f034438df8.mp4


hypha/apply/funds/models/utils.py
```diff
@@ -32,6 +34,18 @@
 LIMIT_TO_REVIEWER_GROUPS = {'groups__name__in': REVIEW_GROUPS}
 
 
+class CustomerTypes(Enum):
```
frjo: UserTypes is better term I think.

hypha/static_src/src/app/src/containers/DisplayPanel/index.js
```diff
@@ -78,6 +79,7 @@ const DisplayPanel = props => {
             <Determination submissionID={submissionID} submission={submission}/> : null}
             {/* <ScreeningOutcome submissionID={submissionID} /> */}
             <StatusActions submissionID={submissionID} />
+            <ReminderContainer submissionID={submissionID}/>
```
frjo: Move this below screening.

frjo (changes requested):
Works as it should but have some things I want changes and added.

* The date picker should use ISO 8601 (or if possible adapt to the users/browser locale).
* The "Assign" select field in the Reminder popup uses the keys and not the values as choices.

Also the new fields etc. for reminders should be added to the Django side of the Reminder UI as well. We also need to work on the backend for sending the extended reminders. But that is coming I assume since this is a draft still.

Parbhat:
@frjo Going the requirements of the feature and it seems that we can skip the suggested [assign](https://github.com/HyphaApp/hypha/issues/1943#issue-604035180) field for now. As not sure how it will be used. Currently, we have an [action](https://github.com/HyphaApp/hypha/blob/a256a25c832fcb9f2bf0b37c4449ac9c86222a70/hypha/apply/funds/models/reminders.py#L30) field that helps to pick the reminder purpose. So, for now, we can replicate the Reminder feature from Django in React app + new fields like Title, description added to both + some other improvements. Sounds ok?

vimal1083:
@frjo Now we have title and description for a reminder, Do you want to add this information in reminder e-mail?

Parbhat:
@vimal1083 It would be good to have the new info in email as well.

Parbhat:
@frjo We can deploy it to test.

frjo (changes requested):
I have removed the time part from reminders in the Django side (only changed widget, not db field), can you do the same for the React side? Then I can get this merged. Nice work!

vimal1083:
@frjo yeah, updated in React

frjo (approved):

hypha/static_src/src/app/src/containers/ReminderContainer/index.js
```diff
@@ -0,0 +1,128 @@
...
+                        <Modal
+                            className={classes.modal} 
+                            open={this.state.open}
+                        >
+                            <>
+                            <ReminderForm 
+                                submissionID={this.props.submissionID} 
+                                closeForm={() => this.setState({open: false})}
+                            />
+                            </>
```
v1k45: Is the fragment block necessary here?

There is only a single component inside `Modal`, we can directly place `ReminderForm` here.

hypha/static_src/src/app/src/containers/ReminderContainer/index.js
```diff
@@ -0,0 +1,128 @@
...
+const withReducer = injectReducer({ key: 'ReminderContainer', reducer });
+const withSaga = injectSaga({ key: 'ReminderContainer', saga });
+  
+  
+export default compose(
+withSaga,
+withReducer,
+withConnect,
+withRouter,
+withStyles(styles)
```
v1k45: Style nitpick:

These lines should be indented.

```suggestion
    withSaga,
    withReducer,
    withConnect,
    withRouter,
    withStyles(styles)
```

hypha/static_src/src/app/src/containers/ReminderContainer/actions.js
```diff
@@ -0,0 +1,25 @@
...
+  data
+});
+
+export const showLoadingAction = () => ({
+	type: ActionTypes.SHOW_LOADING,
+})
+
+export const hideLoadingAction = () => ({
+	type: ActionTypes.HIDE_LOADING,
+})
```
v1k45: Style nitpick:

These two functions have inconsistent indentation. It is indented using tab where as other parts in the part have 2-space indentation.

Same with other occurrences in the changeset.

hypha/static_src/src/app/src/containers/ReminderContainer/index.js
```diff
@@ -0,0 +1,128 @@
...
+    handleModalClose = () => {
+        this.setState({open : false})
+    }
+
+    render(){
+        const { classes } = this.props;
+        return (
+            this.props.reminderInfo.loading 
+            ? <LoadingPanel /> :
+            <div className="reminder-container">
```
v1k45: What do you think about moving the <LoadingPanel /> outside the main `return` block.

We can put an early-return to avoid nested contions.

hypha/static_src/src/app/src/containers/ReminderContainer/selectors.js
```diff
@@ -0,0 +1,24 @@
...
+export const selectFieldsRenderer = state =>
+  state.ReminderContainer ? state.ReminderContainer : initialState;
+
+export const selectReminderContainer = createSelector(selectFieldsRenderer, domain => domain);
+
+export const selectReminders = createSelector(selectReminderContainer, domain => {
+  let reminders = []
+  domain.reminders && domain.reminders.map(reminder => {
+    if(reminders.find(r => r.grouper == reminder.action_type)){
+      const index = reminders.indexOf(reminders.find(r => r.grouper == reminder.action_type))
```
v1k45: We are looping over `reminders` 3 times.

What do you think about using `findIndex` to cut down the loops?

Something like:

```js
const existingReminderIndex = reminders.findIndex(r => r.grouper == reminder.action_type);
// reminder is found.
if (existingReminderIndex != -1) {
  reminders[existingReminderIndex].list.push(reminder)
} else {
  // new reminder.
  reminders.push({ grouper: reminder.action_type, list: [reminder] })
}
```

hypha/static_src/src/app/src/containers/ReminderContainer/index.js
```diff
@@ -0,0 +1,128 @@
...
+                            <ReminderForm 
+                                submissionID={this.props.submissionID} 
+                                closeForm={() => this.setState({open: false})}
+                            />
+                            </>
+                        </Modal>
+                        {this.props.reminders && this.props.reminders.length
+                        ?
+                        this.props.reminders.map(reminders =>
+                        <ReminderList
```
v1k45: Style Nitpick:

We can make this block more easier to read by extracting `this.props.reminders || []`  to a separate variable and then using it here without the extra condition.

Also, it is possible for `reminders` be to anything other than a list after it is loaded?

v1k45 (changes requested):
Looking good, a few nitpicks and suggestions:

hypha/static_src/src/app/src/containers/ReminderContainer/index.js
vimal1083: Without this Fragment, there will be a warning message because of material UI bug. By adding this fragment, it got resolved

v1k45 (approved):
Nice work! LGTM :+1:

PR #2337, opened 2021-04-20
Make duration not required

### LINKED ISSUE #2274: Stop making "project duration" field required by Wagtail system

frjo:
Fixes #2274

Reopening this branch since there is a bug with current forms not working with their change.

```
TypeError
unsupported operand type(s) for &: 'NoneType' and 'bool'
```

https://github.com/HyphaApp/hypha/blob/220edd813c5efab597203758cbf6bd625451e99c/hypha/apply/stream_forms/models.py#L94


frjo:
@Parbhat Needed to revert this since current forms and submissions did not work. Bad testing on my side.

Current forms only needed a resave to start working but for all submitted submissions that is not an option.

Can your team take a look at this?

PR #2384, opened 2021-05-23
Allow running project without webpack/node

### LINKED ISSUE #2250: Make React/Node optional

v1k45:
Fixes #2250 

## Problem

Node/Webpack is needed to access pages which implement part of their functionality in React. 

E.g. Submissions have table view which is rendered using Django, it cannot be viewed without webpack/building js.

## Solution

Add a new `render_bundle` template tag which conditionally renders webpack-loader bundles based on `ENABLE_WEBPACK_BUNDLES` setting.

When disabled, it makes the React/Webpack loaded parts disappear.

## Demo

https://user-images.githubusercontent.com/7877501/119476186-7494ef00-bd6b-11eb-800a-57ac2f938301.mp4

### DIFF
--- a/hypha/settings/base.py
+++ b/hypha/settings/base.py
@@ -660,7 +660,9 @@ REFERRER_POLICY = env.get('SECURE_REFERRER_POLICY',
 
 # Webpack bundle loader
 # When disabled, all included bundles are silently ignored.
-ENABLE_WEBPACK_BUNDLES = True
+if env.get('ENABLE_WEBPACK_BUNDLES', 'true').lower().strip() == 'true':
+    ENABLE_WEBPACK_BUNDLES = True
+
 WEBPACK_LOADER = {
     'DEFAULT': {
         'BUNDLE_DIR_NAME': 'app/',
--- a/hypha/settings/base.py
+++ b/hypha/settings/base.py
@@ -139,7 +139,6 @@ INSTALLED_APPS = [
     'hijack',
     'compat',
     'pagedown',
-    'webpack_loader',
 
     'salesforce',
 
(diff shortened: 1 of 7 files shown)


v1k45:
One observation I had is that the `dev.py` settings uses `webpack-stats.json`, which is only generated on watch mode, i.e. when you're running webpack server.

If someone wishes to have the react components as well, with a one time build, they have to rename the `webpack-stats-prod.json` file. This is a manual process and can be a bit confusing.

We should probably make copy of `webpack-stats-prod.json` as `webpack-stats.json` automatically during building.

hypha/settings/base.py
```diff
@@ -659,6 +658,9 @@
 REFERRER_POLICY = env.get('SECURE_REFERRER_POLICY',
                           'no-referrer-when-downgrade').strip()
 
+# Webpack bundle loader
+# When disabled, all included bundles are silently ignored.
+ENABLE_WEBPACK_BUNDLES = True
```
frjo: I think we should allow this to be set with a env var. Something like:

```
if env.get('ENABLE_WEBPACK_BUNDLES', 'true').lower().strip() == 'true':
    ENABLE_WEBPACK_BUNDLES = True
```

Parbhat:
@vimal1083 Good observation, you can fix it in a separate PR.

frjo:
And welcome to the Hypha team @v1k45!

v1k45:
👋 Glad to be here!

Parbhat (approved):
Looks good

PR #2385, opened 2021-05-24
Updated to Wagtail 2.13

sandeepsajan0:
Update the Wagtail version

1. updated from 2.11.7 to 2.12
2. updated from 2.12 to 2.13


frjo (approved):
@sandeepsajan0  Nice Work, welcome to the Hypha team!

frjo:
Deployed to test at https://test.hypha.app/ now. Needed https://github.com/HyphaApp/hypha/pull/2387 as well or  I got:

`ModuleNotFoundError: No module named 'unidecode'`

Ideas on what's up with that?

Parbhat:
unidecode was [removed](https://github.com/wagtail/wagtail/blob/4daeed1bf5fa1c28eafba36993a43df3cc3a51ec/CHANGELOG.txt#L187) in Wagtail 2.11, @sandeepsajan0 you can confirm if other parts of Hypha uses it.

sandeepsajan0:
Sure @Parbhat  @frjo , I'll check and get back to you.

sandeepsajan0:
@Parbhat Yes, we are using it in [apply/stream_form/blocks](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/stream_forms/blocks.py#L47). According to wagtail documentation, we should use `anyascii` instead of `unidecode`. Should I go ahead and create a separate PR for this one?

Parbhat:
@sandeepsajan0 You can fix it in this PR. Not sure why we didn't noticed it earlier as we are currently on Wagtail 2.11 and unidecode was removed in that version.

frjo:
Good to have the fix in this PR. I did not notice this in my local setup since the unidecode package was already installed by earlier version of Wagtail.

requirements.txt
```diff
@@ -46,3 +46,4 @@ wagtail-cache==1.0.1
 wagtail-purge==0.1
 wagtail==2.13
 whitenoise==5.2.0
+anyascii==0.2.0
```
frjo: @sandeepsajan0 Very minor but can you move this to top so list is sorted please.

sandeepsajan0: Sure @frjo

requirements.txt
```diff
@@ -3,6 +3,7 @@ scout-apm==2.17.0
 sentry-sdk==0.19.5
 
 # Production dependencies
+anyascii==0.2.0
```
Parbhat: We do not need to have it in our requirements as it will be installed from the Wagtail dependencies [list](https://github.com/wagtail/wagtail/blob/8487c67ebb35984f7694537e779583c7d6b92cf4/setup.py#L38).

requirements.txt
```diff
@@ -44,5 +45,5 @@ social_auth_app_django==4.0.0
 tomd==0.1.3
 wagtail-cache==1.0.1
```
Parbhat: It would be good to check if other Wagtail specific packages are compatible with the latest Wagtail 2.13 version.

```shell
wagtail-cache==1.0.1
wagtail-purge==0.1
```

requirements.txt
frjo: Removed it now, thanks.

requirements.txt
frjo: Updated wagtail-cache to latest version.

PR #2393, opened 2021-05-26
Create webpack.static.dev config file for local dev

v1k45:
`gulp build` creates `webpack-stats-prod.json`, which cannot be directly used in local development because Django local settings use `webpack-stats.json` instead of `webpack-stats-prod.json`.

`gulp app` uses dev config, which requires webpack dev server to be running.

This creates a new config to generate `webpack-stats.json` file with production config and uses it in `gulp app` task.

**Testing steps**:
- Delete contents under `hypha/static_compiled`
- View Submissions page, you'll see webapck error
- Run `gulp app`
- Verify that the page works and has react components.

**Proof**:

image
image

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -73,6 +73,7 @@ sass.compiler = require('sass');
 
 // Load webpack config
 var webpackDev = () => require(options.theme.app + 'webpack.dev.config.js');
+var webpackStaticDev = () => require(options.theme.app + 'webpack.static.dev.config.js');
 var webpackProd = () => require(options.theme.app + 'webpack.prod.config.js');
 var webpackAnalyze = () => require(options.theme.app + 'webpack.analyze.config.js');
 
@@ -178,7 +179,7 @@ gulp.task('scripts:production', gulp.series('clean:js', function js () {
 // Build App.
 gulp.task('app', function() {
     return gulp.src(options.theme.app + 'src/')
-        .pipe(webpackStrm( webpackDev() ))
+        .pipe(webpackStrm( webpackStaticDev() ))
         .pipe(gulp.dest(options.theme.app_dest));
 })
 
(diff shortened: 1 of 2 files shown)


hypha/static_src/src/app/webpack.prod.config.js
```diff
@@ -13,6 +13,8 @@ prodConfig.plugins = prodConfig.plugins.concat([
         NODE_ENV: 'production',
         API_BASE_URL: null ,
     }),
+    // Used in local development, where React app is precompiled and webpack server is not actively running.
+    new BundleTracker({filename: './hypha/static_compiled/app/webpack-stats.json'}),
```
vimal1083: here we are creating stats file for dev environment in production command. instead you can define a new gulp task like below

`gulp.task('buildDev', gulp.series(gulp.parallel(gulp.series('styles', 'scripts', 'app'), 'images', 'fonts', 'lint'), 'collectstatic'));
`

v1k45: Interesting. I was thinking of re-using the prod command.

I'll give this a try.

v1k45: @vimal1083 The `app` task produces a `webpack-stats.json` file which is only compatible with the webpack server.

It contains the `publicPath` property which makes webpack-loader load it from webpack server (which is not running). The scripts then fail to load because of this.

An alternative for this is to define a webpack config similar to production, but which only used in the `app` task.

How should I proceed?

vimal1083: Try to parameterize the webpack config, so that we can have same file for both environments. If that is not possible we can have separate webpack config file for dev

v1k45: I ended up going with a separate config, parameterizing the config was making it unnecessarily complicated.

frjo (approved):
Works for me.

vimal1083:
Looks Good

vimal1083 (approved):

PR #2397, opened 2021-05-27
enhance: Use is_apply_staff in replacement of is_staff for users

### LINKED ISSUE #2390: The "user.is_staff" used when it should be "user.is_apply_staff"
The "is_staff" is a Django setting for super admins. The "is_apply_staff" is a Hypha thing telling the system that the user is a member of the role/group "staff".

Are they not confused in some places? Like e.g. these two?

https://github.com/HyphaApp/hypha/blob/a06c1d202082f3dba6b17e71b2cd1a47b9aa828c/hypha/apply/projects/templatetags/approval_tools.py#L13

https://github.com/HyphaApp/hypha/blob/a06c1d202082f3dba6b17e71b2cd1a47b9aa828c/hypha/apply/projects/templates/application_projects/project_detail.html#L161

sandeepsajan0:
Fixes #2390

Replace user's inbuilt `is_staff` method with hypha user's custom `is_apply_staff` property.

### DIFF
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -1008,7 +1008,7 @@ class AssignedReviewersQuerySet(models.QuerySet):
             else:
                 groups = {REVIEWER_GROUP_NAME}
         elif not groups:
-            if reviewer.is_apply_staff:
+            if reviewer.is_staff or reviewer.is_superuser:
                 groups = {STAFF_GROUP_NAME}
             else:
                 groups = {REVIEWER_GROUP_NAME}
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -1008,7 +1008,7 @@ class AssignedReviewersQuerySet(models.QuerySet):
             else:
                 groups = {REVIEWER_GROUP_NAME}
         elif not groups:
-            if reviewer.is_staff or reviewer.is_superuser:
+            if reviewer.is_apply_staff:
                 groups = {STAFF_GROUP_NAME}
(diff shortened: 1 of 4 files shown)


hypha/apply/funds/models/submissions.py
```diff
@@ -1008,7 +1008,7 @@ def get_or_create_for_user(self, submission, reviewer):
             else:
                 groups = {REVIEWER_GROUP_NAME}
         elif not groups:
-            if reviewer.is_staff or reviewer.is_superuser:
```
frjo: I think this is a case where is_staff etc. really should be used.

frjo (changes requested):

hypha/apply/funds/models/submissions.py
sandeepsajan0: I think we are checking the same thing in `is_apply_staff`.  `STAFF_GROUP_NAME` or `is_superuser`.
Or it might be any other custom `is_staff`?
okay, let me check.

sandeepsajan0: Yes wherever we are setting is_staff=True, we are adding STAFF_GROUP_NAME in user groups. 
And in property, is_apply_staff we are checking for both of the things if the user has STAFF_GROUP_NAME or is_superuser. 
I am not getting any edge cases, can you please enlighten it for me.

sandeepsajan0: @frjo I got your point now. I changed it accordingly.

sks444 (approved):
Looks good.

frjo (approved):

PR #2402, opened 2021-05-31
fix: Project table filter's broken dropdown

### LINKED ISSUE #2388: Project table filters has broken drop downs
The project table filters at https://test-apply.hypha.app/apply/projects/ has broken drop downs. Compeer with https://test-apply.hypha.app/apply/submissions/all/.

Most likely a update to filters and/or select2 that has not been implemented.

sandeepsajan0:
Fixes #2388

As I left a [comment](https://github.com/HyphaApp/hypha/issues/2388#issuecomment-851174374) in the issue, these were getting overwritten because of their same names for django-filter. I tried to look into its templates code as well but I found this clue via its html(inspect, select's id and name).

Broken Filters Behaviour:
image

Fixed Behaviour:
image

### DIFF
--- a/hypha/apply/projects/filters.py
+++ b/hypha/apply/projects/filters.py
@@ -43,9 +43,9 @@ class ProjectListFilter(filters.FilterSet):
         (1, 'Behind schedule'),
     )
 
-    fund = Select2ModelMultipleChoiceFilter(label='Funds', queryset=get_used_funds)
-    lead = Select2ModelMultipleChoiceFilter(label='Lead', queryset=get_project_leads)
-    status = Select2MultipleChoiceFilter(label='Status', choices=PROJECT_STATUS_CHOICES)
+    project_fund = Select2ModelMultipleChoiceFilter(label='Funds', queryset=get_used_funds)
+    project_lead = Select2ModelMultipleChoiceFilter(label='Lead', queryset=get_project_leads)
+    project_status = Select2MultipleChoiceFilter(label='Status', choices=PROJECT_STATUS_CHOICES)
     query = filters.CharFilter(field_name='title', lookup_expr="icontains", widget=forms.HiddenInput)
     reporting = filters.ChoiceFilter(
         choices=REPORTING_CHOICES,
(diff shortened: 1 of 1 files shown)


hypha/apply/projects/filters.py
```diff
@@ -43,9 +43,9 @@ class ProjectListFilter(filters.FilterSet):
         (1, 'Behind schedule'),
     )
 
-    fund = Select2ModelMultipleChoiceFilter(label='Funds', queryset=get_used_funds)
-    lead = Select2ModelMultipleChoiceFilter(label='Lead', queryset=get_project_leads)
-    status = Select2MultipleChoiceFilter(label='Status', choices=PROJECT_STATUS_CHOICES)
+    project_fund = Select2ModelMultipleChoiceFilter(label='Funds', queryset=get_used_funds)
+    project_lead = Select2ModelMultipleChoiceFilter(label='Lead', queryset=get_project_leads)
+    project_status = Select2MultipleChoiceFilter(label='Status', choices=PROJECT_STATUS_CHOICES)
```
sks444: @sandeepsajan0, this fixes the ui but breaks the filter functionality, as these fields are not model fields anymore. We may need to use [to_field_name](https://django-filter.readthedocs.io/en/stable/ref/filters.html#to-field-name) param here.

sks444 (changes requested):
@sandeepsajan0, given that filter fields with similar `id` and `name` is the cause of the problem. It would be interesting to know how this html id and name is being generated for that field? Is it coming from django-filter package itself?

I think a stable solution for this would be to have `id` and `name` generated with the set model name prefix. E.g. in our case,  for status field on project filter which has project model set, id could be `project-status` instead of just `status`.  So that no two filters on the same page have same id/name.

If it's coming from the filter package, maybe we can open an issue there, describing the steps to reproduce it.

Until then I think that our fix does the work.

sks444 (approved):
Looks good.

sks444:
We did some research on others ways(https://github.com/HyphaApp/hypha/pull/2402#pullrequestreview-672543803) to avoid this problem. Which looks complicated, so going with renaming the filter fields for now.

frjo (approved):
Excellent work! Nice and clean fix.

Issue #2379, opened 2021-05-18
Add optional vendor creation step to projects


frjo:
This story is part of epic #2352.

> As a contractor I need to provide my banking, tax information to the funder so they can setup me up in their accounting system.

**What the user needs to do**
In Hypha the contractor needs to submit this information:

1. name of organisation or person mentioned on the contract with OTF
2. name of person with authority to sign the contract
3. upload completed w8 or w9 form (depending on their US tax status)
4. bank account holder name
5. bank account number
6. bank account routing number
7. bank account currency
It needs to provide the name and abbreviation. Screenshot from the designs:
![image](https://user-images.githubusercontent.com/1899506/118988419-15a13580-b979-11eb-974c-7f5d5ef3886d.png)

8. address of the bank where the account is located
[A country picker, preferably working like this](https://baymard.com/labs/country-selector)
![Kapture 2021-05-20 at 15 43 59](https://user-images.githubusercontent.com/1899506/118989477-f951c880-b979-11eb-9266-8e85d07d23c8.gif)

9. provide extra information to receive payments if needed - [full information is in the designs](https://www.figma.com/file/6HzFAw7k9zXwuujvyf5uVO/OTF-System?node-id=192%3A2972)

**Designs**
These designs show how the interface should work:
https://www.figma.com/file/6HzFAw7k9zXwuujvyf5uVO/OTF-System?node-id=192%3A2972

**List**
- [ ] Add a setting in "Project settings" (already exist) to turn vendor creation on/off. If on it will be a required step for the applicant/vendor.
- [ ] When vendor creation is on it is a required step before contract and invoicing can happen.
- [ ] On MESSAGES.CREATED_PROJECT send e-mail to applicant informing them that vendor setup is needed.
- [ ] Create a form for adding vendor information. For fields and UX see below.
- [ ] Create a vendor in third party service if exist
- [ ] Update vendor status from third party service.
- [ ] Allow applicant and staff to edit the vendor information.
- [ ] Notify lead when vendor information has been added.

The vendor information like bank account might change over time of the project. We need to keep track of this somehow. I see two options:

1. Have revisions on the vendor model och link revisions to invoices.
2. Copy current vendor information to the invoice and keep the history there.

In a separate app that is OTF only we will be making API calls to create the vendor in a third party system.

ei8fdb:
@frjo @Parbhat 

Here's the designs for the setup new contractor - 
[...]

PR #2404, opened 2021-06-01
Restructure projects app models

sks444:
Moved specific models to specific files. Similar to what we follow through-out Hypha.

Makes it easy to read code and to make changes on specific part of the codebase.

This also allows us to put the [vendor model](https://github.com/HyphaApp/hypha/pull/2401) in a separate file.

Would be doing more of structuring/cleanup as we proceed further but this is a good first step.


frjo:
@sks444 Since no one is using projects we can do what we need.

I think this might be a good PR to rename "payment" to "invoice".

sks444:
@frjo, this pr already has a good amount of changes. Doing that here could mess things up a bit.

I also need some time to understand what's going on with Payment and how the invoicing workflow would be.

Was thinking of finishing the [first version of vendor](https://github.com/HyphaApp/hypha/issues/2379) and then taking up the invoicing related issues. .

frjo (approved):
Nice work!

PR #2410, opened 2021-06-04
Add meta terms in react

### LINKED ISSUE #2241: Ability to set meta terms and filter by them in the React app
Related #2203

vimal1083:
Fixes #2241  

https://user-images.githubusercontent.com/8245654/120791984-81c39200-c552-11eb-981a-ebe76f757f18.mp4

### DIFF
--- a/hypha/apply/api/v1/serializers.py
+++ b/hypha/apply/api/v1/serializers.py
@@ -15,6 +15,7 @@ from hypha.apply.funds.models import (
     ScreeningStatus,
 )
 from hypha.apply.review.models import Review, ReviewOpinion
+from hypha.apply.categories.models import MetaTerm
 from hypha.apply.review.options import RECOMMENDATION_CHOICES
 from hypha.apply.users.groups import PARTNER_GROUP_NAME, STAFF_GROUP_NAME
 
@@ -343,3 +344,17 @@ class CommentEditSerializer(CommentCreateSerializer):
 class UserSerializer(serializers.Serializer):
     id = serializers.CharField(read_only=True)
     email = serializers.CharField(read_only=True)
+
+
+class MetaTermsSerializer(serializers.ModelSerializer):
+    name = serializers.StringRelatedField()
+    children = serializers.SerializerMethodField(
+        read_only=True, method_name="get_children_nodes"
+    )
+    class Meta:
+        model = MetaTerm
+        fields = ("name", "id", "children")
+
+    def get_children_nodes(self, obj):
+        child_queryset = obj.get_children()
+        return MetaTermsSerializer(child_queryset, many=True).data
--- a/hypha/apply/api/v1/serializers.py
+++ b/hypha/apply/api/v1/serializers.py
@@ -371,7 +371,6 @@ class UserSerializer(serializers.Serializer):
 
 
 class MetaTermsSerializer(serializers.ModelSerializer):
-    name = serializers.StringRelatedField()
     children = serializers.SerializerMethodField(
         read_only=True, method_name="get_children_nodes"
     )
--- a/hypha/apply/api/v1/serializers.py
+++ b/hypha/apply/api/v1/serializers.py
@@ -4,11 +4,8 @@ from django_bleach.templatetags.bleach_tags import bleach_value
 from rest_framework import serializers
 
 from hypha.apply.activity.models import Activity
-<<<<<<< 44d17049428a1f67169a8eb29fc5cf5c0b909fa6
-=======
 from hypha.apply.api.v1.screening.serializers import ScreeningStatusSerializer
 from hypha.apply.categories.models import MetaTerm
->>>>>>> feat: Add update submission meta-terms API and add meta-terms to submission details API
 from hypha.apply.determinations.models import Determination
 from hypha.apply.determinations.templatetags.determination_tags import (
     show_determination_button,
--- a/hypha/apply/api/v1/serializers.py
+++ b/hypha/apply/api/v1/serializers.py
@@ -4,6 +4,11 @@ from django_bleach.templatetags.bleach_tags import bleach_value
 from rest_framework import serializers
 
 from hypha.apply.activity.models import Activity
+<<<<<<< 44d17049428a1f67169a8eb29fc5cf5c0b909fa6
+=======
+from hypha.apply.api.v1.screening.serializers import ScreeningStatusSerializer
+from hypha.apply.categories.models import MetaTerm
+>>>>>>> feat: Add update submission meta-terms API and add meta-terms to submission details API
 from hypha.apply.determinations.models import Determination
 from hypha.apply.determinations.templatetags.determination_tags import (
     show_determination_button,
@@ -15,7 +20,7 @@ from hypha.apply.funds.models import (
     ScreeningStatus,
 )
 from hypha.apply.review.models import Review, ReviewOpinion
-from hypha.apply.categories.models import MetaTerm
+
 from hypha.apply.review.options import RECOMMENDATION_CHOICES
 from hypha.apply.users.groups import PARTNER_GROUP_NAME, STAFF_GROUP_NAME
 
@@ -155,13 +160,36 @@ class SubmissionListSerializer(serializers.ModelSerializer):
         return obj.round_id or obj.page_id
 
 
+class MetaTermsDetailSerializer(serializers.ModelSerializer):
+    parent = serializers.SerializerMethodField()
+    parent_id = serializers.SerializerMethodField()
+    class Meta:
+        model = MetaTerm
+        fields = ("id", "name", "parent", "parent_id")
+
+    def get_parent(self, obj):
+        parent = obj.get_parent()
+        if parent:
+            return parent.name
+
+    def get_parent_id(self, obj):
+        parent = obj.get_parent()
+        if parent:
+            return parent.id
+
+
 class SubmissionSummarySerializer(serializers.Serializer):
     summary = serializers.CharField(write_only=True)
+
+
+class SubmissionMetaTermsSerializer(serializers.Serializer):
+    meta_terms = serializers.PrimaryKeyRelatedField(many=True, queryset=MetaTerm.objects.all())
 
 
 class SubmissionDetailSerializer(serializers.ModelSerializer):
     questions = serializers.SerializerMethodField()
     meta_questions = serializers.SerializerMethodField()
+    meta_terms = MetaTermsDetailSerializer(many=True)
     stage = serializers.CharField(source='stage.name')
     actions = ActionSerializer(source='*')
     review = ReviewSummarySerializer(source='*')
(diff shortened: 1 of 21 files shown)


hypha/apply/api/v1/serializers.py
```diff
@@ -154,13 +155,37 @@ def get_round(self, obj):
...
+
+    def get_parent(self, obj):
+        parent = obj.get_parent()
+        if parent:
+            return parent.name
+
+    def get_parent_id(self, obj):
+        parent = obj.get_parent()
+        if parent:
+            return parent.id
```
v1k45: 💭 Would it make sense to have a `parent` object with `id ` and `name` instead of separate fields?

hypha/apply/api/v1/serializers.py
```diff
@@ -343,3 +368,18 @@ class Meta(CommentCreateSerializer.Meta):
 class UserSerializer(serializers.Serializer):
     id = serializers.CharField(read_only=True)
     email = serializers.CharField(read_only=True)
+
+
+class MetaTermsSerializer(serializers.ModelSerializer):
+    name = serializers.StringRelatedField()
```
v1k45: This can be removed. The `name` entry in `Meta.fields` include the field automatically.

hypha/apply/api/v1/urls.py
```diff
@@ -27,6 +28,7 @@
 router.register(r'comments', CommentViewSet, basename='comments')
 router.register(r'rounds', RoundViewSet, basename='rounds')
 router.register(r'screening_statuses', ScreeningStatusViewSet, basename='screenings')
+router.register(r'meta_terms', MetaTermsViewSet, basename="meta-terms")
```
v1k45: **Nitpick**: Inconsistent quotation marks.

```suggestion
router.register(r'meta_terms', MetaTermsViewSet, basename='meta-terms')
```

hypha/static_src/src/app/src/redux/selectors/submissions.js
```diff
@@ -75,6 +75,27 @@ const getSubmissionReminders = createSelector(getCurrentSubmission, submission =
     return submission && submission.reminders || []
 })
 
+const getSubmissionMetaTerms = createSelector(
+    [getCurrentSubmission], submission => {
+        let metaTerms = []
+        if(submission && "metaTerms" in submission){
```
v1k45, line 81: Will there be a case where a submission does not have `metaTerms` property?

From the SubmissionDetailSerializer class it looks like it will be always available.

hypha/static_src/src/app/src/redux/selectors/submissions.js
```diff
@@ -75,6 +75,27 @@ const getSubmissionReminders = createSelector(getCurrentSubmission, submission =
...
+            submission.metaTerms.map(metaTerm => {
+                const existingMetaTermIndex = metaTerms.findIndex(mt => mt.parentId == metaTerm.parentId);
+                if (existingMetaTermIndex != -1) {
+                    metaTerms[existingMetaTermIndex].children.push({'id': metaTerm.id, 'name': metaTerm.name})
```
v1k45: **Nitpick**:

```suggestion
                    metaTerms[existingMetaTermIndex].children.push({id: metaTerm.id, name: metaTerm.name})
```

hypha/static_src/src/app/src/redux/selectors/submissions.js
```diff
@@ -75,6 +75,27 @@ const getSubmissionReminders = createSelector(getCurrentSubmission, submission =
...
+                }
+                else {
+                    metaTerms.push({
+                        parentId : metaTerm.parentId,
+                        parent: metaTerm.parent,
+                        children: [{'id': metaTerm.id, 'name': metaTerm.name}]
```
v1k45: **Nitpick**:

```suggestion
                        children: [{id: metaTerm.id, name: metaTerm.name}]
```

hypha/static_src/src/app/src/redux/selectors/submissions.js
v1k45: 💭  We should consider using lodash in future to simplify array operations like grouping by a particular key. That could also be used for other cases to make the code more readable.

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/saga.js
```diff
@@ -0,0 +1,57 @@
...
+    const url = `/v1/submissions/${action.submissionId}/meta_terms/`
+    try{
+      yield put(Actions.showLoadingAction())
+      yield call(
+        apiFetch, 
+        {
+          path : url,
+          method : "POST",
+          options : {
+            body : JSON.stringify({"meta_terms" : action.data}),
```
v1k45: **Nitpick**:

```suggestion
            body : JSON.stringify({meta_terms : action.data}),
```

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/components/MetaTermTreeView/index.js
```diff
@@ -0,0 +1,129 @@
...
+import { withStyles } from '@material-ui/core/styles';
+
+
+class MetaTermTreeView extends React.PureComponent {
+
+    state = {
+        expanded : []
+    }
+
+    createTreeItemList = (node) => {
```
v1k45: **Nitpick**: What do you think about renaming this to `renderTreeItem`? `createTreeItemList` sounds like the function intends to create a resource.

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/components/MetaTermTreeView/index.js
```diff
@@ -0,0 +1,129 @@
...
+                key={node.id} 
+                classes= {{ 
+                    root: this.props.classes.treeItemRoot, 
+                    selected: this.props.classes.selected,
+                    iconContainer: this.props.classes.iconContainer,
+                  }}
+              >
+                {node.children.map(node => this.createTreeItemList(node))}
+              </TreeItem>
+        }
```
v1k45: **Nitpick**: Missing indentation inside `if` condition.

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/saga.js
```diff
@@ -0,0 +1,57 @@
...
+            }
+          }
+        )
+      let response = yield call(apiFetch, {path : `/v1/submissions/${action.submissionId}`});
```
v1k45:
```suggestion
      let response = yield call(apiFetch, {path : `/v1/submissions/${action.submissionId}/`});
```

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/saga.js
```diff
@@ -0,0 +1,57 @@
...
+      response = yield response.json()
+      const data = yield camelizeKeys(response)
+      yield put(updateSubmissionMetaTerms(action.submissionId, data.metaTerms))
```
v1k45, line 46: Wouldn't the submission already have metaTerms? Looking at `SubmissionDetailSerializer`.

v1k45 (commented):
Looking good overall, some suggestions and questions:

hypha/apply/api/v1/serializers.py
sandeepsajan0: Yes, I think so. I'll remove it.

hypha/apply/api/v1/serializers.py
sandeepsajan0: @v1k45  Yeah that would be the better way but It was the requirement as written here https://github.com/HyphaApp/hypha/issues/2241#issuecomment-842042919

hypha/static_src/src/app/src/redux/selectors/submissions.js
vimal1083: which lodash function would you recommend for this?

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/saga.js
vimal1083: I make sure when user updates meta term, redux store should be updated in sync with database. Sometimes, update to database(POST) may fail, which causes different data in the store & database. So, updated meta term after fetching data from detail API.

hypha/static_src/src/app/src/redux/selectors/submissions.js
vimal1083: Yeah, Initially submission list API will be called, only after the current submission(usually 1st submission in list) set, detail submission API will be called. So, it  is mandatory to check meta terms exist or not.

hypha/apply/api/v1/serializers.py
vimal1083: @sandeepsajan0 , Could you update the API as @v1k45 suggested, I will do the changes in Front end as well

hypha/static_src/src/app/src/redux/selectors/submissions.js
v1k45: We can use [`groupBy`](https://lodash.com/docs/4.17.15#groupBy) for grouping and [`mapValues`](https://lodash.com/docs/4.17.15#mapValues) for mapping the children.

Something like:

```js
_.mapValues(
    _.groupBy(items, 'parentId'),
    children => ({ children: _.map(children, o => _.pick(o, ['id', 'name'])) })
)
```

hypha/static_src/src/app/src/redux/selectors/submissions.js
v1k45: Got it. Thanks for the explanation! I think we should add a short comment explaining this case for future reference.

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/saga.js
v1k45: Got it. This looks good.

Correct me if I'm wrong, but I believe an alternative for this is to directly update the entire submission object in the store instead of just updating the metaTerms property. Is there a reason to not do that?

hypha/apply/api/v1/serializers.py
sandeepsajan0: Sure @vimal1083

hypha/static_src/src/app/src/containers/MetaTerm/containers/MetaTermForm/saga.js
vimal1083: Not required, If you see line no 39, I'm updating the meta term alone.

hypha/apply/api/v1/serializers.py
sandeepsajan0: Done @vimal1083

v1k45 (approved):
Looking great. Tested this and it is working as expected.

---------

One observation which is probably out of scope for this PR:

The button length on the sidebar seems a bit off. I expect "Update Status" to be same as "Meta Terms". Other buttons have the padding similar to Meta terms.

Or is that intentional?

image

vimal1083:
NIce catch, Will fix this now

frjo (changes requested):
Works very well in my testing. The meta term display looks better now in the React app than on the Django side where the parent term repeats.

The for widget for inputing meta terms is however a lot nicer on the Django side.

Possible to get an autocomplete solution like that on the React side as well?

meta_terms

frjo (approved):
This is better. We can iterate on this moving forward if needed.

PR #2418, opened 2021-06-09
Make labels, help_text etc. strings i18n aware

frjo:


frjo:
@Parbhat The Test error is from missing migration, will add them when the rest is in order.

The plan is to have all strings in Hypha translatable. How does this PR look so far?

Parbhat:
@frjo Good start for supporting i18n.

PR #2419, opened 2021-06-09
Add Timestamp for Upload Documents in Submission Details

### LINKED ISSUE #2344: Timestamp for Upload Documents
Please add a timestamp (a date and time0 next to each document uploaded to the application. As a manager, I would like to know when these documents were uploaded.

sandeepsajan0:
Fixes #2344 

As we know, adding the upload timestamp will help someone who will be looking/reviewing the application. So, we didn't change the filename but we added the file last modification/creation date just before the filename on the submission details page so that a Manager or a Reviewer can see when these files got uploaded.

We didn't change the filename because from the user perspective, we are not sure if adding a date to the filename will help but it might create some confusion because of the complex name. 

It is simply looking like this.
image

### DIFF
--- a/hypha/apply/funds/models/mixins.py
+++ b/hypha/apply/funds/models/mixins.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from django.core.files import File
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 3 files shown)


frjo (approved):
Works perfectly!

PR #2422, opened 2021-06-10
Add a CURRENCY_SYMBOL setting

### LINKED ISSUE #2420: Add a setting for currency symbol in base.py
This is to support funding organisations who do not use US Dollar as their funding currency.

frjo:
Fixes #2420


Parbhat (approved):
Looks good. Some changes use `strip` method `settings.CURRENCY_SYMBOL.strip()`, we can directly use the value from the env variable.

Parbhat:
Also, we can make it a simple setting  `CURRENCY_SYMBOL = '$'` instead of an environment variable. As it will not be often changed.

frjo:
@Parbhat I added the ".strip()" command to allow CURRENCY_SYMBOL to have spaces, e.g. "SEK ". That would in most places show up as "SEK 5000". But in some uses it would become "Requested Value (SEK )" without the strip command. Do not know if anyone is going to use space but easy to accommodate it.

I know most organisations implementing Hypha use environment settings and many will need to change the currency. They will only need to do it once when setting up the system but all other settings they need can be done as environment variables.

PR #2428, opened 2021-06-14
Make messages strings translateble.

frjo:


hypha/apply/activity/templates/messages/email/base.html
```diff
@@ -1,9 +1,10 @@
 {% block salutation %}Dear {{ user }},{% endblock %}
```
sandeepsajan0: I think we should add a `trans` tag for "Dear" as well?

frjo: Thanks, fix it now.

sandeepsajan0:
Looks good to me.

sandeepsajan0:
Just an idea/suggestion(it might not be related to this PR): 
1. What if we can pass the language code in urls?
It can help us to solve the "internal user's language issue" 
 >the ability for the “internal user” to choose what language Hypha is displayed in 

Because I think different users in the same org may want to use different languages.

2. I think it will be better to use Babel with `.pot` file to translate in other languages. Like it is [explained here](https://github.com/alexa/skill-sample-python-howto/blob/master/instructions/localization.md)?

frjo:
Django has built in support with "makemessages" (extract all translation in to a po-file) and "compilemessages" (compile all translations in to binary mo-file) commands. Maybe Django uses babel to do this?

Planing to add Djangos "LocaleMiddleware" as well so Django can pick the language the browser wants.

PR #2429, opened 2021-06-15
Fix Multiple Address field error

### LINKED ISSUE #2426: AddressField doesn't work properly when there are two of them on a single form page
Required for https://github.com/HyphaApp/hypha/issues/2379. 

In the [attached design](https://www.figma.com/file/6HzFAw7k9zXwuujvyf5uVO/OTF-System?node-id=192%3A2972) on the last step we need two AddressField. One for `Branch address`, and other for `Intermediary Bank Branch Address`.

I believe we have not used the [AddressField](https://github.com/HyphaApp/hypha/blob/6c10b4add7b074b854e8a08e92f1b80a6f4ddf1e/addressfield/fields.py#L33) twice on a single form before. So this issue has not come up yet.

I was talking to @vimal1083 and looks like this is happening because of some issue with [address_form.js](https://github.com/HyphaApp/hypha/blob/main/addressfield/static/address_form.js) file that we use in the [AddressWidget](https://github.com/HyphaApp/hypha/blob/6c10b4add7b074b854e8a08e92f1b80a6f4ddf1e/addressfield/widgets.py#L88).

To reproduce this we can just add AddressField twice on any of the existing forms or on a new form:

```python
first_address_field = AddressField()
second_address_field = AddressField()
```

When there is one we get these sub-fields:
Screenshot 2021-06-11 at 3 39 04 PM

Which is what we need as well when there are two of them. But adding the second one produces lot unnecessary/repetitive fields on both.

cc: @frjo

vimal1083:
Fixes #2426

### DIFF
--- a/addressfield/static/address_form.js
+++ b/addressfield/static/address_form.js
@@ -74,17 +74,19 @@
     };
 
     $(document).ready(function formReady() {
-        $('.form div.address').addressfield({
-            json: '/static/addressfield.min.json',
-            fields: {
-                country: ".country",
-                thoroughfare: ".thoroughfare",
-                premise: ".premise",
-                locality: ".locality",
-                localityname: ".localityname",
-                administrativearea: ".administrativearea",
-                postalcode: ".postalcode"
-            }
-        });
+        $('.form div.address').each(function() {
+            $('.form div#' + this.id).addressfield({
+                json: '/static/addressfield.min.json',
+                fields: {
+                    country: ".country",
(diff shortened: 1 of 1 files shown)


frjo:
@sks444 If you can confirm that this works I merge it in directly.

@vimal1083 Thanks for the clean and quick fix!

sks444 (approved):
Thanks @vimal1083! 

Works well.

PR #2437, opened 2021-06-18
fix: Make filters' row scrollable, to avoid overflow issue

### LINKED ISSUE #2432: In all submissions, when selecting 2 or more filters, the filter button is no longer clickable
have to manually zoom out of the page to actually click the button.
Screenshot 2021-06-15 09 15 17

Dont know what is the best expected behaviour :-/

cc @ei8fdb and @katiewilson21 for design review

sandeepsajan0:
Fixes #2432 

Peek 2021-06-18 17-31

The filter row was getting overflow, that's why I set `overflow: auto` to make it scrollable horizontally. So, it will be easily accessible and will not break any functionality.

### DIFF
--- a/hypha/static_src/src/sass/apply/components/_form.scss
+++ b/hypha/static_src/src/sass/apply/components/_form.scss
@@ -164,6 +164,7 @@
             padding: 10px 0 30px;
             opacity: 0;
             transition: opacity $transition;
+            overflow: auto;
 
             .filters-open & {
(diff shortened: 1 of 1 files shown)


frjo (commented):
@sandeepsajan0 Quick and working fix. But after talking with @ei8fdb we should, from a UX standpoint, solve this by rebuilding the filter a bit. Will post separate issue for this.

PR #2438, opened 2021-06-24
Add a finance user group, add finance dashboard, give finance access …

### LINKED ISSUE #2389: Create a new "Finance" user role with their own Hypha view
Set up "is_finance" similar to "is_apply_staff". They should only have access to project part of Hypha.

They need a dashboard that shows a table of invoices that have been approved by staff but not by finance.

I think we can reuse the table that is shown on the staff dashboard. We have some updates for that table in another issue.

The project page, the invoices and reports, uploaded files (PDF invoices etc.) and the project communication tab.

They will mostly use Hypha to talk with staff via the communication tab when invoices need corrections.

## UI Design Demo

- [ ] record new demo

## UI Designs

When the finance user logs in they see their dashboard - by default it shows the invoices that need to be reviewed.
4 1 - default view when they log in

If Yodit wants to see all the invocies, she clicks on "Show all invoices", which brings her to the next screen:

The view of "Show all invoices" - 
4 2 - view to show all invoices with all statuses
If Yodit needs to see projects, she clicks on "Projects".

Project statuses shown in this view are for projects that have been funded, are being funded, or will be funded.
4 3 - all projects view

If Yodit the finance user needs to see a specific project, she clicks on the project name, e.g. Maya Summit, which brings her to the next screen:
4 4 - project details view

We've added a "files tab" to the project view. This is shown for **the finance user only**. Here they see all the files that have been uploaded to Hypha that are related to the project. Adding a files tab is being handled by issue #2414.

## Necessary changes

**Dashboard**
- [x] dashboard that shows a table of invoices that have been approved by staff but not by finance.
- [x] When the finance user logs in they see their dashboard - by default it shows the invoices that need to be reviewed.
- [x] 'Submitted date' was when the invoice was submitted 
- [ ] 'Invoice Reference Number' is project name and unique invoice number
- [ ] 'Status'
- [ ] When and which PM/Lead approved the invoice
- [ ]  view of "Show all invoices"
- [ ] In 'Show All Invoices', invoices that have been approved by the finance user and the date invoice has been approved for payment.

**Search & Filters**
- [ ] Search box associated with invoice reference number
- [ ] Search project by name
- [ ] Available filters:
- invoices that have been approved (as a status)
- projects accepted for funding but has not yet started (as a status)
- contracting creation phase or contracting (as a status)
- Filter by Fund
- Filter by Lead

frjo:
…to projects, invoices and reports.

Fixes #2389


hypha/apply/projects/views/payment.py
```diff
@@ -25,19 +25,23 @@
 
 
 @method_decorator(login_required, name='dispatch')
-class PaymentRequestAccessMixin:
+class PaymentRequestAccessMixin(UserPassesTestMixin):
```
frjo, line 28: Rebuilt to use UserPassesTestMixin, looks ok?

hypha/apply/projects/views/report.py
```diff
@@ -1,7 +1,6 @@
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import UserPassesTestMixin
 from django.core.exceptions import PermissionDenied
-from django.http import Http404
```
frjo, line 4: Replaces 404 with PermissionDenied since that is what seems commonly used in Hypha. Any special reason a 404 was used here?

hypha/apply/projects/views/payment.py
Parbhat: Looks good. We can also combine the three if statements into a single if inside the test_func.

hypha/apply/users/decorators.py
```diff
@@ -27,6 +39,10 @@ def is_approver(user):
 
 staff_required = [login_required, user_passes_test(is_apply_staff)]
 
+finance_required = [login_required, user_passes_test(is_finance)]
+
+staff_or_finace_required = [login_required, user_passes_test(is_apply_staff_or_finance)]
```
Parbhat: Typo for finance in `staff_or_finace_required`

hypha/apply/projects/views/report.py
Parbhat: PermissionDenied seems better than 404 here as these pages are already behind the login.

hypha/apply/users/decorators.py
frjo: Thanks! Fixed now I believe.

sandeepsajan0:
Looks good, one thing I noticed. If there is no active PaymentRequest, we are showing just a blank page. I think it will be better if we can show something that helps user to understand that there is no Active Request. A simple message might works. What do you think @frjo?
Screenshot 2021-06-28 at 4 34 42 PM

sandeepsajan0 (approved):

frjo:
@sandeepsajan0 Good catch! Fixed that now.

frjo:
This is deployed to test now.

frjo:
We now need two level Finance approval so another finance admin user group must be created.

PR #2459, opened 2021-07-22
Fix no-results-found bug in react filters

### LINKED ISSUE #2423: In react view, screening status filter option "no status" results in No results found.
Same filter in the non-react view lists submissions correctly.

### LINKED ISSUE #2424: In react view, when a set of filters result in "No results found." in the left pane, the center pane loads the first every application received and the right pane has endless loading gifs
The ideal behaviour would be to load/show No results found. across all panes.

vimal1083:
Fixes #2423 Fixes #2424


frjo (approved):
Nice work!

PR #2466, opened 2021-08-02
Fix Applicant Activity Feed(Review)

### LINKED ISSUE #2461: Applicant Activity Feed
**Describe the bug**

Applicants can see when a reviewer has submitted a review. This should not happen.

**To Reproduce**

1. Submit a review as a staff or reviewer
2. As an applicant, go to the activity feed to see update

You shouldn't be able to see when a reviewer has submitted a review in the activity feed  Hard to scan data formatting  Quickly finding messages from different people is harder when all comments are the same colour even if from different pe

**Priority**

- Low priority 

**Affected roles**

- Applicant
- Staff
- Reviewers

sandeepsajan0:
Fixes #2461 

 Exclude/Hide the New Review activity for Applicant.

### DIFF
--- a/hypha/apply/activity/models.py
+++ b/hypha/apply/activity/models.py
@@ -41,6 +41,13 @@ VISIBILITY = {
 
 class BaseActivityQuerySet(models.QuerySet):
     def visible_to(self, user):
+        # to hide reviews from applicant of existing data
+        from .messaging import ActivityAdapter
+        messages = ActivityAdapter.messages
+        if user.is_applicant:
+            return self.exclude(
+                message=messages.get(MESSAGES.NEW_REVIEW)
+            ).filter(visibility__in=self.model.visibility_for(user))
         return self.filter(visibility__in=self.model.visibility_for(user))
 
     def newer(self, activity):
--- a/hypha/apply/activity/models.py
+++ b/hypha/apply/activity/models.py
@@ -47,7 +47,7 @@ class BaseActivityQuerySet(models.QuerySet):
         messages = ActivityAdapter.messages
         if user.is_applicant:
             return self.exclude(
-                message__in=[messages.get(MESSAGES.NEW_REVIEW), messages.get(MESSAGES.REVIEW_OPINION)]
+                message=messages.get(MESSAGES.NEW_REVIEW)
             ).filter(visibility__in=self.model.visibility_for(user))
 
         return self.filter(visibility__in=self.model.visibility_for(user))
--- a/hypha/apply/activity/models.py
+++ b/hypha/apply/activity/models.py
@@ -41,7 +41,7 @@ VISIBILITY = {
 
 class BaseActivityQuerySet(models.QuerySet):
     def visible_to(self, user):
-        # to hide reviews from applicant of existing data
+        # To hide reviews from the applicant's activity feed
         from .messaging import ActivityAdapter
         messages = ActivityAdapter.messages
         if user.is_applicant:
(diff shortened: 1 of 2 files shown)


frjo:
@sandeepsajan0 Is it not possible to reuse the "visibility" value that we already use for comments? Setting it to "reviewers" and making sure the view respect it should fix the issue.

sandeepsajan0:
@frjo Yes, It is possible but it will only work for new data. For existing data, we will need these changes.

In [this commit](https://github.com/HyphaApp/hypha/pull/2466/commits/1bf4c2c36d64cc8370dd70cafd10546108dcf05a), I made visibility related changes but I removed them later to have this in another PR. Take a look at that commit and if you think it is the right place to have both of the changes then I'll push it here again.

hypha/apply/activity/models.py
```diff
@@ -41,6 +41,13 @@
 
 class BaseActivityQuerySet(models.QuerySet):
     def visible_to(self, user):
+        # To hide reviews from the applicant's activity feed
```
frjo, line 44: Add a TODO about this is just for historic data and not needed for new data efter this.

frjo:
I think we can do both in this PR.

Looking at this it I realise it should effect both MESSAGES.NEW_REVIEW and MESSAGES.REVIEW_OPINION.

Can you add code that set these two actions to visibility "reviewers"? Thanks!

Later on I can create a SQL call that sets visibility to "reviewers" for all historic data as well, then we can remove the custom code.

hypha/apply/activity/models.py
```diff
@@ -41,6 +41,13 @@
...
+        from .messaging import ActivityAdapter
+        messages = ActivityAdapter.messages
+        if user.is_applicant:
+            return self.exclude(
+                message=messages.get(MESSAGES.NEW_REVIEW)
```
frjo, line 50: Please add MESSAGES.NEW_REVIEW and MESSAGES.REVIEW_OPINION.

REVIEW_OPINION is a type of lightweight review.

frjo (changes requested):

sandeepsajan0:
Sure, I'll make the changes.

In review, we have two types of visibility `private` and `reviewers`. If we will have only `reviewers` visibility in activities then it may leads to reviewer's activity feed issue. So, It would be better if we assign `Team` for `private` visiblity and `reviewers` for `reviewers`. What do you think?

frjo:
@sandeepsajan0 The code already adds "team" to MESSAGES.REVIEW_OPINION actions, sorry I did not notice this before. 

So the question id then if MESSAGES.NEW_REVIEW should be set to "team" or "reviewers". Asking @fourthletter this.

frjo:
@sandeepsajan0 Got a reply from @fourthletter, only staff should see MESSAGES.NEW_REVIEW so please set visibility to "team" (not "reviewers" as I wrongly stated above.)

You can then add MESSAGES.NEW_REVIEW to the list at https://github.com/HyphaApp/hypha/blob/45a5cb5eba94e95310f5251c7eb555e341c9e061/hypha/apply/activity/messaging.py#L259-L269

fourthletter:
Thank you @sandeepsajan0  for working on this and following up. 

@frjo Thank you for reaching out because I didn't fully scope out this issue. External Reviewers 1) do not need to see when a review has been submitted because reviews should be anonymous and 2) there is no value in having extraneous items in their activity feed, this info would only add more clutter.

frjo (approved):

frjo:
Deployed to test now.

PR #2470, opened 2021-08-09
archiving this page

ei8fdb:
archiving this page due to migration to docs.hypha.app


frjo:
I think the file should stay, it's a very common practice to include it. It can however be made very brief and then provide a link to the documentation.

ei8fdb:
> I think the file should stay, it's a very common practice to include it. It can however be made very brief and then provide a link to the documentation.

Ok I'll summarise it and cancel the pr.

PR #2486, opened 2021-08-16
Make the submission comment form more compact.

### LINKED ISSUE #1445: Improve submission comment visibility interface
Replace the radio options with checkboxes so the user can pick the role(s) that the comment should be available for. The "All" option can then be removed.

This will need work on the backend to update code that controls the visibility of comments.

comment_visibility_interface

frjo:
Part of #1445


hypha/static_src/src/sass/apply/components/_form.scss
```diff
@@ -335,6 +335,39 @@
...
+
+        .id_message {
+            grid-area: message;
+        }
+
+        .id_visibility_0 {
+            grid-area: visibility;
+        }
+
+        .button {
```
sandeepsajan0, line 365: @frjo  I was also looking at the same issue. These changes look better than mine. Just one thing, we can add `margin-top: 0;` in here, it will be more closer to input area as it is mentioned in issue's comment.

PR #2495, opened 2021-08-20
Swith to django-tinymce from django-tinymce4-lite

sks444:
Related https://github.com/HyphaApp/hypha-issues/issues/1

### DIFF
--- a/hypha/apply/review/fields.py
+++ b/hypha/apply/review/fields.py
@@ -1,7 +1,7 @@
 from django import forms
 from django.forms import widgets
 from django.utils.safestring import mark_safe
-from tinymce import TinyMCE
+from tinymce.widgets import TinyMCE
 
 from hypha.apply.review.options import NA, RATE_CHOICES
 from hypha.apply.utils.options import MCE_ATTRIBUTES_SHORT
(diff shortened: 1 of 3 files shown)


frjo:
@sks444 Far easier than I anticipated. Nice! The TineMCE 5 editor seems to just work, tested it on a couple of submissions, comparing the results.

The extra word count feature found in does not work anymore however. It is a brittle solution that looks up classes and IDs a d replaces/adds text. https://github.com/HyphaApp/hypha/blob/main/hypha/static_src/src/javascript/apply/tinymce-word-count.js. But we can fix that in a separate issue.

Something we could add to this PR is turning of unused TinyMCE plugins.

https://github.com/HyphaApp/hypha/blob/main/hypha/apply/utils/options.py#L3-L24

We do not have buttons for either codesample, code, image or preview. Can we just remove them form the list?

sks444:
Done @frjo

Yeah, it is easier to move as `django-tinymce4-lite` was a fork of `django-tinymce` long time back. It was because `django-tinymce` was not supporting tinymce4 at the time.

I just hope that they didn't change anything major internally. But I can't find anything breaking so we're good to go.

PR #2498, opened 2021-08-20
Use FILE_ALLOWED_EXTENSIONS to validate form upload documents

sandeepsajan0:
Fixes [hypha-issues/2](https://github.com/HyphaApp/hypha-issues/issues/2)

### DIFF
--- a/hypha/public/forms/models.py
+++ b/hypha/public/forms/models.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.core.files.storage import get_storage_class
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
-from django.forms import FileField
+from django.forms import FileField, FileInput
 from django.utils.decorators import method_decorator
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.cache import never_cache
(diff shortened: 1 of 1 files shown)


hypha/public/forms/models.py
```diff
@@ -42,7 +42,7 @@ class FormField(AbstractFormField):
 
 class ExtendedFormBuilder(FormBuilder):
     def create_document_field(self, field, options):
-        return FileField(**options)
+        return FileField(widget=FileInput(attrs={'accept': settings.FILE_ACCEPT_ATTR_VALUE}), **options)
```
frjo, line 45: I changed to FILE_ACCEPT_ATTR_VALUE so the formatting gets correct for the accept attribute.

frjo (changes requested):
This works nicely client side. We would need a check on the backend as well before saving the form.

sandeepsajan0:
We are already having these validations on file fields(on both Multi and Single FileField). These are checking the forms before saving. https://github.com/HyphaApp/hypha/blob/main/hypha/apply/stream_forms/fields.py#L51

@frjo Where else do you think we need it?

frjo:
The "Forms" are not using the uploads field from stream forms but a custom solution. So somewhere here maybe.

https://github.com/HyphaApp/hypha/blob/main/hypha/public/forms/models.py#L73-L93

frjo:
Rebased to get in some fixes so test run as they should.

sandeepsajan0:
@frjo I have tried to add validation in backend with custom clean method in form builder. Let me if it works good enough.

frjo (approved):

Erioldoesdesign:
Can this be merged?

PR #2521, opened 2021-08-31
Update message for empty rounds and labs

### LINKED ISSUE #1879: 'Closed' rounds copy on Dashboard
On the dashboard the 'closed' rounds and labs section reads 'There are no open rounds' when there aren't any closed rounds. It should read 'You have no closed rounds or labs assigned to you'. Likewise the Open tab should read 'You have no open rounds or labs assigned to you'. 

Screenshot 2020-03-24 at 10 47 22

This is what I currently see on the 'Test' environment.

sandeepsajan0:
Fixes #1879 

Just Update the message on the basis of type of round(Open/Closed)

### DIFF
--- a/hypha/apply/dashboard/templates/dashboard/dashboard.html
+++ b/hypha/apply/dashboard/templates/dashboard/dashboard.html
@@ -69,7 +69,7 @@
     {% endif %}
 
     {% if rounds.closed or rounds.open %}
-        {% include "funds/includes/round-block.html" with closed_rounds=rounds.closed open_rounds=rounds.open title="Your rounds and labs" %}
+        {% include "funds/includes/round-block.html" with closed_rounds=rounds.closed open_rounds=rounds.open title="Your rounds and labs" page_type='dashboard' %}
     {% endif %}
 
     {% if projects.table.data %}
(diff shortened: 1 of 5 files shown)


frjo (approved):

frjo (changes requested):
It turns out this text is used in two places.

1. The Dashboard, that needs this new text. (dashboard.html)
2. The submission overview that show all rounds, and should have the old existing text. (submissions_overview.html)

Can we create a template block for the empty text in "round-block-listing.html" and create a new "round-block-listing-dashboard.html" (or maybe a better name?) that extends "round-block-listing.html" and only change the empty text?

sandeepsajan0:
@frjo To call a new "round-block-listing-dashboard.html", we will need to create a copy of "round-block.html" as well. Because we have used include for "round-block-listing.html" in "round-block.html" file.
What if we pass something like `page_type='dashboard'` from the both dashboard.html and submission_overview.html and put message on the basis of that or we can also use an extended block with that to make it more structural?

frjo:
@sandeepsajan0 A page_type variable sounds good.

sandeepsajan0:
I think this message also needs to be updated in submissions. There is an issue with `Closed` and `Open` in message.
It is showing open round message even if I am looking into the closed round section.

image

frjo (approved):

frjo:
@sandeepsajan0 I think we can ignore that for now. But good catching it!

sandeepsajan0:
Okay @frjo,
Please ignore the latest spam review request, I was just testing if reassigning and re-requesting are different. It acts same btw.

PR #2547, opened 2021-09-13
Remove staff reviewers for a submission

### LINKED ISSUE #1575: No way to remove staff reviewer once assigned
**Describe the bug**
The edit reviewers modal box doesn't allow a reviewer to be removed from the list of reviewers. 

**To Reproduce**
1. Go to an application with one or two reviewers assigned
2. Click Reviewers. 
3. Dropdown only allows you to change reviewer to another reviewer, no option to set no one. 

**Expected behavior**
The ability to remove a reviewer from an assigned application. 

**Priority**

**Affected roles**
- Staff
- Reviewers

sandeepsajan0:
Fixes #1575 

Change widget and delete an Assigned Reviewer object if user wants to just unassign that reviewer.
Previously, Select2Widget were look like a checkbox but wasn't an actual checkbox, we were not able to unselect that. 

@frjo "Select a Reviewer" message fine to unassign the reviewer or Should I change it to "-------" or "None"?

### DIFF
--- a/hypha/apply/funds/forms.py
+++ b/hypha/apply/funds/forms.py
@@ -220,7 +220,7 @@ class UpdateReviewersForm(ApplicationSubmissionModelForm):
             self.role_fields[field_name] = data['role']
             self.fields[field_name].initial = assigned_roles.get(data['role'])
 
-        submitted_reviewers = User.objects.filter(
+        self.submitted_reviewers = User.objects.filter(
             id__in=self.instance.assigned.reviewed().values('reviewer'),
         )
 
@@ -230,7 +230,7 @@ class UpdateReviewersForm(ApplicationSubmissionModelForm):
             self.prepare_field(
                 'reviewer_reviewers',
                 initial=reviewers,
-                excluded=submitted_reviewers
+                excluded=self.submitted_reviewers
             )
 
             # Move the non-role reviewers field to the end of the field list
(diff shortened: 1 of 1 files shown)


frjo:
@sandeepsajan0 This works well in my testing, nice fix!

"Select a Reviewer" should be changed as you say. @ei8fdb Do you have a good suggestion what to use?

There is also one situation that we need to handle in some way.

If staff reviewers have been selected *and* submitted reviews and we then remove them without setting new ones the submitted reviews are deleted.

Maybe a check so you can only remove reviewers that have not submitted a review?

sandeepsajan0:
@frjo Thanks for pointing that out. I'll restrict it to reviewers who haven't submitted review.

sandeepsajan0:
@frjo I have filtered the Reviewers who have a review for the same role and submission.
Do we need to add a condition for form error as well? I tried to add form_errors in clean method in my local but wasn't working very well.

frjo (changes requested):
This works but I think we should display a form error messages about it not being possible to remove, just change, reviewers that have reviews.

As it stands the review removal fails silently if the reviewer has reviews.

frjo:
@sandeepsajan0 I committed the change that @ei8fdb suggested. And it all works very well now. Just the test that needs updating it seems. Can you take a look at that?

sandeepsajan0:
Sure @frjo . I'll fix it.

sandeepsajan0:
@frjo I don't know what was the issue. I tried it in my local, it works fine then I just rerun it here and it works. I think there might be some cache issues in CI env.

frjo (approved):

frjo:
Excellent, this is now on test for OTF to check out.

PR #2548, opened 2021-09-13
Make users and funds templates translatable.

frjo:


hypha/apply/users/templates/users/activation/email.txt
```diff
@@ -1,8 +1,10 @@
-{% load wagtailadmin_tags %}
+{% load i18n wagtailadmin_tags %}
 {% base_url_setting as base_url %}
-Dear {{ name|default:username }},
+{% name|default:username as user %}
```
frjo: @Parbhat This seems not to work, suggestions?

I want the translation string below to be the same as in other places to you only need to translate it once.

hypha/apply/users/templates/users/activation/email.txt
```diff
@@ -1,8 +1,10 @@
...
 
-An account on the {{ ORG_LONG_NAME }} web site has been created. Activate your account by clicking this link or copying and pasting it to your browser:
+{% blocktrans %}Dear {{ user }},{% endblocktrans %}
```
frjo, line 5: Want it to look like this and not `{% blocktrans %}Dear {{ name|default:username }},{% endblocktrans %}`.

hypha/apply/users/templates/users/activation/email.txt
frjo: @sks444 or @sandeepsajan0 Suggestion for my issue above?

sandeepsajan0: @frjo  I am checking , I think we need to use 'with' to make it available inside a block. Let me check more options.

sandeepsajan0: @frjo can you try it like this `{% firstof  name username as user %}`.  I think you wants to show `name` and if name is None then `username`, right?

frjo: Thank you @sandeepsajan0, implemented your fix now.

PR #2560, opened 2021-09-23
Make review options translatable.

### LINKED ISSUE #2559: Make review options translatable

ARDC wants to use their own terms for how reviewers rate applications. Current options include "Not so good," which they want to be "Weak Reject" and so on.

The options are defined [here](https://github.com/HyphaApp/hypha/blob/38f040dc0d3dc7920323c75c991632e9325e0f16/hypha/apply/review/options.py#L3-L11).

We'd like to submit a PR that makes these translatable. When merged, this will bring our branch closer to core.

I believe the correct way to do that is in the [lines I referenced](https://github.com/HyphaApp/hypha/blob/38f040dc0d3dc7920323c75c991632e9325e0f16/hypha/apply/review/options.py#L3-L11). But it's possible there's some other appropriate place I haven't considered.

**Priority**

**Affected roles**

- Reviewers

**Ideal deadline**

No deadline, though we plan to submit our patch soon.

chriszs:
Fixes #2559 

Makes the options displayed on the Create Review page translatable. Uses `gettext` instead of `gettext_lazy` because `VISIBILILTY_HELP_TEXT` and `VISIBILITY` are concatenated in [blocks.py](https://github.com/HyphaApp/hypha/blob/edd0879554be163a1943588e867e04eb20f8a61f/hypha/apply/review/blocks.py#L154) and that throws an error. If that causes an issue, I can try to find a workaround.

### DIFF
--- a/hypha/apply/review/options.py
+++ b/hypha/apply/review/options.py
@@ -1,13 +1,15 @@
+from django.utils.translation import gettext as _
+
 NA = 99
 
 RATE_CHOICES = (
-    (0, '0. Need more info'),
-    (1, '1. Poor'),
-    (2, '2. Not so good'),
-    (3, '3. Is o.k.'),
-    (4, '4. Good'),
-    (5, '5. Excellent'),
-    (NA, 'n/a - choose not to answer'),
+    (0, _('0. Need more info')),
+    (1, _('1. Poor')),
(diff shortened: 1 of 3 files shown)


frjo (approved):
Thanks for fixing this missing translation!

PR #2568, opened 2021-09-29
Sage IntAcct Integration

### LINKED ISSUE #1: [META] IntAcct API extension
This story is part of epic https://github.com/HyphaApp/hypha/issues/2352.

If possible keep all IntAcct API calls in a separate app that will only exist on the OTF install.

1. When Vendor is created/updated https://github.com/HyphaApp/hypha/issues/2379, do an API call and create/update the vendor in IntAcct. The IntAcct vendor_id will need to be saved in Hypha.
2. Via a (twice?) daily cron job update the vendor status for non active (pending) vendor. An active vendor has one or more contract attached in IntAcct.
3. When PM/staff view an invoice do an API call to IntAcct to retrieve the contract/deliverables and list them so the PM/staff can select one or more of them to be included in the invoice. There might be more then one contract so PM/Staff need to be able to pick the right one and see the correct deliverables.
4. When invoice is approved by Finance, do an API call and create it in IntAcct.
5. Via a (twice?) daily cron job update the invoice status (pending/approved/paid) for non paid invoices via API call to IntAcct.

sks444:
Fixes https://github.com/OpenTechFund/apply-app/issues/1


hypha/settings/base.py
```diff
@@ -762,3 +762,12 @@
...
 MATOMO_SITEID = env.get('MATOMO_SITEID', False)
+
+# SAGE INTACCT Configuration
+if 'INITIALISE_INTACCT' in env:
+    INITIALISE_INTACCT = env['INITIALISE_INTACCT']
+    INTACCT_SENDER_ID = env['INTACCT_SENDER_ID']
+    INTACCT_SENDER_PASSWORD = env['INTACCT_SENDER_PASSWORD']
+    INTACCT_USER_ID = env['INTACCT_USER_ID']
+    INTACCT_COMPANY_ID = env['INTACCT_COMPANY_ID']
+    INTACCT_USER_PASSWORD = env['INTACCT_USER_PASSWORD']
```
sks444: Need to set these variables to env for it to work.

frjo: I prefer if we set bool True/False like we do in other places in base.py:

https://github.com/HyphaApp/hypha/blob/a5d174ce64950a33745144fcdfa61f01488d3417/hypha/settings/base.py#L731-L733

I also think we should rename INITIALISE_INTACCT to INTACCT_ENABLED. To follow the naming of current vars.

frjo (changes requested):

hypha/settings/base.py
sks444: Thanks @frjo, that is correct, have pushed the changes.

frjo (approved):

sks444:
@frjo, I got this error on sentry while testing on test: `Some of the parameters are wrong`.

While all these errors from the payment service APIs should be handled in our code..we could display a friendly message..or log the errors somewhere..any error on the communication b/w hypha and the intacct api shouldn't break hypha system.

I will be working on this fix. In meantime can you recheck if all the variables are correct. Thanks.

frjo:
I have deployed the latest changes to test now.

sks444:
@frjo, can we deploy this on test? I have sent the env variable that I'm using locally on signal.

frjo:
Deploying to test now.

sks444:
@frjo I suppose we can merge this in now.

I removed all three lines which does the api call. We can add them whenever we need them back.

So now there is no connection b/w hypha and intacct and the code for intacct integration is lying there independently.

PR #2569, opened 2021-09-29
Add an extra choice to available outcomes(determination)

### LINKED ISSUE #2533: Remove default from determination forms
## User story

As a user I would send out the wrong determination type because the default is either 'dismiss', 'accept', or 'need more information' depending on the workflow status. 

## Describe the solution you'd like in Hypha

I would like to select the determination type  I need (dismiss, accept, need more info) before sending out the determination to minimize human error.

Screen Shot 2021-09-03 at 3 29 36 PM

**Priority**

**Affected roles**
- Applicants

**Ideal deadline**
December 2021

sandeepsajan0:
Fixes #2533 

Add description here.


hypha/apply/determinations/models.py
```diff
@@ -99,7 +99,7 @@ class Determination(DeterminationFormFieldsMixin, AccessFormData, models.Model):
         on_delete=models.PROTECT,
     )
 
-    outcome = models.IntegerField(verbose_name=_("Determination"), choices=DETERMINATION_CHOICES, default=1)
+    outcome = models.IntegerField(verbose_name=_("Determination"), choices=DETERMINATION_CHOICES)
```
sks444: Let's not do any db level changes?

Looks like the issue only requires UI/forms level changes.

_changing default would require us to have a discussion on why it was set in the first place..and what could be the implications of changing it now_

hypha/apply/determinations/forms.py
```diff
@@ -472,6 +475,12 @@ def clean(self):
 
         return cleaned_data
 
+    def clean_outcome(self):
+        if not self.cleaned_data['outcome']:
+            self.add_error('outcome', 'Outcome is a required field')
+            return self.cleaned_data
+        return int(self.cleaned_data['outcome'])
```
sks444: I guess this wouldn't be required if we do not remove the default from model field?

sks444:
Looks good @sandeepsajan0.

Let's also do the same [for api ](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/api/v1/determination/utils.py#L17) and make sure that react dashboard has the same behaviour as well.

sandeepsajan0:
@sks444  I checked with one of the submission in my local, and it is working fine in react dashboard as well. Create button is disabled until we will not select a valid choice.
image

sks444:
Thanks @sandeepsajan0!

frjo:
@sandeepsajan0 This works perfectly!

I made the strings translatable and made the empty choice the first in the select drop down. I also added an empty choice to the proposal form dropdown.

Can you take a look and see if I implemented it in a sensible way, thanks!

sandeepsajan0:
Thanks @frjo , Yeah It looks better now.

PR #2588, opened 2021-10-15
String changes in project after workflow walk through.

frjo:
Fixes #2587

PR #2590, opened 2021-10-15
Fix User issue in Project's admin detail template

### LINKED ISSUE #2589: Admin not able to access Project with status `can_make_approval`
Instead of `request.user`, we need to use  only `user` in [project_admin_detail.html](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/projects/templates/application_projects/project_admin_detail.html#L79) template.
Currently, It is showing this error.
image

sandeepsajan0:
Fixes #2589


frjo:
Excellent, just hit this bug in https://github.com/HyphaApp/hypha/pull/2588

PR #2591, opened 2021-10-15
Remove Applicant access from Application form button

### LINKED ISSUE #2584: Applicant should not have access in any way to Approval form
It can stay in the list to show the checkmark but applicant should not have any access to view or edit it.

Skärmavbild 2021-10-13 kl  12 32 48

sandeepsajan0:
Fixes #2584


frjo:
@sandeepsajan0 We also need to add access checks to the views.

hypha/apply/projects/templates/application_projects/includes/supporting_documents.html
```diff
@@ -53,7 +53,9 @@ <h5 class="docs-block__heading">{% trans "Supporting documents" %}</h5>
                 {% if editable %}
                     <a class="docs-block__link" href="{% url 'apply:projects:edit' pk=object.pk %}">
                         {% if object.user_has_updated_details %}
-                        {% trans "Edit" %}
+                            {% if not user.is_applicant %}
+                                {% trans "Edit" %}
+                            {% endif %}
```
sks444: Quoting from the issue description..

>It can stay in the list to show the checkmark but applicant should not have any access to view or edit it.

@frjo just to confirm...are the applicant allow to fill the approval form?

Btw it would be nice check what permissions user should have to do this instead of just relying on user not being the applicant.

@frjo what permission do we need to check for the user who can submit approval form/or what do we call those user? Approver? 

Will the checks happening for [user_can_approve_project](https://github.com/HyphaApp/hypha/blob/e97d7ea1d494a89cd4285961a166e34e2c3a9293/hypha/apply/projects/templatetags/approval_tools.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L18) be enough/correct for this?

sks444:
Also in case applicant are not allow to do any(Fill/Edit/View) the approval form..maybe we can just remove the applicant view from [ProjectEditView](https://github.com/HyphaApp/hypha/blob/e97d7ea1d494a89cd4285961a166e34e2c3a9293/hypha/apply/projects/views/project.py#L644); As looks like [ProjectEditForm](https://github.com/HyphaApp/hypha/blob/e97d7ea1d494a89cd4285961a166e34e2c3a9293/hypha/apply/projects/forms/project.py#L79) attached to [ApplicantProjectEditView](https://github.com/HyphaApp/hypha/blob/e97d7ea1d494a89cd4285961a166e34e2c3a9293/hypha/apply/projects/views/project.py#L626) is only for the Approval form.

hypha/apply/projects/templates/application_projects/includes/supporting_documents.html
sandeepsajan0: @sks444 I also used the same templatetag in [first commit](https://github.com/HyphaApp/hypha/pull/2591/commits/8d78c42f4c17b86c8306bbdb4b8bfda3958e6a4e) and restricted all three actions but then I read the issue description carefully, so I changed it. 
Let @frjo confirm this, I'll change it accordingly.

frjo: It is only staff that are allowed to fill in the project approval form (PAF). And to approve the form a user must be staff + approver role. Staff and Finance are both allowed to view the project approval.

Applicants should not have any access, view/edit/delete or otherwise, to the project approval form.

sandeepsajan0: @frjo I have changed the condition. Can you please check it again, It should work fine now.

sks444 (approved):
Looks good @frjo!

So the approval form "can" be filled in by any user who is staff and not just the lead? But in use case they will mostly be filled in the by the lead.

PR #2593, opened 2021-10-18
Add contracting info action button.

frjo:
Adds an action button to fill in contracting information (vendor setup).

### DIFF
--- a/hypha/apply/projects/templates/application_projects/project_detail.html
+++ b/hypha/apply/projects/templates/application_projects/project_detail.html
@@ -1,6 +1,6 @@
 {% extends "base-apply.html" %}
 
-{% load i18n contract_tools static wagtailcore_tags %}
+{% load i18n contract_tools static wagtailcore_tags approval_tools %}
 
 {% block title %}{{ object.title }}{% endblock %}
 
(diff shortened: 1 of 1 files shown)


hypha/apply/projects/templates/application_projects/project_detail.html
```diff
@@ -153,6 +154,10 @@ <h5>{% trans "Actions to take" %}</h5>
                     {% block admin_actions %}{% endblock %}
                     {% endif %}
 
+                    {% if editable and not project.vendor.user_has_updated_details %}
```
sks444: Looks like the editable variable defined at the top is not carrying here. 

We can define it inside `{% block content %}` or maybe just define it before this line as it is not used any place else on the page.

Let's also check if the [vendor setup is required](https://github.com/HyphaApp/hypha/blob/e97d7ea1d494a89cd4285961a166e34e2c3a9293/hypha/apply/projects/templates/application_projects/includes/supporting_documents.html?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L22).

frjo: Thanks!

sks444 (approved):
Looks good!

PR #2616, opened 2021-10-30
Bypass django forms widget templates

sandeepsajan0:
It is a fix for a [Reset Issue](https://github.com/ResetNetwork/apply-app/issues/129#issuecomment-954761104)

Problem:  Wagtail is using django forms template. And we were overwriting django forms templates.

Solution: Bypass wagtail date and datetime templates directly to django input widget template.

@frjo Please check if it looks good to you. Now we are using wagtail default behaviour for date and datetime templates but there is no placeholder. Let me know if we need placeholder.


frjo (approved):
Perfect.

PR #2620, opened 2021-11-16
Send out vendor setup mail only after lead has been set on a project.

### LINKED ISSUE #2594: Send out contracting info setup mail first when lead has been set.
Not when project is created.

frjo:
Fixes #2594


sks444 (approved):
Looks good @frjo!

PR #2638, opened 2021-11-22
Only allow API token access for round API endpoint.

frjo:


sandeepsajan0:
@frjo Is it okay to have apply API endpoints without APIKeys? I am not sure but if any backend is using the apply APIs with API Keys might get affected.
Btw, I am working to have separate API keys for apply and public APIs. Its implementation is almost done, just trying to test it. So, if this PR changes affects apply APIs in security perspective then we may wait a little. I'll create a PR for the above implementation today.

sandeepsajan0:
@frjo I have created a PR #2641  for separate API Keys for Apply and Public APIs. You may have a look and try it in test env.

frjo:
@sandeepsajan0 Thanks for commenting here and sorry I was late to reply.

The API tokens has not been used for anything since they where implemented. The API has only been used by the React parts of Hypha so far.

Now that we for the first time is going to hand out a token I reconsider the security implications.

PR #2639, opened 2021-11-22
Fix project status view

sks444:
Fixes the filter when doing status view of projects on the projects overview page.

Screenshot 2021-11-23 at 3 46 56 AM


frjo (approved):
Nice!

PR #2641, opened 2021-11-23
Add separate API keys for Apply and Public API endpoints

### LINKED ISSUE #2607: Allow API tokens to be active for only apply or only public API endpoints
Or take it even further and allow API tokens to be active for any combinations of API end points.

sandeepsajan0:
Fixes #2607 

It will restrict the user to access apply site data by public site api key or vice-versa.


frjo:
Closing this since we opt for the solution in https://github.com/HyphaApp/hypha/pull/2638 instead.

This PR works very well but for security reasons we will only use the tokens for public API calls moving forward.

PR #2653, opened 2021-11-26
Require password on email update

### LINKED ISSUE #2643: Require password on account e-mail address change

As a user, when I am logged into my account on a device that I trust and I would like to change my email associated with my account. I will be required to enter my current correct account password in order to confirm the change of email process.

Makes it harder for bad people to take over the account by e.g. session hijack.

Not existing functionality but existing section in the /account area

A solution that includes the logged in users current correct account password on request of email change would also include error pathways for users including:
- If a user enters their password incorrectly
- A user enters an invalid email as a new email (could also include an email that does not have accepted organisational affiliation in the email address)
- If a user spends to long on the page session and 'times out'
- Compatible with any existing 2-factor processes on that account and instance

- Alternative to password entry on email change might be using any current 2-factor process.

**Additional context**

This was part of the recommendations given by the security audit team circa Aug 2021.

**Priority**

- Low priority (annoying, would be nice to see)

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners

**Ideal deadline**

End of Q1 2022

sandeepsajan0:
Fixes #2643 

@sks444 and I discussed about it, we finalise with three points:
 - Redirect user on a page to put password on email change.
 - Session time out of that page when taking too long.(appx 1 min)
 - Send mail to new email for verification after enter correct password.

1st point is implemented, working on the second and third one.


hypha/apply/users/views.py
```diff
@@ -39,6 +43,21 @@ class AccountView(UpdateView):
...
+        slack = form.cleaned_data['slack']
+        user = get_object_or_404(User, id=self.request.user.id)
+        if updated_email and updated_email != user.email:
+            base_url = reverse('users:confirm_password')
+            query_string = urlencode({
+                'updated_email': updated_email,
+                'name': name,
+                'slack': slack
+            })
+            return redirect('{}?{}'.format(base_url, query_string))
```
sks444: @sandeepsajan0, let's not put these variables in url, that would be introducing another security issue while fixing one..rather this could be hidden fields in the password form if possible.

**OR** we could use [Django's cryptographic signing](https://docs.djangoproject.com/en/3.2/topics/signing/), this also allows us to fix the second point in our list..session time out one...e.g. by using [TimestampSigner](https://docs.djangoproject.com/en/3.2/topics/signing/#verifying-timestamped-values) we could know how much time user took to fill up the password and raise error accordingly.

Also, we need to make sure this PasswordForm(related url/view) is only getting accessed on redirect from this view. User should not be able to access it in any other way. We could maybe make the signing parameter in the url mandatory.

sandeepsajan0:
> 2. Session time out of that page when taking too long.(appx 1 min)

2nd point is also done, working on email confirmation part.

hypha/apply/users/utils.py
```diff
@@ -50,3 +51,36 @@ def send_activation_email(user, site=None):
...
+
+    activation_path = reverse('users:confirm_email', kwargs={'uidb64': uid, 'token': token})
+
+    context = {
+        'user': user,
+        'name': user.get_full_name(),
+        'username': user.get_username(),
+        'unverified_email': updated_email,
+        'activation_path': activation_path,
+        'timeout_days': settings.EMAIL_CHANGE_CONFIRMATION_HOURS,
```
frjo: Maybe use PASSWORD_RESET_TIMEOUT_DAYS, it is a Django setting that we already use in other places.

sks444:
- [x] On the accounts page if user changes email they will be redirected on the password page to put password. Other details like `full name` and `slack name` can be updated without password.
     - to validate the redirection is only accessed by the same user we are using signed user email in the request session.
     - also the updated user details are send as signed query variable on the redirection url for further steps.

- [x] Validation of password on password page:
     - Incorrect passwords are not accepted.
     - Also if the user takes too long to put the password. The request will be timed out: User will be redirected back to accounts page and are asked to change email again. This time out is set to `120` seconds by default, which can be configured by setting a different value for env variable: `PASSWORD_PAGE_TIMEOUT_SECONDS`.

- [x] Once user puts the correct password, an email will be sent to their new updated email with a confirmation link
    - Although we have email format validations, this will make sure that user entered their own email.
    - The new email will be successfully updated, once user clicks on the confirmation link.
    - Confirmation link is valid for 3 days by default, which is also configurable by env variable `PASSWORD_RESET_TIMEOUT_DAYS`

sks444 (approved):
@frjo, this is good to go on test.

PR #2669, opened 2021-12-09
Add Rate limit sensitive forms

sandeepsajan0:
Initial Fix [hypha-issues#8
](https://github.com/HyphaApp/hypha-issues/issues/8)

### DIFF
--- a/hypha/apply/users/views.py
+++ b/hypha/apply/users/views.py
@@ -22,8 +22,8 @@ from django.views.generic import UpdateView
 from django.views.generic.base import TemplateView
 from django.views.generic.edit import FormView
 from django_otp import devices_for_user
+from django_ratelimit.decorators import ratelimit
 from hijack.views import AcquireUserView
-from ratelimit.decorators import ratelimit
 from two_factor.forms import AuthenticationTokenForm, BackupTokenForm
 from two_factor.utils import default_device, get_otpauth_url, totp_digits
 from two_factor.views import BackupTokensView as TwoFactorBackupTokensView
@@ -49,8 +49,8 @@ from .utils import send_confirmation_email
 User = get_user_model()
 
 
-@method_decorator(ratelimit(key='ip', rate=settings.DEFAULT_RATE_LIMIT, method='POST', block=True), name='dispatch')
-@method_decorator(ratelimit(key='post:email', rate=settings.DEFAULT_RATE_LIMIT, method='POST', block=True), name='dispatch')
+@method_decorator(ratelimit(key='ip', rate=settings.DEFAULT_RATE_LIMIT, method='POST'), name='dispatch')
+@method_decorator(ratelimit(key='post:email', rate=settings.DEFAULT_RATE_LIMIT, method='POST'), name='dispatch')
 class LoginView(TwoFactorLoginView):
     form_list = (
         ('auth', CustomAuthenticationForm),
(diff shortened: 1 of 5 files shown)


sandeepsajan0:
@sks444 @frjo For now, I put the limit on `/account/` and `/login/` urls with 5 request/min(just to test). If you are testing this in your local, make sure you are using DatabaseCache instead of DummyCache(default for local).

Now, We need to discuss on a few of things.
1. Where to redirect the users when they exceed the limit.(Currently, it is just raising 403).
2. We need to decide a list of the urls with their rate limit, to restrict.

Feel free to add more points and provide your suggestions.

sks444:
@frjo, This adds rate-limit to, login, password change, password-reset, 2FA back tokens, 2FA disable and newsletter subscription.

Currently it is set to use same ratelimit rate at all the places: Which can be configured by changing the value of env variable `DEFAULT_RATE_LIMIT`, which is set to `5 requests/minute` by default. Formatting of this value should be: `3/s`, `3/m`, `3/h` which indicates number of requests per second, per minute and per hour respectively.

There is also the possibility to add ratelimits with multiple rate at the same time, `10/s`, `100/m` in case that is needed for a few specific forms.

```python
@ratelimit(key='user_or_ip', rate='10/s')
@ratelimit(key='user_or_ip', rate='100/m')
``` 

Wagtail forms doesn't have any specific urls/views so this is not used yet on apply forms. Will check more on how we can utilize this package for wagtail forms later. I remember [adding a Recaptcha
](https://learnwagtail.com/tutorials/adding-recaptcha-to-your-contact-forms/) in one of the wagtail projects for the same purpose, maybe we could explore that option as well.

sks444 (approved):
Looks good. Thanks @sandeepsajan0!

One more point of discussion is that it raises 403 on exceeding the rate limit. Best would be to raise the error on the form with a message but that will require editing each of the views, I guess. 

That said, these configuration is only to avoid submission from bots, and rate should be set in  such a way that human users shouldn't exceed the limits, so I guess we are fine here.

@frjo can we also deploy this to test? Thanks!

sandeepsajan0:
>That said, these configuration is only to avoid submission from bots, and rate should be set in such a way that human users shouldn't exceed the limits, so I guess we are fine here.

@sks444 That's a good point. But I think still for the safer side, it would be better to have a custom page or view  to show the message like "Your Request limit has been exceed for the {url} and ..." and handle it like this https://django-ratelimit.readthedocs.io/en/stable/usage.html#exceptions

What do you think?

frjo:
I have updated this for django-ratelimit v4.

Why did we ratelimit the 2FA views that needs login?

Could we add ratelimit for the application form creation (anon users)

sandeepsajan0:
>Why did we ratelimit the 2FA views that needs login?

I think we were trying to rate limit the destructive actions that require password changes, and few of 2FA views require password changes. I have updated the code to ratelimit only those kinds of views.

>Could we add ratelimit for the application form creation (anon users)

I am looking for this. Actually, these are wagtail served pages and we don't have an explicit view/url for them, so it is hard to ratelimit them directly. But let me check our options.

hypha/apply/users/views.py
```diff
@@ -317,6 +316,7 @@ def get_context_data(self, form, **kwargs):
 
 
 @method_decorator(login_required, name='dispatch')
+@method_decorator(ratelimit(key='user', rate='5/m', method='POST'), name='dispatch')
```
frjo: Why not use the "settings.DEFAULT_RATE_LIMIT" here?

PR #2676, opened 2021-12-16
Fix Newsletter open redirect

sandeepsajan0:
Fixes [hypha-issues#5](https://github.com/HyphaApp/hypha-issues/issues/5)


frjo:
This form is part of the public side of Hypha that is to be removed moving forward.

PR #2678, opened 2022-01-05
Update to Django 3.2.12

### LINKED ISSUE #2523: Upgrade to Django 3.2 LTS
Hypha now uses Django 2.2 LTS. Support for that release ends in April 2022. So we should upgrade before that, first quarter 2022 is a good target.

https://www.djangoproject.com/download/#supported-versions

Django 3.2 LTS is supported until April 2024

When updating to Django 3.2 we can also update Wagtail to 2.15+ (2.14 dropped support for Django 2.2).

sandeepsajan0:
Fixes #2523


sandeepsajan0:
Django Form Fields are no more Orderdict in Django3.0 because of python versions.  As it is explained here https://groups.google.com/g/django-developers/c/zL_eMRHanE4. So, we can't use `move_to_end` method [here](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/forms.py#L237) and many other places. 

We got two options, either we change back the fields from dict to Orderdict(in forms init method) or we need to define the `field_order` so that we can move a field to end of dict. 

Which option seem better to you all? 

In my opinion, we may use Orderdict because python dict already introduced `reversed` method and I think it may have `move_to_end` soon?

sandeepsajan0:
Update: Form fields dict has converted back to OrderedDict at a few places. It seems working fine. Some tests were giving errors due to the query crashes with Django3.0 that has been fixed in 3.0.2 and further. So, I upgrade it to 3.0.14. 
Now only [one test](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/tests/test_forms.py#L32) is failing, its executed query number is lesser than expected.

sandeepsajan0:
**Do not merge this PR for now**
I am working on its testing part and I found something interesting, I'll share it soon. Until then don't merge this PR.

requirements.txt
```diff
@@ -28,7 +28,7 @@ django-tables2==2.4.1
 django-tinymce==3.4.0
 django-two-factor-auth==1.13.1
 django-webpack-loader==0.7.0
-django==2.2.26
+django==3.0.14
```
frjo: The goal is Django 3.2.x (LTS). I assume you are taking it step by step to ease the upgrade.

frjo:
Happy that this is on the way! This will allow us to update to latest Wagtail as well..

requirements.txt
sandeepsajan0: Yes, exactly. Its end goal is Django 3.2.

sandeepsajan0:
This PR is now ready to review and merge.

There was an issue in initializing a Form with an updated Django version. 

**Problem**:  [Model Form's fields](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/forms.py#L201) were not getting initialized for fields as empty list[]. Because of uninitialised form, we were short of 4 queries in [ReviewerFormQueries Test](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/tests/test_forms.py#L32). 
**Reason**: They particularly changed this behavior of forms's model_to_dict method with [this commit](https://github.com/django/django/commit/1ffddfc233e2d5139cc6ec31a4ec6ef70b10f87f).
**Solution**: 
1. We change the number of queries in our test. - I discussed this with @sks444 and we decided not to change our tests.
2. We use a custom duplicate model_to_dict method to initialize the form with the same data and same number of queries.

For now, I have implemented the second method.  Please feel free to share your suggestions.

frjo (approved):
Nice first step!

frjo:
We will not deploy this to main branch until we are all the way to 3.2 since other versions between 2.2 and 3.2 are not supported.

sandeepsajan0:
@frjo We already have draft PRs for 3.1 and 3.2 which worked well for me. Let me push the other changes here in a single place(PR).

sandeepsajan0:
I have pushed the changes for 3.2.  But there is still a warning to set the default auto field for some models. Let me fix that first.

sandeepsajan0:
@frjo This PR is ready to review now. I have updated it to 3.2.11 and also resolved all warnings.

frjo:
@sandeepsajan0 I needed to revert your merge. We have updated to node 16 that use a new format for the `package-lock.json` file.

frjo (approved):
Nice work!

PR #2701, opened 2022-02-01
Use environs package to parse  .env variables.

### LINKED ISSUE #2471: Implement "environs" package for nice env parsing of settings
https://github.com/sloria/environs

I like the look of this package. Looks like we could use it to make the base.py settings a bit cleaner and easier to read. Would also allow Hypha on a standard server to pick up .env files automatically.

```
if env.get('EMAIL_USE_TLS', 'false').lower().strip() == 'true':
    EMAIL_USE_TLS = True

```

Could be rewritten as:

`EMAIL_USE_TLS = env.bool("EMAIL_USE_TLS", False)`

as an example.

sandeepsajan0:
Fixes #2471 

This is tested locally but not sure about the production settings file yet.  @frjo Can you please test it on test env first, if it works fine there.
I went through django-environs as well, which is widely used but the environs author already took the [reference](https://github.com/sloria/environs#why-another-library) of django-environs and make it more easy and understandable.


frjo:
Will test it!

Really liking the cleaner look of the settings file, makes it a lot easier to read it.

frjo (approved):

PR #2713, opened 2022-02-09
Parse locale for currency lookup with dash separator

### LINKED ISSUE #2712: Babel currency lookup throws error when region is specified
**Describe the bug**

We have the locale in Hypha set to `en-us`. This caused Babel's locale parsing used in `get_currency_name` calls added in #2693 to throw errors that said it expected only characters.

**To Reproduce**

1. Change `LANGUAGE_CODE` to `en-us`.
2. Open the Hypha apply dashboard in debug mode.
3. You should observe an error message.

**Expected behavior**

This should parse regional language locales, handling more or less how the rest of Hypha/Wagtail does.

**Priority**

- Medium (slowing you down)

**Screenshots**

Screen Shot 2022-02-09 at 2 25 36 PM

Screen Shot 2022-02-09 at 2 25 36 PM

chriszs:
Changes locale parsing for currency lookup to recognize dash separator, consistent with other parts of Hypha.

Fixes #2712

### DIFF
--- a/hypha/apply/projects/forms/vendor.py
+++ b/hypha/apply/projects/forms/vendor.py
@@ -1,7 +1,7 @@
 import datetime
 from operator import itemgetter
 
-from babel.core import get_global, Locale
+from babel.core import Locale, get_global
 from babel.numbers import get_currency_name, get_territory_currencies
 from django import forms
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


sks444 (approved):
Thanks @chriszs!

PR #2715, opened 2022-02-09
Correctly validate Slack field as optional

### LINKED ISSUE #2714: Editing user accounts for non-staff appears to cause error validating non-existent Slack field
**Describe the bug**

When non-staff users try to update their name on the account page, they get the key error seen in the screenshot below. It's coming from [attempting to validate](https://github.com/HyphaApp/hypha/blob/6bc7b8d6cb829cc81c71313a1135f456256907d2/hypha/apply/users/views.py#L70) the Slack field that seems to be removed for all but staff.

**To Reproduce**

1. Go to [account](http://apply.hypha.test:8090/account/)
2. Change your name in the full name field
3. Click "Update profile"

**Expected behavior**

The name to update without an error.

**Priority**

**Affected roles**

- Applicants
- Reviewers
- Partners

**Screenshots**

Screen Shot 2022-02-09 at 6 45 10 PM

PR incoming.

chriszs:
Hypha hides the Slack field for users with non-staff roles. This reflects that by checking for it using an optional `get` (which will default to `None`) instead of trying to access the key directly.

Fixes #2714

### DIFF
--- a/hypha/apply/users/views.py
+++ b/hypha/apply/users/views.py
@@ -67,7 +67,7 @@ class AccountView(UpdateView):
     def form_valid(self, form):
         updated_email = form.cleaned_data['email']
         name = form.cleaned_data['full_name']
-        slack = form.cleaned_data['slack']
+        slack = form.cleaned_data.get('slack')
         user = get_object_or_404(User, id=self.request.user.id)
         if updated_email and updated_email != user.email:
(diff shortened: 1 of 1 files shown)


frjo:
Thanks for this! I have seen this error from time to time but never got so far as to actually fix it.

PR #2726, opened 2022-02-15
Hide the Dashboard "admin" button for users without wagtail admin access

### LINKED ISSUE #2724: "Admin" button appears even if a user does not have wagtail admin rights
**Describe the bug**

When logged in as a user who has access to the admin dashboard but without wagtail admin access the "Admin" button appears even though the user is forbidden to view those wagtail admin pages.  The button should only render for users who are allowed to view the admin pages.

**To Reproduce**

1. Create a user with the "Staff" role (not admin)
2. As a superuser, edit the "Staff" group role so that staff does not have the right to access wagtail admin
3. Navigate to `http://hypha.test:9001/dashboard/`
4. The "admin" button is there

**Expected behavior**

The admin button should ONLY render for administrators / superusers

**Priority**

- Medium : this is confusing for non-technical end users and the button is pretty prominent.

**Affected roles**

- Staff
- Reviewers
- Partners

**Screenshots**

image

 - OS: MacOS
 - Browser Firefox
 - Version 97.0

**Additional context**

I have a fix and will open a PR

slifty:
Fixes #2724

This PR removes the "Admin" button on the dashboard for users who don't actually have access to the administrative interface.

Writing tests for this involved needing to create new factories to generate users with groups that had explicitly modified permissions; if there is a better way to accomplish this that doesn't involve two new factories please let me know.

It also makes a very small fix to a broken test factory (necessary in order to write tests for the functionality of this PR)

### DIFF
--- a/hypha/apply/users/tests/factories.py
+++ b/hypha/apply/users/tests/factories.py
@@ -1,6 +1,7 @@
 import uuid
 
 import factory
+from django.contrib.auth.models import Permission
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group
 from django.utils.text import slugify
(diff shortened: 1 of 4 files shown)


hypha/apply/users/tests/factories.py
```diff
@@ -67,6 +68,32 @@ def groups(self, create, extracted, **kwargs):
             self.groups.add(GroupFactory(name=STAFF_GROUP_NAME))
 
 
+def get_wagtail_admin_access_permission():
```
slifty, line 71: Ideally this would be a fully generic `get_permission_from_perm_string` similar to what is described in [this StackOverflow question](https://stackoverflow.com/a/40892015/159522) but I wasn't sure where to put that kind of code.

frjo:
@slifty I have invited you to the Hypha dev team here on GitHub so you get write access to the repo. 

Could you resubmit this PR directly here in the Hypha repo? Then all the test runs automatically and it easier to deploy to test environments etc.

slifty:
For sure!

Closing this -- will re-open with the new branch.

PR #2728, opened 2022-02-16
Hide admin button for users without admin access

### LINKED ISSUE #2724: "Admin" button appears even if a user does not have wagtail admin rights

slifty:
Fixes #2724

This PR removes the "Admin" button on the dashboard for users who don't actually have access to the administrative interface.

Writing tests for this involved needing to create new factories to generate users with groups that had explicitly modified permissions; if there is a better way to accomplish this that doesn't involve two new factories please let me know.

It also makes a very small fix to a broken test factory (necessary in order to write tests for the functionality of this PR)

(Note: this was originally opened as #2726 but has been moved to this repository instead of a fork to enable CI)

### DIFF
--- a/hypha/apply/dashboard/templates/dashboard/dashboard.html
+++ b/hypha/apply/dashboard/templates/dashboard/dashboard.html
@@ -14,10 +14,12 @@
         {% block page_header %}
             <h1 class="gamma heading heading--no-margin heading--bold">{% trans "Dashboard" %}</h1>
         {% endblock %}
-        <a href="{% url 'wagtailadmin_home' %}" class="button button--primary button--arrow-pixels-white">
-            {% trans "Apply admin" %}
-            <svg><use xlink:href="#arrow-head-pixels--solid"></use></svg>
-        </a>
+        {% if perms.wagtailadmin.access_admin %}
+            <a href="{% url 'wagtailadmin_home' %}" id="wagtail-admin-button" class="button button--primary button--arrow-pixels-white">
+                {% trans "Apply admin" %}
+                <svg><use xlink:href="#arrow-head-pixels--solid"></use></svg>
+            </a>
+        {% endif %}
     </div>
 </div>
 <div class="wrapper wrapper--large wrapper--inner-space-medium">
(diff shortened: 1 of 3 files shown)


slifty:
Apologies for the linting errors; I was running the wrong commands locally

hypha/apply/dashboard/templates/dashboard/dashboard.html
```diff
@@ -14,10 +14,12 @@
         {% block page_header %}
             <h1 class="gamma heading heading--no-margin heading--bold">{% trans "Dashboard" %}</h1>
         {% endblock %}
-        <a href="{% url 'wagtailadmin_home' %}" class="button button--primary button--arrow-pixels-white">
-            {% trans "Apply admin" %}
-            <svg><use xlink:href="#arrow-head-pixels--solid"></use></svg>
-        </a>
+        {% if has_wagtail_admin_access %}
```
frjo: Can we use `{% if perms.wagtailadmin.access_admin %}` directly? Any downsides to that?

frjo (changes requested):

hypha/apply/dashboard/templates/dashboard/dashboard.html
slifty: Fixed up!  I think that's clear enough what it means (the direct-access name is *very slightly* less clear about what we're checking IMO, but not enough to warrant the extra field)

frjo (approved):

PR #2730, opened 2022-02-16
Fix review sidebar item widths

### LINKED ISSUE #2729: Review grid can have overlapping text
**Describe the bug**

Under some circumstances text from the columns in the "Review Sidebar" overlap.

**To Reproduce**

1. Create a submission
2. Create a review with the "Maybe" recommendation
3. View the submission

**Expected behavior**

Score and recommendation text should not overlap

**Priority**

**Affected roles**

- Staff
- Reviewers
- Partners

**Screenshots**

image

 - OS: MacOS 12.1
 - Browser Firefox
 - Version 97.0

**Additional context**

I'll be opening a PR with a patch

slifty:
Fixes #2729

This PR updates the widths of the "review items" sidebar so that the recommendation and score fields won't overlap.

**Before**
image

**After**
image

### DIFF
--- a/hypha/static_src/src/app/src/components/ReviewBlock/index.js
+++ b/hypha/static_src/src/app/src/components/ReviewBlock/index.js
@@ -113,7 +113,7 @@ const ReviewBlock = ({children, recommendation, score}) => {
                     {!isNaN(parseFloat(score)) &&
                         <div>{parseFloat(score).toFixed(1)}</div>
                     }
-                    <div>Actions</div>
+                    <div></div>
                 </li>
             }
             {children}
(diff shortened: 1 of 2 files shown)


frjo:
In the React app there are 4 columns and it's controlled by the same CSS so we need the grid to set 4 columns.

`grid-template-columns: 40% 20% 15% 25%;`

Works but if we replace "Actions" below with something short like "Op." (short for "Operations") we can do

`grid-template-columns: 45% 25% 15% 15%;`

maybe

slifty:
One thought for consideration: removing the "actions" header entirely (since the edit icon might be self explanatory) might do a lot to save space -- though if the actions list can be more than just the edit icon we should reserve space for it.

I spent some time trying to find a more dynamic solution -- a grid layout that would change depending on the number of columns, e.g. something like:

```
45% 35% repeat(2, min-content)
```

but couldn't find a good solution that would create consistent widths between rows.

`45 20 10 25` works.  It's a bummer to lose 25% of the grid width even in cases where the user doesn't have access to the "actions" column, but solves the original issue!

PR #2735, opened 2022-02-22
Make duration field optional and set required None for existing data

### LINKED ISSUE #2274: Stop making "project duration" field required by Wagtail system

sandeepsajan0:
Fixes #2274 and Closes #2337 

Add a custom migration to set required False for existing data, otherwise, it was saved None by default for existing data.


frjo:
This PR has migrations that conflicts with https://github.com/HyphaApp/hypha/pull/2678 so we wait a bit with this one.

sandeepsajan0:
@frjo I have merged the migrations. Looks like #2678 is stable now, so we may proceed with this as well?

frjo:
Excellent, I have deployed it to test now.

frjo:
Deployed failed, I have rebuilt the migration now.

frjo:
Missed the custom migration you built, readded now.

hypha/apply/funds/migrations/0093_set_duration_false.py
```diff
@@ -0,0 +1,24 @@
...
+
+from django.db import migrations
+
+
+def set_duration_required_false(apps, schema_editor):
+    ApplicationSubmission = apps.get_model('funds', 'ApplicationSubmission')
+    for submission in ApplicationSubmission.objects.all():
+        for id, struct_child in enumerate(submission.form_fields):
+            struct_value = struct_child.value
+            if struct_value['field_label'] == 'Duration' and struct_value['required'] is None:
```
frjo: I believe this is looking for fields that has the label set to "Duration". On the test site I get this migration error.

`TypeError: 'RichText' object is not subscriptable`

I assume there exist other kinds of fields on some applications that is named "Duration" but is to a duration field.

Can we make the migration only look for explicit Duration fields?

frjo (changes requested):

hypha/apply/funds/migrations/0093_set_duration_false.py
sandeepsajan0: @frjo I have changed that lookup from `struct_value['field_label']` to `struct_child.block_type`. I think it should work fine now. Can we test it once again?

frjo:
Perfect! Deploying to test now.

frjo (approved):

PR #2768, opened 2022-03-08
Add same validation for max value of requested amount for submission and project

### LINKED ISSUE #2743: Validation of Requested amount on submissions and value on Projects are not the same creating issue
Entering a large value on a submission will result in an error when creating a project from that submission.

```
psycopg2.errors.NumericValueOutOfRange: numeric field overflow
DETAIL:  A field with precision 10, scale 2 must round to an absolute value less than 10^8.
```

https://github.com/HyphaApp/hypha/blob/7418f05fa0a5bea4504beaac761b33988eea622a/hypha/apply/projects/models/project.py#L138-L143

https://github.com/HyphaApp/hypha/blob/7418f05fa0a5bea4504beaac761b33988eea622a/hypha/apply/funds/blocks.py#L36-L45

sandeepsajan0:
Fixes #2743 

Increased the max_digit validation to 20 and put the same for both application submission and to the project.


frjo (changes requested):
This works but needs some adjustments.

1. Please remove the possibility to add decimals in both submissions and projects.
2. The max calculation for the submission amount value is a bit too clever. The max value gets set to "1e+18". Normal users do not get that.

`<input type="number" name="cfae89dc-f327-45f4-80e9-f267c3bd1ec7" value="123456789123456789.00" min="0" max="1e+18" id="id_cfae89dc-f327-45f4-80e9-f267c3bd1ec7">`

Skärmavbild 2022-03-11 kl  09 29 13

sandeepsajan0:
> Please remove the possibility to add decimals in both submissions and projects.

That means we don't need DecimalField in [project model](https://github.com/HyphaApp/hypha/pull/2768/files#diff-6bd530ee29808bcc5507eec23cd8dea9ccadddc9b7283b3b3cbfd17d039d6021R138) anymore? If yes, then we don't need to set any max value or digits anywhere because we can use simply PositiveIntegerField without any limitations.

frjo:
@sandeepsajan0 Yes, remove the ability to add decimal values and use PositiveIntegerField.

frjo (approved):

frjo:
Conflicting migration with https://github.com/HyphaApp/hypha/pull/2568 so we will need to update migration on one of them before merging in to main.

sandeepsajan0:
@frjo #2568 is on hold for now. So, I think we can move forward with this one.

PR #2769, opened 2022-03-08
Fix-2767: Add a query to check review__author if review is in draft

### LINKED ISSUE #2767: Function "in_review_for()" shows a submissions if any of its review is in draft
In some instances staff and reviewers see submissions they already have reviewed in the "Waiting for your review" table in the Dashboard. The function that builds the query is `in_review_for()`.

I checked by removing the ` | Q(reviews__is_draft=True)` part.

This seems to check if any review for the submission is in draft mode. It should check if the users own review is in draft mode instead.

https://github.com/HyphaApp/hypha/blob/b7b70cdf3b3becb2912b2a6c644090de3f7e3d10/hypha/apply/funds/models/submissions.py#L122-L130

The `in_community_review()` has the same issue so fix that one as well.

sandeepsajan0:
Fixes #2767 

Add description here.


frjo (approved):
Works perfectly! This will make staff happy.

PR #2772, opened 2022-03-09
Populate "E-mail" and "Full name" on submissions and disable input for the logged in users

### LINKED ISSUE #2770: For logged in users, populate "E-mail" and "Full name" on new submissions and disable input
Related #2758

We should also disable editing "E-mail" and "Full name" on any existing submissions.

Now users can edit the fields but they get reverted on save to whatever the values are in the user account. This is a bit confusing. Better to disable editing and direct users to edit these values in their user account.

sandeepsajan0:
Fixes #2770 

Add description here.

### DIFF
--- a/hypha/apply/funds/models/applications.py
+++ b/hypha/apply/funds/models/applications.py
@@ -433,6 +433,14 @@ class LabBase(EmailForm, WorkflowStreamForm, SubmittableStreamForm):  # type: ig
     def open_round(self):
         return self.live
 
+    def get_form(self, *args, **kwargs):
+        user = kwargs.get('user')
+        form_class = self.get_form_class(user=user)
+        form_params = self.get_form_parameters()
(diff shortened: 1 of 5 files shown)


frjo:
@sandeepsajan0 Works well in my testing for funds/rounds.

The changes done in `class RoundBase` is needed for `class LabBase` as well to make it works for labs.

There is also one test that is failing since we now pro populate user and mail so they are never empty.

frjo:
@sandeepsajan0 Works well in my testing for funds/rounds.

The changes done in `class RoundBase` is needed for `class LabBase` as well to make it works for labs.

There is also one test that is failing since we now pro populate user and mail so they are never empty.

frjo (changes requested):
Se comment above.

frjo (approved):

PR #2782, opened 2022-03-22
Update required fields error message

### LINKED ISSUE #2771: [Meta] Revise error messages
Users receive error messages within Hypha, but the reason for the error is not always clear. We could try our best to reproduce the errors, but it would be helpful to decide next steps for redrafting error messages.

### User did not include required fields in application form

Screen Shot 2022-02-24 at 8 30 04 PM

### User sent a batch action determination message

Screenshot 2021-11-10 9 23 45 AM (1)

### User sent one determination message

Screen Shot 2022-03-09 at 9 26 21 AM

### LINKED ISSUE #2774: Improv handling of required fields on stream field form creation in Wagtail admin
These include Application forms, Review forms and Determination forms.

1. The best solution would be to prepopulate all required field when e.g a new application form is created. See #2275 #2460 

2. The second best solution is to improve the error messages given when a user try to save a form without a required field. See #2771.

Number 2 should be in place regardless in case a user by mistake removes a required field that has been prepopulated.

sandeepsajan0:
Fixes #2774 , partial fix #2771


frjo:
@sandeepsajan0 Updated the cache version key so now the test run again.

frjo:
@sandeepsajan0 Please try to use a rebase workflow instead of merging.

sandeepsajan0:
okay, yeah it adds an extra commit. I'll use rebase from now on.

@frjo This is ready for review. You may also have a look at the tests and let me know if I am missing something there.

frjo (approved):
Excellent work!

PR #2785, opened 2022-03-23
Fix change email error

### LINKED ISSUE #2784: Changing email on non-staff accounts creates non-null constraint violation
**Describe the bug**

When a non-staff user attempts to edit their email account they ultimately reach a non-null constraint error related to the `slack` field.

**To Reproduce**

1. Log in as a non-staff user
2. Click on profile link (upper right)
3. Note that the "Slack" field is not in the form (this should be the case if it is not a staff user)
4. Change the email address
5. Enter the password confirmation
6. See error

**Expected behavior**

The error should not appear.

**Priority**

- Medium-High (this seems like a significant issue as it is broken functionality to all applicants, reviewers, and partners)

**Affected roles**

- Applicants
- Reviewers
- Partners

**Screenshots**

image

 - OS: MacOS 12.2.1
 - Browser Firefox
 - Version 98.0.1

**Additional context**

@chriszs pointed out this is related to https://github.com/HyphaApp/hypha/issues/2714

The problem comes from the fact that the slack field does not exist in the password edit form for this type of user (it isn't immediately obvious to me where the logic that determines fields in a form lives, since the form view is doing some fancy footwork in rendering fields).

I should note: it isn't clear to me yet why the EmailChangePasswordForm references slack *OR* full_name in the first place, or why `self.user.save` exists in this context since the email is the item being saved and is also the only item that isn't explicitly saved in this logic.  Isn't save already handled in the `ProfileForm`?

One solution would be to only set the slack field if a value is passed.  Another is to remove reference to slack (and full_name) entirely from `EmailChangePasswordForm`.

I will open a PR.

slifty:
Fixes #2784

This PR fixes an issue where non-staff users could not change their email address.

### DIFF
--- a/hypha/apply/users/forms.py
+++ b/hypha/apply/users/forms.py
@@ -105,7 +105,8 @@ class EmailChangePasswordForm(forms.Form):
 
     def save(self, updated_email, name, slack, commit=True):
         self.user.full_name = name
-        self.user.slack = slack
+        if slack is not None:
+            self.user.slack = slack
         if commit:
             self.user.save()
         return self.user
--- a/hypha/apply/users/tests/test_forms.py
+++ b/hypha/apply/users/tests/test_forms.py
@@ -1,7 +1,7 @@
 from django.forms.models import model_to_dict
 from django.test import TestCase
 
-from ..forms import ProfileForm
+from ..forms import EmailChangePasswordForm, ProfileForm
 from .factories import StaffFactory, UserFactory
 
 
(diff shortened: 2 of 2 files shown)


slifty:
One of these days I'll remember to run *all* of the linting steps before opening a PR!

Sorry about that -- fixing the import sort now.

hypha/apply/users/forms.py
```diff
@@ -105,7 +105,8 @@ def clean_password(self):
 
     def save(self, updated_email, name, slack, commit=True):
         self.user.full_name = name
-        self.user.slack = slack
+        if (slack is not None):
```
frjo: Parentis not needed I believe.

slifty:
Based on conversation in the original issue it is possible that the preferred solution will be to remove the not null constraint from the field in question.

I'll keep this PR open in the mean time just in case we feel it is still be valuable / that we still don't want to modify the slack field here in the event that no value is passed regardless of constraint (plus of course there's the fancy *unit test!*).

hypha/apply/users/forms.py
slifty: My javascript / eslint habits creeping in ;)

frjo:
Tested and this fixes the error.

slifty:
I added one other test to get the diff coverage to 100% / it makes sense to ensure that I didn't break the ability to update slack when we want to!

All done with the changes / ready for re-review whenever.

frjo (approved):

PR #2799, opened 2022-04-09
Add filters for role and status, customize templates and add bulk action feature for users

### LINKED ISSUE #2793: Improve filters for user roles in WagTail Admin
## Description

As a sys admin I would like to quickly sort, search, and manage user accounts in the WagTail Admin. 

1) I am able to sort by Name, Email, Status, but not by **Role**

2) Sorting by Name, Email, Status is cumbersome and does not easily filter or hone in on the user accounts I am searching for.

3) Some user accounts are not assigned to a role, making it difficult to understand their permissions (level of access to information) on the platform. 

image

**Priority**

**Affected roles**

- Staff

**Ideal deadline**
October 2022

sandeepsajan0:
Fixes #2793 

Add description here.

### DIFF
--- a/hypha/apply/users/admin_views.py
+++ b/hypha/apply/users/admin_views.py
@@ -41,7 +41,7 @@ class UserFilterSet(WagtailFilterSet):
         return queryset
 
     def filter_by_status(self, queryset, name, value):
-        if value=='active':
+        if value == 'active':
             return queryset.filter(is_active=True)
         elif value == 'inactive':
             return queryset.filter(is_active=False)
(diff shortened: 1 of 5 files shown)


frjo:
@sandeepsajan0 The "THERE ARE 20 USERS" counter show how many users are listed on the current page, not the total number of users.

In all other respects it works perfectly for me.

sandeepsajan0:
Thanks for pointing it out @frjo. I have only 15 users in my local that's why I didn't notice it in my local. Let me fix that.

frjo:
Works perfectly. Just rebased of main as well.

PR #2800, opened 2022-04-11
Fix Group field's bugs

### LINKED ISSUE #2798: Group field is buggy
Group field is currently a **required** section. This requires the applicant to select an option even when the section is not applicable to their proposed scope of work. I propose this section does not need to be marked as **required**

Screen Shot 2022-03-16 at 5 52 49 PM

**Heading Text field** and **Section Text field** when nested within Group Fields do not render or published on the website.

Screen Shot 2022-03-16 at 6 02 56 PM

sandeepsajan0:
Partial fix #2798 

Add description here.


sandeepsajan0:
@frjo This PR contains only a small change that set the required field optional(because if a required field is required so the user doesn't have any option to leave it blank). 
Looks like it still solves one of the mentioned problems in issue #2798.

PR #2804, opened 2022-04-12
Make it possible to enforce 2FA for all users: add env variable and middleware

### LINKED ISSUE #62: Enforce 2FA for all user groups
**Describe Feature Request**

Currently, 2FA is only enforced for the staff user group (via Google login). This request is for it to apply to all user groups and block users who do not have it unabled.

**Requested solution**

- Block access to Hypha if user does not have 2FA enabled.
- Reset password for that user
- User resets password and sets up 2FA in order to access
- Add a setting to turn this on/off. Default will be off so it works just like now. OTF will have it set to on.

**Priority**

- High 

**Affected roles**
- Applicants
- Reviewers
- Partners
- Developers
- General public

**Ideal deadline**
April 18th

sandeepsajan0:
Fixes [#62](https://github.com/OpenTechFund/apply-app/issues/62)

Add description here.


hypha/apply/users/middleware.py
```diff
@@ -13,3 +15,31 @@ def get_message(self, request, exception):
...
+        for url in allowed_urls:
+            if url in path:
+                return True
+        return False
+
+    def __call__(self, request):
+        # code to execute before the view
+        user = request.user
+        if settings.ENFORCE_TWO_FACTOR and user.is_authenticated and not user.is_verified() and not self.is_path_allowed(request.path):
+            return redirect('/account/two_factor/required/')
```
sks444: Let's generate the redirection url based on the name of the url using reverse.

hypha/apply/users/middleware.py
```diff
@@ -13,3 +15,31 @@ def get_message(self, request, exception):
...
+    def is_path_allowed(self, path):
+        allowed_urls = [
+            'login/',
+            'logout/',
+            'account/',
+            'two_factor/',
+        ]
+        for url in allowed_urls:
+            if url in path:
+                return True
```
sks444: Can we directly do `if path in allowed_urls`?

sandeepsajan0: We can but for that we will need to specify all the urls in [allowed_urls](https://github.com/HyphaApp/hypha/pull/2804/files#diff-1ea73a7302b3be891ad19134ba6aab169bb73c38a05c7c5c3967e9fd87a865a4R25). Like we have many urls that contains '/account/' and '/two_factor/' so instead of adding them all, I checked for these sub path in the request path.

hypha/apply/users/middleware.py
sandeepsajan0: Actually, I tried that first but it was giving some errors.

I tried it like this `redirect(reverse('users:two_factor_required'))`( I think it is the correct way, we have used it at many places). But it is showing error `django.urls.exceptions.NoReverseMatch: 'users' is not a registered namespace`. Let me check with it.

sandeepsajan0: I played around it, Looks like we can use name of url only after the [get_response called](https://github.com/HyphaApp/hypha/pull/2804/files#diff-1ea73a7302b3be891ad19134ba6aab169bb73c38a05c7c5c3967e9fd87a865a4R42) that means after view called in middleware. Now sure why it is like that.

sks444:
We are waiting on this for today. We have been discussing avoiding hardcoded urls(both at `redirect` and at `allowed_urls`). If that is not possible then the possibility of writing a few tests cases for this.

We can deploy it on test tomorrow in any case.

frjo (approved):
Works well for me. Just need to know how it works for users the login via Google.

frjo:
The 2FA workflow could use some UI love as well but that is a separate issue.

sandeepsajan0:
@frjo @sks444 I have added the tests as discussed [here](https://github.com/HyphaApp/hypha/pull/2804#issuecomment-1096626449). I have also add a custom template with same ui but without "Go back" button. 
I also removed 'two_factor/' from allowed paths because I noticed all two_factor urls are coming from 'account/'.

Can you both test it again, please?

hypha/apply/users/middleware.py
```diff
@@ -13,3 +21,26 @@ def get_message(self, request, exception):
...
+
+        for sub_path in ALLOWED_SUBPATH_FOR_UNVERIFIED_USERS:
+            if sub_path in path:
+                return True
+        return False
+
+    def __call__(self, request):
+        # code to execute before the view
+        user = request.user
+        if settings.ENFORCE_TWO_FACTOR and user.is_authenticated and not user.is_verified() and not self.is_path_allowed(request.path):
```
sks444: Let's divide the checks into three parts?

```python
if settings.ENFORCE_TWO_FACTOR:
    if user.is_authenticated and not user.is_verified():
        if not self.is_path_allowed(request.path):
```

hypha/apply/users/middleware.py
```diff
@@ -13,3 +21,26 @@ def get_message(self, request, exception):
             return 'Your credentials are not recognised.'
 
         super().get_message(request, exception)
+
+
+class TwoFactorAuthenticationMiddleware:
```
sks444, line 26: Let's add a doc string here:

```
Middleware to enforce 2FA activation for unverified users

To activate this middleware set env variable ENFORCE_TWO_FACTOR as True.

This will redirect all request from unverified users to enable 2FA first. Except the request made on the url paths listed in ALLOWED_SUBPATH_FOR_UNVERIFIED_USERS.
```

sks444 (approved):
Looks good and works well! Excellent work on the tests @sandeepsajan0 👌

Left a few minor nit-pick comments if you want to address them as well.

sandeepsajan0:
@sks444 Docstrings and other changes are done. I don't know why tests are failing on CI while working fine in local.

frjo:
As can been seen in the Video in Slack from @fourthletter it does not work for staff users logging in via Google Oauth2.

They get to the screen instructing them to set up 2FA. Clicking on the button get them in to a redirect loop.

@sandeepsajan0 @sks444 I think the solution is to add code to the middleware to skip users that logged in via Google. Does that sound correct?

The user object has a "social_auth" attribute. Maybe possible to check against that? For normal users it is empty but for staff users logging in via Google it contains a lot of info.

`<QuerySet [{'id': 1, 'user_id': 1, 'provider': 'google-oauth2', 'uid': 'test@example.org', 'extra_data': {'auth_time': 1573377534, 'expires': 3599, 'token_type': 'Bearer', 'access_token': 'd41d8cd98f00b204e9800998ecf8427e'}, 'created': datetime.datetime(2020, 9, 16, 10, 49, 49, 408611, tzinfo=<UTC>), 'modified': datetime.datetime(2020, 9, 16, 10, 49, 49, 427481, tzinfo=<UTC>)}]>
`

sks444:
>@sandeepsajan0 @sks444 I think the solution is to add code to the middleware to skip users that logged in via Google. Does that sound correct?


@frjo, yes, that is correct.

>The user object has a "social_auth" attribute. Maybe possible to check against that? For normal users it is empty but for staff users logging in via Google it contains a lot of info.

Ok, so we can do `user.social_auth.exists()` and if it is True that means user has social auth and we skip the user. This will be False for the user who doesn't have any social auth. cc: @sandeepsajan0

sandeepsajan0:
@frjo @sks444 Yes that makes sense. I updated the code and skip the user with the social auth. @frjo you may deploy it there again just to test it again.

PR #2807, opened 2022-04-13
Only list active users when querying for all reviewers, all staff etc.

frjo:


hypha/apply/funds/models/utils.py
```diff
@@ -26,11 +26,11 @@
...
-LIMIT_TO_STAFF = {'groups__name': STAFF_GROUP_NAME}
-LIMIT_TO_REVIEWERS = {'groups__name': REVIEWER_GROUP_NAME}
-LIMIT_TO_PARTNERS = {'groups__name': PARTNER_GROUP_NAME}
-LIMIT_TO_COMMUNITY_REVIEWERS = {'groups__name': COMMUNITY_REVIEWER_GROUP_NAME}
-LIMIT_TO_REVIEWER_GROUPS = {'groups__name__in': REVIEW_GROUPS}
+LIMIT_TO_STAFF = {'groups__name': STAFF_GROUP_NAME, 'is_active': True}
+LIMIT_TO_REVIEWERS = {'groups__name': REVIEWER_GROUP_NAME, 'is_active': True}
+LIMIT_TO_PARTNERS = {'groups__name': PARTNER_GROUP_NAME, 'is_active': True}
+LIMIT_TO_COMMUNITY_REVIEWERS = {'groups__name': COMMUNITY_REVIEWER_GROUP_NAME, 'is_active': True}
+LIMIT_TO_REVIEWER_GROUPS = {'groups__name__in': REVIEW_GROUPS, 'is_active': True}
```
sks444, line 33: These are being used in model fields..so would require a migration file.

sks444 (approved):
Looks good.

hypha/apply/funds/models/utils.py
frjo: Thanks, will fix that.

PR #2820, opened 2022-04-25
Remove/Hide File field option from Application forms.

### LINKED ISSUE #2102: Merge the two file upload field type in to one.
There are now two stream fields for uploading files, "File field" and "Multiple file field".

We should remove/hide "File field" from being added to new forms. Instead add a setting on "Multiple file field" to set how many fields are allowed. 1-10 or unlimited are my suggestion for options.

Or is it never needed to limit the number of uploads? If that is the case the new setting is not needed.

Lets use the icon now used for "File field" on "Multiple file field" as well when this is worked on.

sandeepsajan0:
Fixes #2102 

This PR contains two major changes:
- Remove/Hide File Field option from the Application Forms in wagtail admin.
   * Create a separate js(form-field-block.js) file similar to wagtail's streamBlock.js
   * Add a telepath adapter to FormFieldBlock to communicate with `form-field-block.js`
- Use the FileField Block's icon for MultiFileField Block.

### DIFF
--- a/hypha/apply/stream_forms/blocks.py
+++ b/hypha/apply/stream_forms/blocks.py
@@ -415,6 +415,10 @@ class FileFieldBlock(UploadableMediaBlock):
     """This doesn't know how to save the uploaded files
 
     You must implement this if you want to reuse it.
+
+    This block option is disabled/hide from the application forms,
+    now it is only working for the existing ones that are already using it.
+    Override the behaviour of FormFieldsBlock(StreamBlock) with 'js/apply/form-field-block.js' to hide its option.
     """
     field_class = SingleFileField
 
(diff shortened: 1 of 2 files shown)


sks444:
@sandeepsajan0, can you put up a comment [here](https://github.com/HyphaApp/hypha/blob/67e216fe0627ef6f62e8ec59adcfae8dd6b4e424/hypha/apply/stream_forms/blocks.py#L414) that Single File Field Block is disabled for the new application forms and only works for the existing ones that are already using it. Also a bit on technical implementation that is done to make it happen.

sandeepsajan0:
@sks444  I just added a comment [there](https://github.com/HyphaApp/hypha/pull/2820/commits/4b9c232dc6208f9e0867df5b8e4e3aeaeeb54499#diff-4091f2ffd7526c0186bbbe9752361ef645e3cb2eb4720c84459cd58df70e157aR418) as you suggested. Please have a look and let me know if it is not enough informative.

sandeepsajan0:
@sks444 @frjo As we discussed, this PR changes have scalability/maintainability issues. So, I looked for the our other options. And looks like max_no restriction can do our work. I tried this restriction in PR #2825, please test that out and let me know if it seems good to you both.

sandeepsajan0:
Closing it as we have #2825 that is working better and more scalable than this PR.

PR #2825, opened 2022-04-26
Disable the file field block by setting max number restriction

### LINKED ISSUE #2102: Merge the two file upload field type in to one.

sandeepsajan0:
Fixes #2102 

With the custom max_no restrictions, it will look like 

image

### DIFF
--- a/hypha/apply/funds/admin_views.py
+++ b/hypha/apply/funds/admin_views.py
@@ -108,6 +108,12 @@ class CreateApplicationFormView(CreateView):
 
 class EditApplicationFormView(EditView):
 
+    def get_form(self):
+        form = super(EditApplicationFormView, self).get_form()
+        single_file_count = sum(1 for block in self.get_instance().form_fields.raw_data if block['type'] == 'file')
+        form.fields['form_fields'].block.meta.block_counts = {'file': {'min_num': 0, 'max_num': single_file_count}}
+        return form
+
     def form_invalid(self, form):
         show_admin_form_error_messages(self.request, form)
         return self.render_to_response(self.get_context_data(form=form))
--- a/hypha/apply/funds/admin_views.py
+++ b/hypha/apply/funds/admin_views.py
@@ -101,6 +101,11 @@ class CopyApplicationFormViewClass(CreateView):
 
 class CreateApplicationFormView(CreateView):
 
+    def get_form(self):
+        form = super(CreateApplicationFormView, self).get_form()
+        form.fields['form_fields'].block.meta.block_counts = {'file': {'min_num': 0, 'max_num': 0}}
+        return form
+
     def form_invalid(self, form):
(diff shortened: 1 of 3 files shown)


sks444:
This looks good but it only works when file field block is select at least once even for creating new application forms.

sandeepsajan0:
@sks444 @frjo I think I found a fix for caching issue as well. It caches the form inside edit handler but it still calls the Create class get_form method. So, we can change the form's meta (max number for files) there. Let me test it and push it.

sandeepsajan0:
@sks444 Can you please test it again, it seems working fine for me now.

frjo:
Deploying to test site now.

hypha/apply/funds/blocks.py
```diff
@@ -194,6 +194,9 @@ class ApplicationCustomFormFieldsBlock(CustomFormFieldsBlock, FormFieldsBlock):
     required_blocks = ApplicationMustIncludeFieldBlock.__subclasses__()
     single_blocks = ApplicationSingleIncludeFieldBlock.__subclasses__()
 
+    class Meta:
+        block_counts = {}
+
```
sks444: Is this still required?

hypha/apply/funds/admin_views.py
```diff
@@ -101,13 +101,24 @@ def get_initial(self):
...
+        return form
+
     def form_invalid(self, form):
         show_admin_form_error_messages(self.request, form)
         return self.render_to_response(self.get_context_data(form=form))
 
 
 class EditApplicationFormView(EditView):
 
+    def get_form(self):
```
sks444, line 120: Add a comment on why this change is required and what it does.

hypha/apply/funds/admin_views.py
```diff
@@ -101,13 +101,24 @@ def get_initial(self):
 
 class CreateApplicationFormView(CreateView):
 
+    def get_form(self):
```
sks444, line 104: Add a comment on why this change is required and what it does.

sks444 (approved):
Looks good. Nice work @sandeepsajan0!

hypha/apply/funds/blocks.py
sandeepsajan0: No, we don't need it. I left it to avoid error but I just tested again by removing it and everything worked fine without this. I'll remove it and will add comments about the change as well.

PR #2827, opened 2022-04-26
Filter reviewer dashboard tables by ReviewerSettings

### LINKED ISSUE #2823: The "Your previous reviews" table is not filters after what users can access
I suggest we continue to list all the users reviews but the ones they can not access are not links. Maybe greyed out? With a tooltip that explains that they no longer have access to it?

frjo:
Fixes #2823

### DIFF
--- a/hypha/apply/dashboard/views.py
+++ b/hypha/apply/dashboard/views.py
@@ -4,7 +4,11 @@ from django.urls import reverse, reverse_lazy
 from django.views.generic import TemplateView
 from django_tables2.views import MultiTableMixin
 
-from hypha.apply.funds.models import ApplicationSubmission, RoundsAndLabs
+from hypha.apply.funds.models import (
+    ApplicationSubmission,
+    ReviewerSettings,
+    RoundsAndLabs,
+)
 from hypha.apply.funds.tables import (
     ReviewerSubmissionsTable,
     SubmissionFilterAndSearch,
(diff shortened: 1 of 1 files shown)


hypha/apply/dashboard/views.py
```diff
@@ -126,6 +130,15 @@ def projects_to_approve(self):
...
 
     def my_reviewed(self, submissions):
+        '''
+        If use_settings variable is set for ReviewerSettings use settings
+        parameters to filter submissions or return only reviewed_by as it
+        was by default.
+        '''
+        reviewer_settings = ReviewerSettings.for_request(self.request)
+        if reviewer_settings.use_settings:
+            submissions = submissions.for_reviewer_settings(self.request.user, reviewer_settings)
```
sks444: @frjo, this is AdminDashboardView this should be done only for [ReviewerDashboardView](https://github.com/HyphaApp/hypha/blob/4b865bb04106ab7a409fdf0dc3d09b75e7602c45/hypha/apply/dashboard/views.py#L175)

frjo: Oops, fixed now.

frjo: But you think the method is ok otherwise?

sks444: Yeah. This looks good. I'm just wondering if this should also be done for [awaiting_reviews](https://github.com/HyphaApp/hypha/blob/4b865bb04106ab7a409fdf0dc3d09b75e7602c45/hypha/apply/dashboard/views.py#L201) and my_flagged.

frjo: Good point!

sks444 (approved):

PR #2831, opened 2022-04-28
Add 'External Reviewer' picklist option to the batch action

### LINKED ISSUE #2805: Add 'External Reviewer' picklist to batch action
## Summary

As a manager I need to assign reviews to the Reviewer group as a batch action. 

## Description

When an application is in the 'External Review' workflow status, or in any workflow status, I can't rely on the user interface to select anyone from the Reviewer group because the third drop down menu is not available.

## Suggested Solution
Screenshot 2022-04-12 9 58 43 AM

## What UI looks like now
image

---
## Internal Approval Process
- [ ] @OpenTechFund/design to add designs to issue
- [ ] @OpenTechFund/devs to review designs for implementation
- [ ] Di to help organise testing/feedback session with OTF users
- [ ] @OpenTechFund/design do testing/feedback session with OTF users
- [ ] OTF to sign-off
- [ ] @OpenTechFund/design to make sure most up to date designs designs are attached to issue

sandeepsajan0:
Fixes #2805 

Add description here.

### DIFF
--- a/hypha/apply/funds/forms.py
+++ b/hypha/apply/funds/forms.py
@@ -348,8 +348,10 @@ class BatchUpdateReviewersForm(forms.Form):
         external_reviewers = self.cleaned_data['reviewer_reviewers']
         submissions = self.cleaned_data['submissions']
         if external_reviewers:
+            # User needs to be superuser or lead of all selected submissions.
             if self.user_cant_alter_submissions_external_reviewers(submissions, self.user):
                 self.add_error('reviewer_reviewers', _("Only Lead can change the External Reviewers"))
+            # If user is trying to change the external reviewers for submissions that doesn't have workflow with external_review stage.
             elif self.submissions_cant_have_external_reviewers(submissions):
                 self.add_error('reviewer_reviewers', _('External Reviewers cannot be selected because of the application workflow'))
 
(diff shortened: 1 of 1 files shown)


hypha/apply/funds/forms.py
```diff
@@ -327,13 +336,25 @@ def __init__(self, *args, user=None, round=None, **kwargs):
...
         return ApplicationSubmission.objects.filter(id__in=submission_ids)
 
     def clean(self):
         cleaned_data = super().clean()
+        external_reviewers = self.cleaned_data['reviewer_reviewers']
+        submissions = self.cleaned_data['submissions']
+        if external_reviewers:
+            # User needs to be superuser or lead of all selected submissions.
+            if self.user_cant_alter_submissions_external_reviewers(submissions, self.user):
+                self.add_error('reviewer_reviewers', _("Only Lead can change the External Reviewers"))
```
frjo: Would it be possible to have the error messages say "External Reviewer" instead of "reviewer_reviewers"?

sandeepsajan0: Actually, 'reviewer_reviewers' is the field name defined [here in form](https://github.com/HyphaApp/hypha/pull/2831/files#diff-8c72fbe57208c4b39ad8032b4678b32ad493b198a736813deae5efaa79b26797R318) and it is coming from there.  I tried to make it similar with existing [UpdateReviewerForm field](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/forms.py#L193). We may change 'reviewer_reviewers' to 'external_reviewers', but i don't think it can be in capitalized form. 
I also tried by passing non-field error but in that case it was showing `__all__` in place of 'reviewer_reviewers'.

PR #2833, opened 2022-04-29
Add env variable to turn hijack/become feature on/off

### LINKED ISSUE #2829: Add global env variable to turn hijack/become feature on/off
HIJACK_ENABLE = env.bool('HIJACK_ENABLE', False)

https://github.com/HyphaApp/hypha/blob/4e8e4d1c5b9ae2e7d81393fc6a39297585394b51/hypha/apply/users/views.py#L160-L172

sks444:
Fixes https://github.com/HyphaApp/hypha/issues/2829

### DIFF
--- a/hypha/apply/users/views.py
+++ b/hypha/apply/users/views.py
@@ -159,7 +159,10 @@ class EmailChangeDoneView(TemplateView):
 
 @login_required()
 def become(request):
-    if not request.user.is_apply_staff:
+    if not settings.HIJACK_ENABLE:
+        raise PermissionDenied()
+
+    if not request.user.is_superuser:
         raise PermissionDenied()
 
     id = request.POST.get('user_pk')
--- a/hypha/apply/users/views.py
+++ b/hypha/apply/users/views.py
@@ -89,7 +89,7 @@ class AccountView(UpdateView):
         return reverse_lazy('users:account')
 
     def get_context_data(self, **kwargs):
-        if self.request.user.is_superuser:
+        if self.request.user.is_superuser and settings.HIJACK_ENABLE:
             swappable_form = BecomeUserForm()
         else:
             swappable_form = None
--- a/hypha/apply/users/views.py
+++ b/hypha/apply/users/views.py
@@ -160,7 +160,7 @@ class EmailChangeDoneView(TemplateView):
 @login_required()
 def become(request):
     if not settings.HIJACK_ENABLE:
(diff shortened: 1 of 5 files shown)


hypha/apply/users/views.py
```diff
@@ -159,7 +159,10 @@ class EmailChangeDoneView(TemplateView):
 
 @login_required()
 def become(request):
-    if not request.user.is_apply_staff:
```
frjo, line 162: I believe this was an error, the form only shows for super admins so staff should not be allowed to use this function. Earlier staff did have access to this function.

frjo:
@sks444 Starting testing this and messed around with it. Seems to be working now, good approach you think?

With selectively adding the path/url I could not get the tests to pass even with the `@override_settings(HIJACK_ENABLE=True)`.

frjo:
Thinking now that adding `HIJACK_ENABLE=True` to `settings/test.py` would have worked.

hypha/apply/users/views.py
```diff
@@ -159,7 +159,10 @@ class EmailChangeDoneView(TemplateView):
...
+    if not settings.HIJACK_ENABLE:
+        raise PermissionDenied()
```
sks444: In theory we should return 404 here. Permission denied is dependent on user and when hijack is disabled it would not work for any user.

..the reason I wanted the urls to not exist when hijack is disabled. But we could also go-ahead with this way.

frjo: Added `HIJACK_ENABLE = True` in test.py, then it work to conditional add the become path. Looks ok?

hypha/apply/users/views.py
```diff
@@ -159,7 +159,10 @@ class EmailChangeDoneView(TemplateView):
 
 @login_required()
 def become(request):
-    if not request.user.is_apply_staff:
+    if not settings.HIJACK_ENABLE:
+        raise Http404(_('Hijack feature is not enabled.'))
```
frjo, line 163: Changed this to a 404 as well. Is it even needed now?

sks444: Not needed. 

_Although we could have it for code readability purposes._

sks444:
Looks good. Let's deploy it on test with 2FA enabled and Hijack disabled so we can test this once there as well.

frjo:
On test site now with settings from above.

PR #2840, opened 2022-05-05
Update overridden users index view to support groups/[gid]/users/ view.

### LINKED ISSUE #2839: The view that list all users belonging to a group returns an error about missing count
The /admin/groups/[gid]/users/ view is broken. 

The overrides index view from wagtail/users/views/users.py needs to be updated I believe.

frjo:
Fixes #2839

This view updated the overridden index view and adjust the results template to handle listing groups users as well.


sandeepsajan0 (approved):
Thanks @frjo I tested it, and it works great. Not sure how did I miss the group part at the time of overriding this.

PR #2842, opened 2022-05-05
Add support for 2FA and Hijack together

### LINKED ISSUE #2830: Show error messages if trying to hijack user that do not have 2FA and ENFORCE_2FA in active
The hijack/become feature gets trapped in a redirect loop if one try to hijack a user that do not have 2FA and 2FA is enforced.

Stop this from happening by instead showing an error message that the user can not be hijacked until 2FA is setup.

sks444:
Fixes https://github.com/HyphaApp/hypha/issues/2830

### DIFF
--- a/hypha/apply/users/forms.py
+++ b/hypha/apply/users/forms.py
@@ -3,6 +3,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import AuthenticationForm
 from django.utils.translation import gettext_lazy as _
 from django_select2.forms import Select2Widget
+from two_factor.utils import default_device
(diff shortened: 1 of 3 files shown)


frjo (changes requested):
We need to have a check if ENFORCE_TWO_FACTOR is true. IF it is not everything should work as before. Only if it is true we should just through the hopes n this PR.

I fully understand that you hesitated to implement this.

Personally I believe a better option is to simple state that hijack and enforce 2FA is incompatible.

sks444:
Thanks @frjo, I completely forgot that we should list all the users when 2FA is not enabled. Can you take a look now.

PR #2843, opened 2022-05-05
Add a feature in wagtail admin to reset 2fa for a user

### LINKED ISSUE #2821: Giver Staff Admin role permission to disable 2fa on user accounts
## Description

As a staff admin, I need the capacity to disable 2fa on user accounts to better manage users' access to the platform. The staff admin role overseas user accounts and troubleshoot technical glitches via the WagTail admin in Hypha. 

## Solution

Staff admin role has permission to disable 2fa for all user accounts except for Admin

sandeepsajan0:
Fixes #2821 
Add description here.


hypha/apply/users/views.py
```diff
@@ -309,5 +311,41 @@ def get_form_kwargs(self):
         return kwargs
 
 
+@method_decorator(permission_required(change_user_perm, raise_exception=True), name='dispatch')
+class TWOFAResetView(FormView):
```
sks444: Just to be consistent with this feature. How about naming `TWOFADisableView`?

hypha/apply/users/views.py
```diff
@@ -309,5 +311,41 @@ def get_form_kwargs(self):
...
+    """
+    View for PasswordForm to confirm the Disable 2FA process.
+    """
+    form_class = TWOFAPasswordForm
+    template_name = 'two_factor/reset.html'
```
sks444: maybe move this template to: `two_factor/admin/disable.html`?

hypha/apply/users/views.py
sks444: Oh.. that name already exists..then maybe call it TWOFAAdminDisableView?

hypha/apply/users/views.py
```diff
@@ -309,5 +311,41 @@ def get_form_kwargs(self):
         return kwargs
 
 
+@method_decorator(permission_required(change_user_perm, raise_exception=True), name='dispatch')
+class TWOFAResetView(FormView):
+    """
+    View for PasswordForm to confirm the Disable 2FA process.
+    """
```
sks444, line 318:
```
View for PasswordForm to confirm the Disable 2FA process on wagtail admin.
```

hypha/apply/users/templatetags/users_tags.py
```diff
@@ -25,3 +26,12 @@ def can_use_oauth(context):
     user = context.get('user')
 
     return can_use_oauth_check(user)
+
+
+@register.simple_tag(takes_context=True)
+def user_2fa_enabled(context):
```
sks444: Instead of using context and then getting the user, can we just pass the user as a variable? I think it is already available in the template?

hypha/apply/users/urls.py
```diff
@@ -88,6 +89,7 @@
         path('two_factor/required/', TWOFARequiredMessageView.as_view(), name='two_factor_required'),
         path('two_factor/backup_tokens/password/', TWOFABackupTokensPasswordView.as_view(), name='backup_tokens_password'),
         path('two_factor/disable/', TWOFADisableView.as_view(), name='disable'),
+        path('two_factor/reset/<str:user_id>/', TWOFAResetView.as_view(), name='two_factor_reset'),
```
sks444: `'two_factor/admin/disable/<str:user_id>/'`?

sks444 (commented):
Looks good. Just a couple of nit-picks.

hypha/apply/users/views.py
sandeepsajan0: Yeah, AdminDisableView sounds good.

hypha/apply/users/templatetags/users_tags.py
sandeepsajan0: Yes, user is there we can use that directly.

sks444 (approved):
Works well.

PR #2851, opened 2022-05-18
Add External Review Form block to fund, for external reviewers

### LINKED ISSUE #2273: Different review forms for staff and reviewers
### Description 

There is currently one review form associated with one proposal. We need separate review forms, one for staff reviewers and another for external reviewers. 

---
## Internal Approval Process
- [ ] @OpenTechFund/design to add designs to issue
- [ ] @OpenTechFund/devs to review designs for implementation
- [ ] Di to help organise testing/feedback session with OTF users
- [ ] @OpenTechFund/design do testing/feedback session with OTF users
- [ ] OTF to sign-off
- [ ] @OpenTechFund/design to make sure most up to date designs designs are attached to issue

sandeepsajan0:
Fixes #2273 

External Reviewer Form is optional, but if it is there in the fund then it should render to the external reviewers.
If an external review form is not there in the fund or funds' workflow doesn't have a stage with the external review then the Internal Review Form will be served.


hypha/apply/review/views.py
```diff
@@ -51,8 +51,16 @@ def get_context_data(self, **kwargs):
...
 
-def get_fields_for_stage(submission):
+def get_fields_for_stage(submission, user=None):
     forms = submission.get_from_parent('review_forms').all()
+    external_review_forms = submission.get_from_parent('external_review_forms').all()
+
+    # Use ExternalReviewForm if submission's stage has external review and external review form is attached to fund.
+    # ExternalReviewForm is only for non-staff reviewers(external reviewers)
+    if submission.stage.has_external_review and external_review_forms:
+        if user and not user.is_apply_staff:
```
sks444, line 61: External forms should only be rendered for reviewers. So instead of checking a user not staff we could check if a user is reviewer..?

hypha/apply/funds/models/utils.py
```diff
@@ -116,7 +116,8 @@ def render_landing_page(self, request, form_submission=None, *args, **kwargs):
     content_panels = AbstractStreamForm.content_panels + [
         FieldPanel('workflow_name'),
         InlinePanel('forms', label=_('Forms')),
-        InlinePanel('review_forms', label=_('Review Forms')),
+        InlinePanel('review_forms', label=_('Internal Review Forms')),
+        InlinePanel('external_review_forms', label=_('External Review Forms')),
```
sks444: If external review form is always gonna be one..ideal way would be to just add a relational field to select a review form..instead of adding InlinePanel which provides the possibility of adding multiple of them.

Although, in terms of UI, this way of adding external review form goes along with what we have for adding other forms. And we also have the validation which raises error if it is more than one. But would be good if we could limit the addition of forms to only one. 

@sandeepsajan0 can you check if it is possible to set max number of items that can be added for a InlinePanel.

sandeepsajan0: Yes, it is possible. It has a `max_num` property that can disable the Add button when the limit is reached. I'll add it for External Review form.

sks444 (approved):
Looks good. Nice work @sandeepsajan0!

@frjo this is ready to be deployed on test.

PR #2856, opened 2022-05-20
Fix TypeError for MessageInline View

### LINKED ISSUE #2855: The "/django-admin/activity/event/" detail view crash with TypeError
TypeError: has_add_permission() takes 2 positional arguments but 3 were given, see:

https://sentry.io/organizations/otf/issues/3288636037/?project=1268368&query=is%3Aunresolved

Removing "has_add_permission" from below resolves the issue but might have side effects. 

https://github.com/HyphaApp/hypha/blob/e560578d087b73cd923f01592b6e74cc953c81bf/hypha/apply/activity/admin.py#L6-L25

sandeepsajan0:
Fixes #2855 

Add description here.

### DIFF
--- a/hypha/apply/activity/admin.py
+++ b/hypha/apply/activity/admin.py
@@ -8,7 +8,7 @@ class MessageInline(admin.TabularInline):
     readonly_fields = ('type', 'recipient', 'content', 'status', 'external_id')
     can_delete = False
 
-    def has_add_permission(self, request):
+    def has_add_permission(self, request, obj):
         return False
 
 


hypha/apply/activity/admin.py
```diff
@@ -8,7 +8,7 @@ class MessageInline(admin.TabularInline):
     readonly_fields = ('type', 'recipient', 'content', 'status', 'external_id')
     can_delete = False
 
-    def has_add_permission(self, request):
+    def has_add_permission(self, request, obj):
```
frjo, line 11: Should we make the same update in the other call to "has_add_permission"? To "has_delete_permission"?

sandeepsajan0: I don't think we need to change it for other calls, because I don't see anything in the release notes regarding this and I checked the code there is no change in BaseModelAdmin, they change it only for InlineModelAdmin(especially for `has_add_permission`). So, I think we are good.

PR #2867, opened 2022-06-02
Integrate django-slack for direct slack messaging from hypha

### LINKED ISSUE #2852: Add ability to send Slack messages directly from Hypha
Hypha can now use an Hubot endpoint to sent messages to Slack. Hubot has not been updated for 3 years https://www.npmjs.com/package/hubot and it would be nice to have this built in to Hypha.

Current endpoint: https://github.com/OpenTechFund/botf

Packages like e.g. https://pypi.org/project/django-slack/ maybe can be of use.

sandeepsajan0:
Fixes #2852 

Add description here.


hypha/apply/activity/messaging.py
```diff
@@ -716,23 +716,28 @@ def slack_channels(self, source, **kwargs):
...
         data = {
-            "room": target_rooms,
             "message": message,
         }
-        response = requests.post(self.destination, json=data)
-
-        return str(response.status_code) + ': ' + response.content.decode()
+        for room in target_rooms:
+            try:
+                slack_message('messages/slack_message.html', data, channel=room)
```
sks444, line 734: @sandeepsajan0 so we don't have to initialise django slack with the env variables and it automatically uses them from the settings?

hypha/settings/base.py
```diff
@@ -524,6 +525,13 @@
 SLACK_DESTINATION_ROOM_COMMENTS = env.str('SLACK_DESTINATION_ROOM_COMMENTS', None)
 SLACK_TYPE_COMMENTS = env.list('SLACK_TYPE_COMMENTS', [])
 
+# Django Slack settings
+SLACK_TOKEN = env.str('SLACK_TOKEN', None)
+SLACK_USERNAME = env.str('SLACK_USERNAME', 'bot')
+SLACK_BACKEND = 'django_slack.backends.CeleryBackend'  # UrllibBackend can be used for sync
+if SLACK_DESTINATION_URL:
+    SLACK_ENDPOINT_URL = SLACK_DESTINATION_URL
```
sks444: @sandeepsajan0, let's get `SLACK_ENDPOINT_URL` directly from env. So replace `SLACK_DESTINATION_URL = env.str('SLACK_DESTINATION_URL', None)` with `SLACK_ENDPOINT_URL = env.str('SLACK_ENDPOINT_URL', None)`, and change `SLACK_DESTINATION_URL` to `SLACK_ENDPOINT_URL` everywhere.

We will have to set `SLACK_TOKEN` anyway for this to work so we can also change `SLACK_DESTINATION_URL` to `SLACK_ENDPOINT_URL` at the env settings.

sks444: This avoids confusion for someone who is looking at this in future.

hypha/apply/activity/messaging.py
sandeepsajan0: No, we don't need to initialise it. It automatically uses SLACK_TOKEN, and other env variables to send the messages.

hypha/settings/base.py
sandeepsajan0: okay, I was thinking of not changing the existing env variables. But yes, we will have to add SLACK_TOKEN so renaming should not be a big issue. I have changed it in the latest commit.

sks444 (approved):
@frjo this is ready to be deployed on test.

Env variables to be set are:
```
SLACK_TOKEN = env.str('SLACK_TOKEN', None)
SLACK_USERNAME = env.str('SLACK_USERNAME', 'bot')
SLACK_ENDPOINT_URL = env.str('SLACK_ENDPOINT_URL', None)
SLACK_DESTINATION_ROOM = env.str('SLACK_DESTINATION_ROOM', None)
```
Optional:
```
SLACK_DESTINATION_ROOM_COMMENTS = env.str('SLACK_DESTINATION_ROOM_COMMENTS', None)
SLACK_TYPE_COMMENTS = env.list('SLACK_TYPE_COMMENTS', [])
```

frjo:
This is now on test and seems to work well. Nice work!

hypha/apply/activity/messaging.py
```diff
@@ -716,23 +716,28 @@ def slack_channels(self, source, **kwargs):
     def send_message(self, message, recipient, source, **kwargs):
         target_rooms = self.slack_channels(source, **kwargs)
 
-        if not self.destination or not any(target_rooms):
+        if not self.destination or not any(target_rooms) or not settings.SLACK_TOKEN:
```
frjo: @sandeepsajan0 Do we need to check for destination/SLACK_ENDPOINT_URL here? It defaults to `https://slack.com/api/chat.postMessage` in  https://github.com/lamby/django-slack/blob/4a6d6579560c507f8d5d13276ba616dab249b704/django_slack/app_settings.py#L13

I think that means it is safe to leave empty in Hypha settings. Maybe add a command in `base.py` that `SLACK_ENDPOINT_URL` defaults to `https://slack.com/api/chat.postMessage` if not set.

Or make the setting in `base.py` directly default to `https://slack.com/api/chat.postMessage` instead of `None`.

hypha/settings/base.py
```diff
@@ -524,7 +525,13 @@
 
 SEND_READY_FOR_REVIEW = env.bool('SEND_READY_FOR_REVIEW', True)
 
-SLACK_DESTINATION_URL = env.str('SLACK_DESTINATION_URL', None)
+# Django Slack settings
+SLACK_TOKEN = env.str('SLACK_TOKEN', None)
+SLACK_USERNAME = env.str('SLACK_USERNAME', 'bot')
```
frjo: Change default to "Hypha"?

hypha/apply/activity/messaging.py
sandeepsajan0: Yes, we set it directly default to `https://slack.com/api/chat.postMessage` in `base.py` then we can remove that check for the destination. Let me change it.

PR #2886, opened 2022-06-13
Add underlines to most links

### LINKED ISSUE #2876: Underlines and links

We have a few (known) places where the universal lack of underlines in links is causing confusion.  The first is in the footer text -- if the site is configured with a link that link has no indication it is clickable.  For instance in this screenshot "Giving" is a link but you would never know:

image

There are existing links across Hypha that appear the same as non-link text; this applies to all of them.

A lot of links in Hypha would be more legible with an underline.  For instance clickable elements in the submissions table, and even in-page navigation such as "Show All".

Might we consider making underline the default for all links and explicitly removing the underline for `.button` based links (e.g. the log out button)

Color alone doesn't do the trick since (A) not everybody can see color and also (B) the blue is used in other places on non-clickable text.

We could instead apply underline just to identified concern points, but I'm worried that will create inconsistency.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- General public

**Ideal deadline**

We'll make this change locally on our own timeline, so no major urgency.

slifty:
This PR adds underlines to most links, keeping them off for primary navigation and buttons.

I wasn't sure if this PR is wanted at this stage but I figure it will be easier to decide if this is the direction Hypha wants by being able to see / work with the difference.

Resolves #2876

### DIFF
--- a/hypha/static_src/src/sass/apply/base/_base.scss
+++ b/hypha/static_src/src/sass/apply/base/_base.scss
@@ -48,7 +48,7 @@ textarea {
 
 a {
     color: $color--primary;
-    text-decoration: none;
+    text-decoration: underline;
 
     &:hover {
         cursor: pointer;
--- a/hypha/static_src/src/sass/apply/components/_button.scss
+++ b/hypha/static_src/src/sass/apply/components/_button.scss
@@ -5,6 +5,7 @@
     background-image: none;
     border: 0;
     box-shadow: none;
+    text-decoration: none;
 
     &:hover {
         cursor: pointer;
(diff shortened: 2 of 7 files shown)


frjo:
Thanks. I have deployed this to test so more people can take a look.

frjo:
This does not work well with the current design unfortunately. Especially on the public part, that I believe only OTF uses, it looks weird in many places.

To be able to commit this to Hypha we need to do it more granularly. Only adding underline to links in specific areas.

Skärmavbild 2022-06-15 kl  08 36 40

slifty:
Got it -- I'll close out this PR and consider opening one for that inverse solution at a later date.

PR #2888, opened 2022-06-14
Upgrade Django bleach to 3.x and bleach to 5.x

### LINKED ISSUE #2803: Update django-bleach and bleach to latest versions.
Version 5 of bleach is not yet compatible with django-bleach 2.x so keep tabs on that.

bleach is a Python module that takes any HTML input, and returns valid, sanitised HTML that contains only an allowed subset of HTML tags, attributes and styles. django-bleach is a Django app that makes using bleach extremely easy.

sandeepsajan0:
Fixes #2803 

Add description here.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,13 +4,13 @@ sentry-sdk==1.5.0
 
 # Production dependencies
 Babel==2.9.1
-bleach==4.1.0
+bleach==5.0.0
 boto3==1.20.22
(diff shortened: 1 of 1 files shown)


sandeepsajan0:
I tested it with submission details pages and a few other pages, all are working fine. But let's use it on test site for a while because we have seen many issues related to bleach.

PR #2890, opened 2022-06-14
Add EMAIL_SUBJECT_PREFIX settings to email subject

### LINKED ISSUE #2775: Implement a EMAIL_SUBJECT_PREFIX setting
The existing setting `EMAIL_SUBJECT_PREFIX` is a standard Django setting. Adding a subject prefix is a common praxis to make it easy to identify e-mails for users.

https://docs.djangoproject.com/en/4.0/ref/settings#email-subject-prefix

It is not used anywhere in Hypha. It was added early on and just stuck around. The default value is "None".

Lets implement it on all e-mails sent out from Hypha.

It could get inserted here:

https://github.com/HyphaApp/hypha/blob/6554683d5756d1932418282ea2d225aa22c9b5ce/hypha/apply/activity/tasks.py#L10-L20

Setting it as empty string should mimic how it works today, no prefix at all. But any organisation using Hypha can decide to add any prefix they like.

sandeepsajan0:
Fixes #2775 

Add description here.

### DIFF
--- a/hypha/apply/activity/tasks.py
+++ b/hypha/apply/activity/tasks.py
@@ -8,6 +8,8 @@ app.config_from_object(settings, namespace='CELERY', force=True)
 
 
 def send_mail(subject, message, from_address, recipients, logs=None):
+    if settings.EMAIL_SUBJECT_PREFIX:
+        subject = str(settings.EMAIL_SUBJECT_PREFIX) + str(subject)
     # Convenience method to wrap the tasks and handle the callback
     send_mail_task.apply_async(
         kwargs={


hypha/apply/activity/tasks.py
```diff
@@ -8,6 +8,8 @@
 
 
 def send_mail(subject, message, from_address, recipients, logs=None):
+    if settings.EMAIL_SUBJECT_PREFIX:
+        subject = str(settings.EMAIL_SUBJECT_PREFIX) + str(subject)
```
sks444, line 12: @sandeepsajan0 you think we need to add `' '` in b/w both strings?

frjo: In Django the default value for this variable is

`EMAIL_SUBJECT_PREFIX = '[Django] '`

with a space at the end. I think best to stick with that for Hypha.

sandeepsajan0: @sks444 Yes, I also thought about adding `' '`, `-` or `:`  in b/w the strings but as Fredrik mentioned we already have space in its default value so we can expect a similar pattern.

PR #2907, opened 2022-07-07
Remove js-hidden class from wrapped block inside the group fields

### LINKED ISSUE #2798: Group field is buggy

sandeepsajan0:
Fixes #2798 (2nd part)

Add description here.


frjo:
Deploying to test now.

PR #2910, opened 2022-07-07
Fix Section Header field render issue

### LINKED ISSUE #2909: Header Block is not rendered correctly in edit view for Stream field forms
The field is getting rendered as a string instead of a block.
image

sandeepsajan0:
Fixes #2909 

Add description here.


sandeepsajan0:
@frjo we can put it on the test to check the functionality of other fields(shouldn't break anything). 
 I took the reference from [application_base.html](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/templates/funds/application_base.html#L56).

PR #2946, opened 2022-08-17
Regression: Fix rendering of help_text for steam_form -> blocks

### LINKED ISSUE #2880: Update to Wagtail 3.x

theskumar:
Fixes #2880 

The current implementation renders the className of the
block in the help section of the block.

This fix addresses that issue and makes use of meta class
to get the help_text as implemented in the wagtail’s 
`stuctBlock`

[help_text = getattr(block.meta, "help_text", None)](https://github.com/wagtail/wagtail/blob/a5f699ffbd7312c260c9c7d31d92b6781753bcfd/wagtail/blocks/stream_block.py#L652-L656)

### DIFF
--- a/hypha/apply/stream_forms/blocks.py
+++ b/hypha/apply/stream_forms/blocks.py
@@ -34,7 +34,6 @@ from .fields import MultiFileField, SingleFileField
 
 class FormFieldBlock(StructBlock):
     field_label = CharBlock(label=_('Label'))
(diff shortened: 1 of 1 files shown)


theskumar:
The issue was that I had these strings as `help_text` in the DB. How I got them is another issue! Probably because I imported the `sandbox.dump` or maybe while using factories.

Screenshot 2022-08-17 at 08 48 51

Closing this for now, as it's a non-issue.

PR #2950, opened 2022-08-18
Fix duplicate comments in in-app feed

### LINKED ISSUE #2929: The same comment show multiple times in in-app notifications

sandeepsajan0:
Fixes #2929 

Fixed two issues:
1. Fixed in-app feed to show only current activities.
2. Add `edit-comment.js` file to the project detail page, to avail the comment edit feature on the project's communication tab.

### DIFF
--- a/hypha/apply/activity/context_processors.py
+++ b/hypha/apply/activity/context_processors.py
@@ -5,5 +5,5 @@ def notification_context(request):
     context_data = dict()
     if hasattr(request, 'user'):
         if request.user.is_authenticated and request.user.is_apply_staff:
-            context_data['latest_notifications'] = Activity.objects.latest().order_by('-timestamp')[:5]
+            context_data['latest_notifications'] = Activity.objects.filter(current=True).latest().order_by('-timestamp')[:5]
     return context_data
--- a/hypha/apply/activity/views.py
+++ b/hypha/apply/activity/views.py
@@ -70,7 +70,7 @@ class NotificationsView(ListView):
 
     def get_queryset(self):
         # List only last 30 days' activities
-        queryset = Activity.objects.latest()
+        queryset = Activity.objects.filter(current=True).latest()
         self.filterset = self.filterset_class(self.request.GET, queryset=queryset)
         return self.filterset.qs.distinct().order_by('-timestamp')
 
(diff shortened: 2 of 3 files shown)


hypha/apply/activity/context_processors.py
```diff
@@ -5,5 +5,5 @@ def notification_context(request):
     context_data = dict()
     if hasattr(request, 'user'):
         if request.user.is_authenticated and request.user.is_apply_staff:
-            context_data['latest_notifications'] = Activity.objects.latest().order_by('-timestamp')[:5]
+            context_data['latest_notifications'] = Activity.objects.filter(current=True).latest().order_by('-timestamp')[:5]
```
theskumar, line 8: The `Activity` model class already has `Meta.ordering`, so there is no need to declare `order_by` here again and if the ordering is used there is no need to use `.latest()`.

```suggestion
            context_data['latest_notifications'] = Activity.objects.filter(current=True).all()[:5]
```

or, if we really want to enforce the ordering here, use: 
```
Activity.objects.filter(current=True).all().order_by('-timestamp')[:5]
```

theskumar (approved):

hypha/apply/activity/context_processors.py
theskumar: It looks like `latest()` is a [custom queryset implementation](https://github.com/HyphaApp/hypha/blob/0288e814da3bf4172c6feb0dee9aa952cab3a537/hypha/apply/activity/models.py#L58-L59), so it makes sense to keep it here. 

That said, why not use an appropriate custom queryset for current activities instead of using `current=True` everywhere?

sandeepsajan0: Yeah, that is also a good idea but It is not that widely used right now, it is only used in one or two places. So, I think it is fine for now.

fourthletter:
@sandeepsajan0 I tested this issue and it was flawless. I made a comment, edited this comment, and the comment only appeared once on the in-app notification. Thank you! 💯 

### Comment on submission 
Screenshot 2022-08-28 7 27 15 PM


### In-App Notification 
Screenshot 2022-08-28 7 26 54 PM

PR #2951, opened 2022-08-22
Add js validation for acceptable file types for all uploads

theskumar:
Fixes - Internally reported

ISSUE: 
1. A user can upload, select and upload an invalid file type after modifying the system-dialog. This patch adds a javascript validation that prevents uploading that file if the extension is not one of the supported types.
2. The submission_id is used to do URL reverse for private files are wrongly put to use the id of the revisions instead of the parent submission. 

The javascript of the django-form-file is updated to validate the file type of the selected file.  https://github.com/mbraak/django-file-form/pull/524 and https://github.com/mbraak/django-file-form/pull/526 are created in the original package; 

we use a temporary version here from the fork: https://github.com/mbraak/django-file-form/compare/0d7896ac596ba2e107bb823b134899c0401fa01a...theskumar:django-file-form:feature/error-message-invalid-files?expand=1

PS: The implementation uses the same behavior used to accept the file during [drag-and-drop](https://github.com/mbraak/django-file-form/blob/896432518b890d7119ff9f454fd66d42ec9ddd8c/frontend/src/drop_area.ts#L93-L95) functionality: 

After this is deployed, the `django_file_form_uploadedfile` table should cleaned of any unaccepted file type already upload.
The management command `python manage.py delete_unused_files` can help, but it doesn't delete files less than one day old, so use manual data removal. 

cc: @fourthletter


frjo:
Deploying this to test now.

@theskumar Nice work on fixing and upstreaming this! Hopefully this will get in to the next relase of django-file-form. If not we will need to move the fork to a Hypha repo.

theskumar:
Thanks! 

Yes, moving to hypha org sounds like a good idea.

theskumar:
@frjo I've tested it at https://test-apply.hypha.app/community-lab/, and the expected behavior looks good.

theskumar:
Added one more patch to the previous solution. https://github.com/mbraak/django-file-form/pull/525

Also, add the functionality to display the invalid file types selected by the user while Dragging-an-dropping and selection via system-dialog. https://github.com/mbraak/django-file-form/pull/526

See the test video below.


https://user-images.githubusercontent.com/236356/186152911-62e3a67a-6820-4acb-99b9-b6c553485d6e.mp4

theskumar:
Also, @mbraak has been kind enough to merge my previous PR, and we can expect a release in the upstream soon. 
- https://github.com/mbraak/django-file-form/pull/523 
- https://github.com/mbraak/django-file-form/pull/524

I've added these two new PR with a patch and an error message feature.
- https://github.com/mbraak/django-file-form/pull/525 
- https://github.com/mbraak/django-file-form/pull/526

frjo:
The latest commits to this PR is now being deployed to test.

frjo (approved):

frjo:
@fourthletter This is on test and awaiting OTF approval.

fourthletter:
@theskumar Thank you SO much for pushing this through last week. I was also able to upload a jpg to the test application 👍

PR #2953, opened 2022-08-26
Add vscode launch config to start django debug server

theskumar:
This configuration allows running the Django server using `python manage.py runserver` and connects to the debugger.

This will be useful to other developers using VSCode to start the Django server and quickly debug the code.


theskumar:
@frjo can you take a look at the CI failing? 
Screenshot 2022-08-26 at 07 31 21

frjo:
It is the CircleCI cache that is acting up, happens from time to time.

theskumar:
Understood!

PR #2952, opened 2022-08-26
Upgrade to python 3.10.

frjo:
Fixes #2901

PR #2954, opened 2022-08-26
Improve testing and development experience

theskumar:
Fixes N/A

- Add `make test` command to lint, test, and report coverage
- Add pytest runner 
  - It automatically set up to use the correct Django settings module
  - Runs tests on all the available CPUs, I was able to run all the tests in total=~30sec using the `make test` command.
  - It provides a cleaner error report, compared to DjangoTestRunner
- Coverage
  - Displayed the report on the terminal after the test finish
  - Disable reporting of coverage for test and migration files
  - Enable coverage reporting for the template files
  - “make test” generates HTML reports to quickly check the coverage of each file
- Update CircleCI to use `pytest` command and CircleCI parallelism=4 (reduction in test time by ~2-3mins from 8-9min to ~5min)


theskumar:
The same cache problem in #2953 seems to be an issue here as well. Are you open to moving the test to GitHub Actions? If this doesn't resolve?

frjo:
I have adjusted the cache settings here https://github.com/HyphaApp/hypha/pull/2952/files. I think that will mitigate the problem.

theskumar:
The coverage is reduced because of the way it's tracked now. 
- Added template `coverage` and removed the `coverage` of migration files.

Also, Django tests now run using parallel circle CI workers (n=4). The test-be now finishes in ~5mins instead of ~9mins.

Screenshot 2022-08-29 at 09 45 35

Makefile
```diff
@@ -12,12 +13,17 @@ lint: sort style
...
 .PHONY: sort-fix
 sort-fix:
-	@echo "Fixing imports with isort" && isort .
+	@echo "Fixing imports with isort" && isort hypha
 
 .PHONY: style
 style:
 	@echo "Checking code style with flake8" && flake8 .
+
+test: lint
```
frjo: Can we move "lint" to its own line here to make it easier to read.

frjo (changes requested):
Excellent work, getting the CI time down is a big plus. Just one very minor nitpick.

Makefile
theskumar: I've updated the "test" target to use other smaller targets in https://github.com/HyphaApp/hypha/pull/2954/commits/a65345a42749c3ced5f9598b6632129bfdb0de47. 

The `test` target now looks like this
https://github.com/HyphaApp/hypha/blob/a65345a42749c3ced5f9598b6632129bfdb0de47/Makefile#L43-L44

This will be maintainable and easier to read.

PR #2963, opened 2022-08-31
Add error message for PAF not configured condition

### LINKED ISSUE #2960: Add a guide message if PAF is not configured to the fund
For now, I am thinking of showing a message on the same page.

"Approval Form is not configured the fund associated with this project/submission. Please configure it first from [Wagtail Admin]('URL to respective fund')."

sandeepsajan0:
Fixes #2960 

Add description here.

### DIFF
--- a/hypha/apply/projects/templates/application_projects/project_admin_detail.html
+++ b/hypha/apply/projects/templates/application_projects/project_admin_detail.html
@@ -48,26 +48,24 @@
 </div>
 {% endif %}
 
-{% if object.requires_approval %}
-    <a
-        {% if not object.can_send_for_approval %}
-            onclick="return false"
-            data-tooltip="
-                {% if not object.lead %}
-                    {% trans "A lead must be assigned" %}
-                {% elif not object.user_has_updated_details %}
-                    {% trans "Project approval form must be completed" %}
-                {% elif object.can_make_approval or object.can_make_final_approval %}
-                    {% trans "Currently awaiting approval" %}
-                {% endif %}"
-        {% endif %}
-        data-fancybox
-        data-src="#send-for-approval"
-        class="button button--bottom-space button--primary button--full-width {% if not object.can_send_for_approval or not user.is_apply_staff %}button--tooltip-disabled{% endif %}"
-        href="#">
-        {% trans "Submit for Approval" %}
-    </a>
-{% endif %}
+<a
+    {% if not object.can_send_for_approval %}
+        onclick="return false"
+        data-tooltip="
+            {% if not object.lead %}
+                {% trans "A lead must be assigned" %}
+            {% elif not object.user_has_updated_details %}
+                {% trans "Project approval form must be completed" %}
+            {% elif object.can_make_approval or object.can_make_final_approval %}
+                {% trans "Currently awaiting approval" %}
+            {% endif %}"
+    {% endif %}
+    data-fancybox
+    data-src="#send-for-approval"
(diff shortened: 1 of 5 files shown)


hypha/apply/projects/templatetags/approval_tools.py
```diff
@@ -40,3 +41,8 @@ def user_can_final_approve_project(project, user):
 @register.simple_tag
 def user_can_edit_project(project, user):
     return project.editable_by(user)
+
+
+@register.simple_tag
+def generate_fund_page_admin_url(page):
```
theskumar:
```suggestion
def fund_page_admin_url(page):
```

Reason: the function is not generating the URL.

Also, curious why not using `{% url "funds_fundtype_modeladmin_edit", page.id %}`, another recommendation here is to use `fund` as function arg instead of a `page`. This function is not capable of generating a fund URL from any `page` object.

hypha/apply/projects/views/project.py
```diff
@@ -725,6 +725,7 @@ def get_context_data(self, **kwargs):
         return super().get_context_data(
             title=self.object.title,
             buttons=self.buttons(),
+            approval_form_exists=True if self.approval_form else False,
```
theskumar, line 728: This looks unnecessary here; it can be easily moved to template logic if `self.approval_form` is passed as context.

hypha/apply/projects/templates/application_projects/project_approval_form.html
```diff
@@ -8,38 +8,53 @@ <h2 class="heading heading--no-margin">{% trans "Editing" %}: {{ object.title }}
...
     </div>
-</div>
+
+{% else %}
+    <div class="wrapper wrapper--light-grey-bg wrapper--form wrapper--sidebar">
+        <div class="wrapper--sidebar--inner">
+            <p>
+                Approval Form is not configured to the fund associated with this project/submission.
+                Please configure it first from
+                <a href="{% generate_fund_page_admin_url object.submission.page %}" target="_blank">Wagtail Admin</a>.
```
theskumar: Consider using the following language. 

```
Approval form not configured. Please add an approval form in the [fund settings](url-to-fund-admin-page).
```

hypha/apply/projects/templates/application_projects/project_admin_detail.html
```diff
@@ -48,26 +48,24 @@ <h4 class="modal__header-bar">{% trans "Approve Contract" %}</h4>
...
+    {% if not object.can_send_for_approval %}
+        onclick="return false"
+        data-tooltip="
+            {% if not object.lead %}
+                {% trans "A lead must be assigned" %}
+            {% elif not object.user_has_updated_details %}
+                {% trans "Project approval form must be completed" %}
+            {% elif object.can_make_approval or object.can_make_final_approval %}
+                {% trans "Currently awaiting approval" %}
+            {% endif %}"
```
theskumar, line 61: Too many logic and permission-related things happening here. Consider moving to a python function that returns 1) `can_send_for_appoval` 2) `reason`

The python function can also be quickly unit tested.

theskumar (changes requested):

hypha/apply/projects/templatetags/approval_tools.py
sandeepsajan0: Yes, adding it directly to the template is better. The `page` is denoting the fund here, we have page and round two fields in the submission table and both have relation to the Page model only.

hypha/apply/projects/views/project.py
sandeepsajan0: As we don't need the approval_form in the template for any other usage so I thought it is better to pass just a boolean instead of the whole form? Maybe we can pass it later depending on the requirement.

PR #2974, opened 2022-09-07
Remove reviewers from list who opinionated but didn't submit review

### LINKED ISSUE #2973: KeyError: <AssignedReviewers: John Doe>
Sentry Issue: [OTF-WEBAPP-BV](https://sentry.io/organizations/otf/issues/3544769973/?referrer=github_integration)

```
KeyError: <AssignedReviewers: John Doe>
(4 additional frame(s) were not displayed)
...
  File "django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "django/views/generic/list.py", line 157, in get
    context = self.get_context_data()
  File "hypha/apply/review/views.py", line 377, in get_context_data
    review = reviews[reviewer]
```

sandeepsajan0:
Fixes #2973 

We already had a check for if a reviewer is opinionated but there is no review(null) added by the reviewer. But there is an edge case that if the reviewer has a draft review (that means not null) and opinionated another review as well, it is getting passed and causing an error. 
Now we have put the check for the draft as well.

### DIFF
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -964,9 +964,9 @@ class AssignedReviewersQuerySet(models.QuerySet):
         ]
         return self.exclude(
             # Remove people from the list who are opinionated but
-            # didn't review, they appear elsewhere
-            opinions__isnull=False,
-            review__isnull=True,
+            # didn't submit a review, they appear elsewhere
+            Q(opinions__isnull=False) &
+            Q(Q(review__isnull=True) | Q(review__is_draft=True))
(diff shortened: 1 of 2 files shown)


theskumar:
Look good, possible to compliment the solution with a test?

frjo (approved):
Nice fix!

PR #2977, opened 2022-09-12
Attached forms displayed in the rounds admin page fix

### LINKED ISSUE #2965: Display of workflow and forms names on rounds in wagtail admin are wrong
Display the round name instead of the name of each workflow/form.

theskumar:
Fixes #2965

Fixed by re-implementing ReadOnlyInlinePanel. Simplified the implementation of ReadOnlyInlinePanel to follow the single responsibility principle. It now just renders all the related items in a list. 

The responsibility of rendering items can be overridden using the panels passed to this ReadOnlyInlinePanel.

Screenshot 2022-09-12 at 07 15 40

Thanks @sandeepsajan0 for reviewing the fix

### DIFF
--- a/hypha/apply/funds/edit_handlers.py
+++ b/hypha/apply/funds/edit_handlers.py
@@ -68,9 +68,6 @@ class ReadOnlyPanel(Panel):
             try:
                 value = getattr(self.instance, self.panel.attr)
             except AttributeError:
-                self.attr = '__'.join(
-                    [self.instance._meta.model_name, str(self.instance.id)]
-                )
                 value = self.instance
 
             if callable(value):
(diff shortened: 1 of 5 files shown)


frjo (approved):
Works perfectly, will put on test.

hypha/apply/funds/edit_handlers.py
```diff
@@ -63,7 +66,7 @@ def render_as_field(self):
 
         def context(self):
             try:
-                value = getattr(self.instance, self.attr)
+                value = getattr(self.instance, self.panel.attr)
             except AttributeError:
                 self.attr = '__'.join(
```
sandeepsajan0: Overall it is working perfectly, and I noticed we should change it as well to `self.panel.attr` because we are not using `self.attr` anywhere(panel's attr are getting used everywhere).

theskumar: This is not needed, so I've removed it.

PR #2966, opened 2022-08-31
Make project approval form similar like ApplicationForms or Reviews Forms

sandeepsajan0:
Fixes #2958 
Add description here.

PR #2979, opened 2022-09-14
Fix PAF admin's related forms issues

sandeepsajan0:
Reported in slack by @fourthletter.

Previously, we were using it as a foreign key so it was hard coded but we changed it in PR #2966 so it needs to be changed.


frjo (approved):
Nice and quick fix.

Deploying to test now.

PR #2980, opened 2022-09-14
Move SendNoticeBlock to required blocks

### LINKED ISSUE #2460: Make "send (determination) notice" a required field in OTF Hypha instance
## Description
On the create determination page for a project (example URL: https://apply.opentech.fund/apply/submissions/XXXX/determination/add/) the primary action button is labelled as Send/Save, as shown in the screenshot below:
 
image

This is controlled by adding a "Send notice" field to the determinaton forms. If the user who creates the determination form forgets to add this field, it won't be shown.

## Current design
The "Send notice" field is in the "Fields" group, as can be seen below:

image

### Wagtail admin section
In the OTF Hypha instance, make the "Send notice" field a "Required" field (i.e move it from the Fields group to the Required group), so that it is shown on the form by default.

### Determination form
In the OTF Hypha instance, move the checkbox to where it is shown below - 

Screenshot 2021-07-30 at 12 10 04

sandeepsajan0:
Fixes #2460


theskumar (approved):

hypha/apply/determinations/blocks.py
```diff
@@ -53,10 +53,23 @@ def get_field_kwargs(self, struct_value):
         return kwargs
 
 
-class SendNoticeBlock(CheckboxFieldBlock):
+class SendNoticeBlock(DeterminationMustIncludeFieldBlock):
+    name = 'send_notice'
+    description = 'Send Notice'
+
+    default_value = BooleanBlock(required=True)
```
frjo: I think this should be `default_value = BooleanBlock(default=True, required=False)`

frjo:
The field say "Send_Notice Field", can we get that to be "Send notice Field". Just looks nicer.

Skärmavbild 2022-09-26 kl  15 59 38

frjo (changes requested):

hypha/apply/determinations/blocks.py
sandeepsajan0: I have updated the default_value field as we discussed.

sandeepsajan0:
> The field say "Send_Notice Field", can we get that to be "Send notice Field". Just looks nicer.
> 
> Skärmavbild 2022-09-26 kl 15 59 38

Yes, we can change it but I think that is a separate issue because I found it at other places as well like full_name field in application forms. It is happening because we are using [name field here](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/utils/blocks.py#L192) and name field can't have spaces in it. 
So, a possible solution is we use [Block's meta's label](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/determinations/blocks.py#L59) instead of the name field whenever the label is present.

frjo (approved):

PR #2989, opened 2022-09-21
Make user e-mail field case insensitive to prevent multiple account with same email

### LINKED ISSUE #2985: User e-mail field is not case incentive
This makes it possible to have one user with `test@example.org` and another with `TEST@EXAMPLE.ORG` etc.

Since e-mail is also the username this need to be fixed.

There is a number of accounts today with duplicate e-mails. How should we handle these @fourthletter.

See also https://github.com/OpenTechFund/issues.otf/issues/34

sandeepsajan0:
Fixes #2985

### DIFF
--- a/hypha/apply/stream_forms/forms.py
+++ b/hypha/apply/stream_forms/forms.py
@@ -1,7 +1,13 @@
 from django import forms
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.forms.fields import EmailField
 from django.forms.forms import DeclarativeFieldsMetaclass
+from django.utils.safestring import mark_safe
 from django_file_form.forms import FileFormMixin
 from wagtail.contrib.forms.forms import BaseForm
+
+from hypha.apply.users.utils import get_user_by_email, is_user_already_registered
 
 
 class MixedFieldMetaclass(DeclarativeFieldsMetaclass):
@@ -41,8 +47,35 @@ class StreamBaseForm(FileFormMixin, forms.Form, metaclass=MixedFieldMetaclass):
 
 
 class PageStreamBaseForm(BaseForm, StreamBaseForm):
-    # Adds page and user reference to the form class
-    pass
+    """ Adds page and user reference to the form class"""
+
+    def clean(self):
+        cleaned_data = super().clean()
+        for field, value in self.fields.items():
+            # email validation of submission form
+            if isinstance(value, EmailField):
+                email = self.data.get(field)
+                if email:
+                    is_registered, _ = is_user_already_registered(email=self.data.get(field))
+                    if is_registered:
+                        user = get_user_by_email(email=email)
+                        if not user:
+                            self.add_error(field, 'Found multiple account')
+                            raise ValidationError(mark_safe('Found multiple account for the same email. '
+                                                            'Please login with the correct credentials or '
+                                                            '<a href="mailto:{}">'
+                                                            'contact to the support team'
+                                                            '</a>.'.format(settings.ORG_EMAIL)))
+
+                        elif not user.is_active:
+                            self.add_error(field, 'Found an inactive account')
+                            raise ValidationError(mark_safe('Found an inactive account for the same email. '
+                                                            'Please use different email or '
+                                                            '<a href="mailto:{}">'
+                                                            'contact to the support team'
+                                                            '</a>.'.format(settings.ORG_EMAIL)))
+
+        return cleaned_data
 
 
 class BlockFieldWrapper:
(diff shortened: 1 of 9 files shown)


hypha/apply/users/models.py
```diff
@@ -87,18 +93,49 @@ def create_superuser(self, email, password, **extra_fields):
...
+        return params
+
     def get_or_create_and_notify(self, defaults=dict(), site=None, **kwargs):
         # Set a temp password so users can access the password reset function if needed.
         temp_pass = BaseUserManager().make_random_password(length=32)
         temp_pass_hash = make_password(temp_pass)
         defaults.update(password=temp_pass_hash)
-        user, created = self.get_or_create(defaults=defaults, **kwargs)
-        if created:
+        user = self.filter(email__iexact=kwargs.get('email')).first()  # case insensitive matching
```
theskumar: use `get_user_by_email` here?

hypha/apply/users/models.py
```diff
@@ -66,6 +69,9 @@ def _create_user(self, email, password, **extra_fields):
         if not email:
             raise ValueError('The given email must be set')
         email = self.normalize_email(email)
+        existing_user = self.filter(email__iexact=email)
```
theskumar: use `get_user_by_email` here?

hypha/apply/users/utils.py
```diff
@@ -7,6 +8,21 @@
 from django.utils.http import urlsafe_base64_encode
 
 
+def get_user_by_email(email):
```
theskumar, line 12: Please add test for this function.

theskumar (changes requested):

hypha/apply/users/models.py
```diff
@@ -87,18 +93,49 @@ def create_superuser(self, email, password, **extra_fields):
 
         return self._create_user(email, password, **extra_fields)
 
+    def _extract_model_params(self, defaults, **kwargs):
```
theskumar, line 100: I"m curious to know why this function needs to be implemented or overridden.

hypha/apply/users/models.py
sandeepsajan0: At the time of user creation, we only need a case-insensitive search.  We can also use `get_user_by_email` but for that, we will need to pass an extra argument and will need to update the logic of the function a bit.  But yes, our logic will only be in one place. Let me try with that logic and see if it looks good.

hypha/apply/users/models.py
sandeepsajan0: It is not overridden, we just copied it from Queryset class(as mentioned in the comment). 
It was inaccessible directly, maybe because of the custom manager(`from_queryset`).

hypha/apply/users/models.py
```diff
@@ -91,18 +97,49 @@ def create_superuser(self, email, password, **extra_fields):
...
+        return params
+
     def get_or_create_and_notify(self, defaults=dict(), site=None, **kwargs):
         # Set a temp password so users can access the password reset function if needed.
         temp_pass = BaseUserManager().make_random_password(length=32)
         temp_pass_hash = make_password(temp_pass)
         defaults.update(password=temp_pass_hash)
-        user, created = self.get_or_create(defaults=defaults, **kwargs)
-        if created:
+        user = get_user_by_email(kwargs.get('email'), sensitive_search=0)  # case insensitive matching
```
theskumar: Please not do this here. I'll provide solution here.

theskumar (changes requested):

hypha/apply/users/models.py
theskumar: Let's not do this

theskumar (changes requested):
Please ensure applicant group is added to the user if they have atleast one application submisison

fourthletter:
Amazing! Thank you 💯 

image

hypha/apply/stream_forms/forms.py
```diff
@@ -41,8 +47,35 @@ def hidden_fields(self):
 
 
 class PageStreamBaseForm(BaseForm, StreamBaseForm):
-    # Adds page and user reference to the form class
-    pass
+    """ Adds page and user reference to the form class"""
+
+    def clean(self):
```
sandeepsajan0, line 52: @frjo I checked this piece of code is not getting used by any review forms or determination forms. Both(Review and determination) forms have their own form classes. So, I think we can move forward with this change. What do you think?

theskumar (approved):
@sandeepsajan0 feel free to remove https://github.com/HyphaApp/hypha/pull/2989#discussion_r987862931 or open a new PR just for removing the unused form.

PR #2991, opened 2022-09-26
Send email digest for activities in Hypha for org level staff

theskumar:
Fixes #2882

This feature complements or provides an alternative to the slack notifications sent by the system to organization staff members.

- [x] 🏃🏼 management command to prepare and send the digest as an email `python manage.py send_staff_email_digest`
- [x] add email settings `ACTIVITY_DIGEST_RECIPIENTS_EMAILS`
- [x] Capture events of the last seven days only. 
- [x] Port all the staff messages sent to staff to the staff notification adaptor

Screenshot 2022-10-06 at 11 06 46@2x

Screenshot 2022-10-06 at 11 08 47@2x

- add mdmail dependency
- add core mail send utils to wrap Django’s send_mail, which takes markdown or text as input.
- add new `ACTIVITY_DIGEST_RECIPIENT_EMAILS` setting

PR #3003, opened 2022-10-11
Introduce archive feature for the submissions(in batch actions)

### LINKED ISSUE #2996: Feature: Support archiving of submission
### User Story
Historic submissions and rounds are not relevant and they become noise on the dashboard. Labs doesn't need archiving they are long lived. Mimizing the amount of data on the platform aligns with better data retention and responsibility practices.

### Proposed Solution

MVP: Archive entire rounds, funds, labs
Long-term goal: archive individual submissions

- Archive submission using true/false (no future date based)
- Send staff notification when some
- Should be able to unarchive a submission
- Add "Archive" button; it should behave the same way as "Delete" button.
- If a submission is archived, then:
  - reviews are also archived
  - determinations are also archived
- 'Staff Admin' and 'Admin' would be able to archive. This functionality is not available to other user groups. 

**Additional context**

Screenshot 2022-10-04 at 09 05 28

**Priority**
- Medium (slowing you down)

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- Developers
- General public

**Ideal deadline**

end of Q4

sandeepsajan0:
Fixes #2996 

- [x] Add archive function to batch action
- [x] Make submission tables archive aware, possibility to toggle between all and active (unrachived)
- [x] Optional permission so only staff admins can see archived submissions.


hypha/apply/activity/adapters/activity_feed.py
```diff
@@ -46,6 +46,7 @@ class ActivityAdapter(AdapterBase):
         MESSAGES.SKIPPED_REPORT: 'handle_skipped_report',
         MESSAGES.REPORT_FREQUENCY_CHANGED: 'handle_report_frequency',
         MESSAGES.BATCH_DELETE_SUBMISSION: 'handle_batch_delete_submission',
+        MESSAGES.BATCH_ARCHIVE_SUBMISSION: 'handle_batch_archive_submission',
```
theskumar, line 49: Hold onto merging this as https://github.com/HyphaApp/hypha/pull/2991 introduces changes in this.

sandeepsajan0:
/rebased with PR #2991. I think, now we may put it to the test.

frjo:
Merged #2991 so changed base back to main for this PR.

hypha/apply/funds/tables.py
```diff
@@ -182,30 +183,46 @@ def render_role_icon(self, value):
         return ''
 
 
-def get_used_rounds(request):
+def get_used_rounds(request=None, queryset=None):
+    if queryset:
```
frjo: The queryset seems to always have a value.

Also on larger datasets all the "if queryset" versions crash the server with a "Proxy Error".

frjo (changes requested):
We hit blocking performance issue on larger datasets with the filter changes.

We have forgotten to add a unarchive batch function (we can do this in a separate PR if easier.)

hypha/apply/funds/tables.py
sandeepsajan0: >The queryset seems to always have a value.

Actually, it is a bit tricky part, we are calling it from `__init__` and [updating the filter's field queryset](https://github.com/HyphaApp/hypha/pull/3003/files#diff-d8cf87355e041ebbd15a4725ce9b462fc6c1e3d752f8c1fee9aeecbd6a56cb98R317) to a value instead of a  callable that's why it is not getting called internally from forms and it always has a queryset. 
In the near future, if somebody removes any of the filter queryset update statement, it will get called internally and execute without queryset. That's why I put a check for queryset but if it is taking time, I think we can use separate methods for [initialization](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/tables.py#L257) and to update the filter's field queryset on the basis of the request's queryset.

sandeepsajan0: Also, queryset here has 2 reference, so it is a bit confusing. I'll change request's or view's 'queryset' to 'dataset' just for inside the table.

frjo (approved):

frjo:
This needs a rebase and migration number update.

sandeepsajan0:
Rebased. @frjo migration number looks fine to me. Is it conflicting with any migration or PR?

frjo:
Main branch has `hypha/apply/funds/migrations/0104_show_deadline.py`

And this PR has `hypha/apply/funds/migrations/0104_applicationsubmission_is_archive.py`

sandeepsajan0:
@frjo Migrations conflicts have been resolved.

frjo:
@theskumar @sandeepsajan0 There is som kind of issue with the migrations in this PR. I can replicate it with the sandbox db that is in the repo. But if I create a new and empty db I do not get this issue.

Any ideas what the problem is?

```
$ git switch feature/archive-submissions
$ git pull
$ dropdb sandbox
$ createdb sandbox
$ pg_restore --verbose --clean --if-exists --no-acl --no-owner -d sandbox public/sandbox_db.dump

DATABASES = {
    'default': dj_database_url.config(
        conn_max_age=600,
        default='postgres:///sandbox'
    )
}

$ python manage.py wagtailsiteupdate hypha.test apply.hypha.test 80
$ python manage.py migrate
```

Trying to access http://apply.hypha.test/dashboard/ gives me this error:

`cursor "_django_curs_4518872576_sync_5" does not exist`, the cursor id changes but same messages.

`In template /Users/frjo/Sites/hypha/hypha/apply/funds/templates/funds/includes/table_filter_and_search.html, error at line 91`

sandeepsajan0:
@frjo Latest changes seem to be working fine for me locally. Let me know if the `cursor error` is now fixed for you as well.

frjo:
@sandeepsajan0 Seems to be working for me as well. Excellent work! How did you manage to track down the issue?

sandeepsajan0:
It was just a combination of guesswork and trial & error with the changes. I also couldn't find an explanation yet why 'an extra distinct was causing the cursor issue'. That's why I wasn't sure and asked you if it is working for you as well. 

Whatever if it is fixed then we may move forward, I'll do a bit more research on it on weekend, and if I'll find something I'll post it here for the record. 
BTW it has something to do with server-side cursor https://docs.djangoproject.com/en/4.1/ref/databases/#server-side-cursors.

sandeepsajan0:
I found an alignment issue on the test, I think it happened because of css update for the 'show archive' button. I'll fix this and will update the PR.

image

PR #3024, opened 2022-10-28
Show reset link directly below password field. Only show Google login…

### LINKED ISSUE #3022: Increase usability of forget password link on login page
### User Story

As a user I would like to see the password update/forget password link in a more prominant position on the login page.

Screenshot 2022-10-27 10 33 38 AM

frjo:
… when configured.

Fixes #3022

### DIFF
--- a/hypha/apply/users/templates/users/login.html
+++ b/hypha/apply/users/templates/users/login.html
@@ -7,9 +7,7 @@
 
 {% block content %}
   <div class="wrapper wrapper--small">
-    {% if wizard.steps.current == 'auth' %}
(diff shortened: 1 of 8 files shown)


theskumar:
Overall looks good. But I noticed the codebase related to the functionality is all spread across different places. e.g., mixed b/w public and apply areas. This might be a problem down the line if we are trying to remove the public site. Just my 2 cents!

Screenshot 2022-11-01 at 10 52 49@2x

theskumar (approved):

frjo:
Good point on public/apply. That login and other pages on the apply site uses public css etc. will complicate things. This PR does not solve that, but does not make new problems either.

PR #3039, opened 2022-11-08
Add feature to download paf as docx file

### LINKED ISSUE #2912: PDF and text/doc output of PAF
## Additional context
This is a sub-issue of #2911  and output of #2904 Re add the PAF stream field form

## Description

Compliance needs to provide contractors and auditors with a contract and the SOW. The SOW contains part of the information in the PAF, and can be generated by the WebApp. 

Design option for Scope of Work PDF output

## Summary

A bundle of information could be exported as pdf files for external stakeholders as a batch action.

## Possible workflows

(Bestest) Option 0. Generated at the same time as the contract:
- Amended to the contract template information
- CA can create a contract with the SOW incorporated into the template (related to #8 )
- CA can export both contract and SOW templates into their document editor
- Contract and SOW will be stored in the system of record/relevant google drive folder

Option 1. The manual process could be something like:
- Finance user goes to Hypha and identifies the projects they need to export as a pdf output
- Finance user asks the Lead Dev for these projects to be exported
- Lead Dev does the export, which creates a number of pdf files
- Finance user downloads/receives pdf files from Lead Dev

Option 2. A more self-service process could be:
- Finance user goes to Hypha and identifies the projects they need to export as a pdf output
- Finance user goes to Hypha to highlight/select the projects 
- Finance user then "exports" the information bundle (contract and scope of work) and downloads the pdf files

## What is required?

0. Generated at the same time as the contract
- Validation that the contractor and auditors get the same SOW - needs OTF to answer
-design: include SOW creation into the contract creation process
-technical: to figure out how to generate the contract template at the same time as the SOW template

2. Minimum process
- design: the Scope of Work output file
- technical: @frjo to figure out how the pdfs get exported

3. Self-service process (images below)
- design: the process so the user can select the projects in Hypha and export the Scopes of Work
- design: the Scope of Work output file
- technical: @frjo and devteam figure out how the pdfs get exported

## Priority

## Ideal deadline
October 2022

sandeepsajan0:
Fixes #2912 

PDF format for PAF:
[PAF-form.pdf](https://github.com/HyphaApp/hypha/files/9999469/testing.2.riview.forms.7.pdf)

DOCX format for PAF:
[PAF-form.docx](https://github.com/HyphaApp/hypha/files/9999479/testing.2.riview.forms.29.docx)

I found one blank line on the top in DOCX format, I am looking if we can remove it. Otherwise, it looks good to me. Please have a look and let me know if we need to change something.


hypha/static_src/src/javascript/apply/simplified.js
```diff
@@ -0,0 +1,19 @@
+(function () {
```
theskumar: `simplified` doesn't should like a semantic name to me. I think it could use better naming.  


Also, for accessibility, should handle the focus with the keyboard as well, and at this point, I would highly recommend using a library for this instead of using custom-made UI. 

https://a11y-guidelines.orange.com/en/web/components-examples/dropdown-menu/#

hypha/apply/projects/views/project.py
```diff
@@ -717,23 +702,33 @@ def get(self, request, *args, **kwargs):
...
+        html = template.render(context)
+
+        buf = io.BytesIO()
+        document = Document()
+        new_parser = HtmlToDocx()
+        new_parser.add_html_to_document(html, document)
+        document.save(buf)
+
+        response = HttpResponse(buf.getvalue(), content_type='application/docx')
+        response['Content-Disposition'] = f'attachment; filename="{project.title}.docx"'
```
theskumar: The filename here should be url encoded, and it would be nice to have a `date` appended to it as well. See below: 

```
20220810-project-title.pdf
```

hypha/apply/projects/views/project.py
```diff
@@ -717,23 +702,33 @@ def get(self, request, *args, **kwargs):
...
-                reverse('apply:projects:document', kwargs={'pk': project.id, 'file_pk': packet_file.id})
-            )
-        return documents_dict
+
+@method_decorator(staff_or_finance_or_contracting_required, name='dispatch')
+class ProjectDetailDocxView(SingleObjectMixin, View):
+    model = Project
+
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
```
theskumar: I would recommend merging the two views related to downloading the pdf and docx together. 

Move the `get_paf_download_context` in the view func itself, it's logic in it is related to the view, so it doesn't need to be extracted out. 

Something on the lines below:

```python
class ProjectExportView(SingleObjectMixing...):
  model = Project

  def get(self, request, export_type):
     project = self.get_object()
     context = ...

     if export_type = "pdf":
        return render_as_pdf(template="template.html", filename=..., context)

     if export_type = "docx":
        context["show_footer"] = False
        return render_as_docx(template="template.html", filename=..., context)
```

theskumar (changes requested):

hypha/apply/projects/views/project.py
frjo: URL encoded is important to support other languages.

Adding date to the file name is a toss for me but I'm not against it.

frjo (changes requested):

hypha/static_src/src/javascript/apply/simplified.js
sandeepsajan0: >simplified doesn't should like a semantic name to me. I think it could use better naming.

Do you mean the file name?

 > I would highly recommend using a library for this instead of using custom-made UI.

I tried using bootstrap and it was working fine but it was changing other things as well like the fonts and all, the page was looking a bit weird/different from other pages. So, I moved forward with the custom UI. I think if we will move to a library we may need to update a few other things as well.

theskumar: Yes. name of the file and the CSS classes you are using. `simplified__dropbtn`. 

You can go with something like `dropdown.js` and HTML classes that start with `dropdown__`

Also, let's get into using a bulky library, but I would encourage you to look at the implementation details of https://blog.hubspot.com/website/html-dropdown

hypha/apply/projects/views/project.py
theskumar: The date is not a deal breaker for me. It is more to improve the user experience of the downloaded file. 

The prefixing of the filename with the date allows for:

1. If the project is exported multiple times, the user doesn't have a cluttered downloads folder. With the same name don't end up overwriting the same file or have file names like `project-xxx.pdf,` `project-xxx (1).pdf`, or `project-xxx (2).pdf`. This happens esp. when someone has auto-saved to a folder on the download.

3. The export content may vary with every export, so for users, it may make sense to have them label it when it's exported. The date provides an indicator around the time of the file without looking at the details of the files themselves. 

(of course, users can keep whatever name they want).

So I would say if there are no downsides to using a prefix, then let's use one to improve the experience. The date instead of `unix-timestamp` is a good compromise of the uniqueness of content as I would imagine the content on the project data would not change a lot daily.

hypha/static_src/src/javascript/apply/simplified.js
theskumar: Edit: checkout https://blog.hubspot.com/website/html-dropdown

hypha/apply/projects/views/project.py
sandeepsajan0: Done!

hypha/apply/projects/views/project.py
sandeepsajan0: Added the date as a prefix and slugify the filename.

hypha/static_src/src/javascript/apply/simplified.js
sandeepsajan0: Added `dropdown.js` and `dropdown.css` as separate new files for dropdown only.

frjo (approved):

theskumar (approved):

PR #3041, opened 2022-11-08
Testing GitHub actions for CI.

frjo:


frjo:
@sandeepsajan0 @theskumar Can you take a look at this error: https://github.com/HyphaApp/hypha/actions/runs/3420529967/jobs/5695470381#step:10:74

Thanks!

theskumar:
Will do.

theskumar:
https://github.com/jerry-git/pytest-split-gh-actions-demo/blob/master/.github/workflows/test.yml

Makefile
```diff
@@ -38,7 +38,7 @@ endif
 .PHONY: py-test
 py-test:
 	@echo "Running python tests"
-	pytest --reuse-db --cov --cov-report term:skip-covered
+	pytest --ds=hypha.settings.test --reuse-db --cov --cov-report term:skip-covered
```
theskumar: The `--ds=...` is picked from the `setup.cfg` https://github.com/HyphaApp/hypha/blob/main/setup.cfg#L23 by pytest; this need not be duplicated here. Can you double-check it?

.test_durations
```diff
@@ -0,0 +1,820 @@
+{
```
theskumar, line 1: Can we add this file to`.gitignore`?

sandeepsajan0:
I was looking into the test failure, I ran the test multiple times locally and got unexpected results every time. Sometimes it passes for all and sometimes gets different errors. 
I tried the same thing here as well, I rerun the 'Hypha CI / test-be (2)'  workflow and it passes this time without any changes. https://github.com/HyphaApp/hypha/actions/runs/3439525991/jobs/5747234058
Looks like it is happening because of the parallelism of the tests with processors, I am not sure about it yet. I am researching a bit more but it doesn't look like a code or test issue.

Update: We are also getting a warning for the same test that was failing last time `/opt/hostedtoolcache/Python/3.10.8/x64/lib/python3.10/copy.py:273: RuntimeWarning: Pickled model instance's Django version is not specified.` It might be something related to cache as well. I am looking into it and will update on it asap.

Makefile
frjo: Will remove it.

.test_durations
frjo: According to the documentation it looks like "pytest-spilt" uses this file to split up the test it the most efficient way.

https://jerry-git.github.io/pytest-split/

theskumar: Got it. 

1. Can we also use `--durations-path` to store in the `.github/` folder.? primarily to avoid cluttering the root folder. 

Looking at the docs, this doesn't indicate `pytest-split` as a replacement for running tests locally with `pytest-xdist`, or am I missing something here? It doesn't have a friendly CLI to split and run all the tests parallelly on the same machine. It seems pytest-split is more suited to CIs with multiple workers. That's why I was suggesting to keep the `.test_durations` file to kept in `.github` folder.

PR #3052, opened 2022-11-20
Remove the React app

### LINKED ISSUE #3074: Remove the React app from Hypha
The React app or "three column view" was built for an organisation that no longer use Hypha. No current implementors use it either, or even know it exist I believe.

No one in the current dev team is a React specialist and the code base is large so removing it simplifies Hypha quite a bit.

We can remove 300+ js files. Instead of 1500+ node packages we are left with "only" 421.

frjo:
Fixes #3074


sandeepsajan0 (approved):

theskumar:
I added some tweaks, and overall looks great! 

- optimize script exec using npm-run-all
- move babel config -> package.json

Things I need to look further. 
- Why run collectstatic as part of the watch?

theskumar (approved):

PR #3075, opened 2022-12-06
Add a separate DRAFT_SUBMISSION event

### LINKED ISSUE #3021: Remove slack alerts and notifications for DRAFT submissions

sandeepsajan0:
Fixes #3021 

NEW_SUBMISSION is getting used to notify staff members on slack and also to notify the applicant. But for draft submission, only the applicant should get the notification, there shouldn't be any recorder activity or slack notification. 
So, added a separate event DRAFT_SUBMISSION that only sends the notification/email to the applicant.


fourthletter:
I still see slack alerts for drafts 

image

fourthletter:
This patch was released last December 2022 but I still receive an alert on the live site

frjo:
Yes, it is not deployed on live yet, only on test.

fourthletter:
Thank you. I created a draft on test and it didn't send a notification. It works great!

PR #3078, opened 2022-12-12
Adding the new Hypha logo and favicons. Will be implemented later on.

frjo:


theskumar:
It would be great to have a version of the logo (colored) available in the SVG format here. 

PS: favicons now can be .svg files.

theskumar (approved):

PR #3045, opened 2022-11-14
Upgrade Wagtail to 4.2

theskumar:
Fixes #3018

- Update requirements - `wagtail` and `wagtail-cache`
- Update `BaseSetting` to `BaseSiteSettings`
- Fix Category StreamBlock at hypha/apply/categories/blocks.py
- Add migration changes
- Add `wagtail_reference_index_ignore = True` to models (See [this issue](https://github.com/wagtail/wagtail/issues/9731) on wagtail for more info)
- Use checkboxes when selecting reviewers for a Round and a Lab
- Update `ReadOnlyPanel` to use `render_html()` method instead of `render_as_object()`

**Screenshots**

Update to UIs
Screenshot 2022-11-14 at 08 46 32@2x

PR #3090, opened 2022-12-20
Fix test with setupTestData(), replace with setUp()

theskumar:
Replace setUpTestData with setUp

`setUpTestData` should not have items that don't support `deepcopy`. e.g. the database model items


theskumar:
This is a PR branched from the https://github.com/HyphaApp/hypha/pull/3045 PR to be able to get to master quickly.

PR #3109, opened 2023-01-09
Update local setup docs

theskumar:
- Remove dependency on python-venv and python-pip
- update how to run test

### DIFF
--- a/docs/developers/localdevelscratch.md
+++ b/docs/developers/localdevelscratch.md
@@ -5,11 +5,9 @@
 Make sure you have these things installed on your system:
 
 * Git
-* Python 3.9.x
-  * python3-venv \(to setup virtual enviroment\)
-  * python3-pip \(to install python packages\)
+* Python 3.10.x
 * PostgreSQL 12.x
-  * libpq-dev \(on Linux at least\)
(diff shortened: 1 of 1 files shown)


docs/developers/localdevelscratch.md
```diff
@@ -366,16 +358,16 @@ Now you should be able to access the sites on [http://hypha.test/](http://hypha.
...
 ```text
-$ python manage.py test --settings=hypha.settings.test
+$ make test
 ```
 
 If you need to rerun the tests several times this will speed them up considerably.
 
 ```text
-$ python manage.py test --parallel --keepdb --settings=hypha.settings.test
+$ py.test --reuse-db
```
frjo: Looks like `make test` runs `make py-test` and that command always use "--reuse-db".

frjo: And I think it is recommended to use `pytest` instead of the old `py.test`.

theskumar: I'll look at them, and update. :)

theskumar: Update the text to be more descriptive. Thanks for the feedback.

Issue #3080, opened 2022-11-21
Remove Public Site Documentation


fourthletter:
### Description 

Hypha's public site will no longer be maintained in the near future. Maintainers will remove documentation on the public site. Example: https://github.com/HyphaApp/hypha-docs/edit/main/docs/contributing/code/implementationdetails.md

theskumar:
I'll cover this as part of the #3110 PR.

PR #3110, opened 2023-01-10
Remove public site

### LINKED ISSUE #3428: Remove the public site from Hypha
The public site of Hypha is to be removed. It was a leftover from when this was a OTF project but no other organisations are using it.

There are some issue that need to be discussed.

Pages like the apply front page, the login page etc. are using the public site templates even do they are apply site pages. Mostly regarding header with nav links and the footer.

Footer is easy since the apply site pages have none. I think that works for the front page etc as well.

What to show in the header/nav?

Other issues we need to resolve?

- [x] https://github.com/HyphaApp/hypha/issues/3080
- [x] https://github.com/HyphaApp/hypha/pull/3432
- [x] Remove Google Translate Widget from Header Nav
- [x] https://github.com/HyphaApp/hypha/pull/3687
- [x] https://github.com/HyphaApp/hypha/pull/3689
- [x] #3694 
- [x] Update urlconf to only have single site and make apply site as default, remove all the code needed to separate out public and apply.

theskumar:
Fixes #3428

Removes the public site and make the application site the only site.

TODO: 
- [x] Documentation update #3080 
- [x] Keep cookiconcent
- [ ] Move (and rename) SystemMessagesSettings to apply (footer_content no longer needed)
- [x] Replace default OTF logo with Hypha logo.
- [x] Make all pages use "base-apply.html" as base. (Front page, login, apply form pages etc.)


theskumar:
@frjo The cookie consent is shared on the public site and apply site. Do we want to keep it for apply or remove it altogether?

frjo:
@theskumar Lets keep the cookie consent feature.

PR #3113, opened 2023-01-10
Upgrade to python 3.11, node 18 and Postgres 14.

### LINKED ISSUE #3112: Upgrade to Python 3.11.x

frjo:
Fixes: #3112


theskumar:
Dockerfile + docs would need to be updated as well.

frjo:
@theskumar We have 13 tests that fails with "RecursionError: maximum recursion depth exceeded". Same on GitHub as when I run locally.

A Goggle search gives some indication that this happens sometimes with python 3.11, maybe some extra checks added?

There is no urgency at all, but when you have time/are bored can you take a look at the errors? Thanks.

theskumar:
@frjo Fixed the issues with the recursion limit. The `setupTestData` classmethod uses recursive deep copy and tries to persist data across different tests, which might be helpful in some cases, it's an anti-pattern in testing, and we should avoid re-using data across tests. 

The `setUp` is run on every test and does a new clean data setup.

I also updated the node to use the current LTS version.

theskumar:
\rebased with main

frjo:
@theskumar I prefer we stick to core tools like npm and pip instead of nvm and pyenv.

People who likes them will know how to use them. For others I think it is better to keep it as simple as possible, there is a lot of stuff to understand in any case.

theskumar:
> @theskumar I prefer we stick to core tools like npm and pip instead of nvm and pyenv.
We can't compare `pyenv` vs `pip` and `nvm` vs `npm` , as even after using the `pyenv` someone would still have to use the `pip` and after using `nvm` you still need to use the `npm`. 

I believe the pyenv and nvm is pretty standard now a days to get the correction version of node and python for a specific project or other project. Do you have any other preference? Most of the devs would require different python/node versions.

I understand there could be concern using it for some, I can update the docs to so they are able to work without it and not make it required thing. Doc can say make sure that they install the right version of python and node, so they can skip it. 

That said, I would want to keep the `.python-version` and `.nvmrc` to remain there so it's able to automatically/complain when I'm not using the right version of python/node.

frjo:
Please keep the files and revise the documentation so pyenv and nvm are a recommendation with a short description of the benefits.

I just do not want to give the impression that  pyenv and nvm are requirements.

theskumar:
@frjo I have update the docs. Now it doesn't feel like pyenv and nvmrc are required.

PR #3102, opened 2023-01-01
Fixes regarding paf notifications 

sandeepsajan0:
Partial fix of #3066 (https://github.com/HyphaApp/hypha/issues/3066#issuecomment-1337463266)

- [x] We should rename "/apply/projects/4/simplified/" to something like "/apply/projects/4/approval/". Also rename "ProjectDetailSimplifiedView".
- [x] The link to the uploads contract in the "Contracts" box should open in a new tab, "target _blank"
- [x] When an applicant uploads a signed contract an e-mail is sent to the applicant but not to the "Compliance Email" address from project settings. I do not think we need the mail to the applicant but need to add one to compliance.
- [x] When a user has approved a PAF I think the PAF should be removed from that users dashboard.(Multiple tabs to show already approved and approval required).
- [x] E-mail should be sent out to 'contracting+approver' user's email address when all "PAF Reviewers Roles" have approved a PAF. The project should also show in a "Final approving requeried" table on the Dashboard for Contracting+Approver users.
- [x] Fix misspelling of "venor", should be "vendor"
- [x] Fix Permissions for contracts(hide buttons).
- [x] Do we need to separate "Compliance Email" setting to one address for finance users and one for contracting users? Or should we send directly to address is user accounts that has the role in question? (We decided email notifications could go to a 'group email' and default to individual emails if group email is not provided)

PR #3115, opened 2023-01-11
Use tabs in multiple tables on the same page

### LINKED ISSUE #3114: Tabs are not functioning properly for multiple tables on same page.
It can be a use case for the dashboard pages or overview pages where we show multiple tables and wants to have tabs in different tables on the same page.

Ex: I tried to update the Finance dashboard, we want to show the invoice table and project awaiting approval table both with tabs.
But when we load the dashboard page only one table is active by default or only one tab is active among all the tabs on the page.

Current behaviour:
image

Expected behaviour:
image

sandeepsajan0:
Fixes #3114

Add description here.

### DIFF
--- a/hypha/apply/dashboard/templates/dashboard/finance_dashboard.html
+++ b/hypha/apply/dashboard/templates/dashboard/finance_dashboard.html
@@ -17,19 +17,46 @@
     </div>
 </div>
 <div class="wrapper wrapper--large wrapper--inner-space-medium">
-    <div class="wrapper wrapper--bottom-space">
+    <div class="wrapper wrapper--bottom-space"
+        role="tablist" aria-label="Invoice Tabs"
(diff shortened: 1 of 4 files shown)


theskumar:
I used alpine to re-implement tabs; I've added some aria-* labels to indicate the purpose of these sections, but it's not fully keyboard ready. 

Alpine only works if served locally if we follow the es6 module system. I'll have to look into it, but the CDN version should be good for us.

You need to put in the real value in the PAF approved section.

sandeepsajan0:
Seems to be working perfectly. Thanks @theskumar.

>You need to put in the real value in the PAF approved section.

Sure, as it is a part of PR #3102 so I'll add the actual value and test it there as well.

sandeepsajan0:
Record: Merging it to the PR #3102.

PR #3118, opened 2023-01-13
Add npm commands to the Makefile.

frjo:


theskumar:
I would recommend having fewer rules or at least having some high-level rules that do things across the backend and frontend together.

- https://rosszurowski.com/log/2022/makefiles

PR #3121, opened 2023-01-17
Project workflow documentation

sandeepsajan0:
Documenting the project's workflow, features, and its implementation.


fourthletter:
On January 17 standup we discussed the final PAF approval will send a notifications/alerts to the Contracting + Approver user group. This include activity feed, slack, email, etc

PR #3129, opened 2023-01-24
Fix email digest - group the emails at Fund/Lab level

theskumar:
Initially, the email digest was grouped based on the email ids.

Now the emails are separated based on the fund or lab they are part of.

Also, include the name of the fund/lab in the email subject.


frjo (approved):

fourthletter:
I received an email digest for several funds with my email. Success! 💥 

image

PR #3132, opened 2023-01-30
Use system fonts for apply site

### LINKED ISSUE #3131: Replace custom fonts with system fonts

Right now, public and apply sites use custom web fonts - 'Nato-sans' and 'montserrat' which add an extra 1.3M download size to the page when loaded initially. It adds to the overall server cost and requires extra bandwidth for the end user. 

Moreover,  it causes re-rendering of the page and adds a flickering effect to the site. 

Replace custom fonts with the fonts available on the user's system. The wagtail site now also uses the `system-fonts`; this will allow for a consistent look across the site and also a familiar font experience. 

Popular websites use System Fonts, including:

- GitHub
- Ghost
- Bootstrap
- Booking.com

The new font stack will be, read more at https://flaviocopes.com/css-system-fonts/
```
-apple-system,BlinkMacSystemFont,Segoe UI,system-ui,Roboto,Helvetica
 Neue,Arial, sans-serif,Apple Color Emoji,Segoe UI Emoji,Segoe UI Symbol,Noto Color Emoji;
```

- Load the fonts over cdn, bandwidth reduction, but not much
- Use a JS-based font loader to avoid the flickering effect; display the content only after fonts are loaded.

**Additional context**

Screenshots from the mac-os rendering:

Screenshot 2023-01-30 at 07 21 08@2x
Screenshot 2023-01-30 at 07 21 32@2x

**Priority**

- Low priority (annoying, would be nice to see)

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- Developers
- General public

**Ideal deadline**

ASAP

theskumar:
NOTE: The font files and `npm -> build:fonts` scripts are still kept as 
they are used by reportlab for rendering PDFs. 

The system font is not removed from the public site to keep up with the OTF branding. 

If not using custom fonts, the pdfs are rendered with
Helvetica. Let’s discuss whether we need to use custom fonts for the PDFs.

Closes #3131


frjo (changes requested):
A good change I think.

Can we remove all "letter-spacing" as well? They worked well with the custom font but I think it better without with the system font.

sandeepsajan0 (approved):
The font looks good to me. Regarding pdf, I think only submission pdf is using the reportlab or the fonts. I am not sure if we are using the submission pdf feature. And if we need that, we can use the same approach as we used for the project pdf.

theskumar:
> The font looks good to me. Regarding pdf, I think only submission pdf is using the reportlab or the fonts. I am not sure if we are using the submission pdf feature. And if we need that, we can use the same approach as we used for the project pdf.

@frjo  Can you confirm if we still need to keep reportlab for generating PDFs? If it's not used, we should clean up the code. Can be taken as another PR. 

---

Update: 

- The public site keeps the old fonts, hopefully, it goes away along with the public site
- letter spacing updated for the apply site

This PR should be good to merge.

hypha/static_src/src/sass/apply/abstracts/_variables.scss
```diff
@@ -64,8 +64,9 @@ $color--correct: $color--mint;
 $color--button-disabled: $color--lighter-blue;
 
 // Fonts
-$font--primary: 'noto-sans';
-$font--headings: 'montserrat';
+$font--primary: -apple-system,BlinkMacSystemFont,Segoe UI,system-ui,Roboto,Helvetica Neue,Arial,
```
frjo: Should there not be quotes around values with spaces?

frjo (changes requested):

hypha/static_src/src/sass/apply/abstracts/_variables.scss
frjo: GitHub have this: `-apple-system,BlinkMacSystemFont,"Segoe UI","Noto Sans",Helvetica,Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji"`

theskumar: Now it says both without quotes and with quotes work for font names with spaces. 

But the spec recommends putting quotes around font names with space, so I'll add them here. Nice catch! 

More at https://stackoverflow.com/a/13752149/782901

frjo (approved):

PR #3155, opened 2023-02-06
Add a 'Update Project Status' button to change project status in between 'in-progress, closing and complete' state

### LINKED ISSUE #3128: Close/Complete a project manually
We have 'Closing' and 'Complete' statuses for a project, but we don't have any way to set these statuses manually for a project. 
Do we really need Closing and Complete as two separate statuses? We may just use them as final status just like Accepted/Declined for Submissions.
We can have a button in 'More Actions'(like submissions) to set the Closing state. And the Complete status button should be available for projects in 'IN Progress' status?

image

sandeepsajan0:
Fixes #3128 

Add description here.

### DIFF
--- a/hypha/apply/projects/forms/__init__.py
+++ b/hypha/apply/projects/forms/__init__.py
@@ -7,6 +7,7 @@ from .payment import (
 from .project import (
     ApproveContractForm,
     ChangePAFStatusForm,
+    ChangeProjectStatusForm,
     CreateProjectForm,
     FinalApprovalForm,
     ProjectApprovalForm,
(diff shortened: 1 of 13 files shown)


frjo:
@sandeepsajan0 Can you take a look at the test errors, thanks.

sandeepsajan0:
@frjo I have fixed the tests. These were failing because we restricted report generation only for IN PROGRESS projects, so just adding the project status worked. 
But I think we need to add a few new tests for CLOSING and COMPLETE status and also need to update a few other tests as well. So, hold on to it for now. I'll mention you again after updating the test.

PR #3156, opened 2023-02-06
Add alternate optional flow where users need to create account and login before applying

### LINKED ISSUE #3064: Add a seperate registration path without application

Allow registrations without applications.  Add a new workflow for that.  If account with that email address already exists, then just route the person through a regular login screen and then bring them back to the rest of the application.

Furthermore, add a configuration option that forces registrations prior to being allowed to create an application

It's an enhancement of the current login process.

When clicking the Apply button, and not currently logged in, continue as normal.  However, add a new button next to "Log In" at the top titled "Register" that allows account creation.  Then, when logged in, can apply as normal.

When configuration to force registration is set, clicking "Apply" will instead take the applicant to the log in screen, from which they can register if they haven't yet.

**Additional context**

This has been completed for a different client, although with the change that registration is always forced.  We would tweak the code slightly to allow the configuration in the case an admin would prefer to continue to allow registration through application.

As such, this feature request is not as much as a request that a feature be developed, but rather that the feature is an acceptable addition to hypha.  A PR would be forthcoming in that case.

**Priority**

It was a high priority for the end user, but it's not a high priority to be in core hypha.

**Affected roles**

- Applicants

frjo:
Fixes #3064


frankduncan:
Resolves #3064 

Seeks to add two features:
* Allowing signing up before submitting an application
* Requiring login before submitting an application

This is enabled by two variables:
* ENABLE_REGISTRATION_WITHOUT_APPLICATION for the former
* FORCE_LOGIN_FOR_APPLICATION for the latter

(As a side note, as this is my first PR of hopefully many from OTS's work with a client on enhancing hypha, any generic code and PR practices that I failed to follow would be greatly appreciated)

theskumar:
I'm off this week but will take this up next week.

hypha/apply/templates/forms/includes/field.html
```diff
@@ -35,26 +37,31 @@
...
+                <output class="form__file-list"></output>
+            {% endif %}
+            {% if widget_type == 'checkbox_input' %}
+                <label for="{{ field.id_for_label }}"></label>
+            {% endif %}
+            {% if widget_type == 'date_input' or widget_type == 'date_time_input' %}
+                </div>
+            {% endif %}
+        </div>
+    {%else%}
```
theskumar:
```suggestion
    {% else %}
```

hypha/apply/templates/forms/includes/field.html
```diff
@@ -35,26 +37,31 @@
...
+        {{field}}
+    {%endif%}
```
theskumar:
```suggestion
    {% endif %}
```

hypha/apply/templates/forms/includes/field.html
```diff
@@ -35,26 +37,31 @@
         <p class="form__help-link"><a href="{{ field.field.help_link }}" target="_blank" rel="noopener noreferrer">{% trans "See help guide for more information." %}<svg class="form__open-icon"><use xlink:href="#open-in-new-tab"></use></svg></a></p>
     {% endif %}
 
-    <div class="form__item">
-        {% if widget_type == 'date_input' or widget_type == 'date_time_input' %}
-            <div class="{{ widget_type }}">
-        {% endif %}
+    {% if not field.is_hidden%}
```
theskumar:
```suggestion
    {% if not field.is_hidden %}
```

hypha/apply/templates/forms/includes/field.html
```diff
@@ -35,26 +37,31 @@
...
+                <div class="{{ widget_type }}">
+            {% endif %}
 
-        {{ field }}
+            {{ field }}
+
+            {% if field.errors %}<h6 class="form__error-text">{{ field.errors.as_text|linebreaksbr }}</h6>{% endif %}
+            {% if widget_type == 'clearable_file_input' or widget_type == 'multi_file_input' %}
+                <!-- <div class="form__file-drop-zone">Drop files here</div> -->
+                <!-- <h1>{{ widget_type }}</h1> -->
```
theskumar: You can remove the comments if not needed.

hypha/apply/templates/forms/includes/field.html
```diff
@@ -19,12 +19,14 @@
                 {% endif %}
             </legend>
     {% else %}
-        <label for="{{ field.id_for_label }}" class="form__question form__question--{{ field_type }} {{ widget_type }}" {% if field.field.required %}required{% endif %}>
-            <span>{{ field.label }}</span>
-            {% if field.field.required %}
-                <span class="form__required">*</span>
-            {% endif %}
-        </label>
+        {% if not field.is_hidden%}
```
theskumar:
```suggestion
        {% if not field.is_hidden %}
```

hypha/apply/users/models.py
```diff
@@ -153,14 +153,20 @@ def get_or_create_and_notify(self, defaults: dict | None =None, site=None, **kwa
...
                 raise IntegrityError("Found an inactive account")
         else:
-            temp_pass = BaseUserManager().make_random_password(length=32)
-            temp_pass_hash = make_password(temp_pass)
+            # Coming from registration without application
+            if 'password' in kwargs:
+                temp_pass_hash = make_password(kwargs.pop('password'))
+            else:
+                temp_pass = BaseUserManager().make_random_password(length=32)
+                temp_pass_hash = make_password(temp_pass)
```
theskumar:
```suggestion
            if 'password' in kwargs:
                # Coming from registration without application
                temp_pass = kwargs.pop('password')
            else:
                temp_pass = BaseUserManager().make_random_password(length=32)
                
            temp_pass_hash = make_password(temp_pass)
```

This can be simplified.

hypha/apply/users/views.py
```diff
@@ -241,7 +278,7 @@ def get(self, request, *args, **kwargs):
         if self.valid(user, kwargs.get('token')):
             user.backend = settings.CUSTOM_AUTH_BACKEND
             login(request, user)
-            return redirect('users:activate_password')
+            return redirect('users:account')
```
theskumar: Curious why this change is needed?

theskumar (changes requested):
Had a quick looks at the code and it looks solid. 

I'll try it out and test it later next week. In mean time, I've left some suggestions for you to add. 

Also, would be great if you can update the `docs/` with the new configurations. Thanks.

frankduncan:
Thanks for the quick review!  Comments to come inline :)  Look forward to more next week.

hypha/apply/users/views.py
frankduncan: This change was based on the workflow that we adopted here, were three things are set:

1. We require login to make an application
2. You can register before making an application (forced by 1.)
3. Local passwords are used, and set during registration (as opposed to using 2FA)

What that means is that upon activating your account, you've already set your password.  So you don't need to do that again.  You should just get redirected to your account page.

I can understand how this change is actually bad when and this change got missed when cherry picking our old internal commits, so it should definitely be removed (since the default behavior should remain unchanged).  But maybe the correct answer is to do some kind of switch depending on whether the user set the password, or it was set to a random one by the system?

What do you think?

theskumar:
Thanks @frankduncan for making changes, I'll be able to review them tomorrow.

hypha/apply/users/views.py
theskumar: I think we can put this as conditional under the new setting introduced by you. So it retains the old behavior and updates only when login/register is required before submission. 

@frjo thoughts?

frankduncan: There's also the path where registration is enabled, but passwords aren't part of it (and are either set later or they use 2FA).  So maybe conditional on both of those, otherwise the user gets sent to the activation page.

So both `WAGTAIL_PASSWORD_MANAGEMENT_ENABLED` and `ENABLE_REGISTRATION_WITHOUT_APPLICATION`

theskumar: @frankduncan That sounds good to me. Let's have that in.

theskumar:
I tested the feature locally, and looks great to me! Will be waiting for https://github.com/HyphaApp/hypha/pull/3156#discussion_r1097364115 to get resolved.

hypha/apply/users/views.py
frankduncan: All set!  Though it turned out to be `WAGTAILUSERS_PASSWORD_ENABLED`, not `WAGTAIL_PASSWORD_MANAGEMENT_ENABLED` for the password fields to be available during registration

theskumar (approved):

frjo:
@frankduncan @theskumar  This works very well in my testing.

PR #3160, opened 2023-02-07
Fix that user_count was missing from groups/n/users view so it broke.

### LINKED ISSUE #3163: View users in this group page only show a 500 error
The page that shows users in a group is broken. The error is "'count' argument to 'blocktrans' tag must be a number."

`{% blocktrans trimmed count count=user_count %}` line in file "apply/users/templates/wagtailusers/users/results.html"

The `user_count` variable is not getting set on the "View users in this group" page.

To access this page go to:

1. Wagtail admin -> Settings -> Groups
2. Click on a group, e.g. "Staff"
3. Click on the green "View users in this group" button in the top right.

frjo:
Fixes #3163

### DIFF
--- a/hypha/apply/users/templates/wagtailusers/users/results.html
+++ b/hypha/apply/users/templates/wagtailusers/users/results.html
@@ -3,11 +3,13 @@
     <div class="users-list__results">
         {% if users %}
             <h2 role="alert">
-            {% blocktrans trimmed count counter=users|length %}
-                There is {{ counter }} user
-            {% plural %}
-                There are {{ counter }} users
-            {% endblocktrans %}
+            {% if user_count %}
+                {% blocktrans trimmed count counter=user_count %}
+                    There is {{ counter }} user
+                {% plural %}
+                    There are {{ counter }} users
+                {% endblocktrans %}
+            {% endif %}
             </h2>
 
             {% if is_searching %}
--- a/hypha/apply/users/templates/wagtailusers/users/results.html
+++ b/hypha/apply/users/templates/wagtailusers/users/results.html
@@ -3,10 +3,10 @@
     <div class="users-list__results">
         {% if users %}
             <h2 role="alert">
-            {% blocktrans trimmed count count=user_count %}
-                There is {{ user_count }} user
+            {% blocktrans trimmed count counter=users|length %}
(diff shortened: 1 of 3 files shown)


hypha/apply/users/templates/wagtailusers/users/results.html
```diff
@@ -3,10 +3,10 @@
     <div class="users-list__results">
         {% if users %}
             <h2 role="alert">
-            {% blocktrans trimmed count count=user_count %}
-                There is {{ user_count }} user
+            {% blocktrans trimmed count counter=users|length %}
```
sandeepsajan0: I think it will show the number of users present on the page not the total number of users.

sandeepsajan0:
@frjo I couldn't figure out the issue, can you please share a screenshot or something where it is breaking? Because on the admin's users' page, it seems to be showing the total number of users to me locally. 
I checked this PR locally, it is showing the number of users present on the page only.

image

hypha/apply/users/templates/wagtailusers/users/results.html
frjo: You are right the fix breaks on normal user list, thanks for catching that.

frjo:
Added an issue explaining the problem now. It is on the "View users in this group" page. Go to groups settings in wagtail admin, click on the Staff group e.g. and then the "View users in this group" button.

frjo:
@sandeepsajan0  It seems "user_count" is not getting set. If you know how to fix that please take over the issue/pr.

sandeepsajan0:
@frjo  thanks for explaining the issue, I'll look into this.

sandeepsajan0:
@frjo Groups are using the `wagtailusers_user` view for showing 'groups/.../users/' that's why it broke. I think we tried to fix it in past by adding a group admin url [here](https://github.com/HyphaApp/hypha/pull/3160/files#diff-39d2d967cd6ae6c055a5bec028703c181c18440e7ca55ef40d1fdc6361d0ebb1L13), but it wasn't working so I override a few things and make it use our custom one(`index`).
I think we can put it to test. 
And I think as wagtail 4 is in progress so we might need to change a few things in that as well. cc.. @theskumar

PR #3162, opened 2023-02-08
Move report permissions to permission file

### LINKED ISSUE #3159: Move Reports permissions to permissions.py file
**Description**:

Reports permissions are at multiple places right now and it is different at every place for the same view.

[ReportAccessMixin](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/projects/views/report.py#L34) allows staff members, finance members, and project owner.

[Dispatch method](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/projects/views/report.py#L64) of report update, details, and other views checks for the report status, timeline, and other report-related things.(Its fine for now but better to move to permissions.py)

[Report templates](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/projects/templates/application_projects/includes/reports.html#L60) are checking for apply staff at multiple places.

As in templates we are showing the report edit, skip, and unskip options only to staff members while in ReportAccessMixin finance and project owner also have the permissions for those views, so we need a little clarification on these permissions. 

**Question**:
Who(user roles) can view the reports?
Who can add/edit/update the reports?

**Proposed solution:**:
Let's keep the permissions in permission.py and use the same in templates and in the views, so that if anyone can't see the buttons can't access the view anyhow and vice-versa.

sandeepsajan0:
Fixes #3159 

Permissions to test:

When Project is in IN PROGRESS state, these actions should be restricted to 'CLOSING' and 'COMPLETE' states.

- [ ] staff can update the report config 
- [ ] applicants/staff can update/add reports

Similarly, when project is in IN PROGRESS, CLOSING, and COMPLETE state.

- [ ] applicants/staff/finance can see 'view' the reports

### DIFF
--- a/hypha/apply/projects/views/report.py
+++ b/hypha/apply/projects/views/report.py
@@ -1,6 +1,5 @@
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import UserPassesTestMixin
-from django.core.exceptions import PermissionDenied
 from django.shortcuts import get_object_or_404, redirect
 from django.utils.decorators import method_decorator
 from django.views import View
@@ -55,8 +54,7 @@ class ReportDetailView(DetailView):
     def dispatch(self, *args, **kwargs):
         report = self.get_object()
         permission, _ = has_permission('report_view', self.request.user, object=report, raise_exception=True)
-        if permission:
-            return super().dispatch(*args, **kwargs)
+        return super().dispatch(*args, **kwargs)
 
 
 @method_decorator(login_required, name='dispatch')
@@ -67,8 +65,7 @@ class ReportUpdateView(UpdateView):
     def dispatch(self, *args, **kwargs):
         report = self.get_object()
         permission, _ = has_permission('report_update', self.request.user, object=report, raise_exception=True)
-        if permission:
-            return super().dispatch(*args, **kwargs)
+        return super().dispatch(*args, **kwargs)
 
     def get_initial(self):
         if self.object.draft:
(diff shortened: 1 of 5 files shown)


hypha/apply/projects/views/report.py
```diff
@@ -47,27 +48,27 @@ def test_func(self):
...
+class ReportDetailView(DetailView):
     model = Report
 
     def dispatch(self, *args, **kwargs):
         report = self.get_object()
-        if not report.current or report.skipped:
-            raise PermissionDenied
-        return super().dispatch(*args, **kwargs)
+        permission, _ = has_permission('report_view', self.request.user, object=report, raise_exception=True)
+        if permission:
```
theskumar: If you use `raise_exception=True`, then the code below will execute only if it passes the test. There is no need for if permission.

hypha/apply/projects/views/report.py
```diff
@@ -47,27 +48,27 @@ def test_func(self):
...
     model = Report
 
     def dispatch(self, *args, **kwargs):
         report = self.get_object()
-        if not report.can_submit:
-            raise PermissionDenied
-        if report.current and self.request.user.is_applicant:
-            raise PermissionDenied
-        return super().dispatch(*args, **kwargs)
+        permission, _ = has_permission('report_update', self.request.user, object=report, raise_exception=True)
```
theskumar, line 67: If you use `raise_exception=True`, then the code below will execute only if it passes the test. There is no need for if permission.

theskumar (changes requested):

PR #3220, opened 2023-02-16
Fix Permissions for Project access(detail view)

### LINKED ISSUE #3205: Contracting should not have access to the project after the contract is approved.
### Description
Contracting should not have access to the project after the contract is approved. Their role in the process is complete

### LINKED ISSUE #3216: The AdminProjectDetailView should show PermissionDenied for unauthorised users

sandeepsajan0:
Fixes #3205 
Fixes #3216 

1. Staff can access the project in every status.
2. Contracting can access projects only in `WAITING_FOR_APPROVAL` and `CONTRACTING` statuses.
3. Finance can access projects only in the `WAITING_FOR_APPROVAL` and `IN_PROGRESS` statuses.
4. Applicant(only project user) can access the project in every status to keep track.
5. Unauthorised users and any other role can not access the project.

### DIFF
--- a/hypha/apply/projects/permissions.py
+++ b/hypha/apply/projects/permissions.py
@@ -153,6 +153,25 @@ def can_view_report(user, report, **kwargs):
     return False, 'Forbidden Error'
 
 
+def can_access_project(user, project):
+    if not user.is_authenticated:
+        return False, 'Login Required'
+
+    if user.is_contracting and project.status in [CONTRACTING, WAITING_FOR_APPROVAL]:
+        return True, 'Contracting can view project only in Waiting for approval and contracting status'
+
+    if user.is_finance and project.status in [WAITING_FOR_APPROVAL, IN_PROGRESS]:
+        return True, 'Finance can view project only  in Waiting for approval and in progress status'
+
+    if user.is_apply_staff:
+        return True, 'Staff can view project in all statuses'
+
+    if user.is_applicant and user == project.user:
+        return True, 'Applicant(project user) can view project in all statuses'
+
+    return False, 'Forbidden Error'
+
+
 permissions_map = {
     'contract_approve': can_approve_contract,
     'contract_upload': can_upload_contract,
@@ -163,4 +182,5 @@ permissions_map = {
     'report_config_update': can_update_report_config,
     'report_view': can_view_report,
     'submit_contract_documents': can_submit_contract_documents,
+    'project_access': can_access_project,
 }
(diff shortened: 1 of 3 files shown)


sandeepsajan0:
@frjo I'll fix the tests after we decide on the permissions. If the above permissions look good to everyone then I can move forward with the tests.  cc_ @fourthletter

hypha/apply/projects/permissions.py
```diff
@@ -39,7 +39,27 @@ def can_upload_contract(user, project):
...
+    if user.is_applicant and user == project.user:
+        return True, 'Applicant(project user) can view project in all statuses'
+
+    return False, 'Forbidden Error'
+
+
 permissions_map = {
     'contract_approve': can_approve_contract,
     'contract_upload': can_upload_contract,
+    'project_access': can_access_project,
```
theskumar, line 185: It would be better to have it called `project_view,` as access it very generic.

theskumar: Also, it would be better to have these strings in `CAPITAL_CASE` as these are more like constants.

sandeepsajan0: I was also thinking to use `project_view` but that ProjectDetailView contains multiple forms(actions/buttons) as well so it is like 'view project + access to those actions'. And if we want to have those forms' view permission separately then it will be just many permission but yes at the same time, it might give us more flexibility over every form/button.

These permissions are not confirmed yet so let's wait for the confirmation, we can work on it after that.

theskumar:
If you looking to add permissions for the lower level permissions and they are configurable i would to have separate permissions for them.
  
​

hypha/apply/projects/permissions.py
theskumar: If you looking to add permissions for the lower-level permissions, and they are configurable, I would like to have separate permissions for them.

hypha/apply/projects/views/project.py
```diff
@@ -665,6 +665,11 @@ class AdminProjectDetailView(
     model = Project
     template_name_suffix = '_admin_detail'
 
+    def dispatch(self, *args, **kwargs):
+        project = self.get_object()
+        permission, _ = has_permission('project_access', self.request.user, object=project, raise_exception=True)
```
frjo, line 670: @sandeepsajan0 Where does this permission get checked, just trying to understand how it works?

sandeepsajan0: @frjo It is getting checked inside `has_permission` method [here](https://github.com/HyphaApp/hypha/pull/3220/files#diff-88c40103819f7fef46a271d3a16f1bf13c7a88a0aae67f6c1ddc613d44435e08R17). Actually, it depends on the argument `raise_exception`, as we passed it as True so if method return Flase(no permission) then `has_permission` will raise PermissionDenied on its own.

PR #3226, opened 2023-02-17
Add htmx and hx-boost for header navs

theskumar:
- Add htmx integration
- Add hx-boost for header nav; it degrades gracefully! 

Configure https://github.com/rstacruz/nprogress to show progress on ajax activities.

Note: django-htmx is not used right now but will be used to add more htmx-based functionalities.

PR #3227, opened 2023-02-17
Update activity notification dropdown to use aplinejs

theskumar:
Also, update the look of the dropdown.

Screenshot 2023-02-17 at 09 39 51


hypha/templates/base-apply.html
```diff
@@ -31,9 +31,7 @@
         <script src="{% static 'js/js.cookie.min.js' %}"></script>
         <script src="{% static 'js/main-top.js' %}"></script>
         <script defer src="https://unpkg.com/alpinejs@3.x.x/dist/cdn.min.js"></script>
```
frjo: Can we include alpinejs as a local file instead of using a CDN?

theskumar: I had tried to include it locally. It doesn't work, as it downloads a few other files and has something to do with how the library is loaded and initialized. I've to research it; we can create a separate ticket to track it. 

PS: This PR doesn't add alpine, it was added before so it's better we solve this in another PR.

theskumar: Screenshot 2023-02-17 at 13 17 13@2x

theskumar: I tried to use the [content of the js](https://unpkg.com/alpinejs@3.11.1/dist/cdn.min.js) file locally, but it didn't work. Do you know if we can do module-based imports for javascript? https://alpinejs.dev/essentials/installation#as-a-module

theskumar: Instead, I've frozen the version of alpine to 3.11.x instead of 3.x. (in https://github.com/HyphaApp/hypha/pull/3227/commits/ffe855ad5a9c4ada3fe692e1f038a40643df64c5)

frjo: Did some testing and we have an outdated "browserslist" that includes dead browsers like IE making babel not use modern presets. This break local alpinejs versions.

Updating to this (babel default I believe) fixes the issue.

~~~~
    "browserslist": [
        "> 0.5%",
        "last 2 versions",
        "Firefox ESR",
        "not dead"
    ],
~~~~

We should update to this in any case.

frjo (changes requested):
When reloading the page the notification box "flash". It does not do that in the current version.

theskumar:
> When reloading the page the notification box "flash". It does not do that in the current version.

The flashing issue was handled by another PR. https://github.com/HyphaApp/hypha/pull/3226/files#diff-991b4d1d98770e27e8db821a366e0c49d5ff3659e38677f7fc87f69f4326bfefR41; I was hoping the htmx PR would merge first! :D 

I have moved the `x-cloak` CSS to this PR now.

hypha/templates/base-apply.html
theskumar: Good catch! The updated browserslist lets alpine work! I've updated the codebase with the local version of alpine and the new browserlist.

theskumar: Good catch! The updated browserslist lets alpine work! I've updated the codebase with the local version of alpine and the new browserlist.

frjo (approved):

PR #3252, opened 2023-02-28
Fix user objects not linked in slack and email messages

### LINKED ISSUE #3172: Slack notification when adding/editing account show <None|Jane Doe>
I guess this is when the user account do not have a Slack name but the code should check for that.

`@joe Joe has edited account for <None|Jane Doe> that now has following roles: Staff, Administrator`

https://github.com/HyphaApp/hypha/blob/921acfd0a89c3b339d1ca23217f0d895e5b98033/hypha/apply/activity/adapters/slack.py#L70-L71

theskumar:
Fixes #3172

The messaging module use model's `get_absolute_url` method
to determine and link a model object to its page.

The user model is linked in the slack messagge template but
didn't have `get_absolute_url()`.

### DIFF
--- a/hypha/apply/activity/tests/test_messaging.py
+++ b/hypha/apply/activity/tests/test_messaging.py
@@ -302,12 +302,12 @@ class TestActivityAdapter(TestCase):
         self.assertIn(submission.phase.display_name, message)
         self.assertIn(old_phase.display_name, message)
 
-    def test_lead_not_saved_on_activity(self):
+    def test_lead_saved_on_activity(self):
         submission = ApplicationSubmissionFactory()
         user = UserFactory()
         self.adapter.send_message('a message', user=user, source=submission, sources=[], related=user)
         activity = Activity.objects.first()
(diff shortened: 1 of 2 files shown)


frjo (approved):

frjo:
@theskumar  The test "test_lead_not_saved_on_activity" fails, did a test rerun but the same test fails so something changed.

theskumar:
I will take a look. Ideally, things shouldn't have changed or been connected.

theskumar:
Found the issue in the here

https://github.com/HyphaApp/hypha/blob/0df974b2fd05a9bad73a048ebb5aee89629974e3/hypha/apply/activity/adapters/activity_feed.py#L208-L216

Based on the name of the test... `test_lead_not_saved_on_activity`, this doesn't correlate why the activity feed should not store the `related` object if it's provided.

I propose to update the test, so that it saves the `related=user` in the `Activity` table when passed to the `send_message`.

PR #3257, opened 2023-03-01
Improve number input field validation

### LINKED ISSUE #3194: Front end validation of number fields etc.
### Description

While entering the amount in the approval form users tend to use alphabets but it only accepts numerical value. 

### Proposed Solution
Number field on form does accept any alphabetical value since right now it takes one there could also be a tip that pops if uses enter alphabets OR help text OR clearly indicate this in the question on the PAF.

image

### LINKED ISSUE #3213: Add placeholder to all input fields, used for simple front end validation css
HTML5 input fields gives us built in validation in browsers. For it to work nicly we need an empty placholder attribute on all input fields.

Then we can add CSS like below to directly show users when input is valid/invalid.

```
    input {
        &:not(:placeholder-shown):valid {
            background-color: lighten($color--correct, 40%);
        }

        &:not(:placeholder-shown):invalid {
            background-color: lighten($color--error, 30%);
        }
    }

```

theskumar:
Fixes #3213 
Fixes #3194 

- Allow the user to enter any input
- Display instant feedback if the input is not a valid number.

It uses alpine and `isNaN` javascript function for validation
Also, move alpine to be included in both public and apply site; forms right now doesn't use `base_apply.html`

Checkout the video below for it's look and feel.

https://user-images.githubusercontent.com/236356/222098169-a0a57d65-eb8f-4851-9b11-29f1860317c6.mp4

### DIFF
--- a/hypha/templates/base-apply.html
+++ b/hypha/templates/base-apply.html
@@ -42,10 +42,10 @@
         <script src="{% static 'js/js.cookie.min.js' %}"></script>
         <script src="{% static 'js/main-top.js' %}"></script>
 
-        <!-- alphine js start-->
+        <!-- alpine js start -->
         <script defer src="{% static 'js/apply/vendor/alpine-3.11.1.min.js' %}"></script>
         <style> [x-cloak] {display: none !important} </style>
-        <!-- apline.js ends-->
+        <!-- alpine.js end -->
 
         {% if COOKIES_ACCEPTED and MATOMO_URL and MATOMO_SITEID %}
         {# we are only expecting strings, so make sure we escape the values #}
--- a/hypha/static_src/src/sass/apply/components/_form.scss
+++ b/hypha/static_src/src/sass/apply/components/_form.scss
@@ -407,6 +407,11 @@
             color: $color--black-50;
             cursor: not-allowed;
         }
+
+        &.invalid {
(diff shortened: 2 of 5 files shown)


hypha/templates/base-apply.html
```diff
@@ -42,11 +42,6 @@
         <script src="{% static 'js/js.cookie.min.js' %}"></script>
         <script src="{% static 'js/main-top.js' %}"></script>
 
-        <!-- alphine js start-->
-        <script defer src="{% static 'js/apply/vendor/alpine-3.11.1.min.js' %}"></script>
-        <style> [x-cloak] {display: none !important} </style>
-        <!-- apline.js ends-->
-
```
frjo, line 49: This should not be removed I think?

frjo (approved):

frjo (changes requested):
Alipinejs got removed in the base-apply template.

hypha/templates/base-apply.html
theskumar: You're right! not sure why I thought `base_apply.html` was inherited from the `base.html`.

theskumar: Updated.

frjo (approved):

PR #3262, opened 2023-03-01
Email on admin account creation

frankduncan:
Fixes #6253

Users receive an email when their accounts are created as part of the application process but NOT when their account is created by an administrator through the administrative interface.

This adds logic to send the activation email after an account has been created

### DIFF
--- a/hypha/apply/users/wagtail_hooks.py
+++ b/hypha/apply/users/wagtail_hooks.py
@@ -1,9 +1,11 @@
 from django.urls import re_path
 from wagtail import hooks
+from wagtail.core.models import Site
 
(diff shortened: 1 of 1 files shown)


frjo:
@frankduncan Thanks! Looks neat and clean. Have you tested how it works with the default workflow where accounts are auto created? Se code below, line 164. As long as we never send out double e-mails to new users I'm all for merging this.

https://github.com/HyphaApp/hypha/blob/4d689e39d0ae58b252b20b5f608ca55a6fc9fe92/hypha/apply/users/models.py#L126-L171

frankduncan:
@frjo Tested and confirmed only one activation email in logs when an application is created.

PR #3263, opened 2023-03-01
fix: allow editing review with old form

### LINKED ISSUE #3254: 500 error when editing a review that's had its form changed
**Describe the bug**

After submitting a review, if the review form is edited removing a score, a 500 error is raised when editing the review.

**To Reproduce**

1. Create a review involving multiple scores
2. Submit the review
3. Edit the review form, removing one of the scores
4. Submit the edit for that review

**Expected behavior**

That the edit review is successful

**Priority**

It was High for our client, as they had a situation where they ran into this and was preventing them from moving forward, but it's fixed on our branch.

**Affected roles**

- Reviewers

**Additional context**

We have a fix for this case, and a PR will be incoming :)

frankduncan:
Closes #3254

Checks for missing form fields, none types in review editing

Further note:

This only fixes the 500 error as it appears.  When doing some testing on it, I noticed that the following also failed when working with a review that had its scores deleted, and then was edited:

* View All reviews on the submission leads to a 500 error
* Viewing the review itself seems to be missing some information (like the new score if I added one)

Because this is issue/PR is us trying to move changes we made to our fork into upstream to align the two codebases, I am submitting it as is.  For the other issues, would it be better to hold off on this fix until we have time to address some of the other things that pop up, or just create issues for them?  I lean toward merging and creating issues just because this PR does remove a 500 error and iterates toward a bug free system.

### DIFF
--- a/hypha/apply/review/forms.py
+++ b/hypha/apply/review/forms.py
@@ -64,7 +64,8 @@ class ReviewModelForm(StreamBaseForm, forms.ModelForm, metaclass=MixedMetaClass)
 
     def save(self, commit=True):
         self.instance.score = self.calculate_score(self.cleaned_data)
-        self.instance.recommendation = int(self.cleaned_data[self.instance.recommendation_field.id])
+        if self.instance.recommendation_field.id in self.cleaned_data:
+            self.instance.recommendation = int(self.cleaned_data[self.instance.recommendation_field.id])
         self.instance.is_draft = self.draft_button_name in self.data
         # Old review forms do not have the requred visability field.
         # This will set visibility to PRIVATE by default.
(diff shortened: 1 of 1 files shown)


frjo:
I can replicate the error. But I do not think this PR is the right solution.

1. Create a review form with 5 fields.
2. Write one review.
3. Modify the review form on the round/lab by removing 2 of the field.

If one tries to edit and save the existing review in current Hypha you get a 500 error, bad.

Editing and saving the review with this PR will delete the removed fields form the existing Review. This is really bad. Content is sacred, Hypha can not delete it without explicit action from users.

frjo:
Further testing revels that Hypha already deletes fields from existing reviews when they are edited if the fields have been removed from the review from on the lab/round. This is a bad bug.

I have confirmed that submissions works as they should nothing gets deleted. 

@theskumar Can you take a look at this? We safe the form for submissions, reviews etc in to "form_fields" so the form always match the "form_data". This works as expected with submissions but not with reviews. 

We should check determinations as well, and Project approval forms since they all are stream field forms and should all work the same.

Tests should be added to make sure the froms work as expected.

theskumar:
@frankduncan Please check if this patch works for you. 

- https://github.com/HyphaApp/hypha/pull/3276

@frjo I believe this PR won't be needed if #3276 works fine, which uses the `form_fields` saved along with the review instead of pulling it from the rounds/labs.

PR #3275, opened 2023-03-09
Add Project SOW Form

### LINKED ISSUE #3209: Add an SOW option
After the PM has filled out the approval form, OTF wants a 'View SOW' option along with 'View Approval Form' in the table. This SOW will essentially just show a part of the entries filled in the SOW form as per [this doc](https://docs.google.com/document/d/1RIV6DDrbq9q7vMVHHUe9xa9nV1RjYsI0/edit).

image

sandeepsajan0:
Fixes #3209 

Add SOW Form Model, add it to wagtail admin, avail for funds and labs


hypha/apply/stream_forms/models.py
```diff
@@ -123,6 +123,7 @@ def get_form_fields(self, draft=False, form_data=None, user=None):
         return form_fields
 
     def get_form_class(self, draft=False, form_data=None, user=None):
+        print("jkjkjkjjk", form_data, draft, user)
```
theskumar: Remove print.

theskumar (changes requested):

hypha/apply/funds/models/forms.py
```diff
@@ -211,9 +211,46 @@ def __str__(self):
         return self.form.name
 
 
+class AbstractRelatedProjectSOWForm(Orderable):
```
theskumar, line 214: Can you please add a docstring on why this Class is needed. IT's not very clear from it's implimentaton.

hypha/apply/funds/models/forms.py
```diff
@@ -211,9 +211,46 @@ def __str__(self):
...
+    class Meta(Orderable.Meta):
+        abstract = True
+
+    form = models.ForeignKey('application_projects.ProjectSOWForm', on_delete=models.PROTECT)
+
+    @property
+    def fields(self):
```
theskumar, line 222: While abstracting a foreign key as a class property, make it feel like an non-expensive access, while in reality it's not. 

I would suggest converting it to `@cached_property` or just not convert it to a property at all.

frjo:
@sandeepsajan0 please take a look at @theskumar comments before we merge in.

hypha/apply/funds/models/forms.py
sandeepsajan0: I have added a docstring there, it is just getting copied from the above forms. Its main function is to attach the SOW forms to the funds and labs.

hypha/apply/funds/models/forms.py
sandeepsajan0: It is the same for other forms, I just copied it from there to keep things similar. Can we work on it in different PR as an improvement for all the forms?

sandeepsajan0:
@frjo It requires some test updates. Let me resolve those three issues also in this same PR and then we move all SOW related things to the main.

sandeepsajan0:
@frjo I think it is ready, we have another PR(in progress, UI related changes) https://github.com/HyphaApp/hypha/pull/3291 on top of this PR.

PR #3276, opened 2023-03-10
Fix editing of review after review form is updated

### LINKED ISSUE #3254: 500 error when editing a review that's had its form changed

theskumar:
While editing a form, do not use the form assigned to the 
Round/lab, instead, use the form used while writing the review
for the first time

This PR also tweaks the look of the view review page; see the screenshot below.

Screenshot 2023-03-10 at 11 05 34@2x

Fixes #3254 
Closes #3263

### DIFF
--- a/hypha/apply/review/models.py
+++ b/hypha/apply/review/models.py
@@ -204,7 +204,7 @@ class Review(ReviewFormFieldsMixin, BaseStreamForm, AccessFormData, models.Model
 
     @cached_property
     def is_updated(self):
-        # Only compear dates, not time.
+        # Only compare dates, not time.
         return self.created_at.date() < self.updated_at.date()
 
 
--- a/hypha/apply/review/templates/review/review_detail.html
+++ b/hypha/apply/review/templates/review/review_detail.html
@@ -14,7 +14,7 @@
     </div>
 </div>
 
-<div class="grid">
+<div style="display: flex; gap: 2rem; padding-top: 2rem;">
     <div>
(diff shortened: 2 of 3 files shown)


frankduncan:
I think the use case that they ran into was that they did some reviews, and then they realized that some things needed to be changed in the forms.  They then wanted reviewers to go back in and update their reviews with the new specifications (otherwise, how would they have run into this issue).

While this fixes the 500 error, and appears to do so in line with the "delete no data" philosophy, we may get a new request along the lines of "allow reviewers to update their reviews after the review form has changed" such that when looking at the reviews in aggregate, they look comparable.  But I'm not 100% on this, and it may be good enough for the table of all reviews to clearly show old and new.  Reviewers also always have the ability to delete their review and submit a new one, so I'll probably just recommend that if it comes up.

At any rate, this does fix the issue in my test cases!

frjo:
The "delete no data" part is non-negotiable. I'm really grateful that you found this bug so we could get it fixed!

frjo (approved):

PR #3279, opened 2023-03-15
Add sentry guide + sentry performance monitoring setup

theskumar:
- Add a guide on how to enable and setup sentry
- Upgrades sentry-sdk to 1.16.0
- Add optional performance monitoring
- Add Browser Javascript Tracking & Tracing

### DIFF
--- a/hypha/templates/base.html
+++ b/hypha/templates/base.html
@@ -211,5 +211,25 @@
         <script src="{% static 'js/main.js' %}"></script>
         <script src="{% static 'js/cookieconsent.js' %}"></script>
         {% block extra_js %}{% endblock %}
+
+        {% block sentry_sdk %}
+        {% if SENTRY_PUBLIC_KEY %}
+        <script
+            src="https://js.sentry-cdn.com/{{ SENTRY_PUBLIC_KEY }}.min.js"
+            crossorigin="anonymous"
+        ></script>
+        <script>
+            Sentry.onLoad(function() {
+                Sentry.init({
+                    environment: "{{ SENTRY_ENVIRONMENT }}",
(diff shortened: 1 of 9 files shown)


hypha/templates/base.html
```diff
@@ -211,5 +211,25 @@ <h4 class="heading heading--no-margin">@{{ settings.utils.SocialMediaSettings.tw
...
+                    environment: "{{ SENTRY_ENVIRONMENT }}",
+                    debug: {% if SENTRY_DEBUG %}true{% else %}false{% endif %},
+                    integrations: [new BrowserTracing()],
+                    denyUrls: {{ SENTRY_DENY_URLS|safe }},
+                    tracesSampleRate: {{ SENTRY_TRACES_SAMPLE_RATE }}
+                });
+             });
+        </script>
+        {% endif %}
+        {% endblock sentry_sdk %}
```
frjo, line 233: Is this not also needed in `base-apply.html`?

Should you have the block name in the end tag?

theskumar: 1. yes. I always forget to add it to the base-apply.html
2. block name, in the end, is allowed, and make things clear which block is getting terminated.

theskumar: Added.

frjo:
@theskumar To me it looks like this could be merged in directly, agree?

theskumar:
I was just testing the user feedback widget. Will mark as ready after

theskumar:
The user feedback widget will be considered later. I need to look deeper into its integration with the new sdk. Good to merge and test now.

frjo:
Personally I find all the chat/feedback widgets that are inserted on so many pages today rather annoying. Having a feedback page that is clearly linked in e.g. the footer is better in my opinion.

PR #3293, opened 2023-03-20
Change help to rich text

### LINKED ISSUE #3288: Change Help documentation to rich text fields

Currently our client wanted to put links inline for the help text that goes with questions.  They have since used some of the other rich text features as well.

Currently the helps have two fields: plain text, and link.

 This should remove the link part of the help section and change the plain text to rich text.

**Additional context**

For current help areas, this will remove the links.  It may be better to provide a migration to combine them, but at this time, it will delete them.

**Priority**

This was a medium, but we changed it in our fork, with change coming soon.

As will previous changes, this will have a PR associated with it.

frankduncan:
Fixes #3288

Changes the help field to rich text.  Of note is that when you have current forms that used both the text and the link, the link is lost to the sands of time.  Maybe we need a transition migration?  But not sure how to do that when working with stream objects.

There's one test change (in its own commit) that I'm not sure on.  Otherwise, everything looks good.


frjo:
@wes-otf How does OTF feel about converting all help text in application forms to rich text fields?

Today we have a plain text field + a help link field.

frjo:
@frankduncan A rebase would help with testing and we would need a migration to convert the current fields to the new one.

old help text + old help link = new help rich text

frankduncan:
@frjo Sounds great!  I've added it to my list and I should have it to you sometime in the next week.

wes-otf:
@frjo @frankduncan sorry for the delayed response but this would be fantastic! This week I've ironically talked with a few staff members at OTF about how it's be great to have some formatting in the help text so I'm certainly an advocate for this.

frjo:
Sounds like we are adding this to Hypha core then.

But see comment above about the migration I belive is a requisite to get it in. I do not think OTF, or any other organisation, want to manually readd all the help text to every existing form.

@wes-otf @frankduncan  Is there any chance we could make the help field use markdown with an editor instead of rich text?

With rich text field you allow staff to put in almost anything, e.g. paste in spagetti HTML from MS Word. 

With markdown we can ensure clean input and still allow for both links and formatting, just as requested.

frankduncan:
Looping in @jim-mcgowan

I lean toward rich text because thus far in my career, the target audience for things like hypha application configuration and target audience for markdown has little overlap.  Furthermore, hypha already supports rich text for submissions, but maybe there's a good drop in wagtail field type that is wysiwyg->markdown.  (If so, why not use that instead of rich text for submissions?)

frjo:
There is a markdown editor in Hypha already, "pagedown" from the Stackoverflow people. It is e.g one of the field option on application forms.

The reason we have a rich text editor is that so applicants can paste in spagetti HTML from MS Word. Preserving tables etc.

We are planing to look in to the possibility of converting that to Markdown before saving it to the DB.

frjo:
The comments tab on submission are using the Markdown editor/field as well.

wes-otf:
+1 to everything @frjo said, I'm also an advocate for Markdown. As mentioned [Pagedown](https://github.com/timmyomahony/django-pagedown) has worked pretty well for us so far and has some nice customization options as well.

frankduncan:
Finally looking at this again!  I can attempt to change it to the markdown form, but I think that building a migration that will work is outside of my ability.  Because of the use of streamfields, and the fact that that this is a change on the low level FormFieldBlock, the migration would have to dip into all instances of all users of that, which is... well, everything, right?  Or am I missing something and this is much easier than it looks?

frankduncan:
Took a stab at using markdown instead, and couldn't figure out a way for it to work.  The pagedown widget didn't come up in editing, and even when using markdown and saving, the help text just displayed a `<p></p>`.  I think that I could use some pointers or help, but otherwise we can just call this feature a deadend and we'll tackle it via an extension!

Did rebase and commit those changes though!

wes-otf:
@frjo maybe we could take a pass at this soon picking up where @frankduncan left off - this actually just came up for OTF where they were hoping for some formatting around help text

frjo:
@wes-otf Do you want to go with RichText or Markdown editor?

1. Should we delete the help link field?
2. How do we handle old help text and help links?

wes-otf:
@frjo I would probably lean toward markdown if possible (because it seems in the long term we'd like to shift everything that direction, not a strong preference). I would also be on the side of deleting the help link field then maybe inserting the link with the current text into the new help text field. Would be curious for your thoughts though

frjo:
See https://github.com/HyphaApp/hypha/pull/4375 for alternative solution.

PR #3149, opened 2023-02-02
New submission all page (beta)

theskumar:
Fixes #3430

- [x] Simplified code
- [x] Use postgres FTS search
- [x] Improve UX for filters
- [x]  Improve submission listing
- [x] Add filter for submission date, OTF have asked about this.
- [x] Add submission ID search, see https://github.com/HyphaApp/hypha/pull/3294
- [x]  🏃🏼 Make sure the batch actions are working 
- [x]  🏃🏼 Permissions check for batch actions
- [x] Ensure search index works after submissions is updated
- [x] ~Integrate Django messages framework with htmx responses~ (Not needed) 
- [x] [Fix Polynomial regular expression used on uncontrolled data](https://github.com/HyphaApp/hypha/security/code-scanning/74)
- [ ] make bulk update status working, currently it's hidden.

PR #3294, opened 2023-03-21
Add search by application number

### LINKED ISSUE #3289: Search by application number via #<app_number>

Just an extension of search

When searching for `#` and the id number of a submission, return just that submission

The client wanted this rather than just typing the submission number in the URL bar.

**Priority**

- Low priority

This was low priority for our client, but is fixed in our fork

As normal, we have PR incoming that solves this.

frankduncan:
Fixes #3289

What's on the tin.  A few notes:

* I preserved the git history from our repository, but can easily squash them all if you'd prefer such a small change be in one commit
* I'm not sure where in the documentation to document this feature

### DIFF
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -369,9 +369,10 @@ class SubmissionFilterAndSearch(SubmissionFilter):
     archived = filters.BooleanFilter(field_name='is_archive', widget=forms.HiddenInput, method='filter_archived')
 
     def search_data_and_id(self, queryset, name, value):
-        if value.strip().startswith("#") and value.strip("#").isnumeric():
-            return queryset.filter(id=value.strip("#"))
-        return queryset.filter(Q(id=value.strip("#"))|Q(search_data__icontains=value))
+        possible_id = value.strip("#")
+        if value.strip().startswith("#") and possible_id.isnumeric():
+            return queryset.filter(id=possible_id)
+        return queryset.filter(Q(id=possible_id)|Q(search_data__icontains=value))
 
     def filter_archived(self, queryset, name, value):
         if not value:
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -369,9 +369,9 @@ class SubmissionFilterAndSearch(SubmissionFilter):
     archived = filters.BooleanFilter(field_name='is_archive', widget=forms.HiddenInput, method='filter_archived')
 
     def search_data_and_id(self, queryset, name, value):
-        if value.isnumeric() and len(value) <= 3:
-            return queryset.filter(id=value)
-        return queryset.filter(search_data__icontains=value)
+        if value.strip().startswith("#") and value.strip("#").isnumeric():
+            return queryset.filter(id=value.strip("#"))
+        return queryset.filter(Q(id=value.strip("#"))|Q(search_data__icontains=value))
 
     def filter_archived(self, queryset, name, value):
         if not value:
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -365,8 +365,13 @@ class SubmissionFilter(filters.FilterSet):
 
 
 class SubmissionFilterAndSearch(SubmissionFilter):
-    query = filters.CharFilter(field_name='search_data', lookup_expr="icontains", widget=forms.HiddenInput)
+    query = filters.CharFilter(method='search_data_and_id', widget=forms.HiddenInput)
     archived = filters.BooleanFilter(field_name='is_archive', widget=forms.HiddenInput, method='filter_archived')
+
+    def search_data_and_id(self, queryset, name, value):
+        if value.isnumeric() and len(value) <= 3:
+            return queryset.filter(id=value)
+        return queryset.filter(search_data__icontains=value)
 
     def filter_archived(self, queryset, name, value):
         if not value:
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -1,5 +1,6 @@
 import json
 import textwrap
+import re
 
 import django_filters as filters
 import django_tables2 as tables
@@ -369,10 +370,10 @@ class SubmissionFilterAndSearch(SubmissionFilter):
     archived = filters.BooleanFilter(field_name='is_archive', widget=forms.HiddenInput, method='filter_archived')
 
     def search_data_and_id(self, queryset, name, value):
-        possible_id = value.strip("#")
-        if value.strip().startswith("#") and possible_id.isnumeric():
-            return queryset.filter(id=possible_id)
-        return queryset.filter(Q(id=possible_id)|Q(search_data__icontains=value))
+        possible_ids = map(int, re.findall('\d+', value))
+        if value.strip().startswith("#") and len(possible_ids) > 0:
+            return queryset.filter(id__in=possible_ids)
+        return queryset.filter(Q(id__in=possible_ids)|Q(search_data__icontains=value))
 
     def filter_archived(self, queryset, name, value):
         if not value:
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -370,7 +370,7 @@ class SubmissionFilterAndSearch(SubmissionFilter):
     archived = filters.BooleanFilter(field_name='is_archive', widget=forms.HiddenInput, method='filter_archived')
 
     def search_data_and_id(self, queryset, name, value):
-        possible_ids = map(int, re.findall('\d+', value))
+        possible_ids = [int(x) for x in re.findall('\d+',value)]
         if value.strip().startswith("#") and len(possible_ids) > 0:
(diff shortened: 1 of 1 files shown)


hypha/apply/funds/tables.py
```diff
@@ -365,9 +366,15 @@ def filter_category_options(self, queryset, name, value):
 
 
 class SubmissionFilterAndSearch(SubmissionFilter):
-    query = filters.CharFilter(field_name='search_data', lookup_expr="icontains", widget=forms.HiddenInput)
+    query = filters.CharFilter(method='search_data_and_id', widget=forms.HiddenInput)
     archived = filters.BooleanFilter(field_name='is_archive', widget=forms.HiddenInput, method='filter_archived')
 
+    def search_data_and_id(self, queryset, name, value):
+        possible_ids = [int(x) for x in re.findall('\\d+',value)]
+        if value.strip().startswith("#") and len(possible_ids) > 0:
```
theskumar: Would prefer to have the keyword `id:` instead of `#` given in hypha will be different kind of main objects like projects, submissions, funds etc. the keyword `id:12` make things very clear. 

The syntax is similar to how github/gmail use filters in the search.

frankduncan: Thanks for the feedback.  I'm going to reach out to our client to see what they think about this.  If they strongly prefer the `#`, then we may want to make that a configured value, or provide some way to override it.

frjo: With "id:" we could then build that out to "title:", "date:" etc.

theskumar: @frankduncan In the new implementation of search in #3149 , the search box allows for searching with just the submission number or `id:<number>`. 

That said, having `#` might not be bad and can be handled as an exception, but if possible let's try keeping a consistent language that can scale.

frankduncan: As talked about in the maintainers meeting (on April 11, 2023), the idea is that `#` is used throughout the application to refer to application numbers, and it makes sense to allow it as a shorthand in the search interface as well.  I believe it's the column heading on the submissions table, and also in various displays, it's shown as `#<number>`

theskumar: I'm not sure I'll be able to explain this well. But I see a problem with using find all the integers then also try to use a different pathway to filter if the search query starts with `#`

So keep this simple, and also for future improvement to the filtering language later. I would like you to use just this rule for now.

If the search query matches the pattern, `#<num>` then do the ID search. It will not support searching for multiple IDs at once and will the search data if `#<num>` is found.

I fundamentally agree it should be able to search for multiple IDs and allow combining with text searches, but the current implementation introduces and encourage bad patterns in the code that will be hard to enhance later on. 

e.g. people searching for `23 34`, which will return the submission with ids 23 and 34 or any submission with searchable content 34 or 23 in them but then it doesn't behave the same way if someone search for `#23 34`. 

Also, while searching for the search_data, it doesn't remove the numerical values against which it's already searched for.

theskumar: If you want to spend sometime on this and do it right, this is an approach I've been playing around with: 

```python
def parse_search_query(search_query):
    """
    Parses Gmail-like or GitHub-like search query string into a dictionary of filters and
    the remaining text.

    Example: "from:johndoe@example.com to:janedoe@example.com subject:hello world #12"
    would be parsed into:
    {
        "filters": {
            "from": ["johndoe@example.com"],
            "to": ["janedoe@example.com"],
            "subject": ["hello", "world"],
            "id": ["12"]
        },
        "text": "hello world"
    }
    """
    filter_pattern = r'(\w+):("[^"]+"|\S+)'
    id_pattern = r'#(\d+)'
    filters = {}
    remaining_text = search_query
    for match in re.finditer(filter_pattern, search_query):
        filter_name, filter_value = match.groups()
        # Remove quotes if present
        if filter_value.startswith('"') and filter_value.endswith('"'):
            filter_value = filter_value[1:-1]
        # Append to existing list if filter already exists
        if filter_name.lower() in filters:
            filters[filter_name.lower()].append(filter_value)
        else:
            filters[filter_name.lower()] = [filter_value]
        # Remove filter from remaining text
        remaining_text = remaining_text.replace(match.group(0), '').strip()

    # Add id filter if present
    for match in re.finditer(id_pattern, remaining_text):
        if "id" in filters:
            filters["id"].append(match.group(1))
        else:
            filters['id'] = [match.group(1)]
        remaining_text = remaining_text.replace(match.group(0), '').strip()
    return {"filters": filters, "text": remaining_text}


def add_search_and_filter_to_submissions(qs, search_query):
    result = parse_search_query(search_query)
    filters = result["filters"]
    text = result["text"]

    print("*" * 30)
    print(result)
    print("*" * 30)

    if "id" in filters:
        qs = qs.filter(id__in=filters["id"])

    if text:
        qs = qs.filter(search_data__icontains=text)

    return qs

```

theskumar: An enhanced version of this `parse_search_query` could also return a django-orm `Q` statement instead of dict of filters, and could understand things like `AND` and `OR` operator but that is all for later and not part of this implementation. 

What I really want to emphasize here is that if we want to do filter with search query, let's have a well-defined grammar.

kfogel: Hi, @theskumar.  

> What I really want to emphasize here is that if we want to do filter with search query, let's have a well-defined grammar.

Completely agree.  The question here is just about whether or not to include a particular idiom in that grammar: "#<number>" as an alternate way of saying "id:<number>" (similarly to how "they will" and "they'll" are both properly within English grammar).

First, I'll summarize, just to make sure I'm understand everything so far correctly:

1. Right now we should handle the case of just *one* "`#<number>`" as the search string: sounds like we all agree on that -- it should be supported, and is equivalent to "`id:<number>`".
2. In the future, we should handle multiple "`#<number>`" in the search string in some way that is both related to the above and well-defined.  But...
3. Because the implementation for that isn't ready yet, and even the semantics are not yet clear (because we don't yet have full Boolean operators, etc), *for now* we don't do any special handling if there are multiple "`#<number>`" in the search string, nor do we do anything special if the search string contains a "`#<number>`" plus other things as well.  In other words, for now all of those cases would just be regular text searches, and would not convert to the semantics of "`id:<number>`" for any of the numbers.

Is that what you're proposing we do for now?  (With further improvements to come in the future, which we can discuss later of course.)

If so, I like it.  But I wanted to make sure that I understand correctly first.

theskumar: @kfogel You are absolutely right.

frankduncan: So does that mean we want to make changes to this PR, or just accept and tackle the next steps via the normal methods?  Though I see a lot of changes from you, so really what I'm asking is: is there any way I can jump in and help?  Or are you on it so I can sit back and watch the magic happen :)

theskumar: Let's modify the current PR and have it merged. 

Because the work I'm doing in #3149 will take sometime and will also not immediately replace the current submission all page. We'll have that available as an opt-in page and run a beta on it. There are various upgrades on how the submission all page works internally, including the search behaviour, if you want to contribute to the filtering side of things you are most welcome in that PR, but I don't want it to do any changes to the current submission search implementation. 

Once the new submission all page is done that includes the new FTS search, I want to start thinking of advance filtering via search query.

frankduncan: The work looks awesome!  What modification to this PR do we want before we merge it?

theskumar: You can update the PR to use this regex `^#(\d+)$` to find the ID and if an ID is present use the id search else go the old search. 

More info about the regex: https://regex101.com/r/wnr8Ew/1

frankduncan: Alright!  commits pushed (and rebased off main).  I think this should match what you're looking for (pun intended)

theskumar (approved):

hypha/apply/funds/tables.py
theskumar: This is perfect! Thank you for being considerate!

frankduncan: Absolutely!  Thank you for being so clear :)

kfogel:
Process question: @frjo, do you review and then do the merge now that @theskumar has given an approval, or does @frankduncan merge, or...?

(I did look at https://github.com/HyphaApp/hypha/blob/main/CONTRIBUTING.md first; it points to https://docs.hypha.app/contributors which is a 404 right now, and I remember a discussion with the doc maintenance team about that... I'll go dig that up and see if the path to fixing that stale doc link can be revived.)

theskumar:
@kfogel, it will be @frjo merging the PR. 

We recently moved about the docs and it now lives at https://docs.hypha.app/getting-started/contributing/code-contributions/

kfogel:
> @kfogel, it will be @frjo merging the PR.
> 
> We recently moved about the docs and it now lives at https://docs.hypha.app/getting-started/contributing/code-contributions/

Thanks, @theskumar.  I've submitted PR #3403 to update `CONTRIBUTING.md` accordingly.

Issue #3297, opened 2023-03-23
Should workflow permissions be handled through wagtail interface


frankduncan:
While working on #3296, I realized that our withdrawing code relied on the applicant having edit permissions on the submission.  This was not a result of changing the wagtail permissions for submissions to "edit", but rather on changing workflow.py here: https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/workflow.py#L225

And changing it to `applicant_edit_permissions`.  This seems like something that should be configured through the admin interface, rather than a code change, but I wanted to open up a discussion about the most appropriate place to enable this.

fourthletter:
I'm sorry we didn't get a chance to discuss this during our Tuesday morning call. We will make a point to discuss these items you've raised during upcoming meetings.

PR #3298, opened 2023-03-23
Add submission withdrawal workflow

### LINKED ISSUE #3296: Allow Applicants to Withdraw a submission

A new workflow and UX to allow applicants to move to a new withdrawn status.  This is for applicants who no longer want to apply, after submitting.

This was seen as a better system than having the applicant reach out and with one of the admin team deleting the submission

**Additional context**

After discussion about this feature, a config toggle should also be added to allow admins of hypha to opt in or out of this feature (set to False initially)

**Priority**

This was originally a high priority, but it was merged into our fork.  So a PR will be incoming

**Affected roles**

- Applicants

frankduncan:
Fixes #3296

What it says on the tin.  When applicants have edit permissions (requires a change to workflow.py, see #3297), they can withdraw applications that have been submitted.  This adds that as a new status for submissions.


hypha/apply/funds/templates/funds/applicationsubmission_detail.html
```diff
@@ -81,6 +81,13 @@ <h5>{% blocktrans with stage=object.previous.stage %}Your {{ stage }} applicatio
                                     <svg class="icon icon--delete"><use xlink:href="#delete"></use></svg>
                                 </a>
                             {% endif %}
+                            {% if request.user|has_edit_perm:object %}
```
frjo: I think adding a `has_withdraw_perm` function is better. It should make sure the user is the applicant, that `ENABLE_SUBMISSION_WITHDRAWAL` is true etc.

hypha/apply/funds/workflow.py
```diff
@@ -343,11 +343,12 @@ def make_permissions(edit=None, review=None, view=None):
                 'ext_internal_review': _('Open Review'),
                 'ext_determination': _('Ready For Determination'),
                 'ext_rejected': _('Dismiss'),
+                'ext_screening_withdrawn': _('Withdraw'),
```
frjo: Why is it not enough with just `ext_withdrawn` in all places? We also need this for all workflows.

hypha/apply/funds/workflow.py
```diff
@@ -343,11 +343,12 @@ def make_permissions(edit=None, review=None, view=None):
...
             },
             'display': _('Need screening'),
             'public': _('Application Received'),
             'stage': RequestExt,
-            'permissions': default_permissions,
+            'permissions': applicant_edit_permissions,
```
frjo: I think withdraw permissions should be separate from edit permissions.

We do not allow applicants to edit submissions that are in review for obvious reasons.

But if we allow an applicant to withdraw an application they should be able to do that in any state up until the submission is approved/declined.

hypha/public/utils/context_processors.py
```diff
@@ -18,5 +18,6 @@ def global_vars(request):
         'ORG_URL': settings.ORG_URL,
         'CURRENCY_SYMBOL': settings.CURRENCY_SYMBOL,
         'GOOGLE_OAUTH2': settings.SOCIAL_AUTH_GOOGLE_OAUTH2_KEY,
+        'ENABLE_SUBMISSION_WITHDRAWAL': settings.ENABLE_SUBMISSION_WITHDRAWAL,
```
frjo: Rebase and move to core app. We have moved most global vars there.

frjo (changes requested):
Overall approach ok but implementation needs work.

frankduncan:
Thanks for the feedback!  Patches incoming

hypha/apply/funds/workflow.py
frankduncan: I think the idea was that we wanted to capture the stage at which it was withdrawn, for the workflow timeline at the top of the submission.  Should we keep it like that, or just put `ext_withdrawn` in the final step (next to accepted, rejected, etc)?

frjo: We do not do that for any other state so I prefer a single *_withdraw" per workflow.

frankduncan: That's fine!  Do you have a preference where in the workflow it goes?  I was going to put it at the end next to accept/reject, but I could also make an argument that it should go at the beginning (INITIAL_STATE)

frankduncan:
Ok!  I believe I've made the correct updates.  I did testing on all the different workflows, but I'm not as familiar with some of the ones we don't use with ARDC, so there may be edge cases I'm not aware of.  Especially the Concept/Proposal, which I believe I did in line with the spirit of those workflows.

frjo (approved):

fourthletter:
A 'withdraw' button has been added 

image

bickelj:
@frjo What would we need to do to get this one in? I was about to make an issue and PR but found this existing one.

frjo:
@bickelj Rebase of current main and address the issues above. I think it makes perfect sense to allow applicants to withdraw an application. If this should be the default or optional I'm not sure but make it optional for now.

hypha/apply/funds/workflow.py
bickelj: This appears completed to me.

hypha/public/utils/context_processors.py
bickelj: This appears done to me.

hypha/apply/funds/templates/funds/applicationsubmission_detail.html
bickelj: Current commits have the checks directly in the `html` file but not a separate `has_withdraw_perm`, so I'll add that function.

bickelj: Whoa, the permissions is quite a bit more complex than I thought it might be. Certainly more than adding one function.

bickelj: I think I see how it works and I am testing out my solution.

hypha/apply/funds/workflow.py
bickelj: In progress.

hypha/apply/funds/templates/funds/applicationsubmission_detail.html
bickelj: I think https://github.com/HyphaApp/hypha/pull/3298/commits/f4ae00168a4651d18a6fe9420a54750f1bdf0480 resolves this.

hypha/apply/funds/workflow.py
bickelj: I think https://github.com/HyphaApp/hypha/pull/3298/commits/f4ae00168a4651d18a6fe9420a54750f1bdf0480 resolves this, let me know what you think.

bickelj:
@frjo I think this is ready for testing again.

hypha/apply/funds/workflow.py
```diff
@@ -420,16 +437,23 @@ def make_permissions(edit=None, review=None, view=None):
...
             },
             "display": _("More information required"),
             "stage": RequestExt,
             "permissions": applicant_edit_permissions,
         },
+        "ext_withdrawn": {
+            "display": _("Withdrawn"),
+            "stage": RequestExt,
+            "permissions": staff_edit_permissions,
+        },
```
frjo: This instance of "ext_withdrawn" should be removed I believe, the correct one exist further down.

hypha/apply/funds/workflow.py
```diff
@@ -377,6 +388,11 @@ def make_permissions(edit=None, review=None, view=None):
             "stage": Request,
             "permissions": no_permissions,
         },
+        "withdrawn": {
+            "display": _("Withdrawn"),
+            "stage": Request,
+            "permissions": staff_edit_permissions,
```
frjo, line 578: The permission should be "no_permissions" for all "withdrawn" states. Same as "rejected". Staff has no point in editing withdrawn applications.

hypha/apply/funds/workflow.py
```diff
@@ -265,6 +269,7 @@ def make_permissions(edit=None, review=None, view=None):
                 "almost": _("Accept but additional info required"),
                 "accepted": _("Accept"),
                 "rejected": _("Dismiss"),
+                "withdrawn": _("Withdraw"),
```
frjo: I do not think we need "withdrawn" in any transitions options. Staff should never withdrawn anything, only the applicant.

hypha/apply/funds/templates/funds/applicationsubmission_detail.html
```diff
@@ -111,6 +111,14 @@ <h5>{% blocktrans with stage=object.previous.stage %}Your {{ stage }} applicatio
                                         {% trans "Delete" %}
                                     </a>
                                 {% endif %}
+                                {% if request.user|has_withdraw_perm:object %}
+                                    <a
+                                        class="font-bold text-red-600 flex items-center hover:opacity-70 transition-opacity"
+                                        href="{% url 'funds:submissions:withdraw' object.id %}">
+                                        {% heroicon_micro "arrow-uturn-down" class="inline me-1 fill-red-600" aria_hidden=true %}
```
frjo, line 144: I like this icon!

frjo (changes requested):
Nice work! Some questions around implementation details.

Good improvement of Hypha.

hypha/apply/funds/templates/funds/applicationsubmission_detail.html
bickelj: What's odd to me is there is some orphaned CSS. Maybe those classes can be cleaned up separately.

hypha/apply/funds/workflow.py
bickelj: I thought (but did not test) that the transitions list was what allowed step A to go to step B, namely it's allowed if step B shows up in step A's transition list. But maybe I am missing the point here, looking more closely.

frjo: Transitions are the options that staff gets when they want to change status.

bickelj: So when I remove `...withdraw...` from the transitions, there is no effect when I apply and then click withdraw. In other words, it does not go into withdrawn status, not good. But this is probably due to the implementation of the `withdraw` method in `SubmissionWithdrawView` where it explicitly looks through the `transitions`: `            for action in obj.workflow[obj.status].transitions.keys()`. Should that `for` loop go away and the `if len(withdraw_actions)` condition go away as well, and do a different kind of lookup to find the action?

bickelj: Hmm. When I remove all the `...withdraw...` from the transitions and change the `SubmissionWithdrawView.withdraw` implementation to this (starting at existing `withdraw_actions`):
```python

        if len(withdraw_actions) == 1:
            action = withdraw_actions[0]
            obj.perform_transition(
                action, self.request.user, request=self.request, notify=False
            )
        elif len(withdraw_actions) > 1:
            raise PermissionDenied(f'In workflow "{obj.workflow}" too many withdraw actions: "{withdraw_actions}"')
        elif len(withdraw_actions) < 1:
            raise PermissionDenied(f'No withdraw actions found in workflow "{obj.workflow}"')
```
then the call to `obj.perform_transition` fails on line 347.

```
  File "/home/everybody/code/ots/ext/hypha-clean/hypha/apply/funds/views.py", line 1611, in withdraw
    obj.perform_transition(
  File "/home/everybody/code/ots/ext/hypha-clean/hypha/apply/funds/models/submissions.py", line 347, in perform_transition
    raise PermissionDenied(f'Invalid "{ action }" transition')
django.core.exceptions.PermissionDenied: Invalid "withdrawn" transition
```

Unless there's another way to advance a workflow, I don't see how this can work. What am I missing?

frjo: My mistake, the transitions are also used to check if the status change is allowed.

So they need to be there but I think we can add custom permissions to them so only applicant are allowed to use them. Example below.

https://github.com/HyphaApp/hypha/blob/f63f0b2a2ea32ed7e4b982f479ebdf7dcc15c0ff/hypha/apply/funds/workflow.py#L402-L408

bickelj: The permissions are already checked twice elsewhere, so I don't think we need a third.

Check one is in the `SubmissionWithdrawView.withdraw` function and check two is whenever any permissions object in `workflow.py`  is checked because `make_permissions` in `workflow.py` defaults `"withdraw"` to `[applicant_can]`. Yes?

hypha/apply/funds/templates/funds/applicationsubmission_detail.html
bickelj: I added issue https://github.com/HyphaApp/hypha/issues/4073 for the CSS thing.

hypha/apply/funds/workflow.py
frjo: This permission hide this status from staff so they do not see it and by accident make use of it.

frjo: I also think we should limit "withdraw" to `applicant_edit_permissions`. 

```
applicant_edit_permissions = make_permissions(
    edit=[applicant_can, partner_can], review=[staff_can], withdraw=[applicant_can]
)
``` 

```
def make_permissions(edit=None, review=None, view=None, withdraw=None):
    return {
        "edit": edit or [],
        "review": review or [],
        "view": view
        or [
            staff_can,
            applicant_can,
            reviewer_can,
            partner_can,
        ],
        "withdraw": withdraw or [],
    }
```

bickelj: > This permission hide this status from staff so they do not see it and by accident make use of it.

Without the suggested change (i.e. while on 5b656bbc2e89c9d93f7ab90e690783667c852d87) locally, I logged in as staff and could not see the Withdraw button. Very well, that's because of a check on the page not displaying a link. So I logged in as the applicant, then copied the "withdraw" link, logged out, logged in as staff, and pasted the withdraw link, clicked the button: `PermissionDenied`:

```
hypha/apply/funds/views.py", line 1599, in withdraw
    raise PermissionDenied
django.core.exceptions.PermissionDenied
```

What code is this? The new view:
```python
class SubmissionWithdrawView(SingleObjectTemplateResponseMixin, BaseDetailView):
#...
   def withdraw(self, request, *args, **kwargs):
#...
        if not request.user.is_applicant:
            raise PermissionDenied #here is line 1599.
```

Would you prefer these two lines in the view to be removed so that the permissions are properly concentrated in the one area? I am making that change speculatively.

When I remove those two lines, then staff indeed are able to withdraw. Also, admin can see the withdraw button. I don't think that would be a problem, though, because an Admin may need to click it on behalf of an applicant.

Hmm, I don't see the ability for the applicant to withdraw when in "Application received" status and with the [above suggestion about limiting withdraw to `application_edit_permissions`](https://github.com/HyphaApp/hypha/pull/3298#discussion_r1698097701). When I undo that change, I can see the withdraw button in that status again as the applicant. But this is solved by adding `withdraw=[applicant_can]` to all the permissions except `hidden_from_applicant_permissions` and `no_permissions`. That makes a bit more sense because a withdrawn application shouldn't be able to be withdrawn again.

Curiously when I add the custom permissions (without specifying a method) it seems the staff can still withdraw (by going to the URL directly). If I add the method, how does it find that method, though? Looking... both of the existing examples are methods on the `ApplicationSubmission` model. I suppose I must add such a method, then. I don't quite understand what this is doing. I see the `django_fsm` library used for workflows. I see that `can_proceed` probably should raise the permission denied error. But I don't see why it doesn't do that.

frjo: @bickelj As staff check the "Update status" button/dialog. There staff will see "Withdraw" unless permission is set to hide it.

hypha/apply/funds/views.py
```diff
@@ -1579,6 +1583,45 @@ def form_valid(self, form):
         return super().form_valid(form)
 
 
+class SubmissionWithdrawView(SingleObjectTemplateResponseMixin, BaseDetailView):
```
frjo: Add

`@method_decorator(login_required, name="dispatch")`

frjo: Is the `SingleObjectTemplateResponseMixin` needed?

hypha/apply/funds/views.py
```diff
@@ -1579,6 +1583,45 @@ def form_valid(self, form):
...
+
+    def post(self, request, *args, **kwargs):
+        return self.withdraw(request, *args, **kwargs)
+
+    def withdraw(self, request, *args, **kwargs):
+        if not settings.ENABLE_SUBMISSION_WITHDRAWAL:
+            raise PermissionDenied
+
+        if not request.user.is_applicant:
+            raise PermissionDenied
```
frjo: Can we use `UserPassesTestMixin` for this check? I think that makes the code easier to read.

frjo: Or if `perform_transition` function does this check we do not need it at all as you mention below.

hypha/apply/funds/workflow.py
bickelj: @frjo I made a lot of noise above which you can read as "the permissions model for workflows is not clear or not working as I would expect on first glance." This is particularly around [the suggestion above](https://github.com/HyphaApp/hypha/pull/3298#discussion_r1696513386) which seemed straightforward to me. I used this snippet for all the `"withdrawn"` keys in `transitions` but somehow staff can still make the transition to withdrawn:
```python

                "withdrawn": {
                    "display": _("Withdraw"),
                    "permissions": {UserPermissions.APPLICANT},
                    "method": "withdraw",
                }
```

Note that staff do not have a withdraw button, however, when they go straight to the `submissions/{id}/withdraw` URL directly they should get permission denied when clicking "withdraw" there. But they don't.

hypha/apply/funds/views.py
bickelj: I removed these two lines locally so that the permissions can be concentrated in the workflow code. However, that means I need to dig into all the subtleties of the workflow permissions model. [See this related comment](https://github.com/HyphaApp/hypha/pull/3298#discussion_r1707262885).

hypha/apply/funds/views.py
bickelj: :shrug: I suppose I could try without once the other issues are resolved.

frjo:
@bickelj I rebased of main and added some permission fixes that seems to work. (I hope it doesn't mess it up for you.)

Please try it out and see how it behaves.

bickelj:
@frjo Thank you for making the change, it didn't mess anything up for me! It looks great!

From a user perspective, applicants cannot withdraw right after submitting the application as expected. If I add `withdraw=[applicant_can]` to `default_permissions`, this problem goes away. I assume we need to add it to most of the others in order for applicants to withdraw from most states, right?

From a programmer perspective, the `permissions` in `transitions` seems to now be honored. Good, except the reason it's honored is it is explicitly looked up in a view. I hoped that the `django-fsm` library (or the Hypha extensions of `django-fsm`) would somehow honor the permissions declared in the workflow even when not explicitly referenced from the view. I mean I thought one of `get_transition`,  `can_proceed`, or `transition` in the `perform_transition` method would cause a `PermissionDenied` but it does not. While I think that specifying permissions in the workflow itself is a more sustainable design, it was not clear to me that additional implementation in the view was a requirement. I still am not sure what `"method"` does there or where else the permissions object in the workflow is used.

frjo (approved):

frjo:
I have added `withdraw=[applicant_can]` to `default_permissions`, makes total sense.

Unsure of adding it to other places. Can we put this on test and see what people think of it?

bickelj:
@frjo Sure, yes, sorry for the delay!

hypha/apply/funds/workflow.py
bickelj: I think this is resolved for all practical purposes.

bickelj:
@frjo While looking at `django-fsm` docs I noticed [django FSM was deprecated in favor of a new viewflow API](https://github.com/viewflow/django-fsm?tab=readme-ov-file#django-friendly-finite-state-machine-support) earlier this year. It probably doesn't affect this PR but I wanted to note it somewhere before I forgot.

frjo:
Created this https://github.com/HyphaApp/hypha/issues/4090

bickelj:
@frjo Do you think this one can be in the next release?

frjo:
I just rebased the PR. It will go to test late this week or next week. So probably the release after the next one.

hypha/apply/funds/workflow.py
```diff
@@ -1104,7 +1268,11 @@ def make_permissions(edit=None, review=None, view=None, withdraw=None):
                 "proposal_almost": _("Accept but additional info required"),
                 "proposal_accepted": _("Accept"),
                 "proposal_rejected": _("Dismiss"),
-                "proposal_withdrawn": _("Withdraw"),
+                "proposal_withdrawn": {
+                    "display": _("Withdraw"),
+                    "permissions": {UserPermissions.APPLICANT},
+                    "method": "withdraw",
+                },
```
bickelj, line 1471: The previous was more concise. What is the advantage of this repetition?

frjo: To give the applicant permission to perform the action.

hypha/apply/funds/workflow.py
```diff
@@ -204,7 +204,7 @@ def make_permissions(edit=None, review=None, view=None, withdraw=None):
             reviewer_can,
             partner_can,
         ],
-        "withdraw": withdraw or [applicant_can],
```
bickelj: This is where I was thinking it avoided the repetition but is that incorrect? Perhaps it was the former 1753 in the view that was duplicative/incorrect before?

bickelj: I meant to say "Line 1753 in the view."

hypha/apply/funds/workflow.py
bickelj: I could have sworn I tested with applicants but I guess you mean in the context of putting the permissions in a single place rather than multiple places.

PR #3302, opened 2023-03-27
Added currency symbol in requested amount field

### LINKED ISSUE #3301: Add currency symbol on requested amount field

Add currency symbol on requested amount field in Application form.

**Additional context**

image

**Priority**

**Affected roles**

- Staff
- Applicants
- Reviewers
- General public

gmurtaza00:
Added currency symbol in Requested amount field in application form. The currency symbol is set in settings/base.py.

image

Closes #3301

### DIFF
--- a/hypha/apply/funds/blocks.py
+++ b/hypha/apply/funds/blocks.py
@@ -38,7 +38,7 @@ class TitleBlock(ApplicationMustIncludeFieldBlock):
 class ValueBlock(ApplicationSingleIncludeFieldBlock):
     name = 'value'
     description = 'The value of the project'
-    widget = forms.NumberInput(attrs={'min': 0})
+    widget = forms.NumberInput(attrs={'min': 0, 'id':'requested-amount'})
 
     class Meta:
         label = _('Requested amount')
--- a/hypha/apply/funds/blocks.py
+++ b/hypha/apply/funds/blocks.py
@@ -38,7 +38,7 @@ class TitleBlock(ApplicationMustIncludeFieldBlock):
 class ValueBlock(ApplicationSingleIncludeFieldBlock):
     name = 'value'
     description = 'The value of the project'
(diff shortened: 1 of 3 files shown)


hypha/apply/funds/blocks.py
```diff
@@ -38,7 +38,7 @@ class Meta:
 class ValueBlock(ApplicationSingleIncludeFieldBlock):
     name = 'value'
     description = 'The value of the project'
-    widget = forms.NumberInput(attrs={'min': 0})
+    widget = forms.NumberInput(attrs={'min': 0, 'id':'requested-amount'})
```
theskumar: Instead of going the js route. I would advice to override the NumberInput template and use that instead. 

Even if we js route, which to me feel little hackish, you should use `class` attribute instead of the `id` to allow for multiple valueblock to exist on the same page.

hypha/static_src/src/javascript/apply/application-form.js
```diff
@@ -2,6 +2,15 @@
 
     'use strict';
 
+    let symbol = $(document.currentScript).attr('symbol');
+    const amount = $('#requested-amount');
+    $(amount[0]).css({paddingLeft: '20px'});
```
theskumar: This is raise an error if `amount` is null. Also this is executed on every page request, not ideal.

theskumar (changes requested):

hypha/apply/funds/blocks.py
frjo: I concur, the js script feels like a hack. This should be possible to fix directly in the template for the widget.

gmurtaza00: `ValueBlock` inherits `ApplicationSingleIncludeFieldBlock`, hence it can only exist once in a form.

frjo:
This approach with js will not make it in to core Hypha.

PR #3309, opened 2023-03-30
Add inline-icon component and add it to svg icon on in listing_base.h…

frjo:
…tml. Tailwind add a display block on all svg as a reset but here we need the svg to be inline.

### DIFF
--- a/hypha/apply/activity/templates/activity/include/listing_base.html
+++ b/hypha/apply/activity/templates/activity/include/listing_base.html
@@ -34,7 +34,7 @@
             {% endif %}
 
             {% if editable %}
-                <div class="feed__comment js-comment" data-id="{{activity.id}}" data-comment="{{activity|display_for:request.user|to_markdown}}" 
+                <div class="feed__comment js-comment" data-id="{{activity.id}}" data-comment="{{activity|display_for:request.user|to_markdown}}"
                 data-visibility-options="{{activity|visibility_options:activity.user}}"
                 data-visibility="{{activity.visibility}}"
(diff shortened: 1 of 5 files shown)


hypha/apply/activity/templates/activity/include/listing_base.html
```diff
@@ -49,7 +49,7 @@
             {% if not submission_title and activity|user_can_see_related:request.user %}
                 {% with url=activity.related_object.get_absolute_url %}
                     {% if url %}
-                    <a href="{{ url }}" target="_blank" class="feed__related-item">
+                    <a href="{{ url }}" target="_blank" class="feed__related-item icon-inline">
                         {{ activity.related_object|model_verbose_name }} <svg><use xlink:href="#arrow-head-pixels--solid"></use></svg>
```
theskumar:
```suggestion
                        {{ activity.related_object|model_verbose_name }} <svg class="inline"><use xlink:href="#arrow-head-pixels--solid"></use></svg>
```

theskumar:
Added some fixes and size adjustment to icons used elsewhere.

Make the size of icon relative so they fit in better at different places.

theskumar:
Affected areas:

Screenshot 2023-03-31 at 09 29 52@2x
Screenshot 2023-03-31 at 09 30 22@2x
Screenshot 2023-03-31 at 09 30 35@2x
Screenshot 2023-03-31 at 09 31 07@2x

theskumar (approved):
self approving ! 🗡️

PR #3313, opened 2023-03-30
Export applications by round

### LINKED ISSUE #3312: Export applications by round
**Is your feature request related to a new functionality? Please describe.**

Staff user should be able to download all the applications in a specific round whether it is open or closed.

**Additional context**

Exported data should be in .csv format.

**Priority**

- Medium 

**Affected roles**

- Staff
- Admin user

gmurtaza00:
Staff/admin user can download all the applications in a specific round whether open or closed. Export option will be visible when we hover over the round on http://hypha.test/apply/submissions/

image

Exported File name will be **Round name.csv**. Exported File will not contain the **round name, application phase and the file fields present in the application form.**

image

Closes #3312


hypha/apply/funds/views.py
```diff
@@ -561,6 +563,56 @@ def get_queryset(self):
...
+
+        if not isinstance(self.obj, (LabBase, RoundBase)):
+            raise Http404(_("No Round or Lab found matching the query"))
+        return super().get_queryset().filter(Q(round=self.obj) | Q(page=self.obj))
+
+    def get(self, request, pk):
+        self.get_queryset()
+        csv_data = self.export_submissions(pk)
+        response = HttpResponse(csv_data.readlines(), content_type="text/csv")
+        response['Content-Disposition'] = 'inline; filename=' + str(self.obj) + '.csv'
```
theskumar, line 614: For the filename, let's use a slug or cleaned versin of the `self.obj`. We can also be explicit on the what to use as the filename instead of relying on the `__str__` version of the object.

I would suggest going with something like `submission-[id]-export.csv` as filename format.

hypha/apply/funds/views.py
theskumar, line 1: Not sure if you already considered using https://django-tables2.readthedocs.io/en/latest/pages/export.html for export data?

gmurtaza00: This method exports the table data but we are exporting applications data by filtering with round and the table contains some extra unnecessary columns.

hypha/apply/funds/views.py
gmurtaza00: `submission-[id]-export.csv` can be used when only 1 submission is exported. Since we are exporting by rounds, naming the file by round name is more appropriate. 
I will update the pr with cleaned version of `self.obj`

theskumar: My bad.. i really meant `round-[id]-export.csv`.

hypha/apply/funds/views.py
theskumar: Gotchaya.

hypha/apply/funds/views.py
gmurtaza00: `round-[id]-export.csv` may mean something to devs but i don't think the admin user have any idea about round-id. Round name would be more meaningful in that case or we can do `round-name_[id]-export.csv`

theskumar: sounds good to me.

gmurtaza00: @theskumar Do you mean this by cleaned version of `self.obj `

image

This is how the file name looks. I tried adding id but the format looked wierd and I think following format is sufficient.
image

fourthletter:
Works great!

image

Issue #3328, opened 2023-04-03
Improving loading of submission detail page


theskumar:
**Is your feature request related to a problem? Please describe.**

There is a lot of data being loaded on the submission detail page, which cause it to be slow. 

**Is your feature request related to an existing functionality? Please describe.**

Display of the submission detail page.

**Describe the solution you'd like**

Since we now have htmx, the "Communication" and "Activity" tab can be loaded on demand and and not load during the initial page load.

**Describe alternatives you've considered**

--

**Additional context**

- Keep the functionality and lock and feel same, this is performance improvement request. 
- I'll work to fix it. 

**Priority**

- Low priority (annoying, would be nice to not see)

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- Developers
- General public

**Ideal deadline**

Date when you'd like to see this accomplished and a reason, if appropriate.

PR #3330, opened 2023-04-04
Load the activities tab of submission and project on demand

theskumar:
The activities tab can have a lot of activities as it’s unpaginated 
and also has an N+1 query to fetch the related objects.

This PR loads the data only after the tab is clicked.

I makes use of the htmx and alpine to observe the window-hash change. 

Notes:
- The communication tab with it's markdown editor, requires some custom initialization logic, which breaks with the htmx loaded content. So it' not included in this PR. 
- On a submission with 5 activities the number of SQL queries reduced from 168 to 148. 

Related #3328

### DIFF
--- a/hypha/apply/activity/services.py
+++ b/hypha/apply/activity/services.py
@@ -1,0 +1,47 @@
+from .models import Activity
+
+
+def get_related_actions_for_user(obj, user):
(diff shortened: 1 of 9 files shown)


frjo (approved):
Nice!

PR #3344, opened 2023-04-07
Added filter to convert decimal review score to whole number.

### LINKED ISSUE #3343: Display individual review score as whole number.

Display individual review score as whole numbers.

**Priority**

- Low priority

**Affected roles**

- Staff
- Admin

gmurtaza00:
Added filter to convert decimal review score to whole numbers.

image

image

image

Closes #3343

### DIFF
--- a/hypha/apply/review/models.py
+++ b/hypha/apply/review/models.py
@@ -180,7 +180,7 @@ class Review(ReviewFormFieldsMixin, BaseStreamForm, AccessFormData, models.Model
 
     @property
     def get_score_display(self):
-        return '{:.1f}'.format(self.score) if self.score != NA else 'NA'
+        return '{:.0f}'.format(self.score) if self.score != NA else 'NA'
 
     def get_absolute_url(self):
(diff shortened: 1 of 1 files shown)


frjo (changes requested):
This seems very subjective what you prefer. I do ot think we will hardcode a change. Making it adjustable with a setting with a defult the same as today would be good I think.

gmurtaza00:
> This seems very subjective what you prefer. I do ot think we will hardcode a change. Making it adjustable with a setting with a defult the same as today would be good I think.

We can make a setting for this but the values specified in options.py file for reviews are in whole numbers. So I guess it better that we display them as whole numbers not decimals.

image

frjo:
Yes, each review question can only be a whole number. A review will in many cases have several so the average will seldom be a whole number.

gmurtaza00:
> Yes, each review question can only be a whole number. A review will in many cases have several so the average will seldom be a whole number.

Got your point. Actually we customized if for ARDC to just calculate sum and not average and it was the follow up issue for that.

PR #3347, opened 2023-04-08
Project's improvements part2

### LINKED ISSUE #3195: Add the same fields to PAF as application forms etc. have.
### Description
The input field mentions bullet points but there’s no option to add bullet points. 

### Proposed Solution
Either update the support text or provide formatting in the field

### LINKED ISSUE #3306: Incorrect link when PAF not configured
When a PAF is not configured and you try to fill in the form Hypha shows this message:

`Approval form not configured. Please add an approval form in the [fund settings]`.

The "fund settings" link goes to: `http://apply.hypha.test/admin/funds/fundtype/edit/[id]/`

The id is correct but the link is not, it should be: `https://test-apply.hypha.app/admin/pages/[id]/edit/`

### LINKED ISSUE #3319: Change Download to View on Contracting documents table
### Description

On the Contract Change table change Download to **View**. The user clicks on View, and the form is rendered into another tab.

image

### LINKED ISSUE #3320: Finance should not see the edit button when reviewing the invoice
### Description

Finance should not see the edit button when reviewing the invoice

image

### LINKED ISSUE #3325: Hide Vendor
### Description

Hide Vendor (no response) from ui. UI was designed for financial software integration that is still pending.

image

### LINKED ISSUE #3327: Send an email when a requests for a change is done in PAF

### LINKED ISSUE #3331: UI fixes: a thread
- [x]  1. Add some space between the body text and `Select lead` as in design
- [x] 2. Make the popup heading left aligned - refer to figma. We need to make this change for all the popupa.
Image

- [ ] 3. As soon as the project is created, the supporting document row shows blue tick, it should be grey. Blue only when documents have been approved, i.e. after PAF approval
- [x] 4. Should we also remove `choose file` since its not doing anything really
- [x] 5. The download option for proposal should either prompt pdf, docx . Else we could probably remove the down icon and use it only when there are options for files
Image

- [x] 6. When uploading supporting documents, the pop-up header might be better labelled `Upload supporting document`
- [x] 7. We also need to put actual categories in the dropdown for user testing
Image

- [x] 8. After uploading supporting documents, we need same formatting for `Download` button which is bold and icon if there's option to download pdf/doc
- [x] 9. Remove action should be red with cross icon, like this 
- [x] 10. Please also left align the texts as shown by vertical line
Image
Reference image:
Image

But also! aren't we going with the design where the required documents are listed on the interface and users can directly upload it without having to use the popup dropown?

- [ ] 11. Can we pls make the font size of sow same at 'otf review' it is too big rn
Image

- [x] 12. There should be a download for the entire Approval form as well, and there should be edit for sow which would just take users to that section in edit mode.
Image

- [x] 13. Please add some space between VP technology and the body text as marked by red line
Image

- [x] 14. Can we please change `Request` to `Submit` so that the terminology is consistent
Image

- [x] 15. Once the PAF has been submitted for approval, we shouldn't allow staff to remove the supporting document
Image

- [x] 16. Please add date when it gets approved ([refer to design](https://www.figma.com/file/cmrfUklaIibgEYRFUIifuE/Hypha-designs?node-id=247-3851&t=Y8TPTBlN6NBR1w1b-4))
Image

- [x] 17. We can confirm this with @fourthletter but i think the SOW should show up on the same page as the Approval form, and should be viewed from the same page but lower. Because the problem that may arise is that people will just approve the PAF without looking at the SOW since they are on separate pages and there's no way to separately approve SOW.
Image
Image

- [x] 18. We don't need two upload contract buttons. i think the lower one should be sufficient
Image

- [x] 19. We need some spacing between the icons and text, everywhere. you can refer to Figma file for this too. We need to check some vertical spacing in few places. Also, check the padding inside Upload button. please check that for all similar buttons.
- [x] 20. If contracting documents are not configured we shouldn't show that row. And the icon shouldn't be blue initially - only after uploading the documents.
- [x] 21. Reupload button should also show once
- [ ] 22. We also don't need to show 'corrections' option when contracting uploads the document? 
- [x] 23. Can we please show the contract date next to the view option([like in the design](https://www.figma.com/file/cmrfUklaIibgEYRFUIifuE/Hypha-designs?node-id=432-3557&t=Y8TPTBlN6NBR1w1b-4)) instead of the label on the left?
Image

- [x] 21. Please make sure that countersigned is written the same way in all the places. Let's go with "Countersigned" instead of CounterSigned or counter-signed or counter sign etc...
- [ ] 22. When the applicant is uploading countersigned contract, the popup header should be `Upload Countersigned Contract`, instead of Upload we can say `Select contract` and then instead of submit it could be `Upload` to make things a bit more clear.
- [x] 23. I think if there is an option to upload contract documents, the `submit contractor documents` should be made available once contract documents have been uploaded else they will get missed causing unnecessary back and forth. @fourthletter can you please confirm regarding point 23?
Image

- [ ] 24. Delete action should be in red, hex: #CD3238 
Image

- [x] 25. Can we please change the label to `Need corrections/amendments to contract?` just to make it explicit and clear?
Image

- [x] 26. Let's please change the Invoice delete button to red, hex: #CD3238 
Image
Image
Also please remove underline from here for the buttons to keep them consistent.

- [x] 27. For the Staff this next step text should be: `Review invoice and take action`
- [x] 28. Also, staff can't upload invoices
Image

- [x] 29. Once the invoice is approved, the finance person should also get an email to process the invoice.
- [x] 30. Should the finance person be able to see invoice before it is approved by the staff? @fourthletter 
Image

### LINKED ISSUE #3335: The Staff should be able to see the contract document status
The staff should be able to see that contracting and the applicant both have uploaded their side of contracts.

Image

Right now it shows like this:

Image

sandeepsajan0:
- Fixes #3306 
- Fixes #3195 
- Fixes #3319 
- Fixes #3320 
- Fixes #3325 
- Fixes #3335 
- Fixes #3327 
- Partial Fixes #3331


sandeepsajan0:
@frjo we can put this on the test so that @abhiXII can test out the things. This PR contains some inline css code at a few places, I'll convert them into css classes after the testing.

PR #3355, opened 2023-04-13
Use both shape and colour to differentiate traffic light symbols.

frjo:
Skärmavbild 2023-04-13 kl  16 47 07


theskumar (approved):
Looks neat!

PR #3361, opened 2023-04-20
Fix email whitespaces and allow email translations

### LINKED ISSUE #3173: Whitespace in project e-mail notification templates
The current templating engine can not filter out whitespacing. E-mail templates must contain not indentation extra line breaks etc. They will all be copies as is to the actual e-mail. Not a good look.

https://github.com/HyphaApp/hypha/blob/af62dddd41362b3f4300ef3784d6d3e190196455/hypha/apply/activity/templates/messages/email/contract_uploaded.html#L11-L32

theskumar:
Due the way the `AdapterBase` for messages is structured, there is not easy to send markdown emails. Right now, it fixes few templates that had horizontal whitespaces issues.

We also use translations while rendering template as string and then remove extra lines introduced due to Django template tags.

Fixes #3173 #3243

### DIFF
--- a/hypha/apply/activity/adapters/emails.py
+++ b/hypha/apply/activity/adapters/emails.py
@@ -14,6 +14,10 @@ from hypha.apply.users.groups import (
     STAFF_GROUP_NAME,
 )
 from hypha.apply.users.models import User
+from hypha.core.mail import (
+    language,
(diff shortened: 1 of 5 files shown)


frjo (approved):
Deploying to test.

PR #3365, opened 2023-04-22
Set margin-block on p-tags since TailWind resets margins.

### LINKED ISSUE #3364: Set margin-block for p-tags, Tailwind resets margins to 0
### Description

Paragraph/line break not working in comments box on Comms tab

frjo:
Fixes #3364

### DIFF
--- a/hypha/static_src/src/sass/apply/base/_base.scss
+++ b/hypha/static_src/src/sass/apply/base/_base.scss
@@ -55,6 +55,10 @@ a {
     }
 }
 
+p {
+    margin-block: 1rem;
+}
+
 ul,
 ol {
     padding: 0;


hypha/static_src/src/sass/apply/base/_base.scss
```diff
@@ -55,6 +55,10 @@ a {
     }
 }
 
+p {
+    margin-block: 1rem;
+}
+
```
theskumar, line 61: Does it make sense to apply this rule to a limited scope and not globally? Happy to look at the place where it is causing issue and apply local changes. 

The default reset are important as are designed to be smooth over cross-browser inconsistencies and make it easier for you to work within the constraints of your design system. -- https://tailwindcss.com/docs/preflight

I looked at the use of `margin-block` which is for providing spacing of the element as per the reading direction of the text. Setting `1rem` in all the directions is not helping at all, it might better if we set `margin: 1rem` and then set `margin-block: 0`.

Tailwind also provides a `prose` class with `typrography` for the text which are dynamically generated and needs to have a look and feel of an article. https://tailwindcss.com/docs/typography-plugin

That all said, due to the historical context of this project, it might make sense to have this done now and not try to go find the individual elements affected.

theskumar: I see it's in the context of the communication comments, if possible let's try fix the spacing only that context. :)

frjo: Earlier we relied on the browser default (margin-block: 1em normally) so if we do not set it globally we will have to set it in a lot of places. Submissions, reviews, determination etc. all become dense and hard to read without paragraph margins.

`margin-block` is for "ltr" languages the same as setting `margin-top` and `margin-bottom`. I have done work with "rtl" pages so appreciate how `margin-block` and `margin-inline` etc. works regardless of language direction.

PR #3390, opened 2023-05-03
Upload button for each project and contract documents categories, Add template to both type of categories

### LINKED ISSUE #3189: Add ability to upload templates for each type of project supporting document
Users can then download the template, fill it in and upload the result.

### LINKED ISSUE #3375: Contract and Project document categories, replace "Recommended minimum" with checkbox
There is no longer a requirement to have the option to require more than one file for each category. Therefor we can replace the "Recommended minimum" (int 0-n) with a checkbox named "Required".

When a category is required (checkbox is checked):

* One file is required
* The category is listed on the project detail pages with a red asterisk*

At all times (category is required or not):

* Only one uploaded file per category is possible.
* All categories are listed on the project detail page. Each with their own upload button.
* When a file is uploaded only a view button shows.
* If user want to replace a file they need to first delete the current file and then upload a new one.

sandeepsajan0:
Fixes #3375 
Fixes #3189 

Changes:
1. Upload button for each category (for both types of documents, projects and contracts)
2. Hide field category select option from the modal and set it programmatically via JS. (for both types)
3. Use required in place of recommended (for both type). Recommended is there for existing categories but not getting used, we may remove or use it later.
4. Add template field(optional) for each category, it can be configured via wagtail and will be shown beside of category name.


frjo:
@sandeepsajan0 This PR needs a rebase after merging of https://github.com/HyphaApp/hypha/pull/3400. Project migrations clash (0075).

sandeepsajan0:
@frjo It is rebased with the main and ready for test deployment.

sandeepsajan0:
@frjo I fixed an issue in js file 'function was not accessible'. Can we put the latest version to the test site.

frjo:
@sandeepsajan0 Deploying to test now.

sandeepsajan0:
@frjo I have fixed the permission for View Template. We can put the latest code to test site and test it out.

frjo:
Deploying to test now.

PR #3412, opened 2023-05-16
Use <relative-time> element to display time

theskumar:
This use https://github.com/github/relative-time-element to make the
display of the datetime accessible, localized, concise and standard
across the site.

This also make use of the custom elements, which is also used by
wagtail 4 at different places in the CMS

This PR also setup a way for us to use ESM modules.

## Screenshots (Before):

Screenshot 2023-05-16 at 08 29 59

Screenshot 2023-05-16 at 08 29 40

Screenshot 2023-05-16 at 08 30 23

## Screenshots (after):
Screenshot 2023-05-16 at 08 27 13

Screenshot 2023-05-16 at 08 28 02

Screenshot 2023-05-16 at 08 28 47

Screenshot 2023-05-16 at 08 29 07

Screenshot 2023-05-16 at 08 30 53


hypha/templates/base-apply.html
```diff
@@ -47,6 +47,9 @@
         <script defer src="{% static 'js/apply/vendor/alpine-3.11.1.min.js' %}"></script>
         <style> [x-cloak] {display: none !important} </style>
         <!-- alpine.js end -->
+
+        <!-- modules -->
+        <script type="module" src="{% static 'js/esm/github-relative-time-element-4-3-0.js' %}"></script>
```
frjo: Can we remove the version number from the file name? A bit cumbersome on updates otherwise.

frjo (changes requested):
Solid improvement, just a small issue regarding version number in file name.

hypha/templates/base-apply.html
theskumar: Done.

package.json
```diff
@@ -6,7 +6,8 @@
     "license": "GPL-2.0",
     "babel": {
         "exclude": [
-            "**/*.min.js"
+            "**/*.min.js",
+            "**/esm/*.js"
         ],
         "presets": [
             [
```
frjo, line 13: Need to add `@github/relative-time-element` to `devDependencies`.

frjo: If we select to use npm install and copy-npm-scripts instead of download-esm.

frjo (changes requested):

package.json
theskumar: Take a look at this post for why we can't use npm for it (yet!), also why simply downloading packages from the cdn will also not work for esm modules. 

https://simonwillison.net/2023/May/2/download-esm/

theskumar: I can add more info if you like to the `download-esm-modules` target to explain why and when to use it. Let me know. 

It's currently supposed to be used manually and rename the files (to remove the version) or if we keep the version number in the html, then update the version number in the html.

frjo: Good arguments for `download-esm`. Can we have the make command rename the file? Printing out a messages about the need of renaming would work as well.

Renaming it manually is not a big thing, but may confuse a new developer, or us in a few month time :-).

theskumar: Added some notes in Makefile and base_apply.html on how to work this setup.

theskumar:
This PR content is already merged to `main` so closing this now.

PR #3415, opened 2023-05-17
#3414: Configured setting for displaying total review score on reviews page

### LINKED ISSUE #3414: Configure displaying total review score

Configure a setting to display total review score instead of average score on reviews page.

**Priority**

- Low priority 

**Affected roles**

- Staff
- Reviewers

gmurtaza00:
Configured a setting for displaying total review score instead of average score. Each organization can customize it accordingly. 

**Setting to consider:** 

- ```DISPLAY_TOTAL_REVIEW_SCORE```

Closes #3414

### DIFF
--- a/docs/setup/administrators/configuration.md
+++ b/docs/setup/administrators/configuration.md
@@ -141,6 +141,12 @@ Possible values are: fund, round, status, lead, reviewers, screening_statuses, c
 
     SUBMISSIONS_TABLE_EXCLUDED_FIELDS = env.list('SUBMISSIONS_TABLE_EXCLUDED_FIELDS', [])
 
+### Score diplayed on reviews page.
+
+If True sum of scores will be displayed otherwise average of scores will be displayed
+
+    DISPLAY_TOTAL_REVIEW_SCORE = env.bool('DISPLAY_TOTAL_REVIEW_SCORE', default = False)
+
 ### Should submission automatically transition after all reviewer roles are assigned.
 
     TRANSITION_AFTER_ASSIGNED = env.bool('TRANSITION_AFTER_ASSIGNED', False)
--- a/docs/setup/administrators/configuration.md
+++ b/docs/setup/administrators/configuration.md
@@ -143,9 +143,9 @@ Possible values are: fund, round, status, lead, reviewers, screening_statuses, c
 
 ### Score diplayed on reviews page.
 
-If True sum of scores will be displayed otherwise average of scores will be displayed
+If 'sum' sum of score will be displayed, if 'avg' average of score will be displayed, empty string is for use-cases where no final summary is desired by any setup.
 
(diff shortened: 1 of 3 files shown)


sandeepsajan0 (commented):
Looks good to me. As it won't affect the current visibility so there shouldn't be any issue but I was thinking to use these kinds of settings as wagtail settings so that these can be tested and changeable easily by any organization. 
What do you think @gmurtaza00 @frjo @theskumar?

sandeepsajan0 (commented):
Looks good to me. As it won't affect the current visibility so there shouldn't be any issue but I was thinking to use these kinds of settings as wagtail settings so that these can be tested and changeable easily by any organization. 
What do you think @gmurtaza00 @frjo @theskumar?

hypha/settings/base.py
```diff
@@ -478,3 +478,5 @@
         debug=SENTRY_DEBUG,
         integrations=[DjangoIntegration()]
     )
+
+DISPLAY_TOTAL_REVIEW_SCORE = env.bool('DISPLAY_TOTAL_REVIEW_SCORE', default = False)
```
frjo: This should be moved up to the Hypa settings section. We want all Hypha settings together in one place.

frjo (changes requested):

frjo:
We want to move a lot of settings in to a Hypha settings section in the Web app. I think we need to decide how that is going to work overall first.

docs/setup/administrators/configuration.md
```diff
@@ -141,6 +141,12 @@ Possible values are: fund, round, status, lead, reviewers, screening_statuses, c
 
     SUBMISSIONS_TABLE_EXCLUDED_FIELDS = env.list('SUBMISSIONS_TABLE_EXCLUDED_FIELDS', [])
 
+### Score diplayed on reviews page.
+
+If True sum of scores will be displayed otherwise average of scores will be displayed
+
+    DISPLAY_TOTAL_REVIEW_SCORE = env.bool('DISPLAY_TOTAL_REVIEW_SCORE', default = False)
```
theskumar: Make it an option instead of boolean. Current boolean make a lot of implicit assumption. It's a either/or situation but not reflected well with the name of the variable. 

Make this setting a take a value, and have it also accept empty for use-cases where no final summary is desired by any setup. 

Suggestion:
```
from typing import Literal

REVIEWS_FINAL_SCORE_METHOD: Literal['sum', 'avg', ''] = env.str('REVIEWS_SUMMARY_METHOD', default='sum')
```

theskumar (changes requested):
Since we are introducing a configuration let's try make it more maintainable. Added by recommendation inline cc: @frjo

docs/setup/administrators/configuration.md
gmurtaza00: @theskumar For cases like empty string, should we return **NA** from calculate_score function?

theskumar: `None` or `NA` seems fine with me.

PR #3416, opened 2023-05-18
Fix max recursion error on vendor setup page

### LINKED ISSUE #3395: Vendor setup is broken with RecursionError
OTF is not using this so it has not been part of testing. At some point we should fix it since it is a neat feature that we put a lot of time into.

```
RecursionError at /apply/projects/1/vendor/
maximum recursion depth exceeded
Request Method:	GET
Request URL:	http://apply.hypha.test/apply/projects/1/vendor/
Django Version:	3.2.19
Exception Type:	RecursionError
Exception Value:	
maximum recursion depth exceeded
```

`hypha/apply/projects/views/vendor.py, line 40, in show_extra_info_form`

sandeepsajan0:
Fixes #3395 

It was a package(django-formtools) issue, they have also mentioned it in [latest version](https://django-formtools.readthedocs.io/en/latest/changelog.html#id1).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,6 +16,7 @@ django-countries==7.5
 django-extensions==3.2.1
 django-file-form==3.4.3
 django-filter==2.4.0
+django-formtools==2.4.1
 django-fsm==2.8.1
 django-heroku==0.3.1
 django-hijack==3.2.6


requirements.txt
```diff
@@ -16,6 +16,7 @@ django-countries==7.5
 django-extensions==3.2.1
 django-file-form==3.4.3
 django-filter==2.4.0
+django-formtools==2.4.1
```
theskumar, line 19: Looks like you adding `django-formtools` as new dependency and not really upgrading. 

Do you think it was removed somehow?

PR #3425, opened 2023-05-29
Add proposal form field to concept determination

### LINKED ISSUE #3049: Show proposal form select field when needed on add determination form
### User Story

OTF enables program managers to select from multiple proposals. This function requires the user to change the status to **invite to proposal**, and select from one of the proposal options. 

Program staff sending out a determination to approve the concept note (first stage) automatically gives applicants the default proposal. In this case, the default proposal is the first proposal associated with the Fund.

### Proposed Solution

Option 1:
Applicants will be required to select a proposal from multiple options when they attempt to 'approve' a concept note by sending out a determination 

Option 2: 
Staff should not be able to 'approve' a concept note by sending out a determination

sandeepsajan0:
Fixes #3049 

1. Add a disabled proposal form field to concept determination.
2. It gets enabled only if the user selects 'Approved' in the determination field.
3. If it is enabled, it is required, else optional.

Update Status is working in the same way as it was working. We have only changed the behavior for the concept's Add Determination.

### DIFF
--- a/hypha/static_src/src/javascript/apply/determination-template.js
+++ b/hypha/static_src/src/javascript/apply/determination-template.js
@@ -22,12 +22,27 @@
         getMatchingCopy(value) {
             if (value === '0') {
                 this.text = document.querySelector('div[data-type="rejected"]').textContent;
+                var proposal_form = document.querySelector('#id_proposal_form');
+                if (proposal_form) {
+                    proposal_form.disabled = true;
+                    proposal_form.required = false;
+                }
             }
             else if (value === '1') {
                 this.text = document.querySelector('div[data-type="more_info"]').textContent;
+                var proposal_form = document.querySelector('#id_proposal_form');
+                if (proposal_form) {
+                    proposal_form.disabled = true;
+                    proposal_form.required = false;
+                }
             }
             else {
                 this.text = document.querySelector('div[data-type="accepted"]').textContent;
+                var proposal_form = document.querySelector('#id_proposal_form');
+                if (proposal_form) {
+                    proposal_form.disabled = false;
+                    proposal_form.required = true;
+                }
             }
             this.updateTextArea(this.text);
         }
--- a/hypha/static_src/src/javascript/apply/determination-template.js
+++ b/hypha/static_src/src/javascript/apply/determination-template.js
@@ -1,4 +1,4 @@
-(function ($) {
+(function () {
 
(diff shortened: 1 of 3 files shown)


hypha/static_src/src/javascript/apply/determination-template.js
```diff
@@ -20,14 +20,27 @@
         }
 
         getMatchingCopy(value) {
+            var proposal_form = document.querySelector('#id_proposal_form');
```
frjo: `const` instead of `var`.

hypha/static_src/src/javascript/apply/determination-template.js
```diff
@@ -20,14 +20,27 @@
...
             }
             else {
                 this.text = document.querySelector('div[data-type="accepted"]').textContent;
+                if (proposal_form) {
+                    proposal_form.disabled = false;
+                    proposal_form.required = true;
+                }
             }
             this.updateTextArea(this.text);
         }
```
frjo, line 46: I think the only jQuery is at the end, the `$(DeterminationCopy.selector()).each((index, el)` part.

If you have time, can you convert that to plain JS and remove all jQuery references from this file.

frjo (changes requested):
Some minor JS comments.

Alos added code that set the initial outcome value based on the action taken. This works for the old static determinations forms but not for the new streamfield ones.

Please see if my implementation is ok.

frjo (approved):

hypha/static_src/src/javascript/apply/determination-template.js
```diff
@@ -20,14 +20,27 @@
...
 
         getMatchingCopy(value) {
+            const proposal_form = document.querySelector('#id_proposal_form');
```
bickelj, line 23: I checked out 3a0b2c290f0aaaeda9ae6f7b7b2cb290a90ba004 and tried `grep -ir id_proposal_form *` hoping to find the template or code that creates an element with `id=id_proposal_form` but I don't see it. Am I missing something or is a change to the template missing? I get the following error on the branch associated with #4142:
```
Uncaught TypeError: proposal_form_field is null
    <anonymous> https://hypha.test/static/js/determination-template.js:53
    EventListener.handleEvent* https://hypha.test/static/js/determination-template.js:51
    <anonymous> https://hypha.test/static/js/determination-template.js:57
```

PR #3429, opened 2023-06-05
Filter only active user for user's queryset

### LINKED ISSUE #3424: The get_users_for_groups() function should only return active users
This function needs to check that the users retrieved are active.

https://github.com/HyphaApp/hypha/blob/84dc72ff9ebc321aec69172189f7e0c05cc84289/hypha/apply/activity/adapters/utils.py#L87-L103

sandeepsajan0:
Fixes #3424 

Add description here.


theskumar (approved):

theskumar:
Suggestion: Introduce a method on the `UserQuerySet` called active. 

so you can use `User.objects.active().filter(...` everywhere we encapsulate the logic what is considered active in a single place.

theskumar (approved):
Maybe in another PR, can you check and replace at other places where `.active()` can be re-used?

PR #3434, opened 2023-06-12
Add settings to show/hide Average review score

### LINKED ISSUE #3394: Add setting to show/hide average score in reviewers box
### Description

Please remove 'average score' from OTF's Hypha

image

sandeepsajan0:
Fixes #3394 

`SHOW_AVERAGE_REVIEW_SCORE`
Set its value default 'true' as it is an existing feature.

### DIFF
--- a/hypha/apply/funds/templates/funds/includes/review_sidebar.html
+++ b/hypha/apply/funds/templates/funds/includes/review_sidebar.html
@@ -4,11 +4,13 @@
     <span>
         {{ recommendation|traffic_light }}
     </span>
+    {% if SHOW_AVERAGE_REVIEW_SCORE %}
     <span class="font-medium">
         {% trans 'Avg. Score'%}: {{assigned_reviewers|average_review_score|floatformat:'1'}}
     </span>
+    {% endif %}
 </div>
-     
+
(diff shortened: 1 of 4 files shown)


theskumar (commented):
Can you please double check if this feature flag is applied everywhere. e.g. 

https://github.com/HyphaApp/hypha/blob/b35bf029557d6d2b622310b357de7e6998234aea/hypha/apply/funds/templates/funds/includes/submission_stats.html#L20-L21

sandeepsajan0:
@theskumar Thanks for pointing that out. It looks like a  custom implementation. I looked for the [template tag](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/review/templatetags/review_tags.py#L47) `average_review_score` everywhere, which is also responsible to calculate average score, but as it is not getting used here that why it got missed. I'll look for all possible occurrences.

sandeepsajan0:
I just saw the implementation, it is different. It only calculates the avg review score of a user(reviewer) across all submissions. While the other one I hide is the average review score by all reviewers for a particular submission.

Actually, it becomes more of a question now. So, @frjo do we want to hide the personal avg score as well on the submissions results stats page?
image

frjo:
@sandeepsajan0 The results view is separate so leave it as is.

PR #3435, opened 2023-06-12
Use logger.warning/error instead of exception for slack errors

### LINKED ISSUE #3423: Fail nicely when configured Slack channel do not exist
~~~~
ChannelNotFound: ChannelNotFound: channel '#@surgesustain' could not be found  File "django_slack/tasks.py", line 12, in send
    return backend.send(*args, **kwargs)
  File "django_slack/backends.py", line 25, in send
    return self.validate(r.headers['content-type'], result, message_data)
  File "django_slack/utils.py", line 20, in validate
    raise klass(result['error'], message_data)
~~~~

sandeepsajan0:
Fixes #3423 

Use logger.error/warning because exception making noise in sentry.

### DIFF
--- a/hypha/apply/activity/adapters/slack.py
+++ b/hypha/apply/activity/adapters/slack.py
@@ -4,6 +4,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.utils.translation import gettext as _
 from django_slack import slack_message
+from django_slack.exceptions import ChannelNotFound
 
 from hypha.apply.activity.adapters.base import AdapterBase
 from hypha.apply.activity.adapters.utils import link_to, reviewers_message
@@ -372,7 +373,10 @@ class SlackAdapter(AdapterBase):
         for room in target_rooms:
             try:
                 slack_message('messages/slack_message.html', data, channel=room)
+            except ChannelNotFound as e:
+                logger.warning(e)
+                return '400: Bad Request'
             except Exception as e:
-                logger.exception(e)
+                logger.error(e)
                 return '400: Bad Request'
         return '200: OK'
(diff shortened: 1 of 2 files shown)


theskumar (changes requested):
@sandeepsajan0 Can you please update the docs around the slack config, and write that if the slack channels are not not found the messages will silently fail, with a error log.

https://github.com/HyphaApp/hypha/blob/main/docs/setup/administrators/configuration.md#slack-settings

hypha/apply/activity/adapters/slack.py
```diff
@@ -372,7 +373,10 @@ def send_message(self, message, recipient, source, **kwargs):
         for room in target_rooms:
             try:
                 slack_message('messages/slack_message.html', data, channel=room)
+            except ChannelNotFound as e:
+                logger.warning(e)
+                return '400: Bad Request'
```
theskumar, line 378: Won't this still lead to an error? whats' the difference b/w the `ChannelNotFound` and the general `Exception` handler, both return 400. Logger won't halt the execution anyway.

sandeepsajan0: @theskumar There wasn't any UI error or execution halt for the user. It is working in the same way as before, we just avoid the sentry noise. Previously we were using the logger.exception and it was hitting sentry, so for ChannelNotFound we just use warning to avoid sentry and error for other exceptions.
And I think returning and logging error is fine here until we implement the 'check channel existence by sending a message' feature. And as this is not a UI error, just logging, so not an issue i think.

frjo:
Since the error does not bother the user I think we keep it as is and close this PR.

PR #3437, opened 2023-06-13
Hide everything for no role user

### LINKED ISSUE #3368: Test to see what users accounts with no roles can access

sandeepsajan0:
Fixes #3368

### DIFF
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -4,6 +4,7 @@ from functools import partialmethod
 from django.apps import apps
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Group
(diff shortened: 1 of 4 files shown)


theskumar (changes requested):
Added a minor comment to make this DRY and performant.

hypha/apply/funds/models/submissions.py
```diff
@@ -567,6 +574,11 @@ def ensure_user_has_account(self):
                     email=email,
                     defaults={'full_name': full_name}
                 )
+                # Ensure applying user should have applicant role
+                applicant_group = Group.objects.get(name=APPLICANT_GROUP_NAME)
+                if applicant_group not in self.user.groups.all():
```
theskumar:
```suggestion
                if not self.user.is_applicant:
```

Use `is_applicant` method

Issue #3442, opened 2023-06-13
Not able to delete the user from wagtail admin


sandeepsajan0:
Even after deleting all the related submissions and projects, the staff admin/staff is not able to delete the user.

It is showing the following error msg on deleting the user:
<img width="800" alt="image" src="https://github.com/HyphaApp/hypha/assets/23638629/8c2f302e-df03-4a53-a792-f3d33646fc51">

wes-otf:
@frjo @sandeepsajan0 Did #3445 fix this/can it be closed? Or is further action required?

frjo:
If I remember correctly this issue is due to some left over items in "actions_*". So the error message are correct and we need to find a way to delete these last items.

sandeepsajan0:
@wes-otf It is kind of partially fixed. As Fredrik mentioned, the problem is the leftover actions for the submissions that already has been deleted.
So basically when we delete a user, we have to delete their submissions first and submissions take care of every other thing like reviews and determination but there is a submission creation action(indirectly connected event) that remains existed/undeleted so because of that action we were not able to delete the user. 
In #3445, we took care of that action and deleted that as well with submission deletion. But this PR only works for new submission deletions while previously deleted submissions' actions are still there in the system. So, for now, we can delete the users who are not connected with those deleted submissions but the ones who are connected we can't delete them.

I think we can create a custom migration that can delete all the actions for non-existing submissions and then I think we will be ready to delete any user from the system.

frjo:
@sandeepsajan0 Excellent summary of the issue. I think the custom migration fix is a good way forward.

wes-otf:
@sandeepsajan0 yeah this was great, thanks for that background! I can take a stab at that custom migration.

wes-otf:
@frjo @sandeepsajan0 Are we only worried about deleting Applicants accounts? Digging through the changes made in #3445 it looks like only `NEW_SUBMISSION` events were handled but when events like `ARCHIVE_SUBMISSION` (even on a deleted submission) or `DELETE_SUBMISSION` exist in the `activity_event` table with id of the intended user to delete, the deletion issue is seen.

For example, I created a submission as an applicant, then on a different admin account deleted the submission. I then attempted to delete said admin account, and ran into `The object you are trying to delete is used somewhere. Please remove any usages and try again!`. 

Is this the intended functionality or should should these events also be deleted/have their `by` ids changed? Maybe this would even somewhat tie into #3441.

frjo:
[...]

Issue #3441, opened 2023-06-13
Deactivate and anonymise user accounts


frjo:
Hypha need a feature for staff admins to deactivate user accounts.

- [ ] Available on user account page in Wagtail admin.
- [ ] Anonymise user name and user e-mail (uuid?)
- [ ] Also anonymise user name and user e-mail on all linked submissions
- [ ] 48 hour window to undo

PR #3445, opened 2023-06-17
Delete submission creation event(NEW SUBMISSION) on submission deletion

sandeepsajan0:
Possible Fix of #3442 

On submission deletion, all other activities, determination, and other things are getting removed but as this `NEW SUBMISSION` event is not directly connected to submission so not getting removed. It is creating issues at the time of user deletion even after submission deletion.

I was also thinking to show the model name, and object name in error itself but user deletion only seems to be dependent on `Activities` and `Events`(user can also be deleted only by removing Activities and events not submissions related to them), and even if we show staff the event or activities ids I don't think staff can easily delete those anyhow via UI. I think we need to decide a better way to delete a user, maybe issue #3441 help us?

### DIFF
--- a/hypha/apply/funds/services.py
+++ b/hypha/apply/funds/services.py
@@ -16,7 +16,7 @@ from django.db.models.functions import Coalesce
 from django.http import HttpRequest
 
 from hypha.apply.activity.messaging import MESSAGES, messenger
-from hypha.apply.activity.models import Activity
+from hypha.apply.activity.models import Activity, Event
 from hypha.apply.funds.models.assigned_reviewers import AssignedReviewers
 from hypha.apply.funds.workflow import INITIAL_STATE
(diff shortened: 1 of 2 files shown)


frjo (approved):

fourthletter:
I agree with your point about https://github.com/HyphaApp/hypha/issues/3441

sandeepsajan0:
@frjo Can we put this PR to the test site? Or it needs rebasing?

frjo:
Deploying to test now.

sandeepsajan0:
@frjo  I found that I missed deleting these events for batch delete action. I have fixed the PR. Can you please put its latest version to the test site?

sandeepsajan0:
@fourthletter It is ready for your testing. You can test this with a new user or with a user whose data hasn't been deleted yet.
I'll create a separate PR to make it work for all users.

Issue #3443, opened 2023-06-15
[Epic] Registration Form


fourthletter:
### Description

As an applicant I need to register and login into the platform. The current registration form does not contain the details I need to register for this site. 

### Example Registration
![image](https://github.com/HyphaApp/hypha/assets/20019656/47ce0a8e-9049-468f-8ce5-3dc80e69fbb3)

### GitHub Issues

- [x] NEW registered users are routed to their dashboard to apply rather than to their profile page
- [x] Always send e-mail after non logged in users apply or create an account #3370
- [x] Alternative workflows for registration https://github.com/HyphaApp/hypha/pull/3156
- [x] remove registration/login from nav bar https://github.com/HyphaApp/hypha/pull/3156
- [x] Update authentication https://github.com/HyphaApp/hypha/pull/3457
- [x]  Add rate limit https://github.com/HyphaApp/hypha/pull/3493

theskumar:
> If a user has an existing account, the user should be routed to the login page
@fourthletter IMO, I would strongly advise not to put this requirement in the current registration flow. The current error is very explicit and widely adopt pattern to say that user account already exist and then user need to manual decide if they want to register using a different email or would like to login using that email.

Are you able to remove auto redirect to login as feature from this request please. 

>  Person registering receives instructions on the length of the password
This is show during the account activations and password change form. So it's not very specific to registration itself. @fourthletter @frjo do you have have specific wording or functionality in mind that you would like to implement.

theskumar:
@fourthletter  Other request in this issue has been added to the PR #3457 please take a look at the screenshot attached, and see if they all make sense. 

PS: That PR requires the https://github.com/HyphaApp/hypha/pull/3432 to be merged first, as it built on top of the changes introduced to the authentication and registration during the nav removal.

PR #3456, opened 2023-06-25
Add question on the register page, if user already have an account

theskumar:
Screenshot 2023-06-25 at 13 57 16@2x

Ref: #3443

### DIFF
--- a/hypha/apply/users/templates/users/register.html
+++ b/hypha/apply/users/templates/users/register.html
@@ -22,7 +22,7 @@
         <button class="link link--button-secondary" type="submit">{% trans "Register" %}</button>
     </form>
     <p>
-        Already have an account? <a href="{% url 'users_public:login' %}">Log in</a>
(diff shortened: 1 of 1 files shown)


hypha/apply/users/templates/users/register.html
```diff
@@ -21,5 +21,8 @@
         {% endfor %}
         <button class="link link--button-secondary" type="submit">{% trans "Register" %}</button>
     </form>
+    <p>
+        Already have an account? <a href="{% url 'users_public:login' %}">Log in</a>
```
frjo: Good addition, please make the strings translatable and then we can merge it in.

frjo (changes requested):

hypha/apply/users/templates/users/register.html
theskumar: Nice catch! Added :)

frjo (approved):

PR #3464, opened 2023-07-04
Replace custom css for form elements on apply with Tailwind

### LINKED ISSUE #3439: Replace custom css for form elements on apply with Tailwind
I noticed that Hypha custom CSS in `_form.scss` stacks on top of Tailwind form styling a bit in places.

I think we can simply remove a lot of the custom code and fall back on stock Tailwind. Maybe adjust colours etc. only.

theskumar:
Fixes: #3439

### DIFF
--- a/hypha/static_src/src/sass/apply/abstracts/_mixins.scss
+++ b/hypha/static_src/src/sass/apply/abstracts/_mixins.scss
@@ -231,18 +231,7 @@
     input[type='checkbox'] {
         margin: 0 auto;
         display: block;
-        width: 20px;
-        height: 20px;
-        border: 1px solid $color--mid-grey;
-        -webkit-appearance: none;
-        -moz-appearance: none;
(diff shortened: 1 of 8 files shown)


frjo:
Nice work!

The old radio and checkbox styles have a blue border. That is in this PR replaced with a more standard grey and they are slightly smaller as well.

I do not have a preference for one over the other but I prefer not to change things to much, unless we have a good reason.

Skärmavbild 2023-07-05 kl  10 26 51
Skärmavbild 2023-07-05 kl  10 27 11

theskumar:
It could be the missing accent color on the check box, let me take a stab at it. Definitely not intensional, i get your point.

theskumar:
Updated in 4eff69c4d
Screenshot 2023-07-05 at 08 51 01@2x

frjo (approved):
Perfect!

PR #3470, opened 2023-07-09
Remove submission-tooltips.js

theskumar:
Use tippy.js instead and css truncate instead of truncation by Django

See: table.py for implementation


frjo:
The only issue I see is that you also get a standard tooltip from the browser. I think it is the "overflow" css that trigger it. There is no tittle attribute on the links.

Skärmavbild 2023-07-10 kl  07 08 08

theskumar:
I'll take a look

theskumar:
Are you using safari, this seems to be know issue in safari and doesn't have an easy solution. 

I'll leave it up to you to if you the extra tooltip is acceptable or not. I don't see this in chrome based browsers. 

https://stackoverflow.com/questions/20974276/prevent-safari-from-showing-tooltip-when-text-overflow-is-hidden-with-ellipsis

frjo:
Yes, it is Safari. Not a big issue I think.

frjo (approved):

theskumar:
@frjo Added fix for safari default title issue at 
https://github.com/HyphaApp/hypha/pull/3470/commits/28cddec82a0a2a8a9142e8d25ad45f73d174d0fe

fourthletter:
Submission title is still truncated, but maybe I don't understand this PR

image

theskumar:
> Submission title is still truncated, but maybe I don't understand this PR

This PR is just internal refactoring to remove the remove the use of `submission-tooltip.js` without changing any existing behaviour. 

If you think the truncation is unnecessary, then that would become a feature request.

theskumar:
If there is no truncation, the tooltip displayed also be not needed here. 

PS: This PR is an effort to clean up some of the old tooling used in this project.

fourthletter:
Thank you for clarifying.

Issue #3448, opened 2023-06-21
Suggestions from the project's user testing


sandeepsajan0:
**Part-1(miro board cards)**:
- [x] PAF waiting for approval table shouldn't show if no project in any tab.
- [x] Show requested by as well in PAF Approvals(sidebar). Right now only Pending and Approved are showing.
- [x] Remove the converted option from the invoice.
- [x] Show activities on the Invoice page.
- [x] Show price in US format.
- [x] Downloaded PAF add dates along with approver names, sync it with platform.
- [x] "Add Invoice' as next step even after invoice creation(remove next step after one invoice).
- [x] Add hyperlinks to comments and use those links for scrolling(request change comments)

**Part-2(figma design changes)**
- [x] #3458
- [x] #3462 
- [x] #3467
- [x] #3469
- [x] Update invoice table column name on project detail page, "Submitted" to "Date Submitted".
- [x] Update "Reporting" section header to "Reports".
- [x] Change Upload Contract modal "submit" button text to "Upload" for Applicant.
- [x] Update Next step text for "Awaiting Contract approval from Staff" to ".... approval by org_short_name"(for applicant role).
- [x] In "Submit for Approval" or "Resubmit" PAF modal, make optional line bold and update the text as per design.
- [x] "Invoice Uploaded" to "Invoice added" in alert/success message text on invoice creation.
- [x] "Create Invoice" to "Add invoice" on invoice form heading.
- [x] Update next step for applicant during PAF approval process, set it to a single/fix text.(see design for text)
- [x] Update Admin bar's "Download" button text fot PAF and SOW to "Download PAF" and "Download SOW".
- [x] Update PAF edit page submit button text to Save.
- [x] Update "Invoice created" to "Invoice added" in activities.
- [x] "View <related_object>" in activity feed instead of just <related_object>. Ex: Contract -> View Contract
- [x] "Edit" and "Delete" to "Edit Invoice" and "Delete Invoice" on the Invoice detail page.
- [x] Update applicant to vendor at every place in projects.
- [x] #3474
- [x] #3479
- [x] #3477
- [x] Update Approval section of PAF detailed page/pdf. Show as per PAF approvals sidebar section(include pending and request changes).
- [x] #3480
- [x] #3494
- [x] #3495
- [ ] - [ ] Make comment required if a user selects request_changes on updating paf status.

frjo:
We can not hardcode US formatting of currency values in Hypha. There must be a setting for it.

[...]

PR #3472, opened 2023-07-10
Update UI, and text for Contracting documents section

### LINKED ISSUE #3458: Update Contracting documents section
Fixes in the Contracting documents section:

- [ ] Replace 'Contracting team' by 'org_short_name' in text.
- [ ] Section update, margin, and tick marks.
- [ ] Update text, add 'Pending' and other things.
- [ ] Final tick to Contract only after submission of contracting documents to staff.
- [ ] Update next step for applicant/Vendor.
image
image

sandeepsajan0:
Fixes #3458  Refs: #3448

For staff:

image

For Vendor:
image

After contract and document upload:
for staff:
image

for vendor:
image

### DIFF
--- a/hypha/apply/projects/templates/application_projects/includes/contracting_documents.html
+++ b/hypha/apply/projects/templates/application_projects/includes/contracting_documents.html
@@ -61,8 +61,8 @@
                     <div class="docs-block__row-inner">
                     </div>
                     <div class="docs-block__row-inner">
+                        <svg class="icon icon--info-circle"><use xlink:href="#info-circle-fill"></use></svg>
                         <a class="docs-block__icon-link" href="#communications" target="_blank">
-                            <svg class="icon icon--info-circle"><use xlink:href="#info-circle-fill"></use></svg>
                                 <span>{% trans "Corrections/amendments?" %}</span>
                         </a>
                     </div>
@@ -144,8 +144,8 @@
                                     {% endif %}
                                     {% if document_category.required %}<span class="text-red-700">*</span>{% endif %}
                                     {% if document_category.template %}
-                                    <a class="pl-2 font-bold underline decoration-dashed" href="{% url 'apply:projects:category_template' pk=object.pk type='contract_document' category_pk=document_category.pk %}" target="_blank">{% trans "View template" %}</a>
-                                    <svg class="icon icon--info-circle pl-0.5"><use xlink:href="#info-circle-fill"></use></svg>
+                                    <svg class="icon icon--info-circle"><use xlink:href="#info-circle-fill"></use></svg>
+                                    <a class="font-bold underline decoration-dashed" href="{% url 'apply:projects:category_template' pk=object.pk type='contract_document' category_pk=document_category.pk %}" target="_blank">{% trans "View template" %}</a>
                                     {% endif %}
                                 </p>
                                 {% if document_category not in remaining_contract_document_categories %}
--- a/hypha/apply/projects/templates/application_projects/includes/contracting_documents.html
+++ b/hypha/apply/projects/templates/application_projects/includes/contracting_documents.html
@@ -50,22 +50,11 @@
             </div>
             {% if contract_uploaded %}
             <div class="docs-block__row-inner">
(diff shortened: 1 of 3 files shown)


theskumar:
@sandeepsajan0 Can you please a screenshot of the UI changes?

sandeepsajan0:
@theskumar I have updated the description with changed UI screenshots. It is as per the design suggested in figma.

@abhiXII you may also have a look at the screenshots and confirm if everything seems fine. 
I left a horizontal line implementation between signed and countersigned statements because I think last time we removed it so that it can look like one body or section. Ex: Approval Form and SOW Form.

hypha/apply/projects/templates/application_projects/includes/contracting_documents.html
```diff
@@ -50,22 +50,11 @@ <h5> {% trans "Are you sure you want to submit contracting documents?" %}</h5>
             </div>
             {% if contract_uploaded %}
             <div class="docs-block__row-inner">
-                <a class="docs-block__icon-link" href="{% url 'apply:projects:contract' pk=project.pk file_pk=contract.pk %}" target="_blank">
+                <a class="font-bold flex items-center w-auto text-left bg-white text-light-blue mr-0 mb-1 p-2.5 border-solid border border-light-blue" href="{% url 'apply:projects:contract' pk=project.pk file_pk=contract.pk %}" target="_blank">
```
theskumar: As the button is quite a re-usable UI element consider using `button button--white button--half-width` or it's variation to keep the look consistent. 

Also, the button styling misses focus and hover state. 

Screenshot 2023-07-11 at 04 29 27@2x

theskumar (changes requested):

hypha/apply/projects/views/project.py
```diff
@@ -319,6 +319,7 @@ def get_context_data(self, **kwargs):
         context['contract_to_sign'] = contract_to_sign
         context['contracts'] = contracts.approved()
         context['contract'] = latest_contract
+        context['org_short_name'] = settings.ORG_SHORT_NAME
```
frjo: Do we need to set this here when we have it in:

https://github.com/HyphaApp/hypha/blob/bf4cd5416d32cb43a358bd202067a63c02378035/hypha/core/context_processors.py#L6-L21

hypha/apply/projects/templates/application_projects/includes/contracting_documents.html
sandeepsajan0: I agree, we have project-actions--white as well but this is a custom button, we have a few other custom buttons as well like Upload, and Reupload contracts. These buttons look a bit different(more padding) from others just to get the user's attention.
 
Actually, these have to be links like other View options but these actions are slightly more important than links and less important than buttons.

I have added focus and hover state with bg and text colour change.

sandeepsajan0:
The hover state for the icon is creating issues for other 'View' links. I'll fix it.
https://github.com/HyphaApp/hypha/blob/4a1255d4f7a810c3f2a4ebd083849e9c08a07fa8/hypha/static_src/src/sass/apply/components/_icon.scss#L126-L128

abhiXII:
Instead of contracting team, we're going to say OTF but I think this is editable on Wagtail so we can do that. 
without the line looks fine too. Thanks

Very minors things I noticed for consistency were: the (i) icon with `corrections/ammendments?' should also be grey and the (i) icon with view template should be on the left of view template text. so that all (i) icons look and feel same on the page.

Also I believe we're not using time anywhere? I'm okay with it being left there but you can see as per the system consistency.

sandeepsajan0:
>Instead of contracting team, we're going to say OTF

It is OTF for vendors/applicants, 'contracting team' is only for staff/team.

I have made the info (i) icon consistent.  For now, let's keep time for the contracts, we can change it anytime if needed. it is just a small change.

@frjo I think we can put its latest version to the test site.

frjo:
Deploying last to test now.

fourthletter:
@abhiXII - thank you for taking a look!

PR #3478, opened 2023-07-12
Update typography to be consistent and scaled properly

theskumar:
Use reference scale from tailwind to update the styles of different
headers.

Overall the fonts for headers were little too large and they are now
little scaled down so they don't look blown up.

Sensible formatting is applied to form elements, esp. help-text is made
smaller.

Adminbar component used on even more pages.

PS: A lot of files are touched and also a lot of code cleanup can still
be done. So for now the best way to test this would be to navigate
through the pages pages and find any obvious issues.


hypha/apply/determinations/templates/determinations/determination_detail.html
```diff
@@ -4,42 +4,43 @@
...
 
 {% block content %}
-<div class="admin-bar">
-    <div class="admin-bar__inner">
-        <a class="simplified__submissions-link" href="{{ determination.submission.get_absolute_url }}">
-            {% trans "Back to submission" %}
-        </a>
-        <h2 class="heading heading--no-margin">{% trans "Determination" %} {% if determination.is_draft %}[{% trans "DRAFT" %}]{% endif %}
-        </h2>
-        <h5>{% trans "For" %}: {{ determination.submission.title }}</h5>
```
frjo, line 14: I think the submission title is useful here. We have it on reviews detail page as well. On reviews it includes applicant name and submission date as well Do not know if that is needed, butI think they should look the same.

Without the submission detail you do not know what submission the determination is for unless you click the "Back to submission" link.

hypha/apply/determinations/templates/determinations/determination_detail.html
```diff
@@ -4,42 +4,43 @@
...
-    </div>
-</div>
+    {% adminbar %}
+        {% slot back_link %}
+            <a class="simplified__submissions-link"
+                href="{{ determination.submission.get_absolute_url }}"
+            >
+                {% trans "Back to submission" %}
```
frjo: On the project side we changed the "Back to …" link text to "View …". I think it would be good to make that change on the submission side as well.

hypha/static_src/src/sass/apply/base/_typography.scss
```diff
@@ -45,32 +42,39 @@ h6,
 h1,
 %h1,
 .alpha {
-    @include responsive-font-sizes(40px, map-get($font-sizes, alpha));
-    font-weight: $weight--black;
+    font-size: 2.25rem;
+    line-height: 2.5rem;
+    font-weight: 400;
```
frjo, line 47: Why not use the variables for weight here as well?

hypha/static_src/src/sass/apply/components/_heading.scss
```diff
@@ -6,16 +6,12 @@
...
-            margin-bottom: 1rem;
-        }
+        margin-top: .25rem;
 
         span {
             &::after {
-                padding: 0 15px;
-                content: '|';
+                padding: 0 .25rem;
+                content: '•';
```
frjo, line 14: I personally prefer the pipe over the bullet but let's see what staff thinks. Most likely they do not even notice the change :-).

hypha/static_src/src/sass/apply/components/_all-submissions-table.scss
```diff
@@ -15,10 +15,8 @@
 
         th {
             &.reviews_stats { // stylelint-disable-line selector-class-pattern
-                color: $color--black-60;
-
                 div {
-                    font-size: 13px;
+                    font-size: 12px;
```
frjo: When possible I prefer to use "rem" consistently for font size now that we are overhauling it.

frjo (changes requested):
Overall I think this is very good changes. Have a few things I believe need fixing and a number of minor comments.

I  approve of the switch over to css variables. We could add some code that converts sass variables to css variables so we have both automatically. I use code like this in many projects:

~~~~
:root {
  // Convert sass variables to css4 variables.
  @each $name, $color in $colors {
    --color-#{$name}: #{$color};
  }

  @each $name, $size in $font-size {
    --fs-#{$name}: #{$size};
  }

  @each $name, $weight in $font-weight {
    --fw-#{$name}: #{$weight};
  }
}
~~~~

hypha/apply/determinations/templates/determinations/determination_detail.html
theskumar: Agreed, updated.

frjo (approved):

hypha/static_src/src/sass/apply/components/_all-submissions-table.scss
theskumar: Updated.

hypha/static_src/src/sass/apply/components/_heading.scss
theskumar: Sure.

hypha/static_src/src/sass/apply/base/_typography.scss
theskumar: I was thinking to move and refractor these in a more meaningful and thoughtful way in another PR. Let me know your thoughts.

theskumar:
Rebased and update some of the layout issues on the 2FA page. 
Backup Codes
Backup Codes (1)

PR #3481, opened 2023-07-13
Fix alignment of upload icon

theskumar:
Screenshot 2023-07-13 at 08 40 37@2x

### DIFF
--- a/hypha/static_src/src/sass/apply/components/_form.scss
+++ b/hypha/static_src/src/sass/apply/components/_form.scss
@@ -108,8 +108,12 @@
                 display: none;
             }
 
+            .form__required {
+                color: $color--white;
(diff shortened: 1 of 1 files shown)


sandeepsajan0 (commented):
I think we need to update the same for the hover state as well.

https://github.com/HyphaApp/hypha/blob/5e9bb5554c26f24a813d640bd41d0deb9aa12408/hypha/static_src/src/sass/apply/components/_form.scss#L111-L112

theskumar:
Fix hover state. Used matching color for button hover and also updated to the font color for `*`

frjo (approved):

sandeepsajan0 (approved):

PR #3482, opened 2023-07-14
Use `wagtailsiteupdate` to set hosts and port

bickelj:
When following docs to set up a development environment, without this change one would see this error:

> Unknown command: 'wagtailupdatesite'. Did you mean
> wagtail_update_index?


frjo:
Thanks for this doc bug fix!

PR #3484, opened 2023-07-16
Remove hx-boot from the primary nav

theskumar:
This fixes the issue where the select2 widget doesn't initialized
properly during the page load.

The select2 internals are dependent on the window.onload and the is not
easy way to switch to activate on htmx.onload event.

Consequently, the hx-boost is removed from the primary navs so the
select2 behaves as expected.


frjo (approved):

frjo:
A pity this is needed but the filter issue must be irritating to staff.

PR #3485, opened 2023-07-16
Rebuild comment visible to option for Staff  and hide for everyone except

### LINKED ISSUE #3486: Rebuild the comment visible to functionality
Remove the "Visible to" functionality for applicants, partners and reviewers. Only staff (staff + finance + contracting) need to have these options.

- [ ] When an applicant post a comment it should be visible to the applicant themself, any partners and to staff.
- [ ] When an partner post a comment it should be visible to the applicant themself, any partners and to staff.
- [ ] When a reviewer post a comment it should be visible to the reviewers and to staff.
- [ ] Remove the partner option from "Visible to". Partners should see the same comments as the applicant themself.
- [ ] The staff default for "Visible to" should be that only "staff" are checked.
- [ ] The "reviewer" and "applicant" checkbox is left empty be default for staff.
- [ ] No other options but the three above are needed for "Visible to".

sandeepsajan0:
Add an option for the finance and contracting team

Fixes #3486

Applicant View:

image

Staff View:

image


sandeepsajan0:
@frjo @theskumar In this PR, I tried to achieve our goal #3486  with the current implementation(with just slight changes).
1.  When an applicant post a comment it should be visible to the applicant themself, any partners and to staff. (no visible to needed)
2. When an partner post a comment it should be visible to the applicant themself, any partners and to staff. (no visible to needed)
3. When a reviewer post a comment it should be visible to the reviewers and to staff. (no visible to needed)
4. Remove the partner option from "Visible to". Partners should see the same comments as the applicant themself.
5. 'Visible to' options are there for Staff/Team only(including finance and contracting).
6. Added help text to 'Visible to' to make it a bit more clear for staff/team.
7. Update visibility text in the comment line. Ex: applicant + OTF team for applicant and applicant + team for staff.

Actually, I also tried Arrayfield(for multiple choices) for visibility but faced some issues because of the existing code and values of the visibility field. It requires some extra work to update db values and update code around visibility. So for now I tried to make it as simpler as possible. 
You both may try it locally and let me know how it feels now.

sandeepsajan0:
@frjo I have fixed the tests. I think we can put it on the test site to test it out for different user roles.

fourthletter:
This was also done for Reviewer -- here is their view

image

fourthletter:
Sandeep - what you've done was great, but I think we need to rework the verbiage for the activity feed. 

We are calling individuals 'Partner', 'Applicant', and even through they are not 'team' it's still confusing for them to see this.

image

PR #3493, opened 2023-07-22
Add RateLimit to registrations

frankduncan:
A vulnerability was introduced with #3064 where registrations could be created en masse using arbitrary email addresses.  An attacker could use this to make hypha auto generate mass emails to those recipients. Adding a ratelimit to the registration should prevent that from being too aggressive.


theskumar (approved):
Looks good to me! 💯

PR #3501, opened 2023-07-28
Add public statuses for invoices

### LINKED ISSUE #3495: Update/Hide invoice statuses on project detail page for Applicant/Vendor.(see design)
As per design sugestions:

Invoice status for applicants/vendors should only be:

    Pending approval
    Approved (when it is approved by staff and finance)
    Declined
    Request for change or more information
    Paid
    Payment failed (new status to be added for finance when a payment goes wrong because of some errors)

These changes has to be done on project detail page.
image

All these statuses need to be designed over the current statuses as wrapper for applicant like APPLICANT_INVOICE_STATUS.

Payment failed is a new status, we don't have it in our current system. Do we need it throughout the system like for staff and finance as well? @fourthletter @frjo

sandeepsajan0:
Fixes #3495 

- [x] Add pubic statuses to invoices.
- [x] Needs to update these statuses for activities as well.
- [x] Fix status history for staff and applicant.
- [x] Remove sidebar activities from invoice detail page(noticed bug on test site).
- [x] Status history section doesn't have any status for applicants. (noticed bug on test site)


fourthletter:
The applicant is still able to see comments and the activity feed with activities completed by staff. This may require a separate GitHub issue.

sandeepsajan0:
@fourthletter We have changed the visibility for a few internal invoice activities only not for all staff activities. Because some activities like invoice transitions should be shown to the applicant. 
And also this will work only for the new activities, old activities are there as it is.

I checked it on the test site, seems fine to me.

Staff activities:
image

Applicant activities:
image

@fourthletter Above activities are from [project 7](https://test-apply.hypha.app/apply/projects/7/) for the same invoice, you may also have a look and let me know if it doesn't look good to you, we can update the things.

PR #3502, opened 2023-07-30
Add pre-commit with black,prettier and djhtml

### LINKED ISSUE #3500: Start using black and prettier for code styling, linting is for real errors only
As far as possible we use whatever defaults black and prettier has. Many projects, including Django, uses these. Will be good to have Hypha code look the same.

Also use djhtml for templates. https://github.com/rtts/djhtml

theskumar:
- Add black, djhtml, prettier
- Add precommit

This PR add configurations for black, djhtml and prettier.

- black is used for formatting python files
- djhtml for formatting html files. Use `{# fmt: off #}` and `{# fmt: on #}` where needed.
- prettier for formatting of sass,css,yml and js files

Note: prettier picks up some of it's config from `.editorconfig`, so it's updated appropriately

Formating of mardown files, minified or vendored js files are ignored.

Formatting used off-the-self rules for better consistency with the community as large.

The formatting is not applied to all the files as it will introduce a lot of merge conflict with
current PRs, instead `pre-commit` config is provide which can be installed by `precommit install` 
locally. It will format only the modified files in the PR/commits. 

Once majority of the current active PRs are either rebased & formatted using precommit or merged, we can
run the `make fmt` on the full repo and activate additional prettier and black checks in our lint rules.

The modified sass files fixes the eslint error that would came up if we were to apply the prettier formatter on them. 

Fixes #3500

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,6 +10,9 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.{yml,yaml,json}]
+indent_size = 2
+
 [*.md]
 trim_trailing_whitespace = false
 
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,4 +1,6 @@
-extends: eslint:recommended
+extends:
+  - eslint:recommended
+  - prettier
 env:
   browser: true
   commonjs: true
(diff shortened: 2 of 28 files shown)


frjo (approved):

theskumar:
Added missing docs and rebased with master

PR #3514, opened 2023-08-05
Remove hardcoded org name, and use configured name

sandeepsajan0:
Used ORG_SHORT_NAME instead of hardcoded org short name in next steps.

### DIFF
--- a/hypha/apply/projects/templatetags/project_tags.py
+++ b/hypha/apply/projects/templatetags/project_tags.py
@@ -34,14 +34,14 @@ def user_next_step_on_project(project, user, request=None):
                 return "Resubmit project documents for approval"
             return "Submit project documents for approval"
         elif user.is_applicant:
-            return "Awaiting project documents to be created and approved by OTF internally. " \
+            return f"Awaiting project documents to be created and approved by {settings.ORG_SHORT_NAME} internally. " \
                    "Please check back when the project has moved to contracting stage."
         if project.paf_approvals.exists():
             return "Changes requested. Awaiting documents to be resubmitted."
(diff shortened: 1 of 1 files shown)


hypha/apply/projects/templatetags/project_tags.py
```diff
@@ -34,14 +34,14 @@ def user_next_step_on_project(project, user, request=None):
                 return "Resubmit project documents for approval"
             return "Submit project documents for approval"
         elif user.is_applicant:
-            return "Awaiting project documents to be created and approved by OTF internally. " \
+            return f"Awaiting project documents to be created and approved by {settings.ORG_SHORT_NAME} internally. " \
```
frjo: These strings should be translatable as well.

hypha/apply/projects/templatetags/project_tags.py
```diff
@@ -29,72 +30,72 @@ def user_next_step_on_project(project, user, request=None):
...
+                return _("Fill in the Approval Form(PAF)")
             if project.paf_approvals.exists():
-                return "Resubmit project documents for approval"
-            return "Submit project documents for approval"
+                return _("Resubmit project documents for approval")
+            return _("Submit project documents for approval")
         elif user.is_applicant:
-            return f"Awaiting project documents to be created and approved by {settings.ORG_SHORT_NAME} internally. " \
-                   "Please check back when the project has moved to contracting stage."
+            return _(f"Awaiting project documents to be created and approved by {settings.ORG_SHORT_NAME} internally. "
```
frjo: I do not think f-strings works with translations, you need to use `.format().` instead.

PR #3520, opened 2023-08-09
Remove custom fields from paf and hide paf & sow attachments in their downloaded versions

### LINKED ISSUE #3518: Remove hard coded fields from PAF
We have a few hardcoded fields on the top of PAF. Some of them get duplicated by the wagtail form fields so decided to remove hardcoded ones.

Hardcoded fields:

- [x] 1. `Proposed start date`
- [x] 2. `Project Proposed end date`
- [x] 3. `Legal name`
- [x] 4. `E-mail`
- [x] 5. `Address`
- [x] 6. `Phone`
- [x] 7. `Value`
- [x] 8. `Sent to Compliance`

image

@abhiXII @frjo @fourthletter Which ones from the above list do we want to remove or just want to use from the wagtail form?

sandeepsajan0:
Fixes #3518

### DIFF
--- a/hypha/apply/projects/templates/application_projects/project_approval_detail.html
+++ b/hypha/apply/projects/templates/application_projects/project_approval_detail.html
@@ -33,50 +33,6 @@
             <article class="wrapper--sidebar--inner">
                 <h4>{% trans "Project Information" %}</h4>
                 <div class="card card--solid">
-                    <div class="grid grid--proposal-info">
-                        <div>
-                            <p class="font-bold">{% trans "Proposed start date" %}</p>
-                            <p>{{ object.proposed_start|date:"DATE_FORMAT"|default:"-" }}</p>
-                        </div>
-
-                        <div>
-                            <p class="font-bold">{% trans "Project Proposed end date" %}</p>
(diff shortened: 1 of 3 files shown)


hypha/apply/projects/templates/application_projects/project_approval_detail.html
```diff
@@ -33,50 +33,6 @@ <h5 class="simplified__subheading">
...
-                        </div>
-
-                        <div>
-                            <p class="font-bold">{% trans "E-mail" %}</p>
-                            <p>{{ object.contact_email|default:"-" }}</p>
-                        </div>
-
-                        <div>
-                            <p class="font-bold">{% trans "Address" %}</p>
-                            <p>{{ object.get_address_display|default:"-"}}</p>
```
frjo, line 59: Address come from vendor model when that is in use. Should it really be removed?

sandeepsajan0: I checked the PAF and 'address' field is coming from PAF as well so we decided to remove it and to show only the PAF fields. We can add it easily if needed, but for now, it looks like a duplicate field [here is image](https://github.com/HyphaApp/hypha/issues/3518#issue-1839350962).

If both of these are different addresses then we need to rename atleast one of them.

PR #3531, opened 2023-08-17
Passwordless login and signup

theskumar:
Fixes #ISSUEID

This PR is depended on #3521 

- [x] Passwordless login
- [x] Passwordless signup 
- [x] Allow user to set a password after going to profile.
- [x] Allow user to change their email even if they don't have an email set.
- [x] Allow user to add their name in the application form if name is not present in the user account.
- [x] Don't display "Dashboard" link if the user does't have permission to access to it.
- [x] Allow to use to setup 2FA without account password.
- [x] Display user content on the login screen, if configured (it is an existing feature)
- [x] If 2FA is enforced, allow the user to submit the application without setting up 2FA
- [x] Add email re-verification option to elevate, sudo mode, apart from password
- [x] Update landing page after application submission, on success it redirects now.
- [x] Update ENABLE_PUBLIC_SIGNUP and FORCE_LOGIN_FOR_APPLICATION to true by default

# Login/Signup Flow

image

## Updated Login Page with Registration Enabled

Screenshot 2023-09-13 at 07 53 06@2x

## After providing the email ID

The messaging is kept neutral to hide if the user is already registered or not. The email will contain more detail, if the account exist or not.

Screenshot 2023-09-13 at 07 57 27@2x

Login email copy
Screenshot 2023-09-13 at 08 08 36@2x

## Signup

New Account Email copy
Screenshot 2023-09-21 at 07 08 52@2x

### Profile Page just after signup 
The user after clicking on the signup link in the email is redirect to homepage. No dashboard is available as the user doesn't have applicant role. If they click on the "profile" button they see this page with open to update profile and setup a password and enable 2FA. 

If the user decide to change the email, password is not asked if not password is set, instead an email is sent to authorize the email change.

Screenshot 2023-09-22 at 08 50 21@2x

### For account with password

Screenshot 2023-10-31 at 7  49 38@2x

After clicking on the "Send a confirmation code to your email" link

Screenshot 2023-10-31 at 8  03 28@2x

Screenshot 2023-10-31 at 7  51 03@2x

### For account without password

Screenshot 2023-10-31 at 7  53 44@2x

## Updated disable 2FA page

It requires "Sudo" mode, instead of password now. 

Screenshot 2023-10-31 at 7  48 01@2x


theskumar:
> If 2FA is enforced, allow the user to submit the application without setting up 2FA

It’s bit tricky because the way middleware for it is designed. https://github.com/HyphaApp/hypha/blob/main/hypha/apply/users/middleware.py#L40-L44
The application forms doesn’t have a url pattern, and would require a db call on each of the request that i’m trying to avoid. 

Currently trying to see if we can `mfa_exempt` a view with a decorator similar to `csrf_exempt` and catch that in the middleware before processing the view.

frjo:
@wes-otf To mimic OTF Live set `ENFORCE_TWO_FACTOR`, `FORCE_LOGIN_FOR_APPLICATION` and `ENABLE_REGISTRATION_WITHOUT_APPLICATION` to true

wes-otf:
Still going through the code with a fine tooth comb to wrap my head around it but after pulling and playing with it for a while all looks/works really good as far as I've seen! I think the flow of everything is very intuitive. Cool stuff!

@frjo just a heads up it looks like `ENABLE_REGISTRATION_WITHOUT_APPLICATION` was swapped with `ENABLE_PUBLIC_SIGNUP`. When I switched these I was able to see the "Register" option.

hypha/settings/base.py
```diff
@@ -134,13 +134,9 @@
 TRANSITION_AFTER_REVIEWS = env.bool("TRANSITION_AFTER_REVIEWS", False)
 
 # Forces users to log in first in order to make an application.  This is particularly useful in conjunction
-# with ENABLE_REGISTRATION_WITHOUT_APPLICATION
+# with ENABLE_PUBLIC_SIGNUP
 FORCE_LOGIN_FOR_APPLICATION = env.bool("FORCE_LOGIN_FOR_APPLICATION", False)
```
frjo: Should we rename and revert this setting?

theskumar: Yes. we can do it, but can also make it backward compatible. But the enable_registration_without_application won't/doesn't make much sense.

frjo: I meant the `FORCE_LOGIN_FOR_APPLICATION` setting. Switch it from default False to default True. I also think it can have a better name, especially with a new default.

frjo: I like the "ENABLE_PUBLIC_SIGNUP" change.

theskumar: Ah. that totally make sense! We can go for it. Can't think of a better name, but going we can potentially just remove it.

frjo:
For the new flow to work we need both `FORCE_LOGIN_FOR_APPLICATION` and `ENABLE_PUBLIC_SIGNUP` to the set to "True". By default they are set to "False".

1. I think there is a case for setting both of these as "True" by default. This is a big change in Hypha but it is what most everyone is using in any case.
2. I think these two settings should be right next to each other in `base.py`.

frjo:
Moved this out of draft/WIP. It works far to well for that.

theskumar:
Sure. that sounds good. Just finalizing the sudo mode/confirm access without password. It will send a code/url in the email to confirm access and elevate the permission.

theskumar:
This is ready to testing now. Update the "Sudo" mode work without password. 

Also, updated 2FA flows to not depend on the passwords. @frjo @wes-otf @Techslammer

theskumar:
Updated description with screenshots

frjo:
Will put this on test after today's testing. This will be part of a very nice Hypha 5.1 release!

theskumar:
@frjo Updated the login with password screen so that the `Forgot your password?` is just below the password field. 

Also, update the context text to be smaller/de-emphasized. 

Screenshot 2023-11-20 at 7  17 43@2x

PR #3533, opened 2023-08-22
Add assign approver option to paf detail page

sandeepsajan0:


sandeepsajan0:

abhiXII:
sorry, a minor observation please: 

let's make the button style same on both pages? the white one with outline is preferable.
image
image

abhiXII:
ad finance, I don't see any options on the PAF page
image

was that purposeful? 🤔 

it shows up after assigning someone 
image

PR #3547, opened 2023-08-31
Update code-style using 'pre-commit run --all-files'

theskumar:
Add black and prettier check to lint 

Install the git hook in your local repo with:

`pre-commit install`

Check and fixes will then be done on each commit. Never get a lint warning again!


frjo (approved):
I just added a missing hyphen to the command in the title.

PR #3550, opened 2023-09-04
Skip unique email message on email update

sandeepsajan0:
Fixes https://github.com/HyphaApp/hypha-issues/issues/17

### DIFF
--- a/hypha/apply/users/forms.py
+++ b/hypha/apply/users/forms.py
@@ -90,6 +90,12 @@ class ProfileForm(forms.ModelForm):
 
         return slack
 
+    def clean_email(self):
+        email = self.cleaned_data['email']
+        if User.objects.filter(email=email).exists():
+            email = self.instance.email  # updated email to avoid email existing message, fix information leak.
+        return email
+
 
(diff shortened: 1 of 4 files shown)


hypha/apply/users/forms.py
```diff
@@ -84,6 +84,12 @@ def clean_slack(self):
 
         return slack
 
+    def clean_email(self):
+        email = self.cleaned_data['email']
+        if User.objects.filter(email=email).exists():
```
theskumar, line 95: Let's not put any business logic in the `clean_email`, any logic should go the view layer or service layer. when you are assigning `email = self.instance.email` things get very confusing down the line.

sandeepsajan0: We have to deal with it in the form else the form will raise the field's unique validation issues. And `validate_unique` is a general method for all fields so I decided to go with something related to the email field only. 

We can just shift this piece of code to another method in services and call that method here but it is just a small code and I have also added a comment for this so looks acceptable. Let me know, if it is really very confusing than we can use that service method.

PR #3564, opened 2023-09-08
Add code-style checks to Github CI Actions

theskumar:
Easier way to check and fix formatting issues if don't want to go
precommit hooks route


hypha/apply/templates/forms/includes/multi_input_field.html
```diff
@@ -1,12 +1,17 @@
 {% load heroicons %}
-<div class="form__item{% if field.help_text %} form__group--wrap{% endif %}{% if field.errors %} form__error{% endif %}{% if not field.initial %} multi-input-field-hidden{% endif %}{% if is_application and field.field.group_number > 1 %} field-group field-group-{{ field.field.group_number }}{% endif %}"{% if is_application and field.field.group_number > 1 %} data-hidden="{% if not show_all_group_fields %}true{% else %}false{% endif %}"{% endif %}" data-multi-field-for="{{ field.field.multi_input_id }}">
+<div
```
theskumar, line 2: In this file the indentation has been been mis-behaving b/w of an extra quotes after "endif"
```
{% endif %}" data-multi-field-for
```

theskumar: Broke the `<div>` attributes into newlines to align things better and also make it readable.

PR #3575, opened 2023-09-18
Fix application form js error with invalid textarea

theskumar:
If the application form contains a textarea and that field is invalid
the JavaScript responsible to focus on the first element with error and
that causes rest of the application-form.js file to not executing,
causing the submit buttom to remain disabled.

### DIFF
--- a/hypha/static_src/src/javascript/apply/application-form.js
+++ b/hypha/static_src/src/javascript/apply/application-form.js
@@ -33,7 +33,12 @@
         var $error_fields = $application_form.find(".form__error");
         if ($error_fields.length) {
             // set focus to the first error field
-            $error_fields[0].querySelector("input").focus();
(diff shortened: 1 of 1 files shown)


hypha/static_src/src/javascript/apply/application-form.js
```diff
@@ -33,7 +33,12 @@
         var $error_fields = $application_form.find(".form__error");
         if ($error_fields.length) {
             // set focus to the first error field
-            $error_fields[0].querySelector("input").focus();
+            const firstErrorField = $error_fields[0];
+            const firstInputEl =
+                firstErrorField.querySelector("input, textarea");
```
theskumar, line 38: Note my self, exclude Tinymce / hidden textareas that are un-focusable.

PR #3577, opened 2023-09-19
Performance improvement on application form

theskumar:
This PR optimizes the code and logic responsible for unlocking
the application form on detecting human activity.

- makes the unlocking action DRY.
- move away from using jquery
- prevent the unlock action from firing on every mouse move, the unlock
  action is performed only once.
- Use more standard key names instead of keycodes.

### DIFF
--- a/hypha/static_src/src/javascript/apply/application-form.js
+++ b/hypha/static_src/src/javascript/apply/application-form.js
@@ -46,27 +46,29 @@
         // Remove the "no javascript" messages
         $(".message-no-js").detach();
 
-        // Wait for a mouse to move, indicating they are human.
-        $("body").mousemove(function () {
-            // Unlock the form.
+        const unlockApplicationForm = function () {
             $application_form.attr("action", "");
             $application_form_button.attr("disabled", false);
-        });
+        };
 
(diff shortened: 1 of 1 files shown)


frjo:
@theskumar  I implemented your performance improvements in to the version I use in my Hugo base theme. That version is vanilla js so I moved does changes over here. Please test and see if I got it right.

frjo (approved):

hypha/static_src/src/javascript/apply/application-form.js
```diff
@@ -1,72 +1,68 @@
...
+                    .setAttribute("aria-describedby", label_id);
+            }
         }
+    });
 
-        // Remove the "no javascript" messages
-        $(".message-no-js").detach();
+    if (errors.length) {
+        // Set focus to the first error field.
+        errors[0].querySelector("input").focus();
```
theskumar: Should we use a if condition here. 

```
const inputEl = errors[0].querySelector("input")
if (inputEl) { 
...
}
``

theskumar:
Minor comment to handle edgecase, reset of the jquery -> vanilla JS looks great!

theskumar:
Look good 🚢

PR #3578, opened 2023-09-19
Make invoice statuses consistent(by {user_role}), update finance 1 to finance, and payment failed status

sandeepsajan0:
Fixes #3567 

image

Finance1 is the username above not the status. Also, some of the statuses are old that's why there are caps inconsistency, it will not happen with a new invoice.

PR #3583, opened 2023-09-27
Add payment failed status, and add user roles to statuses, and update finance1 to finance

### LINKED ISSUE #3567: Invoice statuses inconsistency/confusion
- [ ] Update invoice statuses in a way that these should be consistent as 'status' by 'user_roles'. Ex: Approved by Staff, Resubmitted by Applicant.
- [ ] Add invoice status, Payment Failed, that can be changed after paid. 
- [ ] Rename Finance 1 to Finance as it is confusing if Finance 2 settings are not configured. We can keep Finance 2 as it is, just update Finance 1.

sandeepsajan0:
Fixes #3567 

Copy of #3578


sandeepsajan0:
Duplicate of #3585

PR #3586, opened 2023-09-29
Update status bar to show statuses without hover

sandeepsajan0:
Status Bar with latest designs.

Current status bar(old):
image

Updated status bar(latest):
image

The above status bars are from different submissions so don't get confused with their status steps. The only change is that the statuses are visible in the latest one while in the old one user has to hover over every dot to know its status.

### DIFF
--- a/hypha/static_src/src/sass/apply/components/_status-bar.scss
+++ b/hypha/static_src/src/sass/apply/components/_status-bar.scss
@@ -149,6 +149,15 @@
             opacity: 1;
         }
 
+        &::before {
+            .status-bar__item--is-current & {
+                @include triangle(top, $color--error, 5px);
+                position: absolute;
+                bottom: -10px;
+                left: 5px;
+            }
+        }
+
         // tooltip contents
         &::after {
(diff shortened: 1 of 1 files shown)


theskumar:
Should we keep the triangle in the active status. cc: @abhiXII

abhiXII:
sure @theskumar triangle is good to have if it's not a lot of effort.

theskumar:
Cool. it was there already just removed. @sandeepsajan0 can we keep it?

sandeepsajan0:
Updated status bar with a triangle in active status:
image

frjo:
Latest commit are deploying to test now.

theskumar:
@sandeepsajan0 can you check why these are mis-aligned. 

Screenshot 2023-10-04 at 13 16 06@2x

theskumar:
Also, in the initial screenshot of the issue. There are a lot of stages. 

It should be interesting to see how it looks if all these stages are filled with text labels.

frjo (approved):

PR #3587, opened 2023-09-30
Custom applicant dashboard header

sandeepsajan0:
Updated the applicant dashboard header as per the latest designs.

Current version(old):
image

Updated version(latest):
image

Used a custom code instead of admin bar component because needed custom implementation that is different from other or common admin bars(headers)

### DIFF
--- a/hypha/apply/dashboard/templates/dashboard/applicant_dashboard.html
+++ b/hypha/apply/dashboard/templates/dashboard/applicant_dashboard.html
@@ -4,83 +4,94 @@
 
 {% block title %}{% trans "Dashboard" %}{% endblock %}
 
-{% block content %}
-    {% adminbar %}
-        {% slot header %}{% trans "Dashboard" %}{% endslot %}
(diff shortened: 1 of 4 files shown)


theskumar:
> Used a custom code instead of admin bar component because needed custom implementation that is different from other or common admin bars(headers)

Can you share some info on it, it see it being fundamentally same. I agree the rest of the admin bar will update as a result of this, but it's that expected here?

sandeepsajan0:
As per the latest designs, almost all the headings are centered except applicant dashboard.
image

image

So, we can keep the admin bar component text-centered, and as it is just one place so can have a custom code like submission and project detail pages.

theskumar:
We should able to pass an extra parameter `align=` to check how we want the headers to be displayed, and can default it to be center at a later sage. I can take a look at structuring the admin-bar a better.

frjo:
@sandeepsajan0 We have merged in everything on test now to main so this can be rebased.

frjo (approved):

PR #3600, opened 2023-10-05
Add invoices rows to applicant dashboard with different kind of invoice statuses

sandeepsajan0:
No current version(newly added table to applicant dashboard).

Design:
image

Updated version:
image

### DIFF
--- a/hypha/apply/projects/constants.py
+++ b/hypha/apply/projects/constants.py
@@ -1,21 +1,24 @@
 from django.conf import settings
+from django.utils.translation import gettext_lazy as _
 
 # INT refers Invoice table
-INT_STAFF_PENDING = "Staff pending"
-INT_FINANCE_PENDING = "Finance pending"
-INT_VENDOR_PENDING = "Vendor pending"
+INT_STAFF_PENDING = _("Staff pending")
+INT_FINANCE_PENDING = _("Finance pending")
+INT_VENDOR_PENDING = _("Vendor pending")
 
 # duplicate statuses
-INT_DECLINED = "Declined"
-INT_PAID = "Paid"
-INT_PAYMENT_FAILED = "Payment failed"
+INT_DECLINED = _("Declined")
+INT_PAID = _("Paid")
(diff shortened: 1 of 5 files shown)


hypha/apply/projects/constants.py
```diff
@@ -0,0 +1,35 @@
+from django.conf import settings
+
+# INT refers Invoice table
+INT_STAFF_PENDING = "Staff pending"
+INT_FINANCE_PENDING = "Finance pending"
+INT_VENDOR_PENDING = "Vendor pending"
```
theskumar: Use translations?

sandeepsajan0:

abhiXII:
Here, my feedback would be just to add the white background for rows and remove underline from the invoice numbers. 
image


Q: will we show all the invoices here or top 5 or something with a button to `View all invoices`? Maybe we'll show all where action is required right so the ones with - pending, change requested or failed right?

also are we using the font roboto in designs? somehow the font for $1000 dollar values doesn't match with what I see in design. am I missing something? or maybe it's always been like that and im just noticing it now?

sandeepsajan0:
@abhiXII thanks for the feedback.
>just to add the white background and remove underline from the invoice numbers

The row background is already white it is the body background that is not grey yet that's why its color doesn't look different. We already have a PR for that(#3602), it will look similar to the design after that PR. I think underlining was in the design initially? I checked it is not there now so I'll remove it.
>will we show all the invoices here or top 5 or something with a button to View all invoices?

For now, we are showing all active invoices here because we don't have a separate table for that yet(new navigation and tables). And I think it is fine for now. As soon as we will work on navigation and tables for applicants we can add a `View all invoice` button there with only the five latest invoices.
>also are we using the font roboto in designs? somehow the font for $1000 dollar values doesn't match with what I see in design. am I missing something? or maybe it's always been like that and im just noticing it now?

Not sure but I haven't updated the font, I think it is the same everywhere. I just updated its size and type(like light/bold).

PR #3601, opened 2023-10-05
Update typography of rendered rich text answers

### LINKED ISSUE #3594: Add styling to tables in rendered rich text forms

theskumar:
Fixes: #3594

- Update headings and paragraph to use tailwind/typrography.
- Add border bottom for questions rendering to distinguish it with headings in the rendered answers [1]
- Add styled table [2]

### Preview

Screenshot 2023-10-05 at 09 27 18@2x

### DIFF
--- a/hypha/static_src/src/sass/apply/components/_rich-text.scss
+++ b/hypha/static_src/src/sass/apply/components/_rich-text.scss
@@ -60,6 +60,20 @@
         }
     }
 
+    // This is a hack to make sure a copy pasted table with width property on it
+    // doesn't break the layout. The margin from the table is removed and transferred
+    // to the parent element, because the table is wrapped in a div with overflow: auto
+    &__table {
+        overflow: auto;
+        margin-top: 1.25em;
+        margin-bottom: 1.25em;
+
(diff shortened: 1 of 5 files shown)


hypha/static_src/src/sass/apply/components/_rich-text.scss
```diff
@@ -47,23 +60,12 @@
         }
     }
 
-    &__table {
-        overflow: auto;
```
frjo, line 51: I believe we need to keep this. Some tables are really big/wide and they break the layout otherwise.

frjo (changes requested):

hypha/static_src/src/sass/apply/components/_rich-text.scss
theskumar: I could not reproduce and it looks will only happen if `<table>` has an explicit width defined on it. But I agree it will break the layout and overflow into the sidebar.

I've added it back and also increased the width of the container to be more wide and take the full space.

frjo (approved):

Issue #3427, opened 2023-06-02
Add nav-bar to Apply 


fourthletter:
### Description

A nav-bar will need to be added to the apply page to align with the organization's website, including typology like font and color. As a user, I would need to be rerouted to a webpage on the main website. For example, clicking on 'Fund' would lead to the correct webpage.

### Current nav-bar
![Screenshot 2023-06-02 8 31 38 PM](https://github.com/HyphaApp/hypha/assets/20019656/1548d5fb-1dcf-42c9-a4dc-7c79ae8e5dd9)

### PROPOSED nav-bar
![Screenshot 2023-06-02 8 32 33 PM](https://github.com/HyphaApp/hypha/assets/20019656/91c377e7-c196-4274-be0f-67b1b9b08941)

frjo:
This menu will be completely removed since it is part of the public site. Same thing with the footer.

The logo can continue to link back to https://www.opentech.fund/.

The https://apply.opentech.fund/ site will just be Hypha i.e. and app to handle submission etc.

theskumar:
The open PR to remove the nav bar is at #3432 , and it was decided not to transfer the nav from the org's website should/would not be translated 1:1 to one to hypha. 

This issue should be safe to close and open a new PR, if any further improvement is needed in the primary nav for public facing login/register pages.

wes-otf:
This issue will be partially completed by #3604 but I think it'd be beneficial to keep it around as the HTML nav template still needs to be created based off the WordPress nav theme and entered via wagtail admin.

PR #3604, opened 2023-10-06
Add customizable Header content field

wes-otf:
This is a starting point for #3427. 

With this functionality the wagtail admin panel can be used to add a custom HTML header - very similarly to the footer. **Unlike the footer**, this field is not required and when left blank the default header will be used. This goal is to use this to be able to recreate the WordPress site navbar/header.

## Screenshots
**Using a blank "Header content" wagtail field:**

_The empty "Header Content" Wagtail admin field_
Screenshot 2023-10-06 at 1 32 48 PM

_The usual nav bar_
Screenshot 2023-10-06 at 2 23 19 PM

**Using a custom HTML header (rough-cut, not finalized)**

_The start of the custom HTML "Header content" field_
Screenshot 2023-10-06 at 2 15 05 PM

_The custom header that shows resemblance to the WordPress header (needs more tweaking)_ 
Screenshot 2023-10-06 at 2 14 38 PM

Being that this is my first PR feel free to give any critique you see necessary! Anything is helpful. Cheers!


hypha/templates/base.html
```diff
@@ -72,109 +72,113 @@
...
-                            <svg class="header__icon header__icon--open-search header__icon--open-search-menu-closed icon icon--mobile-menu"><use xlink:href="#magnifying-glass"></use></svg>
-                            <svg class="header__icon header__icon--close-search header__icon--close-search-menu-closed icon icon--mobile-menu"><use xlink:href="#cross"></use></svg>
-                        </button>
-                        <button class="button button--left-space js-mobile-menu-toggle" aria-haspopup="true">
-                            <svg class="icon icon--mobile-menu"><use xlink:href="#mobile-menu-toggle"></use></svg>
-                        </button>
-                    </div>
-
-                    <section class="header__menus header__menus--desktop">
-                        {% cache 3600 navigation__primary current_site %}
```
frjo, line 115: It is only this primary nav that should be replaced. Logo etc. should always be in place.

hypha/public/utils/models.py
```diff
@@ -320,6 +320,12 @@ class Meta:
         ),
     )
 
+    header_content = models.TextField(
```
frjo: Maybe rename this to "Front page navigation".

hypha/public/utils/models.py
```diff
@@ -320,6 +320,12 @@ class Meta:
...
+        "Header content",
+        help_text=_("This will overwrite the default header, html tags is allowed."),
```
frjo: This is good, let the default stay until a custom nav is added.

When we finally scrap the public side of Hypha the default will likely by "My dashboard etc.", as it is on all the apply pages.

frjo (changes requested):

hypha/templates/base-apply.html
```diff
@@ -134,7 +143,7 @@
                 </section>
 
                 <div class="header__button-container">
-                    {% if latest_notifications %}
+                    {% if latest_notifications %}.
```
frjo: A mistake period it look like?

hypha/templates/base-apply.html
```diff
@@ -102,7 +102,12 @@
                 </div>
 
                 <section class="header__menus header__menus--desktop">
-                    {% include "navigation/primarynav-apply.html" %}
+                    {% if settings.utils.SystemMessagesSettings.nav_content|safe and not request.user.is_authenticated %}
```
frjo: I think we can skip the "safe" filter in the if statement.

frjo: For this intermediate version we can skip the "and not request.user.is_authenticated" part as well.

OTF will need this when they launch the new public site. When that happens the "primarynav*.html" will be obsolete.

We need to work on this when we completely remove the public site in any case.

frjo (changes requested):

frjo:
@wes-otf Works well but needs some minor fixes.

hypha/templates/base-apply.html
wes-otf: Ugh - result of a sloppy copy & paste. Should be all set now.

frjo (approved):

PR #3611, opened 2023-10-11
Project Reports as stream field forms

frjo:
With `PROJECTS_ENABLED = True`, Project Report Forms may now be
created in Wagtail Admin and associated with a Fund.

Project reports use the form configured in Wagtail admin. To add or
edit a report form, use `Projects -> Report Forms`. To bind a report
form to a given Fund, use `Apply -> Funds`, set `Project Report Form`.

When projects are enabled, a project shows the Reports in various
project interfaces. A new project report will use the current fields
from the Project Report Form associated with the Fund from which the
Submission/Application was granted. An existing report, when edited,
will use the original fields at the time that report was first filed.
This behavior appears consistent with at least one other similar form.

Thanks to Frank Duncan for the "View" functionality in
hypha/apply/projects/templates/application_projects/report_detail.html
and NonFileFormFieldsBlockFactory for tests that don't need files
as well as re-adding the url for project file lookup by id.

Co-authored-by: Frank Duncan <frankduncan@opentechstrategies.com>


hypha/apply/projects/migrations/0080_auto_20231006_1757.py
bickelj, line 1: I noticed I hadn't renamed this after pushing. I plan to rename it.

hypha/apply/projects/models/report.py
```diff
@@ -166,13 +171,21 @@ def start_date(self):
         return self.project.start_date
 
 
-class ReportVersion(models.Model):
+class ReportVersion(BaseStreamForm, AccessFormData, models.Model):
```
bickelj, line 174: I suspect something simple is wrong here or in `ReportEditForm`.

hypha/apply/projects/forms/report.py
```diff
@@ -1,24 +1,27 @@
...
 
-class ReportEditForm(FileFormMixin, forms.ModelForm):
-    public_content = RichTextField(
-        help_text=_(
-            "This section of the report will be shared with the broader community."
-        )
-    )
-    private_content = RichTextField(
-        help_text=_("This section of the report will be shared with staff only.")
+class ReportEditForm(StreamBaseForm, forms.ModelForm, metaclass=MixedMetaClass):
```
bickelj, line 10: I suspect something simple is wrong here or in `ReportVersion`.

bickelj:
~~Locally I have `PROJECTS_ENABLED = True` and `PROJECTS_AUTO_CREATE = True`.~~ Edit: moot as of 2024-04-18, see [request](https://github.com/HyphaApp/hypha/pull/3611#discussion_r1566220578) and [response](https://github.com/HyphaApp/hypha/pull/3611#issuecomment-2062401995) below.

hypha/apply/funds/models/forms.py
```diff
@@ -268,3 +269,21 @@ class LabBaseProjectApprovalForm(AbstractRelatedProjectApprovalForm):
 
 class LabBaseProjectSOWForm(AbstractRelatedProjectSOWForm):
     lab = ParentalKey("LabBase", related_name="sow_forms")
+
+
+class ApplicationBaseProjectReportForm(AbstractRelatedProjectApprovalForm):
```
frjo: The `AbstractRelatedProjectApprovalForm` should not be reused, create a new one for Reports with a name like `AbstractRelatedReportForm`.

hypha/apply/projects/admin_views.py
```diff
@@ -21,3 +21,11 @@ class CreateProjectSOWFormView(CreateProjectApprovalFormView):
 
 class EditProjectSOWFormView(EditProjectApprovalFormView):
     pass
+
+
+class CreateProjectReportFormView(CreateProjectApprovalFormView):
```
frjo, line 26: Reports are created by applicants/vendors so should not be admin related in any way.

hypha/apply/projects/migrations/0080_auto_20231006_1757.py
```diff
@@ -0,0 +1,1423 @@
...
+            bases=(hypha.apply.stream_forms.models.BaseStreamForm, models.Model),
+        ),
+        migrations.RemoveField(
+            model_name="reportversion",
+            name="private_content",
+        ),
+        migrations.RemoveField(
+            model_name="reportversion",
+            name="public_content",
+        ),
```
bickelj: This probably wants to be two (or three) migrations rather than one, eventually.

hypha/apply/projects/models/report.py
```diff
@@ -12,7 +12,12 @@
 from django.utils import timezone
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
+from wagtail.fields import StreamField
 
+from hypha.apply.funds.models.mixins import AccessFormData
+from hypha.apply.projects.blocks import ProjectApprovalFormCustomFormFieldsBlock
```
frjo: The name is specific about it being used for the approval form. Someone could change it believing it does what the name applies.

Better to create a new class for this use case. It can be identical but with a name explaining what is does.

hypha/apply/projects/admin_views.py
bickelj: This is the report form, though, not the report itself. The admins set up the form the way they would like the applicants to see it.

hypha/apply/funds/models/forms.py
bickelj: Perhaps an AbstractStreamForm that all could inherit? Or maybe another way we can avoid adding so many identical classes with each new feature?

bickelj: In any case, this part seems to be working fine, but I am delighted to have early feedback on this too, the sooner we get it straightened out the better.

hypha/apply/projects/models/report.py
bickelj: Similar to above, maybe we can extract the common code into a truly common/re-usable class?

bickelj: I renamed this class to the more generic thing that it is and added a comment. Does this help resolve?

hypha/apply/projects/admin_views.py
bickelj: Perhaps other roles should be able to edit this besides admins, right? I have yet to get to the exact permissions.

hypha/apply/funds/models/forms.py
bickelj: I updated it to use a near-copy of the other AbstractRelatedXForms.

hypha/apply/projects/migrations/0080_auto_20231006_1757.py
bickelj: Renamed and re-ordered.

bickelj:
I pushed some of the test fixes and the workaround for the stream form file failures. Basically: exclude the file capabilities from the test setup factories. On our branch these are all passing now. Just to close the loop, the test failures are not a high priority issue anymore but they may come up again in a future context.

frjo:
@bickelj Recommend taking a look at https://github.com/HyphaApp/hypha/pull/3822 and https://github.com/HyphaApp/hypha/pull/3823. Make sure you have not the same bug in this implementation.

bickelj:
@frjo After a quick glance at the issue, I think we do not have the same issue, but perhaps we didn't solve it the same way. I will take a look next week to see.

bickelj:
@frjo @wes-otf  I think this is ready enough for review. I could not reproduce the lint error above on `lint-be` task.

frjo:
@bickelj Looking forward to testing this, and get it part of Hypha!

1. Rebase so these commits are on top of current main.
2. There are some merge conflicts, can you please fix these.
3. To handle linting automatically run `pre-commit install`. Then code will be linted when added to git.

See `Makefile` for manual lint etc. commands.

bickelj:
@frjo I see what happened. It was mostly well (rebased, merged) locally and I had pushed the branch [to my GH fork](https://github.com/bickelj/hypha/tree/project-report-forms) but not this branch here. I just re-pushed it here and see I have some migrations to rename.

hypha/apply/projects/forms/report.py
bickelj: Resolved back in December.

hypha/apply/projects/models/report.py
bickelj: Resolved back in December.

bickelj:
@frjo OK, I did some more cleanup but I'm scratching my head about the `hypha/apply/review/migrations/0026_alter_review_visibility.py` being needed in the first place and then the tests adding a 0027 on top of that. I think I only added that because of tests failing here back in January, I don't think we have that in our versions. I am sure it's something simple, but locally I don't see a need for anything new when running `makemigrations`.

wes-otf:
@bickelj I can look deeper into it tomorrow but weirdly enough when I deleted migration 26, renamed the generated migration from `27` -> `26` and changed the dependencies to `("review", "0025_auto_20220722_0844"),` it stopped generating a 27. Seems like a symptom fix though

wes-otf:
tossing the `__pycache__` in the migrations folder may _(or may not)_ have an effect too but if the CI is also seeing it I doubt that's a factor

frjo:
I tested to delete all migrations in this PR and redo them, see https://github.com/HyphaApp/hypha/pull/3845. I think that works but some other tests then fail.

hypha/apply/projects/models/project.py
```diff
@@ -319,10 +335,30 @@ def start_date(self):
...
+            .first()
+        )
+
+        if first_accepted_determination:
+            return first_accepted_determination.created_at.date()
+
+        # If neither an approved contract nor an accepted determination are found, something is wrong and a meaningful
+        # error or warning may be appreciated by future developers.
+        warnings.warn(
+            "No related contracts or determinations were approved/accepted, so no date can be found.",
```
frjo: Please make translatable.

hypha/apply/funds/files.py
```diff
@@ -5,36 +5,63 @@
...
 
     return os.path.join(path, file_name)
 
 
-class SubmissionStreamFieldFile(StreamFieldFile):
-    def get_submission_id(self):
+def generate_submission_file_path(submission_id, field_id, file_name):
+    return generate_private_file_path(
+        submission_id, field_id, file_name, path_start="submission"
+    )
```
frjo: This is used in only one place, can we not use `generate_private_file_path` there as well?

hypha/apply/funds/files.py
```diff
@@ -5,36 +5,63 @@
...
+
+        view_name = "apply:submissions:serve_private_media"
+        kwargs = {
+            "pk": self.get_entity_id(),
+            "field_id": self.field.id,
+            "file_name": self.basename,
+        }
+        if isinstance(self.instance, ReportVersion):
+            view_name = "apply:projects:document"
+        return reverse(viewname=view_name, kwargs=kwargs)
```
frjo, line 61: ReportVersion look a bit crammed in here.

bickelj: In the whole class or just in this method? What do you mean or what other way would you suggest going about this?

hypha/apply/projects/models/project.py
bickelj: Fixed.

hypha/apply/funds/files.py
bickelj: Good idea. I removed the extra function.

bickelj:
> I tested to delete all migrations in this PR and redo them, see #3845. I think that works but some other tests then fail.

Would you recommend this course here? There are some discrete steps in the commits, one creates wagtail form, another creates project form, and it might be easier to follow when broken up.

I guess that's another question: would you expect this whole branch to be squashed (along with the migrations), left as-is with the cleanup commits, or commits edited individually?

frjo:
@bickelj We use "Squash and merge" for all PRs here on GitHub. So it will all become one commit when it lands in the main branch.

bickelj:
@frjo I guess that the migration hiccup is a commonly encountered Django bug and it is cheaper to regenerate migrations than to track down the issue, right?

Regardless, I see the path forward for me:
1. Squash
2. Regenerate migrations
3. Push

bickelj:
@frjo @sandeepsajan0 I think it is [commit:35d32a](https://github.com/HyphaApp/hypha/commit/35d32a76d90e07bc0cb8c74d6f12d9d8bb519bee) that conflicts causing the test failure. I don't think we should require PAF to use Projects. One of the capabilities we added with this feature is to skip any number of Projects stages. For example, a Hypha instance can skip ahead to "Invoicing and Reporting" without doing PAF or Contracting in order to take advantage of the Reporting capabilities.

frjo:
We will discuss reverting PR https://github.com/HyphaApp/hypha/pull/3829.

hypha/apply/projects/models/project.py
```diff
@@ -290,6 +296,15 @@ def create_from_submission(cls, submission, lead=None):
...
             return None
 
+        # The PROJECTS_INITIAL_STATUS env var allows a funder to skip straight to a given Project stage aka status.
+        # Note this is case-sensitive and all the choice keys are lower case. So use a lower-case string setting.
+        status = DEFAULT_PROJECT_STATUS
+        if settings.PROJECTS_INITIAL_STATUS and any(
+            CHOICE[0] == str(settings.PROJECTS_INITIAL_STATUS)
+            for CHOICE in PROJECT_STATUS_CHOICES
+        ):
+            status = settings.PROJECTS_INITIAL_STATUS
```
frjo: @bickelj Can you please remove all code regarding `PROJECTS_INITIAL_STATUS`. We will instead add settings in a separate PR that allow you to disable state by state as needed. More flexible in the long run is our reasoning.

bickelj:
I rebased and removed the `PROJECTS_INITIAL_STATUS` but obviously more work is required due to test failures. I will investigate tomorrow.

bickelj:
I needed to include some more test-related changes, I'm not sure how I missed those other than the passage of time.

hypha/apply/projects/models/project.py
bickelj: @frjo Thank you for this suggestion. Settings is the better choice for the long run. In the process of removal I found some missing test-related commits that are now included.

hypha/apply/projects/migrations/0080_auto_20231006_1757.py
bickelj: We actually ended up squashing it into one.

bickelj:
@frjo This should be ready for testing again.

wes-otf:
@frjo if you don't beat me to it I can give it a go tomorrow

frjo:
I added a minor fix to the submit and save buttons. Added some classes to make it the same as on other forms and updated the text to say "Save draft", as on other forms. This is fixes to old standing issues with report that fits well in this PR I think.

hypha/apply/funds/files.py
frjo: The whole class. It looks like we hacked the class to work with Reports instead of it being built to support different models on purpose.

It works so a change is not urgent but I prefer code that is easy to read in the long run.

frjo (commented):
I have now tested this PR and it works really well. This is how Reports should have been implemented from the start. Brilliant work!

I will put this on the test site next week where more people can hammer on it.

Hopefully I can get it merged in before the end of April.

wes-otf:
Looks & works great from my testing! Thanks for all the contribs & patience with this @bickelj & @frankduncan!

bickelj:
Woohoo! Thank you, @wes-otf and @frjo for reviewing, testing, and merging!

PR #3612, opened 2023-10-12
Pin python-docx version to temporarily fix upstream issues

wes-otf:
With the latest updates to python-docx, older APIs that are relied upon by htmldocx have broken. This pins the version so htmldocx can utilize the old APIs. For more info on this breakage see [here](https://github.com/python-openxml/python-docx/issues/1256#issuecomment-1756839793).


theskumar (approved):
Looks fine by me! 🚢

PR #3615, opened 2023-10-13
Fixed failure to preview rounds in Wagtail Admin

### LINKED ISSUE #3559: Previewing rounds in Wagtail admin result in error
NoReverseMatch 'users' is not a registered namespace

wes-otf:
Closes #3559. 

Updated the app's urls.py to include the url_patterns specified in the user app's url.py.

### DIFF
--- a/hypha/apply/activity/tests/test_messaging.py
+++ b/hypha/apply/activity/tests/test_messaging.py
@@ -48,7 +48,6 @@ class TestAdapter(AdapterBase):
         pass
 
 
-@override_settings(ROOT_URLCONF="hypha.apply.urls")
 class AdapterMixin(TestCase):
     adapter = None
     source_factory = None
--- a/hypha/apply/api/v1/screening/tests/test_views.py
+++ b/hypha/apply/api/v1/screening/tests/test_views.py
@@ -9,7 +9,6 @@ from hypha.apply.funds.tests.factories.models import ApplicationSubmissionFactor
 from hypha.apply.users.tests.factories import ReviewerFactory, StaffFactory, UserFactory
 
 
-@override_settings(ROOT_URLCONF="hypha.apply.urls")
 @override_settings(SECURE_SSL_REDIRECT=False)
 class ScreeningStatusViewSetTests(APITestCase):
     def setUp(self):
@@ -56,7 +55,6 @@ class ScreeningStatusViewSetTests(APITestCase):
 
 
 @override_settings(SECURE_SSL_REDIRECT=False)
-@override_settings(ROOT_URLCONF="hypha.apply.urls")
 class SubmissionScreeningStatusViewSetTests(APITestCase):
     def setUp(self):
         ScreeningStatus.objects.all().delete()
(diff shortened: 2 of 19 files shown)


theskumar:
TBH, I believe this has something to do with the internals of the preview generation view. The view responsible for these forms is very custom. The view responsible for the previews might have changed internally b/w the wagtail version upgrades, causing this issue. 

I don't fully understanding the inner working to two different urls. But you look at logs, the preview view should pick up the `hypha.apply.urls` and not `hypha.urls`. 

That said, another way to resolve this could that once the public part/site is remove we have only only set of ROOT_URLCONF. 

You might also be able to get away by hand-picking the urls and adding them the `hypha.urls`

Screenshot 2023-10-14 at 5  20 30@2x

theskumar:
Digging into more around the url conf issue. This is where the magic happens, and in my opinion we should get rid of this middleware (one the public site is removed), or update this middleware to allow for preview page to use different `urlconf`

https://github.com/HyphaApp/hypha/blob/0a29676e6dcd3a9484c0ba070c872243c66edadd/hypha/apply/middleware.py#L31-L45

wes-otf:
@theskumar This was super helpful! Thanks a ton. Do you think it would be best to wait for the public site to be removed for these changes to be made? It sounds like there's a decent amount is reliant on that.

theskumar:
@wes-otf After discussing with @fourthletter and @frjo we believe it would best to wait till public is removed. 

For now, you check is resolved get resolved if you update ROOT_URLCONF to `hypha.apply.urls` instead of `hypha.urls`. 

I'm not sure what will be the side-effect is we switch out the default root url conf site-wide. If it doesn't break things, it might be wise to do the switch now as part of depreciating the public site.

wes-otf:
@theskumar all sounds good! I updated the `ROOT_URLCONF` as suggested and that also seems to work pretty well. I did a bit of manual testing and didn't seem to see any issues.

wes-otf:
Looks like the tests might need a fix after that though. I'll look into that now.

wes-otf:
Pulling this thread a little more it looks like the main failures from this are coming from the newsletter functionality ([or tests, at least](https://github.com/HyphaApp/hypha/blob/main/hypha/public/mailchimp/tests/test_views.py#L12)) being reliant on the public URL side of things. Is the newsletter something that will be ripped out with the public site?

theskumar:
The newsletter is something won't be part of hypha for sure. So yes it will be ripped out when we remove the public site.

wes-otf:
We can discuss it more in Monday's standup but I think there's bigger database implications for changing the `ROOT_URLCONF` that I haven't fully wrapped my head around yet. I originally thought it was just the tests that were failing (as functionality seemed unchanged) but after doing a full `python3 manage.py reset_db` I'm getting the same error that the CI is showing with:
```
django.db.utils.ProgrammingError: relation "django_content_type" does not exist
LINE 1: ..."."app_label", "django_content_type"."model" FROM "django_co...
```
Going to keep digging around

wes-otf:
Closing this PR to start a new one to fix this issue and remove the public site aspect.

PR #3622, opened 2023-10-20
Fix rendering of stream form fields

theskumar:
How to reproduce:
- open a application submisssion with checkboxes as an form element.
- Get the error below

```
'str' object has no attribute 'help_text'
```

Screenshot 2023-10-20 at 9  43 45@2x


frjo:
The bug activated when the checkboxes field are not set to multiple, i.e. it becomes radios.

This PR fixes it. Nice catch! Wonder how we miss it before, did nobody use radios?

frjo (approved):

PR #3632, opened 2023-10-26
Add comment count to the communications tab.

### LINKED ISSUE #3629: Show how many comments there are on the communication tab

frjo:
Fixes #3629

### DIFF
--- a/hypha/apply/funds/templates/funds/applicationsubmission_detail.html
+++ b/hypha/apply/funds/templates/funds/applicationsubmission_detail.html
@@ -46,7 +46,7 @@
                     </a>
 
                     <a class="tab__item" href="#communications" data-tab="tab-2">
-                        {% trans "Communications" %}
+                        {% trans "Communications" %} ({{ comments.count }})
                     </a>
 
                     <a class="tab__item"


hypha/apply/funds/templates/funds/applicationsubmission_detail.html
```diff
@@ -46,7 +46,7 @@ <h1 class="mb-0 mt-2 font-medium">{{ object.title }}</h1>
                     </a>
 
                     <a class="tab__item" href="#communications" data-tab="tab-2">
-                        {% trans "Communications" %}
+                        {% trans "Communications" %} ({{ object.activities.comments.count }})
```
theskumar: I belive this should be the count of visible comment, no all the comments. I see have a submission with only one comment but the count is 2. 

Screenshot 2023-10-27 at 9  48 05@2x

theskumar (changes requested):

hypha/apply/funds/templates/funds/applicationsubmission_detail.html
frjo: Good catch. Need to filter the count on what comments the user can see. Or we only show it to staff.

frjo:
@theskumar Really simple fixes. Managed to forget that we already have "comments" available in this context.

theskumar (approved):
Nice 🎉

PR #3602, opened 2023-10-05
Add grey background for applicant dashboard

sandeepsajan0:
Current background:
image

Updated version:
image

PR #3634, opened 2023-10-30
Add todo app with task model and integrate todo app APIs with projects part(including PAF process, contracting and invoicing)

### LINKED ISSUE #3628: My Tasks list(Todo app) for every user pending actions
My tasks list UI:
image

**Approach**:
1. Create a separate app/folder for tasks just like activities.
2. Create a Task model to store the pending action temporarily for a user/user role.
3. Add and remove the tasks to the Task model according to the project status and user's actions.
4. Create APIs to add, remove, and retrieve the tasks to the Task model for a user/user role.

**Model**:
Task - statement, user(id), user_group(ids), url, object_type(submission, project), object_id, related_type(contract, paf, invoice, report), related_id(null true), created_at

**APIs**:
add_task_to_user(statement, user, url, object_type, object_id, related_type=None, related_id=None)
add_task_to_user_group(statement, user_group, url, object_type, object_id, related_type=None, related_id=None) 
remove_tasks_for_user(user, task)
remove_tasks_for_user_group(user_group, task)   // remove for all users in that user group)
get_tasks_for_user(user)      // check the user's group and return that user_group tasks as well, ordered with  -created_at.

sandeepsajan0:
Fixes #3628 Closes #3605 

After:
image


hypha/apply/tasks/models.py
```diff
@@ -0,0 +1,90 @@
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    related_content_type = models.ForeignKey(
+        ContentType,
+        blank=True,
+        null=True,
+        on_delete=models.CASCADE,
+        related_name="task_related",
+    )
+    related_object_id = models.PositiveIntegerField(blank=True, null=True)
+    related_object = GenericForeignKey("related_content_type", "related_object_id")
```
theskumar: Thoughts on linking an `activity` object that lead to this task. Can't think of an use-case now, but it will be use to track the source activity that lead to these actionable items.

hypha/settings/django.py
```diff
@@ -18,6 +18,7 @@
     "hypha.apply.review",
     "hypha.apply.determinations",
     "hypha.apply.stream_forms",
+    "hypha.apply.tasks",
```
theskumar: Please rename it to `user_tasks` or simply `todo` app. `tasks` too generic

theskumar: Or `user_actions`

sandeepsajan0: Renamed it to `todo`. `todo` makes more sense than others.

frjo:
@sandeepsajan0 Was this ready for testing?

sandeepsajan0:
@frjo Yes, it is ready for testing.

hypha/apply/todo/views.py
```diff
@@ -0,0 +1,138 @@
...
+    output: None
+    """
+    task = Task.objects.filter(
+        code=code,
+        user=user,
+        related_content_type=ContentType.objects.get_for_model(related_obj).id,
+        related_object_id=related_obj.id,
+    ).first()
+    if task:
+        task.delete()
```
theskumar, line 64:
```suggestion
    Task.objects.filter(
        code=code,
        user=user,
        related_content_type=ContentType.objects.get_for_model(related_obj).id,
        related_object_id=related_obj.id,
    ).delete()
```

Consider using a more optimized version of delete

hypha/apply/todo/views.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+    user_group_matching_tasks = matching_tasks.annotate(
+        group_count=Count("user_group")
+    ).filter(group_count=len(user_group.all()))
+    for group in user_group.all():
+        user_group_matching_tasks = user_group_matching_tasks.filter(
+            user_group__id=group.id
+        )
+    if user_group_matching_tasks.exists():
+        user_group_matching_tasks.delete()
```
theskumar, line 90: This looks is over-writing the `user_group_matching_tasks` in every iteration. You are then removing the last one. Something is not adding up. 

Would be great if you can write a test to check it's going what it should do.

hypha/apply/todo/views.py
```diff
@@ -0,0 +1,138 @@
...
+def get_tasks_for_user(user):
+    user_tasks = Task.objects.filter(user=user).annotate(
+        group_count=Count("user_group")
+    )
+    user_group_tasks = Task.objects.annotate(group_count=Count("user_group")).filter(
+        group_count=len(user.groups.all())
+    )
+    for group in user.groups.all():
+        user_group_tasks = user_group_tasks.filter(user_group__id=group.id)
+
```
theskumar, line 116: Same comment here as https://github.com/HyphaApp/hypha/pull/3634/files#r1421747171

hypha/apply/todo/views.py
sandeepsajan0: It is just finding the tasks with exact user groups(many to many relations) and in the end, deleting those tasks.

I'll add tests for all of these APIs later.  As this PR already has too many changes, I'll add a separate PR for that.

theskumar: It's okay to have the PR as part of this PR itself. Also can you clarify bit more what you mean by `in the end`. 

You are looping over something and just picking the last value for that iteration. I don't see it being predictable. 

image user_group.all() returns `group 1` and `group 2`, you are going be acting on the `group 2` and not do anything for `group 1`. If that is case, you can just `user_group.last()` instead of `user_group.all()`

frjo (approved):

hypha/apply/todo/views.py
sandeepsajan0: Okay, Let me explain the loop logic here. As I mentioned in my last comment it aims to find the tasks with 'exact' user_groups.

Ex: Let's say there are two groups `group1` and `group2` in user_group.
**In line 82:** annotate and filter is making sure that `user_group_matching_tasks` contains all the tasks with two groups(len of 'user_group'), but we are not sure these groups are exclusively `group1` and `group2`.
So for that, we iterate through 'user_group'. In the first iteration, the code filters `user_group_matching_tasks` to include only those with `group1` and updates the collection accordingly. In the second iteration, it ensures that the 'updated' `user_group_matching_tasks` collection also contains `group2`. 

After all iterations, we can be sure that all tasks in `user_group_matching_tasks` have exactly two groups, specifically `group1` and `group2`.

wes-otf:
In testing this worked really well! Only question I had was if we could potentially make the tasks stand out like then do in your screenshot @sandeepsajan0 by changing the background or task color (ie. what was done in #3651) or give them defined borders. 

With contrast (prev screenshot):

285852704-01349224-885d-4fd9-9d92-c3e8719f9d40

In test:

Screenshot 2023-12-15 at 11 56 11

Otherwise this is ready for live!

sandeepsajan0:
@wes-otf I get your point. We need to update the background color for all users' dashboards, we have already updated it for applicants in #3602. I'll add a PR for other dashboards as well.

PR #3638, opened 2023-11-01
Update the query text on all.html

ramojo:
Change the query text to be linguistically correct

Fixes the language used in the clear query section for no results found in hypha/apply/funds/templates/submissions/all.html


frjo:
Thanks for the fix!

ramojo:
> Thanks for the fix!

You're welcome! I'll try to pass more that I come across upstream

PR #3645, opened 2023-11-04
Add "PAFs for review" table for paf approvers

### LINKED ISSUE #3535: New table for PAF Approvers dashboard (PAF for review)
### Description

All PAF submitter and approver (ie Finance, Contracting, Staff, Staff+Approver) will have this table will include the assignee name, date submitted, Fund associated with PAF, and status. 

image

sandeepsajan0:
Fixes #3535 

Right now, we have two tables on the approvers dashboard "PAF waiting for approval" and "PAF waiting for assignee" but as per the latest design we want a new table "PAFs for review" that contains data from both of the above tables and list it in a simple manner.
In this PR we are just adding that new table but not removing the older ones, once we are done with the new table testing and are ready to have it then we can remove the older ones.

image

The date requested column is kind of not clear yet. 
We have multiple dates, one of them is the "Submitted date" which is when a lead/staff submits the PAF for the first time. Another one is "Updated date" which is when staff submits or this approval(not paf) is updated even for change approver/assignee(right now I have used this one).
But I think we might need to maintain one more date as the "date requested" for every approval, which might be different from both of the above dates in sequential workflow(ex: the second approval date_requested will be set as the first approval approved_date) while it will be same as submit_date for parallel workflow.

What do you all think, is the submit_date enough, or should we have date_requested as another as explained above?

**Update**
We decided to use submit_date as date_requested.
Removing the old tables as the new table contains both old tables' data.


hypha/apply/dashboard/views.py
```diff
@@ -94,11 +97,26 @@ def get_context_data(self, **kwargs):
...
         return context
 
+    def paf_for_review(self):
+        if not self.request.user.is_apply_staff:
+            return {"count": None, "table": None}
+        project_settings = ProjectSettings.for_request(self.request)
+
+        paf_approvals = get_paf_for_review(
+            user=self.request.user, project_settings=project_settings
+        )
```
theskumar: In order to make this more testable, I would suggest keeping the project setting out of the picture from the util function. I would also suggest putting the function inside `services.py` instead of `utils.py`. If you think about it is not a common utility but a business function. 

```suggestion
        project_settings = ProjectSettings.for_request(self.request)

        paf_approvals = get_paf_for_review(
            user=self.request.user, is_paf_approval_sequential=project_settings.paf_approval_sequential
        )
```

theskumar: Alternatively, you can initialize the project settings inside the `get_paf_approvals` itself.

hypha/apply/dashboard/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    if project_settings.paf_approval_sequential:
+        all_matched_paf_approvals = list(paf_approvals)
+        for matched_paf_approval in all_matched_paf_approvals:
+            if matched_paf_approval.project.paf_approvals.filter(
+                paf_reviewer_role__sort_order__lt=matched_paf_approval.paf_reviewer_role.sort_order,
+                approved=False,
+            ).exists():
+                paf_approvals = paf_approvals.exclude(id=matched_paf_approval.id)
+
+    return paf_approvals
```
theskumar:
```suggestion
def get_paf_for_review(user, is_paf_approval_sequential):
    """Return a list of paf approvals ready for user's review."""

    paf_approvals = PAFApprovals.objects.annotate(
        roles_count=Count("paf_reviewer_role__user_roles")
    ).filter(
        roles_count=len(list(user.groups.all())),
        approved=False,
    )

    for role in user.groups.all():
        paf_approvals = paf_approvals.filter(paf_reviewer_role__user_roles__id=role.id)

    if is_paf_approval_sequential:
        all_matched_paf_approvals = list(paf_approvals)
        for matched_paf_approval in all_matched_paf_approvals:
            if matched_paf_approval.project.paf_approvals.filter(
                paf_reviewer_role__sort_order__lt=matched_paf_approval.paf_reviewer_role.sort_order,
                approved=False,
            ).exists():
                paf_approvals = paf_approvals.exclude(id=matched_paf_approval.id)

    return paf_approvals
```

theskumar (changes requested):
Added a minor code organization tip.

fourthletter:
> What do you all think, is the submit_date enough, or should we have date_requested as another as explained above?

We decided on November 7 that submit_date is enough and the date_requested is when staff submits the PAF for approval.

sandeepsajan0:
@frjo It is also ready for testing.

abhiXII:
The table columns are clickable but it doesn't do any sorting. Also can we make only the columns with arrows clickable but not the others? 
for example in this table, make only the date requested column clickable and not others? also when someone clicks on the column header, the arrow should also change from pointing down (descending order) to pointing up (ascending order).
image


are we also planning to add `View all PAFs` option on the top right of the table? like this or is that only when there are more PFAs than what can be shown in the table?
image

also because the arrow head looks a bit misaligned, can we use the double arrowhead icon instead? or try to align that one?

sandeepsajan0:
@abhiXII thanks for the feedback.
Yes, looks like sorting has some issues (not working at all). I'll try to fix the arrows as well. 
>are we also planning to add View all PAFs option on the top right of the table?

I think we should show all the `PAF for review` on the dashboard only. In the all PAF table, we can include all other PAFs that might not require your review but someone else review and you may need to look into that.

sandeepsajan0:
I have fixed the sorting issue, only `Date requested` is clickable now, and also aligned the arrows. I went with the single arrows because it also shows the direction of sorting.
@frjo We can put it on test again.

wes-otf:
This looked good in test, I only saw a small issue where making the screen width < 1024px caused the table to act a little weird. I'm not sure how concerned we are about mobile/tablet views for the staff dashboards and perspectives though so this might not matter all that much and can be put to live.

Screen Shot 2023-12-01 at 10 49 19

sandeepsajan0:
@wes-otf Thanks for bringing that up. I also see this issue with other projects and invoice related tables. I'll fix it for `PAF for review` table in this PR but I think we also need a separate issue for other tables. 
image

frjo (approved):

sandeepsajan0:
Updated table view only for tablets or smaller size screens.
image

frjo:
Latest version on test now.

PR #3636, opened 2023-10-31
Added descriptions to group selections in Wagtail admin

wes-otf:
Fixes #3541.

This PR adds descriptive help text to the Wagtail admin edit/create user "Groups" prompt. This was done by creating custom Wagtail forms and overwriting the existing group field, along with adding another field to the [groups.py](https://github.com/HyphaApp/hypha/blob/enhancement/add-desc-to-user-roles/hypha/apply/users/groups.py#L40).

The [definitions](https://github.com/HyphaApp/hypha/issues/3541#issuecomment-1786722987) provided by @fourthletter were shortened just a bit for the sake of appearance. If anyone believes they got too truncated I'm fine with updating them.

## Screenshots
Screenshot 2023-10-31 at 5 07 29 PM

Screenshot 2023-10-31 at 5 07 15 PM

Cheers!

PR #3646, opened 2023-11-06
Show who can see archived submissions

### LINKED ISSUE #3388: Make it clear who sees an archived submission
### Description

As a user I would need to know who gets to see that this submission has been archived. Other submissions have 'team' or 'applicant' with an eye icon.

image

### LINKED ISSUE #3389: Permission improvements of the submission archive feature
Here are some issue with the current implementation:

- [ ] All staff can archive an submission, if they then do not have permission to view archived submission they get a 403. If they should be able to do this a redirect would be nicer.
- [ ] Should all staff be able to archive a submission?
- [ ] Be default no one can see archived submission, SUBMISSIONS_ARCHIVED_ACCESS_STAFF and SUBMISSIONS_ARCHIVED_ACCESS_STAFF_ADMIN are both false be default. Either super admins should have access or SUBMISSIONS_ARCHIVED_ACCESS_STAFF_ADMIN should default to true.
- [ ]  Should we show on archived submission who can access them, maybe in the red border add info if staff or staff admins only can see this?

wes-otf:
Closes #3388 & Closes #3389.

This PR adds an indication as to what user roles can see an archived submission based off of [existing settings](https://github.com/HyphaApp/hypha/issues/3388#issuecomment-1793773642). 

## Screenshots

_With default Hypha settings_
Screenshot 2023-11-06 at 12 04 13 PM

_After updating settings to allow Staff and Staff Admin to see archives_
Screenshot 2023-11-06 at 12 05 02 PM

_After updating settings to only allow Staff Admin (not Staff) to see archives_
Screenshot 2023-11-06 at 12 03 09 PM


hypha/apply/funds/views.py
```diff
@@ -1074,7 +1075,30 @@ def get_context_data(self, **kwargs):
...
+
+        if settings.SUBMISSIONS_ARCHIVED_ACCESS_STAFF_ADMIN:
+            archive_vis_groups.append(TEAMADMIN_GROUP_NAME)
+
+        archive_string = _(
+            "This submission has been archived. This is visible to the roles: {roles}."
+        ).format(roles=", ".join(archive_vis_groups))
+
         return super().get_context_data(
+            archive_string=archive_string,
```
theskumar: This is what I would do make the code more separated by concern. 

1. Create a template tag `join_with_commas`that joins and displays a list of items and separated by a `,` and `and` the code below. It will be useful other places
2. Make a function `get_archived_submission_visibility_groups()`, either in `appy.funds.utils` or `apply.funds.permissions` module
3. The early return with `self.object().is_archive` is not needed, it can be replaced with final return like this

```python 
return super().get_context_data(
    archived_submission_visibility_groups=get_archived_submission_visibility_groups()
    ...
```



```python
def join_with_commas(obj_list):
    """Takes a list of objects and returns their string representations,
    separated by commas and with 'and' between the penultimate and final items
    For example, for a list of fruit objects:
    [<Fruit: apples>, <Fruit: oranges>, <Fruit: pears>] -> 'apples, oranges and pears'
    
    Inspired by: https://stackoverflow.com/a/1242107
    """
    if not obj_list:
        return ""
    l=len(obj_list)
    if l==1:
        return f"{obj_list[0]}"

    return f"{', '.join(str(obj) for obj in obj_list[:l-1])} and {obj_list[l-1]}"
```

hypha/apply/funds/templates/funds/applicationsubmission_detail.html
```diff
@@ -73,7 +73,7 @@ <h1 class="mb-0 mt-2 font-medium">{{ object.title }}</h1>
     {% if object.is_archive %}
         <div class="py-2 bg-red-600 text-white font-bold text-center">
             {% heroicon_outline "lock-closed" aria_hidden="true" size=16 stroke_width=2 class="inline align-baseline mr-1" %}
-            {% trans "This submission has been archived." %}
+            {{ archive_string }}
```
theskumar: Based on the suggested refractor, this code code will look more like. 

```
{% trans "This submission has been archived. This is visible to the roles" %} {{ archived_submission_visibility_groups|join_with_comma }}
```

theskumar (changes requested):
Added notes to make the code more modular and readable.

frjo:
The po and pot files I mostly update in separate PR for practical reasons.

If you update po and pot files in more than one PR you tend to get merge conflicts. No problem keeping them here, good to get them updated.

wes-otf:
@theskumar this is great feedback, thanks a ton! 

I was originally working to make it more sentence-like (ie. "role, role and role") but wondered how that would impact translations if different linguistics were to come into play. If that might be an issue I can refactor my last commit. Definitely makes sense to have it more modular though.

wes-otf:
@frjo Noted with the translation files - would it make more sense to revert them to what's on main and follow up with the altered files? I have translations committed to #3636 that I can revert too and then do all of them at once after it's all merged.

wes-otf:
Reverting this back to a draft as I add features from #3389.

wes-otf:
@frjo @theskumar I believe this is ready to go. The only piece that is still up for discussion would be the behavior of the `403 Forbidden` page when a user without archive access tries to view and archived submission. The options I see are:

- Redirect the user to the submission dashboard
- Create a new "Access Denied" page/view that would allow all instances of `raise PermissionDenied` to be a little more descriptive and clean
- Leave the existing functionality as is

Let me know your thoughts!

frjo (approved):

theskumar:
> @frjo @theskumar I believe this is ready to go. The only piece that is still up for discussion would be the behavior of the `403 Forbidden` page when a user without archive access tries to view and archived submission. The options I see are:
> 
> * Redirect the user to the submission dashboard
> * Create a new "Access Denied" page/view that would allow all instances of `raise PermissionDenied` to be a little more descriptive and clean
> * Leave the existing functionality as is
> 
> Let me know your thoughts!

Explicit is better than implicit. I would suggest going with -- `Create a new "Access Denied" page/view that would allow all instances of raise PermissionDenied to be a little more descriptive and clean.`

wes-otf:
@theskumar All sounds good! The functionality for the 403 page is pretty much mirrored from the 404. It can be customized in Wagtail admin and is triggered when any `django.core.exceptions.PermissionDenied` error is raised. By default, it looks like the following:

Screenshot 2023-11-28 at 8 55 34 AM
-------
With the Wagtail admin prompts:
Screenshot 2023-11-28 at 9 02 15 AM

wes-otf:
This was also tested in the archive context (user with staff role attempted to view an archived submission) and worked well.

wes-otf:
This was tested by @Techslammer & all worked as intended. The latest changes of the 403 page hadn't been deployed to test but that was a minor change that shouldn't change too much. Can retest with that if you'd prefer @frjo, otherwise I'll mark it as ready.

Have a nice weekend y'all!

PR #3658, opened 2023-11-15
Added migration to remove NEW_SUBMISSION events without a submission

### LINKED ISSUE #3442: Not able to delete the user from wagtail admin

wes-otf:
Closes #3442. 

This small migration addition compliments #3445 as it will remove any previous `NEW_SUBMISSION` events that belong to submissions that no longer exist. This will allow for the removal of applicants that don't have any active submission, but can't be deleted due to the `The object you are trying to delete is used somewhere...` error.

Cheers!

### DIFF
--- a/hypha/apply/activity/migrations/0077_remove_deleted_submission_actions.py
+++ b/hypha/apply/activity/migrations/0077_remove_deleted_submission_actions.py
@@ -1,0 +1,32 @@
+from django.db import migrations
+
+from hypha.apply.activity.messaging import MESSAGES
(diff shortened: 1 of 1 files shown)


sandeepsajan0 (approved):
Looks good.
I tested it locally, working great. Thanks @wes-otf

frjo (approved):

PR #3659, opened 2023-11-16
Update 'Waiting for' content for contract upload 

### LINKED ISSUE #3644: Contract upload feature
- [ ] Is it okay if we continue to say "Pending signed contract by Contracting team" here? or should we say "Pending signed contract by OTF" or "Pending signed contract by the project lead" tagging @fourthletter for confirmation
image

- [ ] Also do we still want for the compliance team to be able to upload the contract if they wanted to? or should they be restricted from uploading the contract? in the latter case, we will have to hide the 'upload contract' button for compliance team.

sandeepsajan0:
Fixes #3644 

image

As of now, staff can upload contracts, so the 'waiting for' text is also updated accordingly from "Awaiting signed contract from Contracting team" to "Awaiting signed contract from Staff/Contracting team"(depends on the settings)

### DIFF
--- a/hypha/apply/projects/templatetags/project_tags.py
+++ b/hypha/apply/projects/templatetags/project_tags.py
@@ -146,6 +146,11 @@ def user_next_step_on_project(project, user, request=None):
                         org_short_name=settings.ORG_SHORT_NAME
                     ),
                 }
+            if settings.STAFF_UPLOAD_CONTRACT:
+                return {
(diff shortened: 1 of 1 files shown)


sandeepsajan0:
@frjo We can put it on the test or maybe merge it in as it just contains a text update.

frjo (approved):

PR #3661, opened 2023-11-23
Remove unused ‘report-frequency.js’

theskumar:
It is replaced by alpinejs code and causing js exception on the project page.


frjo (changes requested):
Good to clear out things like this! The reference to it on `project_detail.html` should be removed as well.

theskumar:
ah. missed that while stashing the file. Doing that now!

frjo (approved):

PR #3664, opened 2023-11-24
Add a CustomWagtailUserCreationForm to fix issue with creating users …

### LINKED ISSUE #3649: Can not add new users in Wagtail admin when "Show consent checkbox?" is checked
The form seems to never submit, just toggle between account/role tabs.

I suspect this is a bug from the Wagtail 5 upgrade. We have overriden code in the wagtail admin user section, we likely need to update that code with the Wagtail 5 changes to get it working again.

frjo:
…from Wagtail admin when Show consent checkbox was set.

Fixes #3649


hypha/apply/users/forms.py
```diff
@@ -51,6 +51,11 @@ def __init__(self, *args, **kwargs):
 
 
 class CustomUserAdminFormBase:
+    error_messages = {
+        "duplicate_username": _("A user with that email already exists."),
+        "password_mismatch": _("The two password fields didn't match."),
+    }
+
```
frjo, line 58: Moved it here so it becomes active on the edit form as well.

theskumar: 👍🏼

theskumar (approved):
Works perfectly!

PR #3666, opened 2023-11-26
Fixes and enhancements to project invoice detail page

theskumar:
1. Fix link to "view project page"
2. Add bg-light-grey as there are a log of white cards
3. Fix margin top on the top-most card.
4. Do not show supporting documents section if they are not present

## Before

Screenshot 2023-11-26 at 6  39 02@2x

## After
Screenshot 2023-11-26 at 6  36 37@2x

### DIFF
--- a/hypha/static_src/src/sass/apply/components/_wrapper.scss
+++ b/hypha/static_src/src/sass/apply/components/_wrapper.scss
@@ -202,7 +202,7 @@
         }
 
         .card:first-child {
-            margin-top: 20px;
+            margin-top: 0;
         }
     }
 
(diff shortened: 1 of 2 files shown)


hypha/static_src/src/sass/apply/components/_wrapper.scss
```diff
@@ -202,7 +202,7 @@
         }
 
         .card:first-child {
```
sandeepsajan0, line 204: I think we can just remove this? 
`.card` is used at other places as well but I don't think its first child top margin makes sense anywhere.

wes-otf (approved):

PR #3672, opened 2023-11-29
Restyle the status bar

### LINKED ISSUE #3669: Status bar labels are frequently misaligned/off-center
When looking at the progress/stage indicator of an application, the titles are frequently misaligned and not consistent.

The following are screenshots with lines drawn through the center of the stage points to better show this misalignment. Some are better than others:

Screenshot 2023-11-28 at 11 00 36 AM
-------------
Screenshot 2023-11-28 at 10 54 00 AM(1)
-------------
Screenshot 2023-11-28 at 10 54 46 AM
-------------

Definitely not a mission-critical bug but would still be nice for appearance.

frjo:
Fixes #3669

Make the status bar work with different fonts, viewports etc. and always align.

### DIFF
--- a/hypha/static_src/src/sass/apply/components/_status-bar.scss
+++ b/hypha/static_src/src/sass/apply/components/_status-bar.scss
@@ -44,9 +44,11 @@
         flex: 1;
         position: relative;
         padding-block-start: var(--tooltip-margin-top);
+
+        // The bar for each item.
         border-block-start: var(--bar) solid $color--mid-grey;
 
-        // every items dot
+        // The dot for each item.
         &::before {
             display: flex;
             place-items: center;
@@ -69,6 +71,11 @@
             &::before {
                 background: $color--white;
                 border: calc(var(--dot) / 4) solid $color--tomato;
+            }
+
+            // Fill the bar all the way on accepted/declined.
+            &:last-of-type {
+                border-color: $color--primary;
             }
         }
 
--- a/hypha/static_src/src/sass/apply/components/_status-bar.scss
+++ b/hypha/static_src/src/sass/apply/components/_status-bar.scss
@@ -83,6 +83,7 @@
             border-color: $color--primary;
 
             &::before {
+                color: $color--white;
                 background: $color--primary;
                 border-color: $color--primary;
                 content: "✔";
--- a/hypha/static_src/src/sass/apply/components/_status-bar.scss
+++ b/hypha/static_src/src/sass/apply/components/_status-bar.scss
@@ -1,10 +1,15 @@
 // stylelint-disable  selector-class-pattern
 
 .status-bar {
+    --bar: 3px;
+    --dot: 20px;
+    --triangle: 5px;
+    --tooltip-padding: 10px;
+    --tooltip-max-width: 12ch;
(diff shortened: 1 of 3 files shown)


frjo:
More work needed, this code is used for submissions, projects, on dashboard etc. Only tested on submissions so far.

wes-otf:
@frjo doing user testing it looks like a huge improvement! I did noticell a few weird things come up, like the checks were now a little odd looking and the red progress indicator wouldn't always make it to the needed stage:
Screenshot 2024-01-05 at 17 52 52
Screenshot 2024-01-05 at 17 53 58

I cleared my cache & tested in an incognito window to confirm and the results were the same, but let me know if there's another step I might've missed.

frjo:
@wes-otf Thanks for testing. Fixed the bug where the bar did not get filled on accept/dismissed status.

That the checkmarks are not aligned in the dot is most likely font related. With the system-ui font stack you get different fonts on different plattforms. Needs to be adjusted so it looks nice everywhere.

In your screenshot the checkmark is not white like it is for me, another browser/plattform issue I guess.

wes-otf:
@frjo ahh that totally makes sense. My testing was with Firefox & MacOS if that's helpful. Would it be better to potentially use the heroicons `check` instead of the `✔` character so the appearance is guaranteed regardless of platform?

wes-otf:
I'm definitely not well versed history of this feature so apologies if that's something that's already been tried ^

frjo:
The checkmark should be white for all now.

If we can not easily make it center for all browsers/plattforms, replacing it with a SVG or maybe force it to another font is a good options I think.

frjo:
@wes-otf  Now I think I finally found the issue. The first fix did nothing in Chrome.

Chrome does not set colour on unicode checkmark 2714, it works on 2713. Apparently 2714 is considered an emoji and Chrome does not change colour on these, Webkit does.

frjo:
@wes-otf Do you feel this is good to go now? Planing to make a new release this week.

hypha/static_src/src/sass/apply/components/_status-bar.scss
```diff
@@ -26,189 +31,136 @@
...
+        &--is-complete {
+            border-color: $color--primary;
+
             &::before {
-                left: -10px;
+                font-size: map-get($font-sizes, zeta);
+                font-weight: bold;
+                background: $color--primary;
+                border-color: $color--primary;
+                content: "✓";
```
wes-otf, line 92: All looks good! The only super small thing I saw on test was that the applicant dashboard's status bar still had a black color for the check. Might be a good idea to keep the additions of [49ffdbd](https://github.com/HyphaApp/hypha/pull/3672/commits/49ffdb8d722c9d0337b1f59bf2b9176abeeddae4)

```suggestion
                content: "✓";
                color: $color--white;
```

wes-otf: Screenshot 2024-01-11 at 09 33 05

applicant dashboard in question ^

frjo: Missed that one, thanks for testing!

PR #3673, opened 2023-11-30
Fix typography for submission diff

### LINKED ISSUE #3671: Compare view formatting is sometimes confusing to read
The formatting of the compare view needs some small touch-ups. The issues I've noticed are:

1. Lack of spacing between a heading and the previous data field
2. Inconsistent header sizes for different fields without clear section separation
3. Compare formatting for numbers sometimes has weird wrapping properties
4. If the application has short text fields, the compare view has little to no spacing between the two revisions

----

Screenshot 2023-11-29 at 12 05 56 PM
Screenshot 2023-11-29 at 11 57 03 AM

----

Again, not critical by any means but a "nice to have".

theskumar:
- The numbers don't break
- make the heading consistent, everything gets a h4
- Update the color and style for added/deleted text

Screenshot 2023-11-30 at 10  26 31@2x

closes https://github.com/HyphaApp/hypha/issues/3671


theskumar:
I've able to get another version of compare view working with rich text display. See some of the screenshots below. If there is enough interest, after we merge this, I'll polish this up a bit and send a pull request. It uses [html-diff==0.4.1](https://pypi.org/project/html-diff/)

The styling is borrowed from the Github's markdown diff preview. 

Screenshot 2023-11-30 at 12  23 20@2x

Screenshot 2023-11-30 at 12  26 46@2x

wes-otf (approved):
Looks/works great! Especially the `Proposal Infomation (Old)/(New)` is a super useful addition.

wes-otf:
Two small things after more testing not mentioned in the original issue:
1. Ordered lists get turned to unordered lists and new lines look to be ignored sometimes in Rich Text fields (ie. "Here are some bullets:", "Now for some numbering:" & "Here's a link:"). 
2. Links & their changes are not shown.

In the following screenshots the `Open Tech Fund` link was swapped from https://opentech.fund to https://www.opentech.fund but the compare shows this as unchanged.

The application w/ Rich Text & Markdown:

----
Screenshot 2023-11-30 at 11 05 53 AM

The compare view:

----

Screenshot 2023-11-30 at 11 05 21 AM

frjo:
@wes-otf Good point on the links, we should add some code to handle them.

The submission contains large amount of HTML due to the RichTextEditor and that many people paste in text straight from Microsoft Word etc. To diff the HTML directly is not workable.

What Hypha instead does is to strip all the HTML and compare that instead. Then the code tries to reassemble the text to show it in some resemblance to the original.

This works but is far from optimal. We just discussed today that it might be worth testing to convert the text to markdown instead and compare that.

theskumar:
> We just discussed today that it might be worth testing to convert the text to markdown instead and compare that.

After the call, I spent some more time with it. See my experiment with diffing the html itself and displaying it to user here https://github.com/HyphaApp/hypha/pull/3673#issuecomment-1833682465

If the rich text support is needed, this might be the best way to go.

wes-otf:
I think that would make a lot of sense. Would definitely be interested in your HTML diff @theskumar. Maybe we could do a beta for something like that similarly to the submission view to gauge interest.

PR #3676, opened 2023-12-05
Remove unused pagedown css demo.css

theskumar:


frjo:
Without this file this CSS gets loaded when the Pagedown editor is active: https://github.com/StackExchange/pagedown/blob/32c0bdb121de04e0e956ddde0391022644ff1a76/demo/browser/demo.css

theskumar:
> Without this file this CSS gets loaded when the Pagedown editor is active: https://github.com/StackExchange/pagedown/blob/32c0bdb121de04e0e956ddde0391022644ff1a76/demo/browser/demo.css

Thanks for pointing that out, I still feel not loading the css file in the first would be better. I've updated the pagedownwidget here https://github.com/HyphaApp/hypha/pull/3676/commits/d913750ced7bd7e2eec80a9c1969f314608240fb

frjo (approved):

PR #3685, opened 2023-12-09
Remove unused tomd from dependency

theskumar:
IT's use by to_markdown template tag, which is not in use anywhere in
the codebass. Removing this as it's no longer used


frjo:
It is used in one single place, see end of line 37 below:

https://github.com/HyphaApp/hypha/blob/47613cbddba3e129fd4b5d1c5609711d4e0589b1/hypha/apply/activity/templates/activity/include/listing_base.html#L37-L42

theskumar:
Strange it didn't show up in my search earlier, closing this now.

PR #3686, opened 2023-12-09
Upgrade Wagtail to 5.2.x LTS

### LINKED ISSUE #3674: Update to Wagtail 5.2.x LTS
Should go well with Django 4.2.x LTS

theskumar:
- Update requirements
- Replace classnames with classname
- update wagtail admin page of `settings -> users`

closes #3674

### DIFF
--- a/hypha/apply/categories/admin_helpers.py
+++ b/hypha/apply/categories/admin_helpers.py
@@ -10,11 +10,11 @@ class MetaTermButtonHelper(ButtonHelper):
             return
         return super().delete_button(pk, *args, **kwargs)
 
-    def prepare_classnames(self, start=None, add=None, exclude=None):
-        """Parse classname sets into final css classess list."""
-        classnames = start or []
-        classnames.extend(add or [])
-        return self.finalise_classname(classnames, exclude or [])
+    def prepare_classname(self, start=None, add=None, exclude=None):
+        """Parse classname sets into final css classes list."""
+        classname = start or []
+        classname.extend(add or [])
+        return self.finalise_classname(classname, exclude or [])
 
     def add_child_button(self, pk, child_verbose_name, **kwargs):
         """Build a add child button, to easily add a child under meta term."""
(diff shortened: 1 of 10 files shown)


wes-otf:
Hey, looks like all went pretty well! The only thing I noticed was that it look like the add users button, header & search functionality was removed somehow here (could be an issue with my setup?). The user view got more compact though, which I think looks slick.

**Before (main)**
Screenshot 2023-12-11 at 11 05 27


----

**After**

Screenshot 2023-12-11 at 10 55 45

wes-otf:
This only looks to be an issue with the users view, the groups view (`/admin/users/`) has the header, add group & search button.

theskumar:
@wes-otf Nice catch. I've updated the users views and view for groups. Assuming the only customization to the users listing views was the fact that search needs to account for `full_name` instead of `first_name` and `last_name`.

frjo:
@wes-otf @theskumar The roles and status filters in the sidebar is gone with this PR.

Skärmavbild 2024-01-22 kl  13 36 55

frjo (changes requested):
I assume the filters should still be there.

theskumar:
Filtering should be back. It was little weird that the `filtering` logic has been skipped in the implementation of the user's setting page.

frjo:
Filters works perfectly now. I notices a minor issue with column sorting. It works for the "Email" column but not for the "Display name" column. The header is a link and all and clicking it reloads the page but no sorting seems to happen.

But if you first click "Email" header and then directly click "Display name" the query is added to the url and sorting does work.

theskumar:
> Filters works perfectly now. I notices a minor issue with column sorting. It works for the "Email" column but not for the "Display name" column. The header is a link and all and clicking it reloads the page but no sorting seems to happen.
> 
> But if you first click "Email" header and then directly click "Display name" the query is added to the url and sorting does work.

In the view there is no `asc` and `desc` ordering. The field is either in sorted or unsorted state. 

The reason for the behavior you see with the "Display Name" is because the list is by default ordered by "display name", clicking on the "Display Name" was suppose to remove the ordering, but there is not sorting to remove, so it does nothing. 

I've remove the initial `<a href` so it become clickable only after some other sort is in effect. Clicking on the "email" sort will enable the sorting on the display name.

frjo (approved):

PR #3694, opened 2023-12-12
Remove public site from management commands

wes-otf:
This PR will remove the public site from management commands and fix the IDs that are used when the apply site is used to reflect that of production.

### DIFF
--- a/hypha/apply/home/migrations/0002_add_apply_homepage.py
+++ b/hypha/apply/home/migrations/0002_add_apply_homepage.py
@@ -29,7 +29,7 @@ def create_homepage(apps, schema_editor):
 
     # Create a site with the new homepage set as the root
     Site.objects.create(
-        hostname="apply.localhost", root_page=applyhomepage, is_default_site=False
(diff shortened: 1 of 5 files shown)


hypha/apply/home/migrations/0002_add_apply_homepage.py
```diff
@@ -29,7 +29,7 @@ def create_homepage(apps, schema_editor):
 
     # Create a site with the new homepage set as the root
     Site.objects.create(
-        hostname="apply.localhost", root_page=applyhomepage, is_default_site=False
+        id=2, hostname="apply.localhost", root_page=applyhomepage, is_default_site=False
```
theskumar, line 32: This migration will only run for fresh setup and not those where the migrations has already been run. 

Is the intension here to fix these IDs for the existing setups? If so, it would need a fresh migration with custom python code.

wes-otf:
@frjo @theskumar what are the thoughts on this? anything else need to happen to get it over the line?

wes-otf:
This was closed by #3738

PR #3697, opened 2023-12-15
Remove public from root URLs

### LINKED ISSUE #3559: Previewing rounds in Wagtail admin result in error

wes-otf:
Fixes #3559. This will remove the public URLs from the root `urls.py` file, and fix the round preview failure via the removal of the previous middleware "hack" to inject the `hypha.apply.urls` url conf. This relies on #3689 and #3687.


wes-otf:
weird, I can't get that `tus_upload` to fail locally. Not sure what's causing that.

theskumar:
Closing this as it will be covered in #3738

PR #3698, opened 2023-12-18
PDF stamping of approved invoices 

### LINKED ISSUE #3539: Invoice has name of approver and date stamp
### Description

Manipulate PDF invoice submitted by vendor with the approver information and date stamp

theskumar:
- [x] Add approved by staff to the approved invoice pdfs
- [x] Add inovice approver and date

Screenshot 2024-05-20 at 6  33 05@2x

Closes: https://github.com/HyphaApp/hypha/issues/3539

### DIFF
--- a/hypha/apply/projects/views/payment.py
+++ b/hypha/apply/projects/views/payment.py
@@ -6,6 +6,7 @@ from django.contrib.auth.models import Group
 from django.core.exceptions import PermissionDenied
 from django.db import transaction
 from django.shortcuts import get_object_or_404, redirect
+from django.template.loader import render_to_string
 from django.utils import timezone
 from django.utils.decorators import method_decorator
 from django.utils.safestring import mark_safe
@@ -35,6 +36,7 @@ from hypha.apply.todo.views import (
 )
 from hypha.apply.users.decorators import staff_or_finance_required
 from hypha.apply.users.groups import STAFF_GROUP_NAME
+from hypha.apply.utils.pdfs import html_to_pdf, merge_pdf
 from hypha.apply.utils.storage import PrivateMediaView
 from hypha.apply.utils.views import (
     DelegateableListView,
@@ -121,6 +123,7 @@ class ChangeInvoiceStatusView(DelegatedViewMixin, InvoiceAccessMixin, UpdateView
             and self.request.user.is_finance_level_1
             and self.object.status == APPROVED_BY_FINANCE
         ):
+            self.object.save()
             messenger(
                 MESSAGES.APPROVE_INVOICE,
                 request=self.request,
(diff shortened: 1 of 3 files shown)


theskumar:
@sandeepsajan0 @frjo Are you good with denormalizing and add the `approved_at` and `approved_by` to the invoice model? 

The code to update them will go in here: 

https://github.com/HyphaApp/hypha/blob/1b401334914f013b7d30e9bf4c2838361ee7c8f7/hypha/apply/projects/views/payment.py#L81-L87

theskumar:
Added the logic to store the approver and date at https://github.com/HyphaApp/hypha/pull/3698/commits/f426d46f5be088b3228dca94c155789ceb28128c

Let me know if this something that need to be improved upon? e.g. displaying both finance1 and finance2 info etc.

hypha/apply/projects/models/payment.py
```diff
@@ -183,6 +183,13 @@ class Invoice(models.Model):
         verbose_name=_("Invoice amount"),
     )
     status = FSMField(default=SUBMITTED, choices=INVOICE_STATUS_CHOICES)
+    approved_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+    )
+    approved_at = models.DateTimeField(null=True, blank=True)
```
sandeepsajan0: Can we rename it to `final_approval_...` or something that doesn't confuse it with internal approvals like staff approval or finance approval(in case of extended workflow)? Because we have to stamp the pdf only with final approval(finance approval or finance2 approval).

Or we may directly look out for the final approval activity as the [activity](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/activity/models.py#L98) already has the date and user.

hypha/apply/projects/views/payment.py
```diff
@@ -84,6 +86,9 @@ def form_valid(self, form):
             and self.request.user.is_finance_level_1
             and self.object.status == APPROVED_BY_FINANCE
         ):
+            self.object.approved_by = self.request.user
```
sandeepsajan0: It is not the right place to update approved_by or approved_at because here we are just notifying the finance members to approve the invoice.

It requires adding another condition to check if the status is `Approved by Finance`(without extended).

theskumar:
@sandeepsajan0 thanks for feedback. I would like wait for the feedback on the content sample itself and then tackle both at once. 

- https://github.com/HyphaApp/hypha/issues/3539#issuecomment-1860480909

theskumar:
Updated the code to make use of activity feed. It will list all the approvals with timestamps.

hypha/apply/projects/views/payment.py
```diff
@@ -416,12 +419,35 @@ def dispatch(self, *args, **kwargs):
...
+        # if not, then it's for invoice document
+        if (
+            self.invoice.status == APPROVED_BY_STAFF
+            and self.invoice.document.file.name.endswith(".pdf")
+        ):
+            if activities := Activity.actions.filter(
+                related_content_type__model="invoice",
+                related_object_id=self.invoice.id,
+                message__icontains="Approved by",
+            ).visible_to(self.request.user):
```
sandeepsajan0, line 436: It will pick all the Approved by ... statuses. If an invoice has been back and forth between finance and staff then it might show multiple `Approved by Staff` but I think only latest one should be fine. We might need distinct activities.  Also, we don't want to add this for invoices with status `Approved by finance`?  And we might need to hide it from the applicant.
image

hypha/apply/projects/views/payment.py
```diff
@@ -121,6 +123,7 @@ def form_valid(self, form):
             and self.request.user.is_finance_level_1
             and self.object.status == APPROVED_BY_FINANCE
         ):
+            self.object.save()
```
sandeepsajan0, line 126: I think it was getting saved in `form_valid()` in line100?

wes-otf (approved):

PR #3702, opened 2023-12-27
Make default logo and favicons Hypha branded

frjo:


frjo:
@wes-otf When this gets deployed to OTF Live we will need to add OTF logo to the Wagtail admin site settings.

Can be tested on OTF test to make sure the images fit well.

It is generally a good idea to have images with 2x the resolution needed. If the logo is displayed as 200x50 pixels, the image should be 400x100 pixels. This makes it look nice and sharp on high resolution displays (any Apple device e.g.)

wes-otf:
Sounds good to me!

PR #3727, opened 2024-01-17
Fix errors for compare revisions page(JSONDecodeError and Invalid literal int error)

### LINKED ISSUE #3709: Editing a submission leads to duplicate entries in the "revisions" section, one of which causes an error
When a proposal has been edited, it will sometimes create a duplicate entry in the revisions section. One of these duplicates (typically the oldest entry) will result in an error of `invalid literal for int() with base 10: ''` at [hypha/apply/funds/blocks.py](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/blocks.py#L200) when attempting to open it, due to a lack of a `duration` field.

sandeepsajan0:
Fixes #3709

### DIFF
--- a/hypha/apply/funds/blocks.py
+++ b/hypha/apply/funds/blocks.py
@@ -51,6 +51,8 @@ class ValueBlock(ApplicationSingleIncludeFieldBlock):
         icon = "decimal"
 
     def prepare_data(self, value, data, serialize):
+        if not data:
+            return data
         return format_number_as_currency(str(data))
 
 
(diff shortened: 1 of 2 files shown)


frjo (approved):

frjo:
@wes-otf Can you please test this since you have a good grasp of the issue.

wes-otf:
@frjo totally! is it on test yet? I think that's going to be the easiest way to confirm all is amended

frjo:
Deploying to test now, will take a couple of minuts.

wes-otf:
looks like it works to me!

wes-otf (approved):

PR #3728, opened 2024-01-17
Add dir attribute to the html tag, ltr/rtl.

### LINKED ISSUE #3726: Improve RTL support
- [x] Add configurable dir attribute to base templates.

frjo:
Fixes #3726

### DIFF
--- a/hypha/templates/base-apply.html
+++ b/hypha/templates/base-apply.html
@@ -1,7 +1,8 @@
 {% load i18n static wagtailcore_tags wagtailimages_tags navigation_tags util_tags hijack cookieconsent_tags %}<!doctype html>
 {% wagtail_site as current_site %}
 {% get_current_language as LANGUAGE_CODE %}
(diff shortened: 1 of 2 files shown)


frjo:
@frankduncan This add basic rtl support to Hypha. Django can autodetect the direction so no settings needed, if you e.g. set the language to "ar" the direction will be "rtl" as it should be.

frjo:
Hypha has a sass function "opposite-direction" to handle ltr/rtl but that is an old way of doing it. You must also explicitly make use of it everywhere and that is not so now.

Much better is to replace all use of things like "margin-right" with "margin-inline-end", "padding-top" with "padding-block-start" etc. With these you get rtl support without doing anything, the browser just does the right thing.

theskumar (approved):
Nice 💯

frankduncan:
Ah!  This will be great when hypha as a whole is set to an RTL language (and the localization file is created with all the translations).  We'll still need a solution when it has to do both simultaneously.  Something like #3293 would help with that.

PR #3732, opened 2024-01-18
Improve rtl support with modern css

### LINKED ISSUE #3731: Use -block* and -inline* css attributes instead of top/right/bottom/left
This improves rtl support among other things.

frjo:
Fixes #3731

Improves rtl support among other things. Replaces attributes like top/left/margin-right/padding-left as well as tailwind classes with `*-inline*` and `*-block*` version.

`padding-inline-start` is for ltr the same as `padding-left` but for rtl it is the same as `padding-right`. The browser just does the right thing.

For tailwind we should always use `ms-1` instead of `ml-1` and `me-1` instead of `mr-1` etc. The s/e stands for start/end.

## Test Steps
 - [ ] Look at multiple pages and confirm that the styling is not off/look weird anywhere.

### DIFF
--- a/hypha/apply/activity/templates/activity/include/listing_base.html
+++ b/hypha/apply/activity/templates/activity/include/listing_base.html
@@ -4,7 +4,7 @@
     <div class="feed__pre-content hidden lg:block">
         <p class="feed__label lg:py-2 feed__label--{{ activity.type }}">
             {% if  activity.type == 'comment' %}
-                {% heroicon_mini "chat-bubble-left" class="inline align-text-bottom mr-2" aria_hidden=true %}
+                {% heroicon_mini "chat-bubble-left" class="inline align-text-bottom me-2" aria_hidden=true %}
             {% else %}
                 {{ activity.type|capfirst }}
             {% endif %}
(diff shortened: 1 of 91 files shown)


theskumar:
Nice. 

Small doubt, for the icons, since the icons itself are not moving to the right on the text, does it make sense to update the margin on the icon from right to left. 

As per current logic, I believe this is what is going to happen.

ltr setup
```
[icon][margin][text]
```

rtl setup, it will become:
```
[margin][icon][text-reversed]
```

We really need a space b/w the icon and the text. 

Also, what would be best way to test the code for a rtl language?

theskumar:
Ignore my comment above. I tested with `dir=rtl` with english itself, and looks like browser also swaps the icon position, and does exactly how it should happen. 

which is

```
[text][margin][icon]
```

💯

theskumar (approved):

theskumar:
Looking good. Let's get this merged in. 

Can you also create another ticket to find other minor issues related to `rtl` support and fix them.

Techslammer:
Feedback from test: "The 'copy questions to clipboard' doesn't really align with anything. I could see it better fitting to the direct right of the title and centered, or maybe directly under. Not a priority by any means"
Screenshot 2024-02-09 3 54 01 PM

Techslammer:
Feedback from test: also not a priority, but the 'requesting to receive services' choice might be missed, would be better if directly under the other two options (Rapid Response Round - Jan 2024)
Screenshot 2024-02-09 3 54 47 PM

Techslammer:
Feedback from test: more space between the 'addressed problems' in RRF could make the options easier to read (edited) 
Screenshot 2024-02-09 3 55 18 PM

Techslammer:
Above are super minor so we can move this forward to deploy live. Let me know if that is the case.

frjo:
@Techslammer Please create issues for these things. They are unrelated to this PR.

The "the 'requesting to receive services' choice" is in an option you add to application forms in Wagtail admin. The forms on test are very likely different form them on live.

Techslammer:
@frjo When this is deployed live, i'll check in with PMs to review if those issues are still there and if so, will create issues on GH

PR #3736, opened 2024-01-25
User with Staff role can approve contracts

### LINKED ISSUE #3735: If a user has staff and contracting roles they can not approve contracts
The `can_approve_contract` can be a little simplified to allow anyone with staff role to approve contracts. User that has no staff roles falls down to the default return value that is `False`.

frjo:
Fixes #3735

### DIFF
--- a/hypha/apply/projects/permissions.py
+++ b/hypha/apply/projects/permissions.py
@@ -34,7 +34,7 @@ def can_approve_contract(user, project, **kwargs):
     if not user.is_authenticated:
         return False, "Login Required"
 
-    if user.is_apply_staff and not user.is_contracting and not user.is_applicant:
+    if user.is_apply_staff and not user.is_applicant:
         return True, "Only Staff can approve the contract"
 
     return False, "Forbidden Error"


frjo:
ARDC had an issue with this. A user with contracting + staff role was not allowed to approve contracts.

I think it should be ok to let any user with staff role approve contracts.

hypha/apply/projects/permissions.py
```diff
@@ -34,7 +34,7 @@ def can_approve_contract(user, project, **kwargs):
     if not user.is_authenticated:
         return False, "Login Required"
 
-    if user.is_apply_staff and not user.is_contracting and not user.is_applicant:
+    if user.is_apply_staff:
         return True, "Only Staff can approve the contract"
```
theskumar, line 38: If removing `is_applicant` then maybe ensure that the user approving is not applicant or partner on the project? I am assuming it's going be rare but just thinking around _least privilege_ point of view.

frjo: Good point.

sandeepsajan0 (approved):
Back then it was the requirement to hide it from Staff+contracting role because the user with Staff + contracting role can upload the contract and also can approve the contract but as of now, we have provided an option of uploading the contract to the staff as well so I think it is good to loosen up the approve permission as well.

PR #3738, opened 2024-01-29
Remove public site

theskumar:
## Description

- Keeps the migrations to delete the public tables
- Updates and adds a global footer (admin configurable)
- Update the layout of homepage

Fixes #3110 #3559

## Test Steps

 - [ ] Remove the public site, go to "Wagtail Admin" > "Pages", set the "Apply Home Page" as the default page of the site. Remove the public site page
 - [ ] Do regression of the whole site


frjo:
I have merged in all "Remove public…" PRs before this one now so this can be rebased of main.

frjo:
I merged in the icon PR so this can be rebased.

I also see this error:

~~~~
WARNINGS:
?: (urls.W003) Your URL pattern 'account/two_factor/setup/complete/' [name='two_factor:setup_complete'] has a name including a ':'. Remove the colon, to avoid ambiguous namespace references.
?: (urls.W005) URL namespace 'apply:projects:reports' isn't unique. You may not be able to reverse all URLs in this namespace
~~~~

frjo:
So nice seeing all the delete file statements in this PR!

theskumar:
> I merged in the icon PR so this can be rebased.
> 
> I also see this error:
> 
> ```
> WARNINGS:
> ?: (urls.W003) Your URL pattern 'account/two_factor/setup/complete/' [name='two_factor:setup_complete'] has a name including a ':'. Remove the colon, to avoid ambiguous namespace references.
> ?: (urls.W005) URL namespace 'apply:projects:reports' isn't unique. You may not be able to reverse all URLs in this namespace
> ```

These two warning doesn't seem to b/c of changes in the PR. If it make sense I'll update them in this PR or have them resolved in another PR.

theskumar:
Unable to get the failing test reproduced locally. Will try it tomorrow.

hypha/templates/base-apply.html
```diff
@@ -1,102 +1,62 @@
...
-        <header class="header">
-
-            <div class="header__inner wrapper wrapper--large">
-                <a href="{{ settings.utils.SystemMessagesSettings.site_logo_link|default:"/" }}" aria-label="Home link">
-                    {% if settings.utils.SystemMessagesSettings.site_logo_default %}
-                        {% image settings.utils.SystemMessagesSettings.site_logo_default width-215 as logo_default %}
-                        <img class="header__logo header__logo--desktop"
-                             width="215"
+{% extends "base.html" %}
+{% load i18n static wagtailcore_tags heroicons %}<!doctype html>
```
frjo: Need to add `wagtailimages_tags` here.

hypha/apply/home/templates/apply_home/includes/apply_listing.html
```diff
@@ -1,23 +1,35 @@
...
-                {{ page.title }}
-            </h4>
-            {% if page.description %}
-                <h6 class="listing__teaser">
-                    {{ page.description|markdown|bleach }}
-                </h6>
-            {% endif %}
-        </div>
-        <a class="listing__button" href="{% pageurl page %}">{% trans "Apply" %}</a>
+  <div class="flex justify-between items-center py-8">
```
frjo: A "gap" would be good here I think. To get some space between the text and the button.

frjo (changes requested):

hypha/templates/base-apply.html
theskumar: Since this page doesn't use any `{% image ..}` tag `wagtailimages_tags` is not needed on this page. Are you getting any error?

hypha/apply/home/templates/apply_home/includes/apply_listing.html
theskumar: Totally agree. I believe it got lost in the rebasing :D

theskumar: Added.

hypha/templates/base-apply.html
theskumar: Nice catch! Added

wes-otf:
@theskumar really curious what your fix is for the tests. When I tried to fix the fund previews back in Oct. I kept getting [the same issue](https://github.com/HyphaApp/hypha/pull/3615#issuecomment-1772817453). It seemed like the unit tests were dependent on my local database for whatever reason and when I would do a full reset of that database I was start to see the failures recreated locally.

wes-otf:
Once this issue is cornered we also should  probably add unit test refactoring to the roadmap for this year. All the tests should run isolated and unaffected by the database (if that is the case)

wes-otf:
@theskumar This fixed all issues for me: https://github.com/HyphaApp/hypha/commit/6de8fff45ca52392ccf24e2fe96c960aba19301a. Might just be the start to a larger issue but it seems like there's some objects that are being initialized too early in the API app. Seems like wrapping them in functions does the trick? Curious to hear your thoughts or if you have another solution already cooked up.

theskumar:
Django did had start up issues in very old versions before they introduced the `apps.py` config option with `ready()` function in that would be excecuted after all the model classes are loaded. So that could be it.

frjo (approved):

theskumar:
@wes-otf you saved me hours. The solutions seems to be working and accessing the models before its initialized is a very bad idea in general. The loading of models and apps are very unpredictable. 

That why Django introduced https://docs.djangoproject.com/en/5.0/ref/applications/#django.apps.AppConfig.ready to you can perform any on_load activities here. 

I believe there are more code in the forms.py in certain apps that is initialized like this, instead of in the `__init__.py` looks like the fixes things for now. But we can re-visit and cleap all module level queries.

frjo:
This doc need updating at some point as well:

https://github.com/HyphaApp/hypha/blob/50693eb9d95865925b24028911cf37ad4269a529/docs/getting-started/contributing/code-contributions.md#L19-L34

theskumar:
> Reversing URLS

Updated.

theskumar:
Looking into the test case failing. Let's hold on to merging before that. This something to do with the updated `Sites` setting.

frjo:
Can it be that the initial migrations created a public and apply site and set the public as default. When testing against this we get errors.

wes-otf:
@theskumar glad it worked!! Was a bit of a shot in the dark, weird stuff.

theskumar:
> Can it be that the initial migrations created a public and apply site and set the public as default. When testing against this we get errors.

Yes. You are absolutely right. updated the code in https://github.com/HyphaApp/hypha/pull/3738/commits/b788eec4c591313cf4fbbafcd889e2d2c2b68c4d

Re-created by creating a fresh db, ran migration, without adding the sandbox db data

frjo:
Deploying this to test now.

frjo:
I have set the apply site as default on test now.

Skärmavbild 2024-02-06 kl  12 57 20

PR #3744, opened 2024-02-01
Visibility enhancements for partners

### LINKED ISSUE #3733: Comments with a visibility of `all` don't have a visibility indication
I believe this is intended functionality but we've seen a few issues where it can be confusing to know what the visibility is on a comment (that has visibility set to all) without the visibility indicator.

**Priority**

**Affected roles**

- Staff
- Staff Admin

### LINKED ISSUE #3737: Partners should be able to change visibility options on comments
When a partner is added to an application, they should be able to change who can see comments they leave (**just** staff, or both staff and the applicant). There maybe be a need for the partner to share sensitive information of just communicate with OTF staff.

wes-otf:
## Description

Closes #3737 and Closes #3733. This adds a new visibility setting of `applicant partners`. This addition turned out to be a bit unwieldy because of how situation it is. The main constraints I was designing for was in the case of a lab/fund that was using the partner role for vendors. I was trying to also account for the possibility that multiple vendors will be assigned to a single submission (which I believe is an edge case). 

The main concern was applicants/vendors inputting sensitive data into the communications tab and having each other see such. For example: if a vendor uploads banking information an applicant (or other vendor) should not be able to see this information and vice versa. 

Also tried to include a few docstring additions in lieu of the documentation updates.

### Setup

 - [ ] As an **Applicant** submit a new application to a fund/lab.
 - [ ] Leave two comments with the visibilities of `Staff only` & `Partners`
 - [ ] As a **Staff Admin** add a Partner account (that you can access) to the submission
 - [ ] Leave three comments with the visibilities of: `Applicant`, `Partners`, & `Applicant & Partners`.
 - [ ] As the previous mentioned **Partner** account, navigate to the comms. tab of the submission.
 - [ ] Leave three comments with the visibilities of: `Applicant`, `Partners`, & `Staff only`.

### Testing Partner Visibility

 - [ ] As the previous mentioned **Partner** account, navigate to the comms tab of the submission with these comments.
 - [ ] Ensure the only visible comments are the ones that were left with the visibility of `Partners` & `Applicant & Partners`
 - [ ] Ensure the comment the current account left with the visibility of `Staff only` is also visible
 
### Testing Applicant Visibility

- [ ] As the **Applicant** of the submission, navigate to the comms tab.
- [ ] Ensure the only visible comments are the ones that were left with the visibility of `Applicant` & `Applicant & Partners`

### Testing Staff Visibility

- [ ] As **Staff Admin** navigate to the comms tab.
- [ ] Ensure all previously made comments comments are visible.

### Testing `all` label (#3733)

- [ ] As **Staff Admin** navigate to the comms tab.
- [ ] Make a comment with visibility of `All`
- [ ] Ensure there is a visibility indicator with an eye icon labeled `all` (see below for example).
Screenshot 2024-02-01 at 14 20 48


hypha/apply/activity/models.py
```diff
@@ -189,38 +213,93 @@ def __str__(self):
...
+        Returns:
+            A list of tuples in the format of:
+
+            [(<visibility string>, <visibility display string>), ...]
+        """
+        if submission_has_partner:
+            if user.is_applicant:
+                return [
+                    (APPLICANT_PARTNERS, VISIBILITY[PARTNER]),
+                    (APPLICANT, VISIBILITY[TEAM]),
```
wes-otf: For the applicant, it didn't make much sense to me to set the choice to `Applicant`. I figured this would remove the ambiguity.

hypha/static_src/src/sass/public/components/_button.scss
wes-otf, line 1: This will also likely be a merge conflict once #3738 comes through but I couldn't build w/o it so will fix after.

frjo:
This is what I see when logged in as a partner. The option "Partner" I assume makes it possible for the applicant to see the message as well. I think this should be made more clear.

Skärmavbild 2024-02-02 kl  09 38 34

wes-otf:
That's really odd, this should be what you see:
Screenshot 2024-02-02 at 08 04 53

are you any other roles aside from partner? and is the partner assigned to this application?

frjo:
Yes, the user was applicant + partner, when I uncheck applicant I to see this.

Skärmavbild 2024-02-02 kl  14 24 07

frjo:
But having partner and applicant role is not that uncommon I think. As soon as you apply for anything you get the applicant role.

I think that if you have the partner roles and are assigned as a partner to a submission you should be treated as a partner. If you have a applicant role should not effect in this case. Sounds logical?

wes-otf:
@frjo Yeah that makes total sense and would likely simplify things a bit for staff too. That last commit disallows applicants to be partners on their own application and checks if a user is a listed partner on the application before showing the partner visibility. I think this might be more inline with that logic.

frjo (approved):
Nice!

PR #3750, opened 2024-02-06
Fix url namespace warnings.

### LINKED ISSUE #3748: Fix URL namespace warnings
We should fix these warnings that keeps poping up.

```
?: (urls.W003) Your URL pattern 'account/two_factor/setup/complete/' [name='two_factor:setup_complete'] has a name including a ':'. Remove the colon, to avoid ambiguous namespace references.
?: (urls.W005) URL namespace 'apply:projects:reports' isn't unique. You may not be able to reverse all URLs in this namespace
```

The first is just a renaming I think.

The other looks like a bug, the "vendor/" path is named "reports", same name as the actual path "reports/".

frjo:
Fixes #3748.

To silence the remaning Django check warning one can add this to `local.py`:

`SILENCED_SYSTEM_CHECKS = ["urls.W003"]`

### DIFF
--- a/hypha/apply/projects/urls.py
+++ b/hypha/apply/projects/urls.py
@@ -146,30 +146,4 @@ urlpatterns = [
             )
         ),
     ),
-    path(
-        "vendor/",
-        include(
-            (
-                [
-                    path("", ReportListView.as_view(), name="all"),
-                    path(
-                        "<int:pk>/",
-                        include(
-                            [
-                                path("", ReportDetailView.as_view(), name="detail"),
-                                path("skip/", ReportSkipView.as_view(), name="skip"),
-                                path("edit/", ReportUpdateView.as_view(), name="edit"),
-                                path(
-                                    "documents/<int:file_pk>/",
-                                    ReportPrivateMedia.as_view(),
-                                    name="document",
-                                ),
-                            ]
-                        ),
-                    ),
-                ],
-                "reports",
-            )
-        ),
-    ),
 ]


hypha/apply/users/urls.py
```diff
@@ -174,7 +174,7 @@
     path(
         "account/two_factor/setup/complete/",
         RedirectView.as_view(url=reverse_lazy("users:backup_tokens"), permanent=False),
-        name="two_factor:setup_complete",
+        name="setup_complete",
```
theskumar: When I was looking at this issue. The `two_factor:setup_complete` was referenced from the hypha code so was assumed this namespace might be getting used by the library itself, so didn't decide to update it. 

If this works and don't causes any issue to the our workflow. This looks good to me.

sandeepsajan0: I checked the workflow, everything seems to be working fine for me.

hypha/apply/projects/urls.py
```diff
@@ -168,7 +168,7 @@
                         ),
                     ),
                 ],
-                "reports",
+                "vendor",
```
sandeepsajan0: As it is not getting used anywhere, I think we can just remove this whole path.
I looked into the commit history as well and it is added [here](https://github.com/HyphaApp/hypha/commit/c2a3598ab98bbc741a6319ca26e41ac8e88611d7#diff-d63538e8198ab657b1878c6f6f4b998bf30824d6ad414a594131e740f7ee34ceR61) when forms were getting arranged to multiple files. It doesn't make any sense in that commit as well, It might be the preparation to add vendor urls but later those urls were [added separately](https://github.com/HyphaApp/hypha/commit/9a2d4840e3dfdb8c432cf012acc5f761e70d5095#diff-d63538e8198ab657b1878c6f6f4b998bf30824d6ad414a594131e740f7ee34ceL40).

Issue #3537, opened 2023-08-23
[Finance Role] Invoice Table on Project Page


fourthletter:
### Description

As finance I need to know the vendor name, date of when the invoice has been submitted, the period of performance, the amount on the invoice, and the invoice status (ie approved by program staff), and name of the Lead

- [ ] We will check with Finance which default sorting order do they prefer.
- [ ] Add sorting to date, Lead, and dollar amount columns, and perhaps status
- [ ] We will check with finance the purpose and need for Invoice date

Multiple select will have an update status batch action to update status to: Approve, Decline, Payment returned, Paid

![image](https://github.com/HyphaApp/hypha/assets/20019656/47d67fd0-1b5b-4f9a-a669-8fb79be9ddac)

frjo:
@di @sandeepsajan0 I guess it is here OTF want US date format?

If that is the case I propose the following:

1. Create a list of the places where OTF need US date format (maybe only here?)
2. Create a new date format setting with a logical name. (INVOICE_SHORT_DATE_FORMAT or PROJECT_SHORT_DATE_FORMAT or something better?)
3. If the new date format setting *is not* set Hypha uses the Django date settings as now.
4. If the new date format setting *is* set Hypha uses it for the specific pages/views (see 1).

PR #3756, opened 2024-02-07
Add status update batch action to invoices table

sandeepsajan0:
## Description

Fixes part of #3537

image

## Test Steps

 - [ ] Invoices table should be accessible to staff and finance
 - [ ] Staff and Finance both should have the option to select the invoices and update their statuses(depending on their statuses)
 - [ ] Check the statuses order like finance shouldn't be able to `paid` an invoice that is not `Approved by finance` yet.
 - [ ] Play around and provide your feedback.


wes-otf:
In testing when there were two invoices that had the status `Paid`, selecting both of these and attempting updating the status to `Payment failed` results in a 500 with:

```
sequence item 0: expected str instance, NoneType found
```

At [activity_feed.py](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/activity/adapters/activity_feed.py#L170). More info in sentry but it still seems to update the status. Seemed also specific to these statuses as others would update w/o issue. This was done as the finance role.

sandeepsajan0:
@frjo Its latest version is ready to put on test. 
@wes-otf There were issues with older invoices because we added invoice_number recently to the invoice model and as nobody was using the project/invoices so we didn't handle it for the older ones and let it None.
image

Although this was working fine for invoices with invoice_number so it wouldn't make any issue on prod still I have added a check to resolve this issue for older invoices as well.
It seems working fine now, you may try it again after the latest code deployment.

wes-otf:
For some reason now whenever I go to `/apply/projects/invoices/` on test I get a 500 page cause by the error: 
```
Missing staticfiles manifest entry for 'js/apply/jquery.fancybox.min.js'
```
Which is weird, was there an issue with `collectstatic` on the deploy or something?

frjo:
@wes-otf I missed doing a rebase for this PR, we have recently flatten the dir structure in another PR.

Should be fixed now.

PR #3761, opened 2024-02-09
Fix the 2FA flow

### LINKED ISSUE #3757: Can not set up 2FA, works in v5.3.1
`django.urls.exceptions.NoReverseMatch: 'two_factor' is not a registered namespace`

My guess is that we have a small hickup in the URLs after removing public.

We should add a test to make sure this does not get broken agin.

Replicate by going to your account page, click "Enable 2FA" and on the next page click "I've installed a 2FA App". Then you get the above error.

I notices that the 2FA pages need some margins as well. One example here:

Skärmavbild 2024-02-07 kl  13 33 22

theskumar:
## Description

- Update page styling
- Clean up internal code
- add tests

Fixes #3757

## Test Steps

 - [ ] goto profile 
 - [ ] Enable 2FA
 - [ ] logout and try login, use both code and backup code
 - [ ] disable profile


frjo:
Nice work! It looks good now and I had no issue with the functionality.

Did we not have a "sudo" like feature for things like disabling 2FA, change e-mail etc.?

theskumar:
To check the sudo, goto application local storage and remove the elevate key. Since you are testing things after you just logged in. You might be in the sudo mode already

frjo:
I now see the elevate function when trying to access backup codes etc. but not disabling 2FA on my account page.

If I try to disable 2FA for a user in Wagtail admin I do see it.

PR #3769, opened 2024-02-12
Update primary nagivation dropdown UX

theskumar:
## Description

- Fix the dropdown being hidden on the beta all page
- Use mouseover for quick access
- Update over states
- Update the urls to that any external urls can also be added via
  settings, removes dependeny of `reverse` while specifying urls
- How active page in the dropdown as well, a dark-blue border
- Update the focus outline to be a simple dashed-border, instead of bulky outline

## Test Steps

 - [ ] Checkout the primary navigation in the top-header

### DIFF
--- a/hypha/core/templates/core/navigation/primarynav-apply.html
+++ b/hypha/core/templates/core/navigation/primarynav-apply.html
@@ -4,30 +4,53 @@
         <ul class="nav nav--primary" role="menubar">
             {% apply_nav_items request.user as nav_items %}
             {% for item in nav_items %}
-                {% url item.url as target_url%}
-                <li class="nav__item" role="presentation">
+                <li class="nav__item"
+                    role="presentation"
+                    x-data="{open: false}"
+                    x-on:mouseover.away="open = false"
+                >
                     {% if item.sub_items %}
-                        <div class="relative inline-block" x-data="{open: false}">
-                            <a class="nav__link {% if target_url in request.path %}nav__link--active{% endif %}"
-                               href="{% url item.url %}"
+                        <div class="relative inline-block">
+                            <a class="nav__link {% if item.url in request.path %}nav__link--active{% endif %}"
+                               href="{{item.url}}"
                                aria-label="{% trans "Menu Item" %}"
                                aria-haspopup="menu"
                                aria-expanded="false"
                                role="menuitem"
+                               @mouseover="open = true"
                                @click.prevent="open = ! open">
                                 {{ item.title }}
-                                {% heroicon_outline "chevron-down" aria_hidden="true" size=16 stroke_width=3 class="inline align-baseline mr-1" %}
+                                <span :class="open ? '[&>svg]:translate-y-1': ''">
+                                    {% heroicon_outline "chevron-down" aria_hidden="true" size=16 stroke_width=3 class="inline align-baseline mr-1 transition-transform" %}
+                                </span>
                             </a>
-                            <div x-cloak x-trap="open" x-on:keydown.escape="open = false" x-show="open" x-transition @click.outside="open = false" class="absolute block bg-white min-w-max shadow-lg z-1 mt-7 ">
+                            <div
+                                x-cloak
+                                x-trap.noautofocus="open"
+                                x-on:keydown.escape="open = false"
+                                x-show="open"
+                                x-transition
+                                @click.outside="open = false"
+                                class="min-w-48 absolute block bg-white shadow-xl z-20 mt-4 border rounded-sm"
+                            >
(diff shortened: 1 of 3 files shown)


hypha/core/templates/core/navigation/primarynav-apply.html
```diff
@@ -4,30 +4,53 @@
...
+                                               {% endif %}"
+                                    >
                                         {{ sub_item.title }}
                                     </a>
                                 {% endfor %}
                             </div>
                         </div>
                     {% else %}
-                        <a class="nav__link {% if target_url in request.path %}nav__link--active{% endif %}" href="{{ target_url }}" role="menuitem">
+                        <a class="nav__link {% if target_url in request.path %}nav__link--active{% endif %}" href="{{ item.url }}" role="menuitem">
```
frjo: `target_url` should be replace with `item.url` here as well I think.

hypha/core/templates/core/navigation/primarynav-apply.html
```diff
@@ -4,30 +4,53 @@
...
                             </a>
-                            <div x-cloak x-trap="open" x-on:keydown.escape="open = false" x-show="open" x-transition @click.outside="open = false" class="absolute block bg-white min-w-max shadow-lg z-1 mt-7 ">
+                            <div
+                                x-cloak
+                                x-trap.noautofocus="open"
+                                x-on:keydown.escape="open = false"
+                                x-show="open"
+                                x-transition
+                                @click.outside="open = false"
+                                class="min-w-48 absolute block bg-white shadow-xl z-20 mt-4 border rounded-sm"
```
frjo: `min-w-48` seems to not generate any styles? Why not use `min-w-max`? Having menu item break on two lines is undesirable.

frjo:
Nice works! Open on hover, or not, is a personal preference. I will ask the other maintainer in the meeting tomorrow what they prefer.

Two minor items that would be nice to resolve:

1. The dotted outline looks a bit off I think. I see it uses `focus-visible` so it should only show for keyboard navigation. But I still see it using only the mouse/trackpad.
2. When you hover over the longes menu item the bold text makes the with of the menu increase a tiny bit. 

Skärmavbild 2024-02-12 kl  13 07 59

frjo (changes requested):

hypha/core/templates/core/navigation/primarynav-apply.html
theskumar: `min-w-48` ensure the width of two dropdown are similar. It won't break them in two lines as it's not setting any max width. That said, I should consider added generous max-width so the dropdown doesn't look weird for a link with too many words. 

Let me know if it all sounds reasonable?

hypha/core/templates/core/navigation/primarynav-apply.html
theskumar: 🖖🏼

theskumar:
> Nice works! Open on hover, or not, is a personal preference. I will ask the other maintainer in the meeting tomorrow what they prefer.
> 
> Two minor items that would be nice to resolve:
> 
> 1. The dotted outline looks a bit off I think. I see it uses `focus-visible` so it should only show for keyboard navigation. But I still see it using only the mouse/trackpad.
> 2. When you hover over the longes menu item the bold text makes the with of the menu increase a tiny bit.
> 
> Skärmavbild 2024-02-12 kl 13 07 59

Which browser are you using, this is what i see: 

Screenshot 2024-02-12 at 12  33 18@2x

theskumar:
The visible-focus seems to be not working with `x-trap` and mouseover, already spent sometime trying to figure out a solution, but no luck. I'll give it another try.

hypha/core/templates/core/navigation/primarynav-apply.html
theskumar: This is could be b/w you are not using the latest tailwind, which introduced `min-w-*`

theskumar:
@frjo any feedback on the nav getting opened while on hover? The outline issue is only happening if the mousehover is used. If mouseover is not desired, there is need to chase after the focus issue.

theskumar:
Update and fixed the outline issue on safari, it should look better now. 

Screenshot 2024-02-15 at 10  01 57@2x

frjo (approved):

frjo:
Latest version on test now.

wes-otf:
Hey! Added functionality looks good to me, I'm just a little confused about what purpose the dotted border serves? I didn't see it changing on test at all, just stayed on `All Submissions`/`All Projects`. On Firefox 121 I can't seem to use keyboard nav at all with Hypha if that's what this is in reference to?

PR #3777, opened 2024-02-19
Add head end and body end include files. Empty by default to make it easy to insert custom content

frjo:
Allow implementors to add `templates_custom/includes/head_end.html` and/or `templates_custom/includes/body_end.html` to easily insert things like tracking code etc. in to `base.html`.


frjo:
You can already easily add `templates_custom/base.html` to customise it. But then you need to track and sync any future changes to `base.html`. I think this PR allow for a more easily maintainable solution.

Issue #3269, opened 2023-03-07
The category filter on the submission table crash the system when used on large data set


fourthletter:
Description

The category field does not present accurate information. 

```
'SystemExit: 1

SELECT DISTINCT "funds_applicationsubmission"."form_fields", "funds_applicationsubmission"."form_data" FROM "funds_applicationsubmission" WHERE NOT (("funds_applicationsubmission"."next_id" IS NOT NULL OR "funds_applicationsubmission"."is_archive"))

SELECT "categories_option"."id", "categories_option"."sort_order", "categories_option"."value", "categories_option"."category_id" FROM "categories_option" INNER JOIN "categories_category" ON ("categories_option"."category_id" = "categories_category"."id") WHERE "categories_category"."filter_on_dashboard" ORDER BY "categories_option"."sort_order" ASC'
```

frjo:
@theskumar If you first filter down to a small data set it works. But trying it on 5000+ will not work. The limit may be lower that that even.

The filter seems to resource intensive. Any ideas what to do about that?

Assiged you the issue I think is related in Sentry.

theskumar:
Given the complexity and abstraction involved in the current implementation, I can't say for sure there is an easy to handle this. The table has a few  N+1 queries, which can be solved by: 
1. caching/ de-normalizing 
2. making more minor calls for the data only when needed. For this, the things that are hidden in subtables. 
3. Simplify the table itself.
4. Optimize the queries. 

I do have a cleaner implementation in #3149. If this bug is not urgent, I'll try to cover up as much as possible in that PR, or if it's too urgent, I can start fixing things in the current implementation.

theskumar:
@frjo Also, it would be nice to enable the Sentry Performance/Tracing enabled so we can pinpoint and see meaningful performance-related improvements or degrades.

PR #3778, opened 2024-02-19
Fixed category labels not being populated

### LINKED ISSUE #3741: Beta submissions view categories label is rendered as an ID rather than a title
All bug bounty props to @Techslammer. When a category is filtered on in the new beta `All Submissions` view, the filter label appears at the category ID rather than the category name.

## Screenshots

*selecting the category*
Screenshot 2024-01-30 at 15 22 56

*the filter label*
Screenshot 2024-01-30 at 15 21 11

wes-otf:
## Description

Fixes #3741. Adds a new tag to get the label of the category from the ID

## Test Steps

 - [ ] Navigate to the beta view
 - [ ] Select a new category to filter on
 - [ ] When a chip pops up to indicate an active category filter, ensure it is the same as the label that was previously selected, not a number.


hypha/apply/funds/templates/submissions/all.html
```diff
@@ -1,6 +1,6 @@
 {% extends base_template %}
 
-{% load i18n static %}
+{% load i18n static  filter_tags %}
```
frjo: Double space.

frjo:
I think this PR look good but the feature over all is not useful it its current state. See https://github.com/HyphaApp/hypha/issues/3269

I do not think category filtering active on OTF live?

frjo (approved):

wes-otf:
@frjo Ah that issue is new to me. It works on live with categories that have lower numbers of submissions but I am seeing that `SystemExit` with other categories.

theskumar:
Looking at the Traceback, the `SystemExit` seems to be a performance issue trying to filter by category, we can have another  ticket to resolve that, the filter by category which uses django-filter should be optimized so it doesn't loop-over all the submissions, trying to find and filter by category individually. I can take a stab at it. 

Screenshot 2024-02-21 at 6  15 00@2x

hypha/apply/funds/templatetags/filter_tags.py
```diff
@@ -0,0 +1,19 @@
...
+    """Get the category string label of a selected category
+
+    Args:
+        label_id:
+            The ID of the label to get the string representation of
+
+    Returns:
+        A representation of the label
+    """
+    return str(Option.objects.get(id=label_id))
```
theskumar: This looks good, but would generally suggest dealing with it in view.py code itself, that way it's easier to fetch all the category names once, instead of here it will 5 db calls if there are 5 categories selected, just to get the names of all the 5 categories. 

I generally see template tags causing hidden db queries.

frjo:
@theskumar Would be really nice if you can improve the performance of the category query. I can see how this filter can be useful for staff.

hypha/apply/funds/templatetags/filter_tags.py
wes-otf: Makes sense! I was actually trying to figure out if I could cache the result of the initial query somehow but popping the login into `view.py` might make things easier.

wes-otf:
that last push should be a lot more efficient ^

theskumar (approved):
💯

PR #3780, opened 2024-02-21
Update overview sections to use them as filter

### LINKED ISSUE #3771: Overview section for count and quick access.
The overview section should work as a filter instead of referring to a new page.
image

sandeepsajan0:
## Description

Fixes #3771 

Submissions:
image

Projects:
image

Invoices:
image

Loading state:
image

## Test Steps

 - [ ] Check the All Submissions, All Projects, and Invoices pages.
 - [ ] Check if overview sections(in ss above) are clear and can filter the data by selecting any one of them.
 - [ ] Overview sections should be easily accessible and able to filter the data correctly.

### DIFF
--- a/hypha/apply/funds/templates/funds/includes/status-block.html
+++ b/hypha/apply/funds/templates/funds/includes/status-block.html
@@ -4,7 +4,7 @@
         <ul class="status-block">
             {% for status, data in status_counts.items %}
                 <a class="status-block__item {% if data.is_active %}status-block__item--active{% endif %}" href="{{ data.url }}">
-                    <h5 class="status-block__title text-black">{{ data.name }}</h5>
+                    <p class="status-block__title text-black"><strong>{{ data.name }}</strong></p>
                     <p class="status-block__info text-gray-400">{{ data.count }} {{ type }}</p>
                 </a>
             {% endfor %}
--- a/hypha/apply/funds/templates/funds/includes/status-block.html
+++ b/hypha/apply/funds/templates/funds/includes/status-block.html
@@ -3,11 +3,10 @@
     <section class="section">
         <ul class="status-block">
             {% for status, data in status_counts.items %}
-                <li class="status-block__item">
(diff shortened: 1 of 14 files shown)


sandeepsajan0:

hypha/apply/funds/templates/funds/includes/status-block.html
```diff
@@ -3,11 +3,10 @@
     <section class="section">
         <ul class="status-block">
             {% for status, data in status_counts.items %}
-                <li class="status-block__item">
-                    <h5 class="status-block__title">{{ data.name }}</h5>
-                    <p class="status-block__info">{{ data.count }} {{ type }}</p>
-                    <a class="status-block__link" href="{{ data.url }}">{% trans "View" %}</a>
-                </li>
+                <a class="status-block__item {% if data.is_active %}status-block__item--active{% endif %}" href="{{ data.url }}">
+                    <h5 class="status-block__title text-black">{{ data.name }}</h5>
```
frjo: Overusing heading tags is a left over from the old Hypha design. I think we should take the opportunity here to update it. I think just replacing it with "strong" tag would be god, making it a bit smaller. The p-tag below could get the standard body font size. I think that would look nicer and take up less space.

@abhiXII What do you think?

frjo:
Added a note regarding the size of the text in the overview filter. Other than that it works perfectly in my testing, nice work!

frjo:
I suggest we delete this css line as well.

https://github.com/HyphaApp/hypha/blob/37b2f494a1ec871f425719ff8f13978ee6aea57a/hypha/static_src/sass/components/_status-block.scss#L49

frjo:
And we can use the strong tag **or** set weight here for the status-block__title:

https://github.com/HyphaApp/hypha/blob/37b2f494a1ec871f425719ff8f13978ee6aea57a/hypha/static_src/sass/components/_status-block.scss#L57-L61

PR #3781, opened 2024-02-21
Flatten static src dirs

frjo:
## Description

Removes the "src" and "apply" sub directory levels resulting in a more flat directory structure for static assets.

No changes in functionality should be a result of this. It is only to make it easier for developers to move around the code base. 

## Test Steps

 - [ ] Check that js and css files build in order.
 - [ ] Check that js and css files still loads in order on all pages.


theskumar (approved):
Looks good!

Issue #3788, opened 2024-02-28
When editing an existing submission the save/submit button say "Preview and submit"


frjo:
Staff can not save when editing an existing submission. The button say "Preview and submit" but you are never given the option to submit.

Applicants can save but the submission does not seem to progress when it should.

frjo:
@wes-otf Can you confirm this issue? If so it is a quite bad bug.

frjo:
I will discuss this with @sandeepsajan0 tomorrow and hand of any info we dig up here so @wes-otf can take over.

wes-otf:
Can confirm, exactly what you said in the issue. Am currently updating OTF's applicant guidebook today and found a whole host of issues with the preview that was missed on my own dev testing and on user testing.

Going to devote solid time to looking at this tomorrow but I think relying on the draft stage to preview the application is what's biting me here. I think there might need to be another stage potentially added that is dedicated to previews.

PR #3789, opened 2024-02-29
Disable submission preview until a fix for the bugs are in place.

frjo:
Related #3788

Sets `SUBMISSION_PREVIEW_REQUIRED` to False by default and hides the submission preview buttons.

We need a permanent fix for the preview bugs but until then this hides the functionality.


sandeepsajan0 (approved):
Looks good. Preview is no more there and submit is working fine.

frjo:
My testing show the same, everything works with this in place.

Issue #3793, opened 2024-03-01
Batch determination `KeyError` for non-Staff user


bickelj:
**Describe the bug**

Given a user not having the "Staff" role, e.g. "Admin" or "Staff Admin", when submitting a Batch Determination, then a `KeyError` (shown via an HTTP 500 error) occurs.

**To Reproduce**

1. Log in as a non-Staff user (or remove the Staff group from your user in Wagtail Admin)
2. Visit `apply/submissions/all/`
3. Select a couple of submissions having the same status
4. Click "Status" button (at the top of the table)
5. Select "Dismiss" from the dropdown
6. Click "Update"
7. Click "Send"

See the 500 error.

**Expected behavior**

Either a permission error or a successful update to the submissions' status.

**Priority**

- Low priority (annoying, would be nice to not see)

**Affected roles**

- Non-Staff who do Batch Determination
- Developers
- Testers

**Desktop (please complete the following information):**

 - OS: GNU/Linux, PureOS 10
 - Browser: Firefox 115
 - Version: v5.3.1

**Additional context**

Found and/or refined by @tris-ots, @frankduncan, and me.

wes-otf:
Hey, thanks for submitting this! What user role were you exactly? I tried this as `Applicant` & `Partner` and couldn't get the submissions page to render, just hung at a white screen (which should also prob result in a 404). It also isn't in the nav bar at that point.

bickelj:
@wes-otf Admin

bickelj:
@wes-otf See the suggested fix in #3796. If there's a better way we are all for it. Have a great weekend!

wes-otf:
Closed via #3796.

PR #3796, opened 2024-03-01
Let non-Staff (e.g. Admin) do batch determination

bickelj:
When an Admin who otherwise could make determinations on submissions does a batch determination then it should succeed. This commit allows it to succeed or throws a permissions error otherwise.

Issue #3793 Batch determination `KeyError` for non-Staff user

### DIFF
--- a/hypha/apply/determinations/forms.py
+++ b/hypha/apply/determinations/forms.py
@@ -132,7 +132,8 @@ class BaseBatchDeterminationForm(BaseDeterminationForm, forms.Form):
         widget=forms.ModelMultipleChoiceField.hidden_widget,
     )
     author = forms.ModelChoiceField(
-        queryset=User.objects.staff(),
+        # Permissions should be handled by the view rather than the QuerySet here.
+        queryset=User.objects.active(),
         widget=forms.ModelChoiceField.hidden_widget,
         required=True,
     )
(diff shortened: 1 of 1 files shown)


wes-otf (approved):

wes-otf:
All looks good to me! Any additional thoughts @frjo?

frjo:
Any performance issues with using ` queryset=User.objects.all()`? That can potentially be a lot of posts but I'm unsure when/how this gets loaded.

Feels a bit lax as well, but that is maybe a none issue.

@theskumar Any thoughts on this?

bickelj:
@frjo In the cases I looked at, the queries should be by primary key. E.g. L601/L603 in the diff and also in my SQL log file I think this is one of the queries on that set.
> SELECT "users_user"."id", "users_user"."password", "users_user"."last_login", "users_user"."is_superuser", "users_user"."is_staff", "users_user"."is_active", "users_user"."date_joined", "users_user"."email", "users_user"."full_name", "users_user"."email_preference", "users_user"."slack", "users_user"."drupal_id" FROM "users_user" WHERE "users_user"."id" = 338620 LIMIT 21

I could be incorrect about this, though.

theskumar:
> Any performance issues with using ` queryset=User.objects.all()`? That can potentially be a lot of posts but I'm unsure when/how this gets loaded.
> 
> Feels a bit lax as well, but that is maybe a none issue.
> 
> @theskumar Any thoughts on this?

I believe it will not cause any performance issue, but we should filter/allow only the active users instead of all the users.

frjo:
@bickelj Can we use `queryset=User.objects.active()` instead?

bickelj:
> @bickelj Can we use `queryset=User.objects.active()` instead?

@frjo That makes sense. I'll try it and see.

bickelj:
@theskumar @frjo Active-only change made.

PR #3805, opened 2024-03-13
Display "View Live" button only if round is live in admin

### LINKED ISSUE #3794: `View Live` on a round in Wagtail always results in a 404
When given the option to view a round live, it always results in a 404 due to the live round existing at the fund's slug. For cases like these, the `View Live` button should bring the user to the fund's URL. Are there cases where the content would exist at the round's slug?

https://github.com/HyphaApp/hypha/assets/145372368/76705215-513a-4670-8062-0d52766ab0e4

theskumar:
## Description

Currently the "view live" button on the round detail admin page, links
to the round pages with the url containing fund slug followed by slug of
the round page.

The round can be accessed only via the fund url and the fund url serves
the active round, so the current link leads to 404.

This fixes the admin "view live" button so that it's displayed only if
the round is live and also links it correctly to the fund slug

Fixes #3794

## Test Steps

 - [ ] Goto Admin > apply > rounds
 - [ ] Open one of the active rounds, based on start and end date
 - [ ] Ensure it displayed "Live" button on the top right corner.
 - [ ] If the page is not live, the button should be hidden

### DIFF
--- a/hypha/apply/funds/models/applications.py
+++ b/hypha/apply/funds/models/applications.py
@@ -5,6 +5,7 @@ from django import forms
 from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
 from django.core.exceptions import ValidationError
+from django.core.handlers.wsgi import WSGIRequest
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
 from django.db.models import (
@@ -233,7 +234,7 @@ class RoundBase(WorkflowStreamForm, SubmittableStreamForm):  # type: ignore
     )
     sealed = models.BooleanField(default=False)
 
-    def get_url(self) -> Optional[str]:
+    def get_url(self, request: Optional[WSGIRequest] = None) -> Optional[str]:
         """Generates the live url, primarily used in the wagtail admin for the "view live" button.
 
         Returns:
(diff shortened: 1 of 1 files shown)


hypha/apply/funds/models/applications.py
```diff
@@ -232,6 +233,18 @@ class RoundBase(WorkflowStreamForm, SubmittableStreamForm):  # type: ignore
     )
     sealed = models.BooleanField(default=False)
 
+    def get_url(self) -> Optional[str]:
+        """Generates the live url, primarily used in the wagtail admin for the "view live" button.
+
```
wes-otf: The only issue I had was when I went from my dashboard (`/dashboard/`) to the Wagtail admin dashboard (`/admin/`) I got the error `RoundBase.get_url() got an unexpected keyword argument 'request'`. This seems to fix it just by adding the optional request kwarg. All looks good otherwise!

```suggestion
    def get_url(self, request: Optional[WSGIRequest]=None) -> Optional[str]:
        """Generates the live url, primarily used in the wagtail admin for the "view live" button.

        Args:
            request: The request that will be passed by Wagtail when loading the admin interface, currently unused.
```

theskumar: Nice catch, I thought I could get away without passing all the original parameters :D

Adding them.

theskumar: Done, https://github.com/HyphaApp/hypha/pull/3805/commits/b416c2d3a61356793b32df82509d041b7725d36b

PR #3807, opened 2024-03-14
Allow unassign staff reviewers only for inactive reviewers

### LINKED ISSUE #3553: Reviewer update form error when existing staff reviewer account is deactivated
This is an edge case. A staff (PM) is actively reviewing submission but quits and the account is marked as inactive.

Opening a reviewer update form on a submission that that person was assigned to review now shows "---" instead of the name of the account, since it is inactive.

A review might already be submitted so the organisation do not want to reassign it to another staff (PM).

Trying to make any changes to the form, e.g. adding external reviewers, will result in an error that a staff reviewer can not be unset, only reassigned since an review exist.

sandeepsajan0:
## Description

Fixes #3553 

## Test Steps

 - [ ] Assign a staff reviewer to a submission.
 - [ ] Submit the review
 - [ ] Set assigned reviewer as inactive via admin.
 - [ ] Try to update reviewers or unset that inactive reviewer.

### DIFF
--- a/hypha/apply/funds/forms.py
+++ b/hypha/apply/funds/forms.py
@@ -310,7 +310,7 @@ class UpdateReviewersForm(ApplicationSubmissionModelForm):
             ).last()
             if (
                 assigned_reviewer
-                and not cleaned_data[field]
+                and (not cleaned_data[field] and assigned_reviewer.reviewer.is_active)
                 and assigned_reviewer.reviewer in self.submitted_reviewers
             ):
                 self.add_error(


wes-otf:
In testing this worked well, only issue I saw is that when trying to change from a reviewer that is **not** inactive and pressing `Update`, the prompt closes without any sign of validation issue. Only upon re-opening the `Update Reviewers` prompt do you see the validation error, but even then it keeps the previously selection option which makes it a bit confusing. This error should not allow the closing of the prompt window.

This screenshot was after I had changed the current reviewer to `---`, pressed `Update` and re-opened the prompt:
Screenshot 2024-03-21 at 13 19 06

Could be out of scope for this PR though. I can make a new issue if so.

sandeepsajan0:
@wes-otf Yes, you are right. It is the behavior of all the modals so I think we can fix this as a separate issue. I'll create an issue for this and will look into it.

**Update:** I have added a new issue here https://github.com/HyphaApp/hypha/issues/3834

wes-otf:
Great! Should be ready to go then

PR #3815, opened 2024-03-19
Reduce the lazy load time for applicant submission on dashboard

theskumar:
## Description

Right now, the my submission for the applicants on their dashboard is
loaded after 1s of page load which is too much delay.

This PR reduces the delay so it's the loading via htmx is triggered as
soon as the page is finished loading. Also, the PR reduces the loading
preview/skeleton to 3 so there is must of jumpiness before/after the
submissions are loading, this is based on the fact that on an average
there we only few submission by a single applicant.

## Test Steps

 - [ ] Login as applicant
 - [ ] Notice the difference in the loading of "My submission" section, it should be much after now

### DIFF
--- a/hypha/apply/dashboard/templates/dashboard/applicant_dashboard.html
+++ b/hypha/apply/dashboard/templates/dashboard/applicant_dashboard.html
@@ -44,7 +44,7 @@
                     <h2 class="font-light flex-1">{% trans "My submissions" %}</h2>
                 </div>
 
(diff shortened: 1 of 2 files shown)


frjo (approved):

wes-otf (approved):

wes-otf:
This is great, all for a speedier load! Just out of curiosity what's the historical context here? Why was it necessary to have the delay previously?

theskumar:
The load delay at other places is 50ms, just enough to let UI settle down a bit. Since the "My submission" section is at the top, removing the delay altogether is just fine here.

PR #3821, opened 2024-03-19
Removed registration form, cleaned up login flow

### LINKED ISSUE #3813: Login and signup form improvements
front_page
login_or_signup
create_your_account
login

wes-otf:
## Description

Fixes #3813. Main changes:

- [x] Got rid of the registration form in favor of using the `/auth/` view. Having two places where the user can potentially register is confusing. 
- [x] Cleaned up different aspects of the login flow that were mentioned in #3813, like inconsistency in terminology (`Login` vs `Log in`), removal of buttons where they don't need to be, etc.
- [x] Added a button on the password-ed login to bring the user back to the passwordless log in/register view.
- [x] Have buttons in column rather than row on *both* mobile & desktop view and made them the same width (`18 rem`). 
    - Should we center the login page for desktop? This was done as it felt awkward to have everything on the left side while the buttons crept to the center
- [x] Added icons to `Log in with <ORG> email` & `Log in or register via email` for consistency
- [x] Moved login buttons to their own templates as some were being reused
- [x] Added missing translation blocks 

## Screenshots

**Desktop Views**

*Homepage*
Screenshot 2024-03-19 at 17 40 58

----

*Passwordless Login*
Screenshot 2024-03-19 at 17 09 34

----

*Password-ed Login*
Screenshot 2024-03-19 at 17 09 53

----

**Mobile Views**
*Homepage*

----

*Passwordless Login*

----

*Password-ed Login*

## Test Steps

 - [ ] Ensure you're logged out
 - [ ] On the home page, confirm a `Log in` button appears in the top right corner
 - [ ] Select the `Log in` button, and confirm you get taken to the passwordless login page (`/auth/`)
 - [ ] Select the `Log in with password` button and confirm you get taken to the password-ed page (`/login/`)
 - [ ] Confirm there is a button at the bottom of the page reading `Log in without password`
 - [ ] Select this button and confirm you get taken to the passwordless login page (`/auth/`)


frjo:
Nice work, and fast!

I found three issues I think we can fix in this PR.

On the front page we show a login button and you can signup with that form but I think the button should express that. Maybe "Login or Signup" when `ENABLE_PUBLIC_SIGNUP` is set?

Skärmavbild 2024-03-20 kl  09 47 52

The button back to the e-mail only login should take `ENABLE_PUBLIC_SIGNUP` in to account.

Skärmavbild 2024-03-20 kl  09 47 17

The password reset form needs work as well.

Skärmavbild 2024-03-20 kl  10 00 49

wes-otf:
@frjo great catches, thanks for taking a look! All should be resolved now, along with some CSS cleanup & addition translation tags added.

It's weird stylelint flagged my variable use with that `media-query-no-invalid`. Some quick research [suggests that rule should only apply to plain CSS](https://github.com/stylelint-scss/stylelint-scss/issues/954#issuecomment-1925926260)

frjo:
I rebased this PR and made some minor changes to text of a button.

hypha/templates/includes/passwordless_login_button.html
```diff
@@ -4,5 +4,5 @@
     href="{% url 'users:passwordless_login_signup' %}{% if next %}?next={{next}}{% endif %}"
 >
     {% heroicon_mini "envelope" size=18 class="inline align-text-bottom me-1" aria_hidden=true %}
-    {% trans "Log in " %}{% if ENABLE_PUBLIC_SIGNUP %} {% trans "or" %} {% trans "Sign in" %}{% endif %} {% trans "via" %} {% trans "e-mail" %}
+    {% trans "Log in without password" %}
```
wes-otf, line 7: I like this a lot better, should we indicate that this is where you would also go to signup though as that isn't given anywhere else?

wes-otf: (if enabled)

frjo: My thinking was that if you clicked your way to the password login form you are most likely not needing to create an account.

wes-otf: fair enough, all good by me!

PR #3825, opened 2024-03-20
Minor fixes to submission all table and status bar on small screens

frjo:
Before:
Skärmavbild 2024-03-20 kl  10 19 52

After:
Skärmavbild 2024-03-20 kl  14 39 17


theskumar (approved):
💯

theskumar:
@frjo looks like the FE lint is breaking, rest looks fine to me.

PR #3832, opened 2024-03-25
Add `APPLICANT_PARTNER` visibility to staff comment visibility

### LINKED ISSUE #3831: Staff cannot view comments left with `APPLICANT_PARTNER` visibility
When an applicant comments on an application with `APPLICANT_PARTNER` visibility, this comment will not be visible to staff.

wes-otf:
Fixes #3831. Somehow this slipped through both me & user testing but all works well now. Also swapped so staff comment logic gets evaluated first, thus if staff submits an application (usually for testing) then goes to make a comment to another application they won't be stuck with applicant options for visibility.

## Test Steps

 - [ ] As `User Testing Applicant` create a new application and submit it.
 - [ ] As staff, add `User Testing Partner` as a partner on the created application.
 - [ ] As `User Testing Applicant` leave a comment on the application with `Partner` visibility
 - [ ] As staff, ensure the comment that was left is visible.

### DIFF
--- a/hypha/apply/activity/models.py
+++ b/hypha/apply/activity/models.py
@@ -224,7 +224,7 @@ class Activity(models.Model):
             A list of visibility strings
         """
         if user.is_apply_staff:
-            return [TEAM, APPLICANT, REVIEWER, PARTNER, ALL]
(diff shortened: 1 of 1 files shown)


frjo (approved):

wes-otf:
@frjo can this be merged? ideally would like to deploy this Monday morning to get this patched ASAP

PR #3833, opened 2024-03-26
Rename flag app to bookmark.

### LINKED ISSUE #3028: Rename Flag app to Bookmark
### Description 

The Staff flagged feature/dashboard was developed as a "to-do list" for users. Staff could bookmark/save applications to their own dashboard and on a dashboard for other staff members. 

### Challenges

- [ ] Design has not been user tested and feedback has never been collected on the functionalities. 
- [ ] Determine whether this feature needs to be retired because there has not been a clear use case to refine the design, e.g., usability testing

Screenshot 2022-11-02 3 12 42 AM

frjo:
Fixes #3028

## Test Steps

 - [ ] Test so you can set user and staff bookmark on a submission.
 - [ ] Check so the bookmarked submission show up in the bookmark tables on the dashboard
 - [ ] Check so the staff bookmark view is accessible in the main menu.
 - [ ] Check so staff bookmarked submission is listed in the view.


hypha/apply/bookmarks/migrations/0002_rename_flags_bookmarks.py
```diff
@@ -0,0 +1,15 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("bookmarks", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.RenameModel("Flag", "Bookmark"),
+        migrations.RunSQL("INSERT INTO bookmarks_bookmark (SELECT * FROM flags_flag);"),
+        migrations.RunSQL("DROP TABLE flags_flag;"),
```
frjo, line 14: @theskumar A better way to do this than raw sql?

theskumar: This looks simple enough that will run on all the database types. 

Or you can use the RunPython, using the sub-queries: 

https://docs.djangoproject.com/en/5.0/ref/models/expressions/#subquery-expressions

theskumar: Also, since the model is renamed, curious if it's needed to drop the original flags table.

theskumar:
Curious, it `starred` would be a better term, it will allow for nice emoji uses at different places. Bookmarks doesn't really have a symbolic representations. Just a thought!

hypha/apply/bookmarks/migrations/0002_rename_flags_bookmarks.py
frjo: The test are failing since the flag tables will not exist on a new db.

frjo:
Closing this, see https://github.com/HyphaApp/hypha/pull/3843. We handle renaming is separate PR if needed.

PR #3841, opened 2024-03-28
Make `is_staff` true when a user is Staff Admin

### LINKED ISSUE #3782: Selecting `View Message Log` on submission redirects to the user's dashboard
When looking at application details and selecting `View Message Log` on the right, a client-side error of `Uncaught ReferenceError: NProgress is not defined` occurs, redirecting the user to their dashboard.

Screenshot 2024-02-23 at 15 52 15

wes-otf:
Fixes #3782. Rather than having `is_staff` set to true when a user's email is the org's domain, this sets it true when a user is Staff Admin or a Superuser. Also includes a migration to handle existing cases where Staff Admins are not `is_staff`.

## Test Steps

 - [ ] As a Staff Admin, navigate to a submission
 - [ ] Ensure there is a `View message log` button on the right hand side of the submission tabs
 - [ ] Ensure clicking `View message log` brings up the Django Admin message log
 - [ ] Create a new account with the role of Staff Admin, become this user and repeat steps 2 & 3
 - [ ] Edit an existing account to add the role of Staff Admin, become this user and repeat steps 2 & 3


wes-otf:
I was originally trying to add an `is_staff`property to `User` that would be an alias to `is_apply_staff_admin` and was intended to overwrite the `AbstractUser` attribute which worked fine for existing users, but I ran into issues when a new user was created as `is_staff` was null in the database which violated rules. 

This seemed like a more elegant solution but @theskumar let me know if this conflicts with your plans for permissions.

frjo (approved):

frjo:
Rebase and then we can put on test.

theskumar:
> I was originally trying to add an `is_staff`property to `User` that would be an alias to `is_apply_staff_admin` and was intended to overwrite the `AbstractUser` attribute which worked fine for existing users, but I ran into issues when a new user was created as `is_staff` was null in the database which violated rules.
> 
> This seemed like a more elegant solution but @theskumar let me know if this conflicts with your plans for permissions.

Let's fix the functionality first and we can think about re-organizing it under a proper permission framework.

wes-otf:
sounds good! should be ready then.

PR #3843, opened 2024-03-29
Fix user and staff flagg views. Add staff flag view to main menu.

frjo:
Fixes these issues:

* The flagg views are not updated with the new adminbar component.
* Staff flagged view are not included in the main menu.
* Unflagging submissions did not work in the user and staff flagged views, only on dashboard.
* Staff flagged filter was missing from new all-beta view.

## Test Steps

 - [ ] Check that "Staff flagged" is in the main menu.
 - [ ] Check that it works to unflagg submissions in all flag views.


hypha/apply/funds/views_beta.py
```diff
@@ -124,6 +124,8 @@ def submission_all_beta(
     if "flagged" in search_filters:
         if "@me" in search_filters["flagged"]:
             qs = qs.flagged_by(request.user)
+        if "@staff" in search_filters["flagged"]:
```
theskumar, line 127: @frjo please also update the documentation at https://docs.hypha.app/user-guides/applications/submission-search/

theskumar (approved):
Just a minor note around updating the user guide, rest of the functionality looks good to me! Tested it locally

hypha/apply/funds/views_beta.py
frjo: Good point, fixed that now.

PR #3844, opened 2024-03-30
Make all submission tables respects the SUBMISSIONS_TABLE_EXCLUDED_FIELDS setting

### LINKED ISSUE #3758: All tables that make use of BaseAdminSubmissionsTable does not take SUBMISSIONS_TABLE_EXCLUDED_FIELDS in to account
An example is the "Submissions waiting for your review " table on staffs dashboard. It displaces fields that are set to be excluded in the `SUBMISSIONS_TABLE_EXCLUDED_FIELDS` setting.

frjo:
Fixes #3758

## Test Steps

 - [ ] Check that tables on the dashboard do not contain columns that have been excluded.

### DIFF
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -4,6 +4,7 @@ import re
 import django_filters as filters
 import django_tables2 as tables
 from django import forms
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db.models import F, Q
 from django.urls import reverse
@@ -97,7 +98,7 @@ class SubmissionsTable(tables.Table):
     class Meta:
         model = ApplicationSubmission
         order_by = ("-last_update",)
-        fields = (
+        fields = [
             "title",
             "phase",
             "stage",
@@ -105,8 +106,10 @@ class SubmissionsTable(tables.Table):
             "round",
             "submit_time",
             "last_update",
-        )
-        sequence = fields + ("comments",)
+        ]
+        sequence = fields + [
+            "comments",
+        ]
         template_name = "funds/tables/table.html"
         row_attrs = {
             "class": make_row_class,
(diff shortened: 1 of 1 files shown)


wes-otf:
@frjo is this deployed? On the test instance I still see `Organization name` under `Your previous reviews` despite `SUBMISSIONS_TABLE_EXCLUDED_FIELDS=meta_terms,organization_name` in heroku but I could be misunderstanding what the target is here:
Screenshot 2024-04-17 at 10 10 10

Screenshot 2024-04-17 at 10 10 27

frjo:
Seems to be a bug in my implementation. Does not have any effect on the views "…waiting for your review/…previous reviews".

But it does work on e.g. "Staff flagged submissions"

@theskumar @sandeepsajan0 Any ideas here?

hypha/apply/funds/tables.py
```diff
@@ -188,8 +191,16 @@ class Meta(SubmissionsTable.Meta):
...
             "reviews_stats",
             "organization_name",
-        )
-        sequence = fields + ("comments",)
+        ]
+
+        # Remove fields that have been added to the exclude fields settings.
+        fields = [
+            x for x in fields if x not in settings.SUBMISSIONS_TABLE_EXCLUDED_FIELDS
+        ]
```
sandeepsajan0: It doesn't seem to be working that way. I looked and tried to update its base columns and sequence in `__init__` then it works.
here is the code:
```    
    def __init__(self, *args, **kwargs):

        excluded_columns = getattr(settings, "SUBMISSIONS_TABLE_EXCLUDED_FIELDS", [])
        for column_name in excluded_columns:
            if column_name in self.base_columns:
                del self.base_columns[column_name]
        self.sequence = [column_name for column_name in self._meta.fields if column_name not in excluded_columns]```

sandeepsajan0:
Rebased and added the commented code. @frjo You may test it again, it should work as expected now.

hypha/apply/funds/tables.py
sandeepsajan0: I've added the code.

frjo:
Closing this in favour of future direction where tables on dashboards are replaced with lists that show the info people has most use of.

PR #3846, opened 2024-04-03
Add draft determinations to todo list

### LINKED ISSUE #3828: Add determination draft to todo list

It is a step towards using todo list for submissions. We can start with drafts like determination and review drafts.

If a staff has a draft determination for an open application then it should be listed in todo list. And it should get removed if the application is closed/accepted or on determination submission.

**Priority**

**Affected roles**

- Staff

sandeepsajan0:
Fixes #3828 

image

## Test Steps

We should test it for multiple cases and try to break it.

 - [ ] Add a draft determination to any submission.
 - [ ] Check out the dashboard, it should listed in my task section.
 - [ ] Submit it and it should be gone from the dashboard.
 - [ ] Another case, somebody else saves the current draft as a draft then it should be on both user's dashboards.
 - [ ] On submission by any of them, it should be gone from both of the dashboards.
 - [ ] Any other cases left, lets try that out.


sandeepsajan0:
@frjo Can we put this on the test? Review draft PR has a use case that will depend on the determination so I am thinking of avoiding conflicts.

wes-otf:
This just about passed user testing! the only issue they noticed was when the application name was longer (so full message took up more than one line) it looks like the border around the task status doesn't adjust well:
Screenshot 2024-04-11 9 54 58 AM

sandeepsajan0:
I see, I'll fix it. But I think we might also need to truncate the title here as it can be more problematic with multiline titles. @frjo @wes-otf What do you think, should we truncate only the title or limit the whole message max up to 2 lines?

sandeepsajan0:
Fixed `draft` label UI for multiline messages.

Single line message:
image

Multiline message:
image

sandeepsajan0:
WIth Trucated title and tooltip:
image

wes-otf:
the latest w/ the tooltip wasn't deployed to test yet but the fixes to the draft label worked! I think this is probably good to go.

wes-otf:
from that screenshot the tooltip definitely looks a lot better than the multi-line. nice work @sandeepsajan0!

PR #3864, opened 2024-04-09
Make the open rounds beta filter the default

### LINKED ISSUE #3857: `Open` rounds should be the default for the `Rounds` filter in beta
In the beta submissions view, `Open` rounds should be the default open tab rather than `Closed`

**Edit**: Unpublished rounds that were not closed also do not appear in either list, I would expect these to probably appear in closed

wes-otf:
Fixes #3857. Swaps the `Open` & `Closed` tabs in the beta round filter. This also ensures that the closed tab will be selected if there are no open rounds.

## Test Steps

 - [ ] Navigate to the all submissions beta view as Staff
 - [ ] Ensure `Open` rounds is the default tab in the filter prompt

### DIFF
--- a/hypha/apply/funds/templates/submissions/submenu/rounds.html
+++ b/hypha/apply/funds/templates/submissions/submenu/rounds.html
@@ -17,22 +17,26 @@
 
         <div
             role="tablist"
-            x-data="{tab: 'closed_rounds'}"
+            {% if open_rounds %}
+                x-data="{tab: 'open_rounds'}"
+            {% else %}
+                x-data="{tab: 'closed_rounds'}"
+            {% endif %}
         >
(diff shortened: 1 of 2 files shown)


hypha/apply/funds/templates/submissions/submenu/rounds.html
```diff
@@ -17,23 +17,27 @@
 
         <div
             role="tablist"
-            x-data="{tab: 'closed_rounds'}"
+            {% if open_rounds %}
+                x-data="{tab: 'open_rounds'}"
+            {% else %}
+                x-data="{tab: 'closed_rounds'}"
+            {% endif %}
```
theskumar, line 24: 💯

theskumar:
Left a note at #3857, if that's not an issue, then this PR looks good me, implementation wise.

PR #3868, opened 2024-04-10
Fix error message alignment for change password form

### LINKED ISSUE #3842: Password error formatting
Small but wanted to document it somewhere, the error prompt should be right next to the password field but instead there's quite a bit of space between the two. Should the error appear on the first prompt instead of the confirmation too? Not sure if it matters.

Screenshot 2024-03-29 at 14 01 19

theskumar:
Inline error messages

Screenshot 2024-04-10 at 5  58 42@2x

Fixes https://github.com/HyphaApp/hypha/issues/3842

## Test Steps

 - [ ] Try changing the password and use different password while retyping

### DIFF
--- a/hypha/apply/templates/forms/includes/field.html
+++ b/hypha/apply/templates/forms/includes/field.html
@@ -56,7 +56,11 @@
 
                 {{ field }}
 
-                {% if field.errors %}<h6 class="form__error-text">{{ field.errors.as_text|linebreaksbr }}</h6>{% endif %}
+                {% if field.errors %}
+                    <h6 class="form__error-text">
+                        {{ field.errors.as_text|linebreaksbr }}
+                    </h6>
+                {% endif %}
                 {% if widget_type == 'clearable_file_input' or widget_type == 'multi_file_input' %}
                     <output class="form__file-list"></output>
                 {% endif %}
--- a/hypha/apply/templates/forms/includes/field.html
+++ b/hypha/apply/templates/forms/includes/field.html
@@ -57,9 +57,9 @@
                 {{ field }}
 
                 {% if field.errors %}
-                    <h6 class="form__error-text">
+                    <p class="form__error-text">
                         {{ field.errors.as_text|linebreaksbr }}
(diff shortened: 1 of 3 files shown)


frjo:
Can we have the error message pop up on the "New password" field instead of the "New password confirmation" field?

hypha/apply/templates/forms/includes/field.html
```diff
@@ -56,7 +56,11 @@
 
                 {{ field }}
 
-                {% if field.errors %}<h6 class="form__error-text">{{ field.errors.as_text|linebreaksbr }}</h6>{% endif %}
+                {% if field.errors %}
+                    <h6 class="form__error-text">
```
frjo: Can we get rid of the "h6" as well. The original Hypha theme misuse headings quite a lot.

frjo (changes requested):

hypha/apply/templates/forms/includes/field.html
theskumar: Sure

theskumar:
> Can we have the error message pop up on the "New password" field instead of the "New password confirmation" field?

let me take a stab at it. I believe it's the Django's default password change form.

hypha/apply/templates/forms/includes/field.html
theskumar: Updated.

theskumar:
> Can we have the error message pop up on the "New password" field instead of the "New password confirmation" field?

I cross checked and the validation logic comes from the default ChangePassword View/Form from Django. So I would try avoid changing the position of the error message unless you have a strong opinion/need for it. Also, in the latest version of Django the code related to validation of passwords has been updated, so I would advice we don't implement a custom validate logic which will then need to be maintained. 

Thinking out loud, the validation error messages in the screenshot is shown only after both the passwords match, so I think it might be too confusing to users.

wes-otf (approved):

frjo (approved):

PR #3869, opened 2024-04-10
WIP: Make it easy for staff to close rounds & filter on unpublished rounds

### LINKED ISSUE #3867: Closed rounds should include unpublished rounds
Make all filters include unpublished rounds when search for closed rounds. 

This include unpublished rounds:

https://github.com/HyphaApp/hypha/blob/8e144552f3f561033b93bfc1c02670e80387557d/hypha/apply/funds/tables.py#L581-L600

This does not include unpublished rounds:

https://github.com/HyphaApp/hypha/blob/8e144552f3f561033b93bfc1c02670e80387557d/hypha/apply/funds/views_partials.py#L92-L122

This does not include unpublished rounds:

https://github.com/HyphaApp/hypha/blob/8e144552f3f561033b93bfc1c02670e80387557d/hypha/apply/funds/admin_helpers.py#L84-L100

wes-otf:
Fixes #3867. Closed rounds now include unpublished rounds

## Test Steps

 - [ ] As Staff Admin...
 - [ ] Unpublish an existing live round in Wagtail admin
 - [ ] Navigate to https://test-apply.hypha.app/admin/funds/round/ and ensure the unpublished round shows up when the  `Closed` filter is selected on the right hand side
 - [ ] Navigate to the [beta/list submission view](https://test-apply.hypha.app/apply/submissions/all-alt/) and ensure that the unpublished round appears under the `Rounds` > `Closed` filter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,24 +1,66 @@
-docs_build/
+# Hypha
 /hypha/static_compiled
-__coverage__/
+
+# Django
+*.log
+*.mo
+*.pot
 *.pyc
-.DS_Store
-*.swp
-/venv/
-/static/
-/media/
-/tmp/
-/.vagrant/
-/Vagrantfile.local
+__pycache__
+db.sqlite3
+media/
+static/
+tmp/
+
+# Node
 node_modules/
-/npm-debug.log
-/.idea/
-.cache/
-.eslintcache
+npm-debug.log
+
+# Backup files
+*.bak
+
+# If you are using PyCharm
+# User-specific stuff
+.idea/**/workspace.xml
+.idea/**/tasks.xml
+.idea/**/usage.statistics.xml
+.idea/**/dictionaries
+.idea/**/shelf
+
+# AWS User-specific
+.idea/**/aws.xml
+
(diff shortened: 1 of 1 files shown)


sandeepsajan0 (commented):
If I unpublished an open round it is shown in both tabs open and closed on the staff dashboard. I think It is because previously open and close both were filtered via date and now we have added not-live to closed so date-wise it is open but it is not live so it is closed as well.

image

image

sandeepsajan0:
I am not sure but l think it might be confusing for staff because right now rounds are closed only if their end date has been passed. But now in admin, open unpublished rounds will be shown in the closed tab, technically it is correct but it might be confusing. 
 
Do we know the use case why staff is unpublishing a round, do they want to close it or just want to take it down for some time? Because if it just for some time and want to reopen it again then a separate filter `unpublished` might help better otherwise closed rounds will be a long list.

wes-otf:
@sandeepsajan0 I agree, I think it might be better to move toward `unpublished` being a filter and giving more clarity/making it easier to just close a round

PR #3881, opened 2024-04-18
Skip PAF Approval process if no PAFReviewerRole assigned

### LINKED ISSUE #3839: If no PAF Reviewer Roles created skip project approval
This makes the system more adaptable to different use cases.

sandeepsajan0:
Fixes #3839 

##  Assumptions

An organization only chooses one option either go with PAFReviewerRoles or without that. Changing in between might cause the issue with `Internal Approval` status projects. We will fix this in another issue if needed.

## Test Steps

 - [ ] Remove all PAFReviewerRoles from wagtail admin
 - [ ] Select a project in draft state, and fill paf
 - [ ] There should be a `Continue to next status` button in Actions to take. Also check `To do` message in the sidebar.
 - [ ] After confirmation, the project should move to the contracting stage.
 - [ ] There shouldn't be any `Internal Approval` stage.

Don't forget to add those PAFReviewerRoles back because we have already chosen that option and to avoid the issue explained in assumption above.

### DIFF
--- a/hypha/apply/projects/templates/application_projects/project_admin_detail.html
+++ b/hypha/apply/projects/templates/application_projects/project_admin_detail.html
@@ -4,14 +4,24 @@
 
 
 {% block admin_assignments %}
-    {% if object.is_in_progress %}
+    {% user_can_update_project_status object user as can_update_status %}
+    {% if can_update_status %}
         <div class="js-actions-sidebar sidebar__inner sidebar__inner--light-blue sidebar__inner--actions {% if mobile %}sidebar__inner--mobile{% endif %}">
 
             <h5>{% trans "Actions to take" %}</h5>
 
-            {% user_can_update_project_status project user as can_update_status %}
+            {% user_can_skip_pafapproval_process project user as can_skip_paf %}
+            {% if can_skip_paf %}
+                <a data-fancybox data-src="#continue-to-next-phase" class="button button--white button--full-width button--bottom-space" href="#">{% trans "Continue to next status" %}</a>
+
+                <div class="modal" id="continue-to-next-phase">
+                    <h4 class="modal__project-header-bar">{% trans "Continue to next stage" %}</h4>
+                    <p>{% trans "Have you filled the PAF and sure to move to the next stage, this action can not be reverted" %}</p>
+                    {% trans "Continue" as submit %}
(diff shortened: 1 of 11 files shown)


frjo (approved):
Nice work!

wes-otf:
Following the test steps this didn't seem to work on the test instance, I removed all approver roles from users but the project still went to internal review. Let me know if I misunderstood a step of this:
Screenshot 2024-05-21 at 17 06 31
Screenshot 2024-05-21 at 17 06 44
Screenshot 2024-05-21 at 17 07 09

sandeepsajan0:
@wes-otf You don't have to remove 'approval roles from user' instead just remove PAFReviewerRoles from Project Settings.
It should be in `Project > Project Settings > PAFReviewerRoles`. 
image

hypha/apply/projects/templates/application_projects/project_admin_detail.html
```diff
@@ -4,14 +4,24 @@
...
             <h5>{% trans "Actions to take" %}</h5>
 
-            {% user_can_update_project_status project user as can_update_status %}
+            {% user_can_skip_pafapproval_process project user as can_skip_paf %}
+            {% if can_skip_paf %}
+                <a data-fancybox data-src="#continue-to-next-phase" class="button button--white button--full-width button--bottom-space" href="#">{% trans "Continue to next status" %}</a>
+
+                <div class="modal" id="continue-to-next-phase">
+                    <h4 class="modal__project-header-bar">{% trans "Continue to next stage" %}</h4>
+                    <p>{% trans "Have you filled the PAF and sure to move to the next stage, this action can not be reverted" %}</p>
```
wes-otf: What are your thoughts on the verbiage here being changed to something like `Please ensure the Project Form is completed and you are ready to proceed to the next stage. This action cannot be reverted.` for a bit more clarity & to use the new "Project Form" vs "Project Approval Form" language?

wes-otf:
hahahaha thanks for that assist, I had a feeling I wasn't taking the right steps there. All looks good though! Left a minor comment about wording but perfect functionality wise.

hypha/apply/projects/templates/application_projects/project_admin_detail.html
frjo: Updated the text now.

PR #3893, opened 2024-04-19
Visibility fixes for partners that are also applicants

### LINKED ISSUE #3892: Partners that hold an applicant role have issues viewing comments with `Partner` visibility
Something I didn't consider previously was that given partners can also be applicants (and often are), there are issues that come with this as some partners were having applicant visibility on applications they were partners on. This is because the `APPLICANT` role is considered first in the following (L233):
https://github.com/HyphaApp/hypha/blob/7ed9d365d9dbbaf0103cf8baa5a1970a5725f41f/hypha/apply/activity/models.py#L216-L236

even with this mediated and `PARTNER` being evaluated first, the partner would have the visibility of a partner on applications they submitted.

wes-otf:
Fixes #3892. This solution was a bit unruly (especially ugly after linting) but it seemed like the logical approach without refactoring the whole `visibility_for` to be a normal method rather than a class method. This fix evaluates if the partner/reviewer is the author of the source submission first before evaluating the their role. The reviewer role was also lumped in here because even though I know we advise against a reviewer also being an applicant, it still could happen without oversight.

I tried to keep the code as clean as I could and reuse QuerySets, but let me know if y'all see any way I can optimize!

## Test Steps

 - [ ] Follow the same test steps as #3744, but ensure the partner user also has the role of `Applicant`
     - Only steps regarding the partner account's visibility need to be followed

### DIFF
--- a/hypha/apply/activity/models.py
+++ b/hypha/apply/activity/models.py
@@ -2,6 +2,7 @@ import os
 import uuid
 from typing import List, Optional, Tuple
 
+from django.apps import apps
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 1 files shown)


frjo (approved):

wes-otf:
@frjo if you get the time can you give this a test? I'd like to get this patch in soon

frjo:
Works well in my testing.

PR #3896, opened 2024-04-22
Utilize emails as display name on comments when no name is set

### LINKED ISSUE #3895: No display name is shown on comments if a user doesn't have their name set
If a user has left their name field blank in their profile, comments will only show up with the name of ` (<ROLE>)`

In the following, only `Test Partner` has their name set:
Screenshot 2024-04-22 at 05 08 47

wes-otf:
Fixes #3895. Quick one I noticed while working on the partner comment stuff. Comments will now display with the user email where no full name is set.

## Screenshots
Screenshot 2024-04-22 at 05 20 24

## Test Steps

 - [ ] Clear the `Full Name` field on https://test-apply.hypha.app/account/ and select `Update Profile`
 - [ ] Navigate to a submission that has comments from the current user account
 - [ ] Ensure that the current email appears on the comments instead of an empty name

### DIFF
--- a/hypha/apply/users/models.py
+++ b/hypha/apply/users/models.py
@@ -226,21 +226,26 @@ class User(AbstractUser):
     def get_full_name(self):
         return self.full_name.strip()
 
-    def get_short_name(self):
-        return self.email
+    def get_short_name(self) -> str:
+        """Gets the local-part (username) of the user's email
+
+        ie. hyphaiscool@hypha.app returns "hyphaiscool"
+        """
+        return self.email.split("@")[0]
 
     def get_display_name_with_group(self) -> str:
         """Gets the user's display name, along with their role in parenthesis
 
         If the user has a full name set that will be used, otherwise pulls the email.
         """
-        display_name = self.full_name if self.full_name else self.email
+        display_name = str(self)
         is_apply_staff = f" ({STAFF_GROUP_NAME})" if self.is_apply_staff else ""
         is_reviewer = f" ({REVIEWER_GROUP_NAME})" if self.is_reviewer else ""
+        is_partner = f" ({PARTNER_GROUP_NAME})" if self.is_applicant else ""
         is_applicant = f" ({APPLICANT_GROUP_NAME})" if self.is_applicant else ""
         is_finance = f" ({FINANCE_GROUP_NAME})" if self.is_finance else ""
         is_contracting = f" ({CONTRACTING_GROUP_NAME})" if self.is_contracting else ""
-        return f"{display_name.strip()}{is_apply_staff}{is_reviewer}{is_applicant}{is_finance}{is_contracting}"
+        return f"{display_name}{is_apply_staff}{is_reviewer}{is_applicant}{is_partner}{is_finance}{is_contracting}"
 
     @cached_property
     def roles(self):
--- a/hypha/apply/users/models.py
+++ b/hypha/apply/users/models.py
@@ -229,13 +229,18 @@ class User(AbstractUser):
     def get_short_name(self):
         return self.email
 
-    def get_full_name_with_group(self):
+    def get_display_name_with_group(self) -> str:
+        """Gets the user's display name, along with their role in parenthesis
+
+        If the user has a full name set that will be used, otherwise pulls the email.
+        """
+        display_name = self.full_name if self.full_name else self.email
         is_apply_staff = f" ({STAFF_GROUP_NAME})" if self.is_apply_staff else ""
         is_reviewer = f" ({REVIEWER_GROUP_NAME})" if self.is_reviewer else ""
         is_applicant = f" ({APPLICANT_GROUP_NAME})" if self.is_applicant else ""
         is_finance = f" ({FINANCE_GROUP_NAME})" if self.is_finance else ""
         is_contracting = f" ({CONTRACTING_GROUP_NAME})" if self.is_contracting else ""
-        return f"{self.full_name.strip()}{is_apply_staff}{is_reviewer}{is_applicant}{is_finance}{is_contracting}"
+        return f"{display_name.strip()}{is_apply_staff}{is_reviewer}{is_applicant}{is_finance}{is_contracting}"
 
     @cached_property
     def roles(self):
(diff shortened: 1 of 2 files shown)


wes-otf:
branched from my other PR on accident, should be all set now.

frjo (approved):
Nice fix.

theskumar:
Is there any privacy concern with emails being display. Is so we can try showing only first few characters and the domain. 

e.g. for `saurabh@example.com` -> `sau...@example.com`

wes-otf:
Hey! In OTF's usage I don't see a scenario where there could be a concern of privacy. Most of the time the email address is accessable via the application overview anyway as far as I know. Let me know if there's a scenario you think this could be an issue though

theskumar:
> Hey! In OTF's usage I don't see a scenario where there could be a concern of privacy. Most of the time the email address is accessable via the application overview anyway as far as I know. Let me know if there's a scenario you think this could be an issue though

I think i was primarily thinking around the emails of reviewers (internal and external). But then i as also thinking stricter and more privacy friendly implementation without a use-case.

I used something like this for the long filenames, where the use-case was more of UX but here it might act as good UX (for long emails) & privacy by default feature.

just to be sure, I don't have a strong argument/opinion for this. We go with what you decide, I'm just get worried when PII get exposed to a wider audience. 

https://github.com/HyphaApp/hypha/blob/d647118ee827eb1c30ff50620ac221a78726591c/hypha/apply/utils/templatetags/apply_tags.py#L30-L40

wes-otf:
ahh I see, that totally makes sense! I'm absolutely all for preserving privacy wherever possible. 

I chatted with staff a bit today and I guess per their bylaws OTF's external reviewers shouldn't even be able to interact with the applicants (more in #3907). If we move forward and make it so only staff can see comments from external reviewers, then only internal reviewers are to be considered, which would [always be staff](https://github.com/HyphaApp/hypha/blob/fb3356810d27fa64334acab161eae381540c112c/hypha/apply/funds/forms.py#L446) and thus appear as the [Org's long name](https://github.com/HyphaApp/hypha/blob/fb3356810d27fa64334acab161eae381540c112c/hypha/apply/activity/templatetags/activity_tags.py#L30-L35) in comments rather than any other display name. I'll ask around and see what folks think though, especially if there's a need for partners' identities to be concealed or anything.

frjo:
Works well in my testing.

wes-otf:
Chatted with staff & Slammer around this - currently the only time someone who isn't staff would be seeing a user's email would be in the case of a partner and in OTF's use case that's fine, the partner & applicant are expected to have this info already. I think this might be worth a revisit down the road though as we continue to expand and rethink the comment section.

wes-otf:
For now I think this is ready to go though unless there's any objections

theskumar:
When a new comment is posted, the applicant get this email. `hypha/apply/activity/templates/messages/email/comment.html`. If the name of the commenter is not set, the email contains the email of the commenter. Should that be fixed? Maybe use more generic phrase. like `comment by Staff` instead or `comment by xyz@email.com`

frjo:
@theskumar Good point!

How do you all feel about this suggestion?

We let the short name be just the first part of the e-mail address.

```
    def get_short_name(self):
        return self.email.split("@")[0]

```

We then replace most uses of "full_name" with "display_name" throughout Hypha.

If the user has not set a name we have something to display, but we do not leak the full e-mail address unnecessarily.

hypha/apply/users/models.py
```diff
@@ -229,13 +229,18 @@ def get_full_name(self):
...
+        """Gets the user's display name, along with their role in parenthesis
+
+        If the user has a full name set that will be used, otherwise pulls the email.
+        """
+        display_name = self.full_name if self.full_name else self.email
         is_apply_staff = f" ({STAFF_GROUP_NAME})" if self.is_apply_staff else ""
         is_reviewer = f" ({REVIEWER_GROUP_NAME})" if self.is_reviewer else ""
         is_applicant = f" ({APPLICANT_GROUP_NAME})" if self.is_applicant else ""
         is_finance = f" ({FINANCE_GROUP_NAME})" if self.is_finance else ""
         is_contracting = f" ({CONTRACTING_GROUP_NAME})" if self.is_contracting else ""
```
frjo, line 247: We should likely add "Partner" here.

hypha/apply/users/models.py
```diff
@@ -229,13 +229,18 @@ def get_full_name(self):
...
-        return f"{self.full_name.strip()}{is_apply_staff}{is_reviewer}{is_applicant}{is_finance}{is_contracting}"
+        return f"{display_name.strip()}{is_apply_staff}{is_reviewer}{is_applicant}{is_finance}{is_contracting}"
```
frjo: I think we can remove "strip()" from this line. It is already done in "get_full_name".

frjo (changes requested):

wes-otf:
All great catches! The short name seems to work nice & will even tighten up the options in the top right now that it doesn't need to include an email domain:
<br>
*before*
Screenshot 2024-05-08 at 14 15 52

*after*
Screenshot 2024-05-08 at 14 16 16

other changes should also be more in-line with what was mentioned here

hypha/apply/users/models.py
```diff
@@ -226,21 +226,26 @@ def __str__(self):
...
+        """
+        return self.email.split("@")[0]
 
     def get_display_name_with_group(self) -> str:
         """Gets the user's display name, along with their role in parenthesis
 
         If the user has a full name set that will be used, otherwise pulls the email.
         """
-        display_name = self.full_name if self.full_name else self.email
+        display_name = str(self)
```
wes-otf, line 241: figured this was better than duplicating the str dunder logic

hypha/apply/users/models.py
```diff
@@ -226,21 +226,26 @@ def __str__(self):
...
         is_apply_staff = f" ({STAFF_GROUP_NAME})" if self.is_apply_staff else ""
         is_reviewer = f" ({REVIEWER_GROUP_NAME})" if self.is_reviewer else ""
+        is_partner = f" ({PARTNER_GROUP_NAME})" if self.is_applicant else ""
```
frjo: Should be `self.is_partner`

frjo (approved):

sandeepsajan0 (approved):

PR #3898, opened 2024-04-23
Refractor primary nav and dashboard, hide projects info if not enabled

theskumar:
Also, refractor and allow for importing permission functions from anywhere and not just limit to be imported from the
`hypha.users.decorators` module

This PR also doesn't display project info on the dashboard if the projects is not enabled

For the stats block at the top of the dashboard, if the projects is not abled displays the count of flagged submissions and the count of previously reviewed submissions

Simplify the nav generator logic. 

## Test Steps

 - [ ] set PROJECTS_ENABLED=false in your .env file
 - [ ] Check that the projects related info is not displayed in the primary navigation or on the user's dashboard


theskumar:
Apologies for mixing up two type of change in a single PR, 1) how permissions method is imported 2) PROJECTS_ENABLED flag

frjo (approved):

wes-otf:
This looks good to me! Not sure the implications of this/in scope of PR but should we add a settings conditional here so Wagtail settings for Projects don't come up if not enabled?
https://github.com/HyphaApp/hypha/blob/587c6c7a2f0519c748499c3139ad9f3e261231d6/hypha/apply/projects/wagtail_hooks.py#L5

theskumar:
> This looks good to me! Not sure the implications of this/in scope of PR but should we add a settings conditional here so Wagtail settings for Projects don't come up if not enabled?
> 
> https://github.com/HyphaApp/hypha/blob/587c6c7a2f0519c748499c3139ad9f3e261231d6/hypha/apply/projects/wagtail_hooks.py#L5

I think that's an very nice suggestion, and we would probably provide a cleaner experience to admins. It will open up other things to hide in admin as well, as i believe the admin was never thought/build that way, e.g. entries in group -> permissions, PAF form options in the fund/round, etc. 

Not sure it's all worth the effort. Thoughts?

wes-otf:
yeah that's true, probably out of scope here. I'll write up an issue and toss it on the backlog - if we're ever itching for work we can pick it up.

PR #3905, opened 2024-04-30
Add draft submission of an applicant to their todo list

sandeepsajan0:
Fixes #ISSUEID

image

## Test Steps

 - [ ] Create a draft as an Applicant.
 - [ ] Check my task list in dashboard and that draft should be there as a task to submit.
 - [ ] After submit of the draft, task should be remove from my task list.


frjo:
@sandeepsajan0 Migration conflict so need a rebase before putting on test.

PR #3913, opened 2024-05-03
Email notifications on new comments for partners

### LINKED ISSUE #3871: Partner comment notifications
Partners should get notifications for comments on applications they're assigned to & applicants should get notifications when partners comment. There may be more examples of where notifications should be sent too.

wes-otf:
Fixes #3871. Allows partners that are assigned to an application to also get email notifications when a new comment is made that has applicable visibility. There was some refactoring that took place to allow for email messages to be customized on a per-recipient basis rather than having one be generated for all recipients.

## Test Steps

 - [ ] Assign one or more partners to an application
 - [ ] As staff, comment on the application with the visibility of `Partners`
 - [ ] Ensure only partners receive the notification email
 - [ ] As staff, comment on the application with the visibility of `Applicants`
 - [ ] Ensure only the applicant receives the notification email
 - [ ] As staff, comment on the application with the visibility of `Applicants & Partners`
 - [ ] Ensure both the applicant & partners receive the notification email
 - [ ] Try some of the same logic but become partner and/or applicant and ensure email notifications have a similar result as the previous steps

### DIFF
--- a/hypha/apply/activity/adapters/emails.py
+++ b/hypha/apply/activity/adapters/emails.py
@@ -424,8 +424,7 @@ class EmailAdapter(AdapterBase):
             recipients: List[str] = []
 
             comment = kwargs["related"]
-            partners = self.partners(source)
-            if partners:
+            if partners := [*source.partners.values_list("email", flat=True)]:
                 if comment.visibility == PARTNER:
                     recipients = partners
                 elif comment.visibility in [APPLICANT_PARTNERS, ALL]:
@@ -470,14 +469,6 @@ class EmailAdapter(AdapterBase):
             if source.phase.permissions.can_review(reviewer)
             and not reviewer.is_apply_staff
         ]
-
-    def partners(self, source) -> List[str]:
-        """Get all partner emails from the source (if any)
-
-        Returns:
-            list: emails of all partners on the source
-        """
-        return [partner.email for partner in source.partners.all()]
 
     def partners_updated_applicant(self, added, removed, **kwargs):
         if added:
--- a/hypha/apply/activity/adapters/emails.py
+++ b/hypha/apply/activity/adapters/emails.py
@@ -418,26 +418,31 @@ class EmailAdapter(AdapterBase):
             return user.email
 
         ApplicationSubmission = apps.get_model("funds", "ApplicationSubmission")
-        if message_type == MESSAGES.COMMENT and isinstance(
-            source, ApplicationSubmission
-        ):
-            recipients: List[str] = []
+        Project = apps.get_model("application_projects", "Project")
+        if message_type == MESSAGES.COMMENT:
+            # Comment handling for Submissions
+            if isinstance(source, ApplicationSubmission):
+                recipients: List[str] = []
 
-            comment = kwargs["related"]
-            if partners := [*source.partners.values_list("email", flat=True)]:
-                if comment.visibility == PARTNER:
-                    recipients = partners
-                elif comment.visibility in [APPLICANT_PARTNERS, ALL]:
-                    recipients = partners + [source.user.email]
-                else:
-                    recipients = [source.user.email]
+                comment = kwargs["related"]
+                if partners := [*source.partners.values_list("email", flat=True)]:
+                    if comment.visibility == PARTNER:
+                        recipients = partners
+                    elif comment.visibility in [APPLICANT_PARTNERS, ALL]:
+                        recipients = partners + [source.user.email]
+                    else:
+                        recipients = [source.user.email]
 
-            try:
-                recipients.remove(comment.user.email)
-            except ValueError:
-                pass
+                try:
+                    recipients.remove(comment.user.email)
+                except ValueError:
+                    pass
 
-            return recipients
+                return recipients
+
+            # Comment handling for Projects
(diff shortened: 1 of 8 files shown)


hypha/apply/activity/tests/test_messaging.py
```diff
@@ -497,14 +507,64 @@ def test_no_email_private_comment(self):
         self.assertEqual(len(mail.outbox), 0)
 
     def test_no_email_own_comment(self):
-        application = ApplicationSubmissionFactory()
-        comment = CommentFactory(user=application.user, source=application)
+        submission = ApplicationSubmissionFactory()
```
wes-otf, line 510: unrelated to this PR but updated for verbiage consistency w/ rest of the tests

frjo (approved):

hypha/apply/activity/adapters/emails.py
```diff
@@ -410,6 +416,30 @@ def recipients(self, message_type, source, user, **kwargs):
...
+
+        ApplicationSubmission = apps.get_model("funds", "ApplicationSubmission")
+        if message_type == MESSAGES.COMMENT and isinstance(
+            source, ApplicationSubmission
+        ):
+            recipients: List[str] = []
+
+            comment = kwargs["related"]
+            partners = self.partners(source)
+            if partners:
```
theskumar:
```suggestion
            if partners := self.partners(source):
```

More concise.

hypha/apply/activity/adapters/emails.py
```diff
@@ -441,14 +471,27 @@ def reviewers(self, source):
             and not reviewer.is_apply_staff
         ]
 
+    def partners(self, source) -> List[str]:
+        """Get all partner emails from the source (if any)
+
+        Returns:
+            list: emails of all partners on the source
+        """
+        return [partner.email for partner in source.partners.all()]
```
theskumar: This could simplified to `source.partners.values_list('email', flat=True)`, and then you might also not need this extra function. 

https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.values_list

hypha/apply/activity/adapters/emails.py
theskumar: Alternatively, more sematic

```python
if partner_emails  := self.get_partner_emails(source):
   ...
```

or,
```python
if partner_emails := source.partners.values_list('email', flat=True):
   ...
```

hypha/apply/activity/templates/messages/email/partners_update_partner.html
```diff
@@ -1,7 +1,7 @@
 {% extends "messages/email/base.html" %}
 
-{% load i18n %}
-{% block salutation %}{% trans "Dear Partner," %}{% endblock %}
+{% load i18n activity_tags %}
+{% block salutation %}{% trans "Dear" %} {% comment_saluatation_name %},{% endblock %}
```
theskumar: IMO, we should just use the recipient name always, given the next sentence is `You have been added as partner...`

Thoughts?

sandeepsajan0 (approved):
Works great!!

hypha/apply/activity/adapters/emails.py
wes-otf: I did this via comprehension because line 432 needs a python list to append to: `recipients = [source.user.email]` but I think you're totally right in terms of this function being able to be removed. I also replaced the comprehension with `[*source.partners.values_list('email', flat=True)]`, not too sure which is better in terms speed & resources though

hypha/apply/activity/templates/messages/email/partners_update_partner.html
wes-otf: my primary concern was I didn't want the user's email to be repeated by using the user `__str__` as that also felt redundant but actually with the `get_short_name` change in #3896 I just tossed `comment_salutation_name` in favor of using the string representation - seemed to work well in local tests!

hypha/apply/activity/adapters/emails.py
wes-otf: I always forget about the walrus operator - great suggestion!

wes-otf:
thanks for testing y'all! all the feedback is fantastic @theskumar

wes-otf:
this should be all set, had a few kinks but added more unit tests to ensure they don't come back

hypha/apply/activity/adapters/emails.py
theskumar: `values_list` gives an iterable, so i don't see any benefit to de-struct it and make it a list again. 

```
In [1]: [*[1,2,3]]
Out[1]: [1, 2, 3]
```

theskumar: I'm not sure about the performance, but if you really need a pure list to work with. 

```python
list(source.partners.values_list('email', flat=True))
```

is more python and readable to me.

wes-otf: to be able to add `source.user.email` to the recipient iterable it needs to be a pure list rather than a queryset but totally agreed on readability

wes-otf:
can we get this in? staff are pinging me saying they're having a hard time getting replies from partners without reaching out to them

PR #3920, opened 2024-05-08
Disable project related functionality if PROJECTS_ENABLED=False

### LINKED ISSUE #3606: Staff can access the /apply/projects url even if the PROJECTS_ENABLED=False
**Describe the bug**

While the feature flag PROJECTS_ENABLED=False, the staff user can still visit the project urls directly and access them. 

Found it while the test started to fail at https://github.com/HyphaApp/hypha/actions/runs/6466044430/job/17553298947#step:7:267

This test were passing earlier as the ROOT_URLCONF is by default set to 'hypha.urls' and if you update the ROOT_URLCONF to hypha.apply.urls you can see that test fail. This test case also missing the positive test case, it only have negative testcase.

**To Reproduce**

1. Login as staff.
2. Have "PROJECTS_ENABLED=False" in settings, this should be default if not enabled.
3. Visit `/apply/projects` or in local https://localhost:9001/apply/projects
4. See projects overview

**Expected behavior**

**Priority**

- Medium (slowing you down)
- **Low priority (annoying, would be nice to not see)**

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- Developers
- General public

**Additional context**

This currently breaks the test run at #3531

theskumar:
- do not register project urls
- do not register project menu in wagtail
- do not enable contracting/finance dashboard

Fixes #3606

Project settings and vendor settings will still be present, I could not find a way to hide/disable them conditionally. 

Screenshot 2024-05-08 at 4  16 29@2x

## Test Steps

 - [ ] Set PROJECTS_ENABLED=False
 - [ ] check dashboard, wagtail admin and finance roles
 - [ ] general site-wide checks
 - [ ] try visit the urls related to project/finance directly

### DIFF
--- a/hypha/apply/dashboard/templates/dashboard/dashboard.html
+++ b/hypha/apply/dashboard/templates/dashboard/dashboard.html
@@ -113,7 +113,7 @@
         {% if my_reviewed.table.data %}
             <div class="wrapper wrapper--bottom-space">
                 {% trans "Your previous reviews" as review_heading %}
-                {% include "funds/includes/table_filter_and_search.html" with filter=my_reviewed.filterset filter_action=my_reviewed.url search_term=search_term search_action=my_reviewed.url use_search=True use_batch_actions=False heading=review_heading %}
+                <h2 class="heading heading--normal">{{ review_heading }}</h2>
                 {% render_table my_reviewed.table %}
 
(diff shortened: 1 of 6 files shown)


frjo (approved):

wes-otf:
slick! works well.

PR #3921, opened 2024-05-08
Update sandbox database

wes-otf:
the Sandbox site has been updated to Hypha 5.8.0 & now the database dump reflects that


frjo:
@wes-otf  Merged. Now you need to rebase "sandbox" from main again and deploy so the new db is in place.

PR #3926, opened 2024-05-14
Add a Remeber me checkbox to login that extend session time.

### LINKED ISSUE #3897: Lower session time and add "Remember me" option to log in form
For improved security we should lover the default session time.

- [x] Set default `SESSION_COOKIE_AGE` to 12 hours.
- [x] Set default `SESSION_COOKIE_AGE_LONG` to two (2) weeks.
- [x] Add "Remember me" checkbox to login form, Hypha then uses `SESSION_COOKIE_AGE_LONG`.

Some questions:

1. Can we make it work with passwordless login?
2. If you login with Google, what session time should we set?

frjo:
Fixes #3897

"Remember me" is only available when logging in with password. I see this as an "advanced" feature. The passwordless login flow is easy so not a big hassle to do it once a day.

Organisation that wants to can set `SESSION_COOKIE_AGE` to a custom value, e.g. set it to two weeks to reagin the old behaviour.

To disable "Remember me" set `SESSION_COOKIE_AGE` to the same value as `SESSION_COOKIE_AGE_LONG`

This PR also:

* removes unused code is_public_site
* removes unused code register_extra_text
* removes all use of link-button* classes. Buttons are now only handled by the button class.
* add a tiny border radius to all buttons (this is something we implement on all elements in Hypha)

## Test Steps

 - [x] When logging in the session time should be 12 hours when remember me is not checked and 2 weeks when it is.


theskumar:
I believe this will add this only to login with password view, should it be added to passwordless login as well?

theskumar:
Probably little late here, but I believe using an "incognito" or "guest" mode in the browser is probably the safest way to use hypha if it's not used on their own computer. Security + Privacy. 

Wondering if it would make sense to use a verbatim like this in the form as well? 

Screenshot 2024-05-15 at 11  53 08@2x

frjo:
I'm still thinking about how to implement it for passwordless. I assume we need to add something to the login link we send via e-mail that indicated the "remember me" part.

Better solution?

hypha/apply/users/forms.py
```diff
@@ -14,20 +15,21 @@
...
 
-    remember_me = forms.BooleanField(
-        label=_("Remember me"),
-        help_text=_(
-            "On trusted devices only, keeps you logged in for a longer period."
-        ),
-        required=False,
-        widget=forms.CheckboxInput(),
-    )
+    if settings.SESSION_COOKIE_AGE < settings.SESSION_COOKIE_AGE_LONG:
```
theskumar: Make it less than and equal to?

frjo: Fixed now, thanks.

wes-otf:
@frjo testing this now, did you get any solutions for the passwordless side of things? I personally like the url/query param solution, removes the burden from us of having to store any unneeded data/weird user states.

wes-otf:
worked well! verified cookie age is as expected

PR #3928, opened 2024-05-17
Show Task list instead of activities in notification dropdown

### LINKED ISSUE #3887: Make the notification bell display a users task list instead of the activation feed

sandeepsajan0:
Fixes #3887 

image

## Test Steps

 - [ ] …


theskumar:
Why not re-use the same code/view for the dropdown content? 
Screenshot 2024-05-19 at 9  43 59@2x

move the dropdoown little up, see the current implementation as an example. 
Screenshot 2024-05-19 at 9  46 22@2x


Screenshot 2024-05-19 at 9  45 58@2x

sandeepsajan0:
@theskumar I thought that the dashboard look might take too much space in the dropdown but it looks fine. I have updated its UI.

image

frjo:
Could we test using something like `button--narrow` class on the "View" buttons? And/or `button--white`? Other ideas?

Visually they take over a bit when they are that big and that many.

sandeepsajan0:
@frjo I used button--narrow and button--transparent both. How is it looking now?

image

wes-otf:
This is killer! Much better use of the notification bell. @Techslammer & I had discussed some improvements earlier this week that can build on this PR, I'll chat about them a bit in Monday's call.

wes-otf:
for this PR it would actually be cool if the notification bell got inverted in color when clicked like what happens with the account drop down:
Screenshot 2024-06-12 at 15 51 15

PR #3930, opened 2024-05-20
Add application submission id everywhere, with an optional prefix

### LINKED ISSUE #3830: Allow adding prefix to IDs of submissions per fund/lab

No.

The IDs of the submissions is auto-incremented number right now. This is not ideal for referencing an application outside the hypha software. Having an optional prefix/shortcode, like `FUND24-<num>` allows for an easily identifiable identifier for application which can be referenced outside the hypha system without confusion.

- In admin, an ID prefix can be added
- After the prefix is set, then all the submissions made to that round/lab will store a `public_id` constructed using the prefix and the original db id. 
- The `public_id` will be searchable and displayed everywhere instead of submission -> id.

Generate the submission id dynamically instead of storing it. This is not ideal, as it can be modified very easily, and won't lead to a permanent reference number. 

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- Developers
- General public

**Ideal deadline**

This Month

theskumar:
This PR allow adding an option prefix to the submission ID's that are auto generated. 

The prefix can be added by going to round or lab settings page. 

This PR also ensure the prefix or application ID displayed along with title almost everywhere for each reference. Since the title of application is not unique, this helps add uniqueness to it. 

The autolinking of submission ids in communication has be updated to account for application id prefix

Fixes https://github.com/HyphaApp/hypha/issues/3830

Screenshot 2024-05-20 at 3  38 36@2x

Screenshot 2024-05-20 at 4  43 55@2x

Screenshot 2024-05-20 at 4  44 32@2x
Screenshot 2024-05-20 at 4  45 05@2x

## Test Steps

 - [ ] general checks across the site
 - [ ] check notification generate - slack, email, etc.


hypha/apply/funds/models/submissions.py
```diff
@@ -507,6 +510,10 @@ class Meta:
     def is_draft(self):
         return self.status == DRAFT_STATE
 
+    @property
+    def title_with_id(self):
```
sandeepsajan0: Can we use this method to provide 'title' or 'title with public_id' based on a feature flag? Some org might not want to show the application count or not at least to the applicant.

theskumar: Great suggestion, however, if possible I would try keep the number of feature flags to minimal, unless any adaptor thinks it's a deal break for them. 

@wes-otf @frjo If you happen to present this in adopter meeting, can we try get an idea if someone would want to hide this info? The ids have been and will always be available in the url.

hypha/apply/funds/templates/funds/includes/rendered_answers.html
```diff
@@ -25,13 +25,13 @@ <h5 class="text-base">{% trans "E-mail" %}</h5>
...
     {% if object.get_address_display != "-" %}
-        <div class="grid__cell--span-two">
+        <div>
             <h5 class="text-base">{% trans "Address" %}</h5>
             {{ object.get_address_display }}
         </div>
     {% endif %}
     {% if object.get_organization_name_display != "-" %}
-        <div class="grid__cell--span-two">
+        <div class="">
```
frjo: Remove the class attribute.

hypha/apply/funds/models/submissions.py
frjo: Did not sound like a need to hide exist. Could we provide a template for this? Making it easy to adopt the output.

You mentioned that some wanted to have the prefix first.

frjo (changes requested):
Do we support search for prefix or number (\#123)?

hypha/apply/funds/models/submissions.py
theskumar: Updated in https://github.com/HyphaApp/hypha/pull/3930/commits/0e406525490d9609c20cf2d0850f23314561adcb

theskumar:
Getting the submission prefix in the url as well is lot of work also requires decent testing. I think it would better to tackled in another pr.

frjo:
@theskumar I think this needs rebuild migrations and a rebase. Merged in a number of PRs today.

theskumar:
@frjo updated! :)

frjo (approved):

hypha/apply/activity/adapters/slack.py
```diff
@@ -31,51 +31,55 @@ class SlackAdapter(AdapterBase):
...
-            "A new submission has been submitted for {source.page.title}: <{link}|{source.title}> by {user}"
+            "A new submission has been submitted for {source.page.title}: <{link}|{source.title_text_display}> by {user}"
         ),
         MESSAGES.UPDATE_LEAD: _(
-            "The lead of <{link}|{source.title}> has been updated from {old_lead} to {source.lead} by {user}"
+            "The lead of <{link}|{source.title_text_display}> has been updated from {old_lead} to {source.lead} by {user}"
         ),
         MESSAGES.BATCH_UPDATE_LEAD: "handle_batch_lead",
         MESSAGES.COMMENT: _(
             "A new {comment.visibility} comment has been posted on <{link}|{source.title}> by {user}"
```
wes-otf, line 41: Should this be:
```suggestion
            "A new {comment.visibility} comment has been posted on <{link}|{source.title_text_display}> by {user}"
```
That's only one I noticed that was missing on slack

wes-otf:
Super small issue from my testing with the slack comment notif but otherwise works great!

PR #3931, opened 2024-05-20
Allow staff to edit project title

### LINKED ISSUE #3927: Allow staff to edit project name and title

sandeepsajan0:
Fixes #3927 

image

## Test Steps

 - [ ] …


frjo (approved):

frjo:
@sandeepsajan0 Could we add project user edit to the same form? Best to do in a separate PR or can we do it here?

Project name should only be editable if "Vendor" is not used I think. If Vendor is used the Vendor name should be used instead of Project name in most/all places.

wes-otf:
This works well! In my opinion maybe we write up a new issue so this change shows up in the activity feed too? May be useful for staff to see that a project name was changed from X -> Z and by who. Let me know what y'all think

frjo:
Yes, adding an activity about the name change is a good idea. Will add an issue about that.

PR #3950, opened 2024-05-29
Move addressfield into hypha and update jquery.addressfield.js.

### LINKED ISSUE #3939: Update Hypha addressfield
- [ ] Update address information if needed
- [ ] Update components if available
- [ ] Move inside of `hypha` directory.

frjo:
Fixes #3939

## Test Steps

 - [ ] Test so the address field on application forms still work as it should. For both new and existing submissions


wes-otf:
Testing was good! It looks like this is no longer [maintained](https://github.com/tableau-mkt/jquery.addressfield) & we're still using a release from 2016 - is this something we should be looking for an alternative for?

theskumar (approved):
Looks good to me.

PR #3955, opened 2024-05-30
Update Screening UI & UX

### LINKED ISSUE #3873: Rebuild Screening (thumb up/down) feature to use htmx
This function on the submission detail page now uses jQuery ajax. Would be nice to convert it to htmx.

Skärmavbild 2024-04-16 kl  09 52 51

https://github.com/HyphaApp/hypha/blob/main/hypha/static_src/javascript/screening-status.js

https://github.com/HyphaApp/hypha/blob/main/hypha/apply/api/v1/screening/views.py

https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/templates/funds/includes/screening_status_block.html

theskumar:
- Use htmx
- Update/fix behaviour
- code cleanup - js, api, python views/forms

Fixes https://github.com/HyphaApp/hypha/issues/3873

https://github.com/HyphaApp/hypha/assets/236356/befc068a-864e-4ae2-a571-268f3fe288f7

---

Screenshot 2024-06-03 at 9  18 57@2x

Screenshot 2024-06-03 at 9  19 53@2x

---

Screenshot 2024-06-03 at 9  16 08@2x

## Test Steps

 - [ ] Open a submission
 - [ ] check the behaviour for both applicant and staff
 - [ ] Check the status is displayed correctly in both "List" and "table" view of submission listing.


theskumar:
Still need to update the test cases, but please review ui and ux of screening.

frjo:
Screening decisions can be long so I think we need to allow wrapping. Can we also make the padding for the options the same as for the heading "Options".

Skärmavbild 2024-05-31 kl  13 02 28

This text is not needed I think.

Skärmavbild 2024-05-31 kl  13 34 02

frjo:
If you pick a none default option it is saved to the database in the submission all tables the submission still show as "Awaiting screening".

The reason is that these functions are used:

https://github.com/HyphaApp/hypha/blob/8b69c168dc33323bb049e180640d092ddaa6935a/hypha/apply/funds/models/submissions.py#L946-L956

frjo:
I really like the UI for this. It feels slick and very functional!

hypha/apply/funds/templates/funds/includes/screening_status_block-button.html
```diff
@@ -0,0 +1,90 @@
...
+             x-cloak
+             role="menu"
+             x-show="isOpen || openedWithKeyboard"
+             x-transition.origin.top
+             x-trap="openedWithKeyboard"
+             @click.outside="isOpen = false, openedWithKeyboard = false"
+             @keydown.down.prevent="$focus.wrap().next()"
+             @keydown.up.prevent="$focus.wrap().previous()"
+        >
+            <header class="subheading border-b px-3 py-2 flex justify-between items-center gap-2">
```
frjo, line 70: py-n classess add padding-right/left. Better to use classes that add padding-inline-start/end.

frjo (changes requested):

hypha/apply/funds/templates/funds/includes/screening_status_block-button.html
theskumar: Since the padding left and right is same, would it matter?

theskumar:
> Screening decisions can be long so I think we need to allow wrapping. 

I'll enable wrapping. 

> Can we also make the padding for the options the same as for the heading "Options".

The padding left is to allow for the ✔️ mark to show up. It follow the same behaviour as the dropdown in submission list view. I would suggest we keep as it is. Unless you think otherwise. Let me know, I'll remove.

Update: Looks like there was bug on the padding. Updated the code in https://github.com/HyphaApp/hypha/pull/3955/commits/6cae3c36df01ca1f71e6e49b8103cd09735ac93a

Screenshot 2024-05-31 at 7  46 53@2x

theskumar:
> If you pick a none default option it is saved to the database in the submission all tables the submission still show as "Awaiting screening".

Should we update the submission table to display the currently selected screening option instead?

theskumar:
> This text is not needed I think. https://github.com/HyphaApp/hypha/pull/3955#issuecomment-2142012064

Removed in https://github.com/HyphaApp/hypha/pull/3955/commits/0e696d964bbf064cc31fbf6c3eb3b21c2fefce65

theskumar:
Managed to update the test-cases, made them work with the new htmx flow.

hypha/apply/funds/templates/funds/includes/screening_status_block-button.html
```diff
@@ -0,0 +1,96 @@
+{% load i18n heroicons %}
+
+<div class="relative"
+     x-data="{ isOpen: false, openedWithKeyboard: false, leaveTimeout: null }"
+     @mouseleave.prevent="leaveTimeout = setTimeout(() => { isOpen = false }, 250)"
```
wes-otf: super nit-picky but I think I'd almost prefer no timeout with the mouse leave. To me it feels a little a little lethargic/cluttered when trying to move between the two screening options. Super small detail and may be my own personal preference though.

https://github.com/HyphaApp/hypha/assets/145372368/2bcc9185-dad6-49a9-9f5e-f6038acb05a8

wes-otf:
This is came out so good! In testing it was really nice to use, fun in way. Nice work @theskumar!

hypha/apply/funds/templates/funds/includes/screening_status_block-button.html
theskumar: @wes-otf  I agree with you. I've also felt the same way at times.

The timeout on a menu is generally set due to accessibility concerns. For example, some users may have motor control limitations, so it's beneficial to prevent the menu from disappearing immediately after the mouse leaves it.

That being said, a 250ms delay is likely too long in this case. Reducing it to around 100ms would likely provide a better balance between accessibility and responsiveness.

theskumar: @wes-otf was not able to play the video earlier. The mouse timeout seems fine to me. 

I believe what you're showing is b/c of the fade-in and fade-out animation duration. It's a necessary evil, but let me find a way around it so it's less distracting.

theskumar: Updated the transition a bit and also added some better hover and active states to the buttons. The transition looks much better without the `leavetimeout` but I was think let's keep it if the updated transition looks good.

https://github.com/HyphaApp/hypha/assets/236356/dadadc3b-dc92-4dbf-a16f-2c077a81e2b5

frjo (approved):

hypha/apply/funds/templates/funds/includes/screening_status_block-button.html
wes-otf: This is great! I totally get the accessibility concerns, I think you found the perfect medium at 100ms though. Seems a lot snappier.

PR #3956, opened 2024-05-30
Optionally skip applicant upload contract step by default

### LINKED ISSUE #3954: Let contract uploader skip the countersigning step by default

Not every funding organization uses Hypha for signing and countersigning. They might have the already-signed contract in another system but would like a copy in Hypha. To streamline the process of uploading such a contract, a Wagtail option to help skip the countersigning step is helpful.

This is in the Projects functionality, e.g. `apply/projects/1/` where `1` is in the `Contracting` state but does not have any contracts uploaded.

A Wagtail option that makes it easy to upload once instead of twice.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Staff
- Applicants
- Contracting

bickelj:
Add Wagtail setting to skip vendor contract upload

Hypha tracks whether a contract has been signed and countersigned depending on which role uploaded it and the context of the upload. This change creates a new Wagtail setting called `upload_countersigned_contract`. When set to `True`, whenever a contract upload occurs, the UI shows that the document needs to be countersigned and the document gets marked as countersigned. The net effect is that there is no need for two contract uploads, only one. When combined with the `staff_upload_contract` set to `True`, staff can upload a single contract document without other users' actions.

Fixes #3954

See notes in #3954. This is a different implementation based on #3953 and not based on the work from #3938. I am pushing this to have a record of it here. Perhaps there is something here that was missed in #3938 or vice versa. It is a slightly different approach.


bickelj:
Regarding the test failures, I don't expect this to be merged as-is, again the point is to offer the code for comparison.

bickelj:
I spent a few minutes to better integrate this with the newer approach, at least for the sake of advancing immediately to the Invoicing and Reporting stage. However, it does not look like `signed_by_applicant` is `True`, but perhaps it should be since other things might follow from that :shrug:.

bickelj:
I see the one failing test and I poked at it for a few minutes but I think it highlights a question of how we want to integrate and what should happen with `signed_by_applicant`. When staff uploads a countersigned contract it truly is `signed_by_applicant`, right?

bickelj:
@frjo Should I take this closure as a rejection of the premise or just of this implementation?

I have an open question above for the team as to tracking countersigned contracts:  When staff uploads a countersigned contract it truly is `signed_by_applicant`, right?

PR #3962, opened 2024-06-04
Allow staff admin to impersonate other users

### LINKED ISSUE #3961: Allow Staff Admin to impersonate other users
Allow Staff Admin to impersonate other users by enabling this functionality in the Hijack module of Hypha.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Staff Admin

bickelj:
When `HIJACK_ENABLED = True` and Staff Admin can see Wagtail admin and Staff Admin can view users, allow Staff Admin to impersonate.

Fixes #3961

## Test Steps

- Set `HIJACK_ENABLED = True`.
- As Admin, allow Staff Admin to add/update/remove Users.
- As Staff Admin user, find an `Impersonate` button now available in Wagtail admin
- Impersonate a user by clicking the button.

### DIFF
--- a/hypha/apply/users/templates/wagtailusers/users/list.html
+++ b/hypha/apply/users/templates/wagtailusers/users/list.html
@@ -1,4 +1,4 @@
-{% load i18n l10n wagtailusers_tags wagtailadmin_tags %}
+{% load i18n l10n wagtailusers_tags wagtailadmin_tags hijack %}
 <table class="listing">
     <thead>
         <tr>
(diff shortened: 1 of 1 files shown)


frjo:
Are you using Hijack in production? This is not something I want to encourage in Hypha.

The current implementation is only ment for development and testing, where it is quite handy.

bickelj:
@frjo It appears we do use hijack in prod.

frjo:
@bickelj That is not something that will be encourage in Hypha. If staff can act as other users there is no accountability and no meaningful auditing can be done.

PR #3965, opened 2024-06-04
Fix styles to use inline/block instead of left/right etc.

frjo:
Some left/right/top/bottom had slipped past.


theskumar (approved):
Nice!

PR #3968, opened 2024-06-06
Use Dialog box inplace of fancybox modal, with htmx and alpinejs

### LINKED ISSUE #3391: Convert all dialogs/modals to use the <dialog> tag and htmx instead of fancybox lib
Hypha mainly use the fancybox lib for dialogs and modals. It works well. The dialog-tag nowadays have good support in all modern browsers.

The dialog-tag will:

* Simplify the code.
* Improve accessibility.
* One less thing that relies on jQuery.

Good articles on the subject:

* http://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
* https://blog.webdevsimplified.com/2023-04/html-dialog/

sandeepsajan0:
Fixes #3391 

https://github.com/HyphaApp/hypha/assets/23638629/64b6babe-ff0f-4b98-a880-7517a6207887

Before:
image

After:
image

## Test Steps

 - [ ] Check all the modals in the Action to take section.
 - [ ] Check their functioning.
 - [ ] Check that the value must get update accordingly, like on update lead, only lead value should get updated instead of reloading the whole page/context.
 - [ ] Check Update Status different cases.


hypha/templates/base.html
```diff
@@ -65,6 +65,7 @@
         <style> [x-cloak] {display: none !important} </style>
         <!-- Alpine Plugins: installed with npm -->
         <script defer src="{% static 'js/vendor/alpine-focus.min.js' %}"></script>
+        <script src="//unpkg.com/@fylgja/alpinejs-dialog" defer></script>
```
theskumar: Use local version please.

requirements.txt
```diff
@@ -30,6 +30,7 @@ django-storages==1.14.2
 django-tables2==2.7.0
 django-tinymce==3.7.1
 django-two-factor-auth==1.16.0
+django-user-agents==0.4.0
```
theskumar, line 33: Needed?

hypha/apply/projects/forms/project.py
```diff
@@ -85,7 +85,7 @@ class ProjectCreateForm(forms.Form):
         label=_("Select Project Lead"), queryset=User.objects.all()
     )
 
-    def __init__(self, instance=None, user=None, *args, **kwargs):
+    def __init__(self, *args, instance=None, **kwargs):
```
theskumar, line 88: I believe the `*args` will be just before the `**kwargs`, the old code seems right, is this ruff doing the formatting?

sandeepsajan0: It created problems on passing positional args like self.request before instance at the time of initiation. And similar method is used at many places in funds/forms.py so I used this method. I think both methods are fine, does it create any problem?

requirements.txt
sandeepsajan0: Not sure but couldn't find any other better way yet. May be we can remove this with updated UI for mobile?
For now, we need this to avoid rendering same HTML twice, that'll cause conflict with alpine js dialog boxes.

hypha/apply/funds/templates/funds/includes/dialog_form_base.html
```diff
@@ -0,0 +1,53 @@
...
+    class="form {{ extra_classes }}"
+    method="post"
+    hx-post="{{ url }}"
+    hx-target="{{ target }}"
+    hx-swap="innerHTML"
+    hx-select="{{ target }}"
+    id="{% if form_id %}{{ form_id }}{% else %}{{ form.name }}{% endif %}"
+    enctype="multipart/form-data"
+    {% if action %}action="{{ action }}"{% endif %}>
+<!--    hx-on::after-request="if(event.detail.successful){leadForm = false, console.log(leadForm)}",>-->
```
sandeepsajan0: @theskumar Have you used after-request event? Although it is working fine and even leadForm is set as False but dialog box is not closing. If it can work and close the dialog box on setting leadForm false then we can remove the '@click' trigger from submit button. It'll help in error cases.

hypha/apply/funds/forms.py
```diff
@@ -208,7 +198,7 @@ def clean_submissions(self):
 class UpdateReviewersForm(ApplicationSubmissionModelForm):
     reviewer_reviewers = forms.ModelMultipleChoiceField(
         queryset=User.objects.reviewers().only("pk", "full_name"),
-        widget=Select2MultiCheckboxesWidget(attrs={"data-placeholder": "Reviewers"}),
+        widget=Select2MultiCheckboxesWidget(attrs={"data-placeholder": "Select..."}),
```
frjo: Can we not delete all the ` widget=Select…` lines now that we use choice.js?

hypha/templates/base.html
```diff
@@ -65,6 +65,7 @@
         <style> [x-cloak] {display: none !important} </style>
         <!-- Alpine Plugins: installed with npm -->
         <script defer src="{% static 'js/vendor/alpine-focus.min.js' %}"></script>
+        <script src="{% static 'js/vendor/alpine-dialog.min.js' %}"></script>
```
frjo: Add a "defer" to the script tag so it loads late.

frjo (changes requested):

hypha/apply/funds/templates/funds/includes/dialog_form_base.html
theskumar: You can find some nice pattern in this sample code.

https://github.com/geoffbeier/django-htmx-modal-form?tab=readme-ov-file

I think you should use `oob` to swap any other page on the element and use `target=this` for dismiss the dialog box, if the response is 204.

hypha/apply/funds/forms.py
sandeepsajan0: Yes, I think we can delete these.

hypha/apply/projects/forms/project.py
theskumar: i think it's fine if it works, we can tackle it later.

hypha/apply/funds/templates/funds/includes/dialog_form_base.html
theskumar: https://github.com/geoffbeier/django-htmx-modal-form/blob/tailwind/movie_collection/templates/index.html#L69-L80

frjo:
Skärmavbild 2024-06-20 kl  10 07 33

hypha/apply/funds/templates/funds/includes/admin_primary_actions.html
```diff
@@ -1,26 +1,48 @@
...
 {% if object.is_archive %}
-    <a data-fancybox
-       data-src="#unarchive-submission"
-       class="button button--bottom-space button--primary"
-       href="#">
-        {% heroicon_outline "lock-open" aria_hidden="true" size=16 stroke_width=2 class="inline align-baseline me-1" %}
-        {% trans "Unarchive Submission" %}
-    </a>
+    <div x-data="{ unarchiveForm: false }">
+        <a class="button button--bottom-space button--primary" @click="unarchiveForm = true" hx-get="{% url 'funds:submissions:unarchive' pk=object.pk %}" hx-target="#unarchive_form">
```
frjo: Since no "src" attribute is needed can we change from a-tag to button-tag?

sandeepsajan0:
@theskumar Can you please look into choice.js selectable dropdowns issue? Other things like UI shift and close via htmx event have been fixed but I couldn't find a way yet to put those dropdowns on the top of the dialog box.

requirements.txt
theskumar: Is this something that would be helpful? https://github.com/HyphaApp/hypha/commit/a892b7d9314237d0f41260d974563d513c909bf7#diff-991b4d1d98770e27e8db821a366e0c49d5ff3659e38677f7fc87f69f4326bfefR8

theskumar: > Not sure but couldn't find any other better way yet. May be we can remove this with updated UI for mobile?
For now, we need this to avoid rendering same HTML twice, that'll cause conflict with alpine js dialog boxes.

what is this used currently?

theskumar:
I've started an alternate implementation that fixes the choices.js overflow issue, and also has some other goodies. There were some hard-choices b/c of which I had to let go `<dialog>`, more on it to come later, i.e. either the dropdown in the dialog will be visible or the dialog will have proper scrolling, we need to have both. 

Right now, I'm going over each of the modal and make sure they work as expected. I've updated the pattern of modal creation and closing to be more simpler. 

You can checkout locally and compare a diff with main at https://github.com/HyphaApp/hypha/compare/enhancement/gh-3991-htmx-dialog-alt?expand=1

frjo (approved):

theskumar:
Superseded by https://github.com/HyphaApp/hypha/pull/4053

closing this PR

Issue #3979, opened 2024-06-10
Create command to sanitize database


bickelj:
**Is your feature request related to a problem? Please describe.**

As developers, there are times that recreating an issue requires recreating a data condition, and to do that most quickly involves restoring a customer database. However, we would rather use fake data than real data to minimize the risk of exposing sensitive data.

**Describe the solution you'd like**

A django command that replaces all the identifiable information in the database with fake text.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Developers

frjo:
Excellent feature!

theskumar:
Might be related and/or alteast can use the same underlying logic when someone implements the feature blow: 

- #3441

PR #3980, opened 2024-06-10
Sanitize database script

bickelj:
Issue #3979

## Test Steps

- As developer, restore a database, e.g. sandbox database
- Run the django command `sanitize_database`
- Run the django command `createsuperuser` (users names/emails may be overwritten)
- Log in as the superuser (just created above) and see sanitized data


bickelj:
I should note that this script works in some cases but not all. Regardless, it is a developer script, not customer-facing, and it is a new script, so I see it as low risk to merge.

bickelj:
@frjo Is this one OK to merge?

PR #3984, opened 2024-06-12
Staff role should not suppress reviewer emails

### LINKED ISSUE #3983: User is not receiving email notifications when assigned to review an application
**Describe the bug**

User has several roles including staff and admin. Email notifications for review are not happening.

**Expected behavior**

Emails should go out.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Staff

bickelj:
Reviewer email notification for assigned application behavior corrected.

Fixes #3983

## Test Steps

Set up a dummy email SMTP server or use file-based email.

- Update a user to be both Staff and Admin
- Assign that user to review an application/submission
- See emails go through to that user

### DIFF
--- a/hypha/apply/activity/adapters/emails.py
+++ b/hypha/apply/activity/adapters/emails.py
@@ -438,7 +438,6 @@ class EmailAdapter(AdapterBase):
             reviewer.email
             for reviewer in source.missing_reviewers.all()
             if source.phase.permissions.can_review(reviewer)
-            and not reviewer.is_apply_staff
         ]
 
     def partners_updated_applicant(self, added, removed, **kwargs):


frjo:
We need some way to configure this. OTF uses Slack and their staff do not want duplicate info via e-mail.

I'm thinking a settings `"STAFF_NOTIFICATION_METHOD" = email|slack|…`

bickelj:
@frjo That sounds simple enough. But then looking under the hood, it is not obvious if/where we track state of "I sent X type of message or not" so this might be more work than it seems.

frjo:
@bickelj All notifications to staff are sent via Slack currently. That is the reason for `and not reviewer.is_apply_staff`.

We plan, hopefully this year, to make all notifications go via e-mail as default. Staff notifications via Slack will be optional then.

If you think it is worthwhile you can implement the setting for this particular case. But might be best/easiest to wait for us fixing it for all notifications.

bickelj:
@frjo Since this is a one-line merge conflict, I think we can leave this alone for now. I was trying to think of a durable, better, more general way of honoring such a setting but it eluded me. My best thought was something like "pass the list of enabled messengers in kwargs to every messenger" and/or "pass the list of already-processed messengers in kwargs to every next messenger" but any actions taken on those would be under the assumption that the list is in priority order, which is flimsy.

PR #3991, opened 2024-06-15
Add tips for forced push and history search in shell.

frjo:


theskumar (commented):
Nice! This is super useful, I use up-arrow all the time for history search. 

Recently started using zsh-auto-suggestions and loving it. 

https://github.com/zsh-users/zsh-autosuggestions

PR #3992, opened 2024-06-16
Give the task bell the same hover state as the user menu

### LINKED ISSUE #3987: Invert notification bell when hover/clicked
This replicated the behaviour of the new user menu.

frjo:
Fixes #3987

https://github.com/HyphaApp/hypha/assets/191392/2c99a3c4-130c-4cbb-b47a-945ef70bdcd9

### DIFF
--- a/hypha/templates/base-apply.html
+++ b/hypha/templates/base-apply.html
@@ -100,7 +100,7 @@
                 {% endif %}
             </section>
 
-            <div class="header__button-container flex h-10 gap-4">
+            <div class="header__button-container flex h-10 gap-5">
                 {% comment %} Todo List {% endcomment %}
                 {% if request.user.is_authenticated and request.user.is_apply_staff %}
                     <div
@@ -108,7 +108,7 @@
                         x-init="$watch('open', value => { if (value) { document.getElementById('id-task-list').dispatchEvent(new Event('htmx:fetch')); } })"
                     >
                         <a href="{% url "todo:list" %}"
-                           class="p-2 flex items-center hover:opacity-70 transition-opacity"
+                           class="flex p-2 rounded-full text-black hover:bg-black hover:text-white active:bg-black active:text-white focus:bg-black focus:text-white"
                            aria-label="{% trans "Task List" %}"
                            aria-haspopup="task_list"
                            aria-expanded="false"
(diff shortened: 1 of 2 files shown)


theskumar:
Probably unrelated to this PR, but noticed this whitespace, unexpected? 
Screenshot 2024-06-16 at 9  40 30@2x

hypha/templates/base-apply.html
```diff
@@ -100,23 +100,23 @@
...
+            <div class="header__button-container flex h-10 gap-5">
                 {% comment %} Todo List {% endcomment %}
                 {% if request.user.is_authenticated and request.user.is_apply_staff %}
                     <div
                         x-data="{open: false}"
                         x-init="$watch('open', value => { if (value) { document.getElementById('id-task-list').dispatchEvent(new Event('htmx:fetch')); } })"
                     >
                         <a href="{% url "todo:list" %}"
-                           class="p-2 flex items-center hover:opacity-70 transition-opacity"
+                           class="flex p-2 rounded-full text-black hover:bg-black hover:text-white active:bg-black active:text-white focus:bg-black focus:text-white"
```
theskumar: Did you mean to keep only one `focus:text-white`? Remove `focus:bg-back`?

theskumar:
The use of `active` css is causing some side-effects (see the video below), maybe try using the alpine variable to style the button while the dialog is open using `:class=` 

- When the dialog is open, try hovering over the primary navigation.
- If you click on the bell icon itself, the black bg remain active.

I noticed some minor improvements that can be done. 
- use standard `ring-1` for active state, i.e. a border around to indicate the button is active or focused when you use keyboard or mouse. 
- Remove the `end-4` and use `end-0` in the tasks list dialog, let it align similar to the dialog box alignment of the profile button. 
- Use `transition-colors` on the buttons
- To close the dialog boxes you must click somewhere else, include a "close" botton in the dialog -> panel? 
- I think using gap-4 instead of gap-5 is fine. gap-5 feels odd spacing, we rarely use it.

https://github.com/HyphaApp/hypha/assets/236356/9e78341f-a87d-44ba-a508-b072dc54324c

hypha/templates/base-apply.html
```diff
@@ -100,23 +100,23 @@
                 {% endif %}
             </section>
 
-            <div class="header__button-container flex h-10 gap-4">
+            <div class="header__button-container flex h-10 gap-5">
```
theskumar: `gap-4` is probably fine here, an even number `1em` that we use everywhere.

hypha/templates/base-apply.html
theskumar: You can add `transition-colors` as well here

wes-otf:
Works well in testing! The only thing I noticed is with the latest fixes the icons don't stay inverted when the menu is active. If this is intentional then all should be ready to go!

PR #3995, opened 2024-06-19
Update Wagtail to v5.2.5

frjo:
Also fixes one instance of "The `classnames` kwarg for MenuItem is deprecated - use `classname` instead.".


frjo:
@theskumar I still get the "RemovedInWagtail60Warning" above when going to Wagtail admin but I do not find any more instance of `classnames`. Ideas?

theskumar:
> @theskumar I still get the "RemovedInWagtail60Warning" above when going to Wagtail admin but I do not find any more instance of `classnames`. Ideas?

@frjo probably a third-party library still using it, that registers a menuitem.

PR #3997, opened 2024-06-19
Use x-cloak on user-menu popover to hide it during initial load

theskumar:
Also, use `ms-2` on the admin icon margin

Update html for better git diff and readability, linebreak on each property

## Test Steps

 - [ ] on the initial page load the user-menu doesn't flicker (i.e show up and hides immediately


wes-otf (approved):

wes-otf:
nice! was just going to write up an issue for this

frjo (approved):

PR #4005, opened 2024-06-24
Fix editing comment updates the comment visibility

theskumar:
Editing a comment which is for internal team changes it's visibility to applicant + staff

How to reproduce:
- Login as staff
- Add a comment and then edit the same comment
- Refresh the page and see the visibility becomes Applicant + staff

When editing the comment, I didn't see staff getting a UI to update the visibility so not allowing the visibility change in the commit edit api seems to be the right thing to do as well

## Test Steps

 - [ ] login as staff and leave and edit the comment, make sure the visibility is not updated


frjo (approved):
Good catch on this one.

wes-otf (approved):
works great!

PR #4007, opened 2024-06-26
Update node modules and node version

frjo:


frjo:
I have not updated eslint to v9. There is a new config format, some breaking changes etc. so decided to hold off on that for a bit.

theskumar (approved):

PR #4008, opened 2024-06-26
Rename PAF Reviewer Roles to Project Reviewer Roles in admin UI

### LINKED ISSUE #3910: Rename PAF reviewers roles to Project reviewer roles
We have renamed "Project approval form" to "Project form" so this is a followup change.

sandeepsajan0:
Fixes #3910 

image

## Test Steps

 - [ ] …


sandeepsajan0:
@frjo I have updated it in admin UI only where it has to be set. Do you think we should rename the model name or everywhere in the code as well?

frjo (approved):

frjo:
Did a search for "paf" and there are many instances in the code! I think we can let them be. The UI that staff uses is now fixed with this PR.

wes-otf (approved):

PR #4019, opened 2024-06-28
Add invoice date and paid date to invoice and invoices lists and tables

### LINKED ISSUE #3537: [Finance Role] Invoice Table on Project Page

sandeepsajan0:
Fixes #3537 

**Invoice date**:
image

image

image

image

**Paid Date**:

image

## Test Steps

 - [ ] …


frjo:
Migration conflict with another PR on test so we wait to test this for a bit.

wes-otf:
Just had a demo meeting with finance and they liked it! When the pay date is set via the update status modal, where is that info displayed? I couldn't find where that was listed after it was set.

wes-otf:
two other quick notes: 

1. do we need the timestamp on the `Invoice Date` field or could it only be the date (like the `Pay Date` field)? 
2. the `Update status` button formatting breaks when in the staff admin role, while for finance it looks fine. Not sure if this is a result of this PR or something else, though:
Screenshot 2024-08-30 at 10 38 04

frjo:
1. I think we can use `DateField` instead of `DateTimeField` for both `paid_date` and `invoice_date`.
2. @wes I think you was the "applicant" as well if you could see the delete button. According to the code only applicants see the delete button in the table.

https://github.com/HyphaApp/hypha/blob/1372124eb9f5d2a4d203515d7c35522039d3c217/hypha/apply/projects/templates/application_projects/includes/invoices.html#L47-L53

hypha/apply/projects/models/payment.py
```diff
@@ -182,6 +182,8 @@ class Invoice(models.Model):
         null=True,
         verbose_name=_("Invoice amount"),
     )
+    invoice_date = models.DateTimeField(null=True, verbose_name=_("Invoice date"))
+    paid_date = models.DateTimeField(null=True, verbose_name=_("Paid date"))
```
frjo: I think we can use `DateField` for both these fields.

frjo (changes requested):

frjo (approved):

PR #4024, opened 2024-07-02
Update the django-tinymce package, now includes TinyMCE 6.

frjo:
## Test Steps

 - [ ] Go to an application or other form that uses the rich text editor. Confirm that all features needed are present and that the formatting works as expected.


wes-otf:
from what I saw all seems to work fine!

PR #4025, opened 2024-07-02
Fix login redirects

### LINKED ISSUE #4022: Redirect after auth fails when using methods other than email magic link
One of the issues I've seen with #4009 is that staff are logged out each day, then they click a link while logged out (ie. from slack) & login with OAuth, they are redirected to their dashboard rather than the link they clicked. When selecting `Log in with password` the redirect URL params are not preserved, so that also does not allow for redirection.

I know stuff with OAuth is a bit weird so would be curious to hear thoughts here.

wes-otf:
Quick one that fixes #4022! Pretty sure the only issue with the redirects was the `next` variable was never updated to the `redirect_url` variable. Tested locally it works fine between the passwordless/passworded logins but couldn't test with social login. Everything I've seen online indicates this should work fine, though.

## Test Steps

 - [ ] Log out of Hypha
 - [ ] Try to go to a specific Hypha URL (ie. a submission)
 - [ ] Ensure that you are redirected to the URL `https://<hypha url>/auth/?next=<step two path>`
 - [ ] Ensure clicking `Log in with password` redirects to the URL `https://<hypha url>/login/?next=<step two path>`

### DIFF
--- a/hypha/templates/includes/org_login_button.html
+++ b/hypha/templates/includes/org_login_button.html
@@ -1,7 +1,7 @@
 {% load i18n heroicons %}
 <a
     class="button button--secondary button--login mb-4"
-    href="{% url "social:begin" "google-oauth2" %}{% if next %}?next={{ next }}{% endif %}"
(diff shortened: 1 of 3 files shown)


frjo (approved):
Nice!

PR #4026, opened 2024-07-03
Task is not removed for staff while project moved to next stage

### LINKED ISSUE #3946: When Project Form is submitted by someone that isn't the lead, to do notification stays in lead's todo
If the Project Form is submitted by a user that isn't the project lead, the `Project ... is waiting for PAF submission` notification will remain in the lead's To-Do list despite the Project Form being submitted. This was done with no PAF Reviewers set so I wonder if it's related to #3881?

sandeepsajan0:
Fixes #3946 

## Test Steps

 - [ ] …

### DIFF
--- a/hypha/apply/projects/views/project.py
+++ b/hypha/apply/projects/views/project.py
@@ -666,6 +666,20 @@ class SkipPAFApprovalProcessView(DelegatedViewMixin, UpdateView):
         project.status = CONTRACTING
         response = super().form_valid(form)
 
+        # remove PAF submission task for staff group
+        remove_tasks_for_user_group(
+            code=PROJECT_SUBMIT_PAF,
+            user_group=Group.objects.filter(name=STAFF_GROUP_NAME),
+            related_obj=self.object,
+        )
+
+        # remove PAF rejection task for staff if exists
+        remove_tasks_for_user_group(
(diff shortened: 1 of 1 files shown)


wes-otf:
Hey @sandeepsajan0 I don't think the tasks show up in my task list anymore - I just create a new project and make sure the fund has a valid Project Form set & I'm the lead right?

sandeepsajan0:
@wes-otf Sorry for the late response, I missed it somehow.
>I just create a new project and make sure the fund has a valid Project Form set & I'm the lead right?

Project creation creates a task for all staff members(exact) and I think your account might have other roles like staff admin, try/check it with any staff account. Ideally, only staff members will be dealing as project leads and we have some hard/exact role checks in project side because of PAF reviewer roles.

wes-otf:
@sandeepsajan0 this worked for me! my only thoughts are that only staff that are leads on a project should have it appear in their task list. At least with how many projects OTF has I could see it bogging down the tasks lists. @frjo @Techslammer any thoughts?

frjo:
@wes-otf If OTF want to change who gets the task I think a new issue is the place for that.

If this PR removes the task as it should we might approve it.

wes-otf:
that sounds good to me! I'll open another issue

PR #4030, opened 2024-07-04
Use correct perm check for Administration link in user menu. Use same icon for Administration button as in user menu

frjo:


theskumar (approved):
Thank you, I forgot to update the icons in the button, used cog so it looks less busy in the user menu.

PR #4032, opened 2024-07-04
Enable group stream boxes when in draft and edit mode

### LINKED ISSUE #4031: When editing a saved draft with a required, unselected group form type, application is unable to be completed
**Describe the bug**

In an application with a **required** group style form, such as

* Group Start with "Did you run into issues" and Yes/No radios
* Text area "If yes, please explain"
* Group End

If a draft is saved when nothing has been entered, then later editing that draft does not allow those radio buttons to be selected, the textarea always shows up, and the form cannot be completed.

**To Reproduce**

See above.  Bug appears in either labs or funds

**Expected behavior**

The group style forms to work as if the submission had never been saved as draft.

**Priority**

- High - But a bandaid fix has been deployed (and will be commited in a PR)

**Affected roles**

- Applicants

 - OS: All
 - Browser: All

frankduncan:
Fixes #4031

They are disabled when editing a proposal in general, but in the case of a draft, the users sometimes need to be able to edit them in order to finish an application.

This fix may be incomplete, as it does not allow the group forms to have the full functionality of hiding the followup forms when "No" is selected, and showing them when "Yes" is selected.

### DIFF
--- a/hypha/apply/funds/templates/funds/application_base.html
+++ b/hypha/apply/funds/templates/funds/application_base.html
@@ -58,9 +58,9 @@
                     {% for field in form %}
                         {% if field.field %}
                             {% if field.field.multi_input_field %}
-                                {% include "forms/includes/multi_input_field.html" with is_application=True %}
+                                {% include "forms/includes/multi_input_field.html" %}
                             {% else %}
-                                {% include "forms/includes/field.html" with is_application=True %}
+                                {% include "forms/includes/field.html" %}
                             {% endif %}
                         {% else %}
                             {% if field.group_number > 1 %}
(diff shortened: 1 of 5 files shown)


frjo:
@wes-otf Field groups only work when filling in an application form the first time.

I do not remember why we did it this way.

Do you/OTF see a reson for not having them active all the time (create/draft/edit)?

wes-otf:
sorry, this slipped by me. but not that I'm aware of! If anything it would make make sense to me to have them all around

frjo:
@wes-otf This is now on test.

wes-otf:
this works nice! I'll write up another issue but I think if the group fields are closed on submission the fields under them should be cleared - could make things weird for a reviewer if someone was still able to submit answers for a field that was supposed to be hidden but curious what you think @frjo

PR #4040, opened 2024-07-10
Remove tasks from task list

sandeepsajan0:
Fixes Manual task removal issue. 
Two colors trash because of single user task and user group task.
Screenshot 2024-07-25 at 11 20 15 AM

Msg on removing single-user task.
image

Msg on removing group task.
image

## Test Steps

 - [ ] …


sandeepsajan0:
Another way as Fredrik suggested to make the trash visible only by hovering on the row:

https://github.com/user-attachments/assets/0d4eb3fd-4487-41e9-86bb-10d5627f824c

@frjo @theskumar @wes-otf Let me know if it looks better than the screenshots in PR description. Also, we need to think about differentiating between single user task and group task for deletion like right now I use light red trash and dark red trash with different msg and tooltip.

sandeepsajan0:
Updated: It shows up on hover
image

frjo (approved):

hypha/apply/todo/templates/todo/todolist_item.html
```diff
@@ -12,5 +12,8 @@
             </span>
         {% endif %}
     </div>
+    {% if task.user %}
+        <button class="ms-2 invisible group-hover:visible" data-tooltip="Remove" hx-delete="{% url 'todo:delete' pk=task.id %}" hx-target="this" hx-swap="none" hx-confirm='{% trans "Are you sure you want to remove this task? It will remove this task from your task list." %}'>{% heroicon_micro "trash" class="inline me-2" aria_hidden=true %}</button>
```
wes-otf:
```suggestion
        <button class="ms-2 sm:invisible sm:group-hover:visible" data-tooltip="Remove" hx-delete="{% url 'todo:delete' pk=task.id %}" hx-target="this" hx-swap="none" hx-confirm='{% trans "Are you sure you want to remove this task? It will remove this task from your task list." %}'>{% heroicon_micro "trash" class="inline me-2" aria_hidden=true %}</button>
```

Mobile users can't access the remove button if using a touch screen. Also worth making sure it plays well with #4057 and scales for all mobile screens. On the test instance it looked to break around a screen width of 384px

wes-otf:
small fixes around usage on mobile but other than that all looked good in testing! once those are set I think we can merge it. nice work @sandeepsajan0!

sandeepsajan0:
@frjo @theskumar As we discussed in dev meeting use x-mark or checkmark instead of trash icon with both primary and black and white buttons. Here are the possible options we can choose:
1. All blue/primary buttons
image
2. Black and white/ transparent buttons
image

3. x-marks with primary buttons in place of trash icon
image
4. x-mark with transparent buttons(h-24 and w-24)
image

5. check icon with primary buttons
image
6. check mark with transparent buttons
image

frjo:
I lean towards option 4.

wes-otf:
I also like 4! For me it weirdly makes more sense to have the `x` on the right of the `View` button too. That might be my personal preference though

sandeepsajan0:
>For me it weirdly makes more sense to have the x on the right of the View button too.

@wes-otf At first, I put it there but then we thought it is better not to make it directly visible or attention seeker so decided to show the x mark only on hover. And if we show it in right on hover there will be a little UI shift so we decided to show it left of View button.

sandeepsajan0:
@wes-otf I hide the draft or task type for smaller screen(<640px), it looks good now even with 375px.

image

For bigger screens:
image

sandeepsajan0:
@frjo It is ready for re-testing with x-mark and transparent buttons and is usable on mobile.

frjo:
On test now and I thinks it looks and works really well.

One small change, can we have the tool tip say "Remove task" instead of only "Remove"? This makes in clear that the submission, determination etc. is not removed, only the task itself.

Skärmavbild 2024-08-07 kl  13 49 34

PR #4042, opened 2024-07-10
No longer look at WAGTAILUSERS_PASSWORD_ENABLED and ENABLE_PUBLIC_SIGNUP in activation

### LINKED ISSUE #4041: When WAGTAILUSERS_PASSWORD_ENABLED and ENABLE_PUBLIC_SIGNUP are enabled, it's impossible to set a password
**Describe the bug**

When WAGTAILUSERS_PASSWORD_ENABLED and ENABLE_PUBLIC_SIGNUP are set to true, the "Set Password" logic redirects to "user:acount" rather than "user:activate" making it so you can never actually set the password.

This is a vestige of the old registration/password system that was removed in favor of passwordless reigstration.

**To Reproduce**

1. Set WAGTAILUSERS_PASSWORD_ENABLED and ENABLE_PUBLIC_SIGNUP to true in your local.py
2. Create new account
3. Go to account page, click "Set Password"
4. Follow link in email
5. Arrive at account page

**Expected behavior**

Arrive at the password setting page.

**Priority**

- High

**Affected roles**

- New Users

**Additional context**

The "activation" flow may no longer be needed at all.  Just have "Set Password" go to the update password page instead of doing the activation dance in email.  The endpoint is the same.

However, there may be reasons to keep the current flow, so the incoming PR just fixes this specific bug.

frankduncan:
When both of these are set, the "Set Password" in the user account page link always redirects to the user:account page, preventing the user from ever setting a password.

Fixes #4041

### DIFF
--- a/hypha/apply/users/views.py
+++ b/hypha/apply/users/views.py
@@ -286,15 +286,10 @@ class ActivationView(TemplateView):
         if self.valid(user, kwargs.get("token")):
             user.backend = settings.CUSTOM_AUTH_BACKEND
             login(request, user)
-            if settings.WAGTAILUSERS_PASSWORD_ENABLED and settings.ENABLE_PUBLIC_SIGNUP:
-                # In this case, the user entered a password while registering,
-                # and so they shouldn't need to activate a password
-                return redirect("users:account")
-            else:
-                url = reverse("users:activate_password")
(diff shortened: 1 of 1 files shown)


frjo:
Good fix I believe but I would advice against letting staff mess with users passwords, not a good security practice.

frankduncan:
So would I, but it's not up to me :)

PR #4043, opened 2024-07-10
Speed & appearance improvements to the results page

### LINKED ISSUE #3941: Results page fixes & enhancements
The results view (`/apply/submissions/result/`) is something staff has voiced they would like to utilize more to see how funds have preformed and get more overall stats on. There are a few current issues that could use some TLC:

- [ ] **Backend overload** - When initially loading the page or removing all filters, the load speed either crawls or crashes Hypha
- [ ] **Intuitive-ness** - The UI flow could use a bit of a face lift in how data is presented and filters are used (ie. does it make sense to have a submission search functionality?)
- [ ] **Bugs**
    - [ ] Round amount numbers to two decimal places
    - [ ] Filtering by drafts always shows 0 results

wes-otf:
Fixes #3941. This makes some speed improvements to the results page along with fixes to the appearance of some currency values. Some of the biggest improvements come with the initial loading of the `/submissions/result/` with no filters/query params applied, the `.values()` for all submissions were being calculated twice. After some crude benchmarking with very large (OTF scale) datasets, it seems to be ~26% faster. Other improvements were just not loading fields that were irrelevant to the data presented.

Testing of initial loading of `/submissions/result/` was as follows:

**Unoptimized**

| Trial | SQL Time (ms) | SQL Queries |
| ----- | ------------- | ----------- |
| 1     | 3715.86       | 32          | 
| 2     | 4429.34       | 33          |
| 3     | 4449.12       | 33          |
| 4     | 4615.63       | 33          |
| 5     | 4852.89       | 33          |

*Average SQL Time:* 4412.57

**Optimizied**

| Trial | SQL Time (ms) | SQL Queries |
| ----- | ------------- | ----------- |
| 1     | 3006.05       | 32          |
| 2     | 3448.47       | 32          |
| 3     | 3315.12       | 32          | 
| 4     | 3317.39       | 32          |
| 5     | 3137.71       | 32          |

*Average SQL Time:* 3244.95

Hopefully this leads to less crashing on production. If you see anything else that can be optimized let me know!

## Test Steps

 - Ensure the `/submissions/result/` view... 
     - [ ] is snappier and doesn't crash on load
     - [ ] has proper decimal amounts & currency symbols


wes-otf:
I also removed the search functionality for now. It seems like it's hard for the user to know what data is being pulled in that (ie. if you search `wes@test.org` are you getting stats from applications where `wes@test.org` is a applicant, mentioned within the application, etc). This also allowed for deferring of the `search_document` & `search_data` fields which can be huge text fields.

frjo:
@wes-otf I discovered a bug in my original code with the use of `intcomma` so I attempted to fix it.

I guess I started using `intcomma` and later added babel currency conversion. They do not work together. That why we needed to remove commas but that only works for some locales.

Hopefully this fix makes it work for all locales.

wes-otf:
I think this is ready for test if you also agree @frjo

frjo (approved):

frjo:
Filtering on lead and picking Abhishek results in this error https://otf.sentry.io/issues/5681941380/?project=5823154&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=0

Seems to be a amount that is badly formed and babel choke on it.

frjo:
I think I found the issue. If none of the listed submissions have a amount you get the above error.

I added an amount to https://test-apply.hypha.app/apply/submissions/681/ and now it work

* https://test-apply.hypha.app/apply/submissions/result/?lead=1011&query=&archived=

frjo (changes requested):
Allow for the option when none of the filtered submissions have a amount set.

wes-otf:
thanks for testing this! When the there's no fields the total value gets passed to `format_number_as_currency` as `None`, so catching this and swapping it to `0` seems to fix this issue both when there are no applications with the Requested Amount field & when the field exists on applications but it just not populated.

frjo (approved):

PR #4051, opened 2024-07-12
Add a reporting table for projects

### LINKED ISSUE #4050: Add a richer Report table for Projects

Project Reports. The current `reports` table only shows submitted reports.

To stay on top of reports that are due, and not just submitted, a more complete filterable/sortable table of reports is needed.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Staff

bickelj:
Fixes #4050

What a small line for such a large set of changes.  Those supporting changes include:

* Creating a view for the new Reporting Table
* Creating a current_report() method in report_config that is slightly differnt than current_report_due (see inline doc)
* Adding a ProjectQuerySet method that does some intense subquerying to be able to express the current_report in sql so that the status can be filtered against

## Note from the cherry-picker/integrator

I see that the summary tables for reports was removed from `apply/projects` and that the current dropdown in the main navigation header points to `apply/project/reports`. This new table appears under `apply/project/reporting` and is linked to in the original fork in which this change occurred. As for integration here, the question is: do we want to replace the original table or augment off to the side somehow? Where should the link to this new table live?

## Test Steps

URL is `apply/projects/reporting`.

 - [ ] Get a project to the reporting phase, and it should show up in the table
 - [ ] Filter the table by not started and make sure it shows up
 - [ ] Set the project's report due to various kinds, and the report due date in table should change correspondingly
 - [ ] Save a report as a draft, which should move it to in progress on the table
 - [ ] Filter the table by in progress to make sure it shows up!
 - [ ] Complete the report, which should move it to submitted status on the table
 - [ ] Filter the table by submitted to make sure it shows up.

### DIFF
--- a/hypha/apply/projects/tables.py
+++ b/hypha/apply/projects/tables.py
@@ -2,6 +2,7 @@ import json
 import textwrap
 
 import django_tables2 as tables
+from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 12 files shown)


bickelj:
@frjo Is this one ready for merge?

hypha/apply/projects/tables.py
```diff
@@ -246,6 +247,51 @@ def order_end_date(self, qs, desc):
...
+        return format_html("<span>{}</span>", value)
+
+    current_report_submitted_date = tables.Column(
+        verbose_name="Submitted date", accessor="current_report_submitted_date__date"
+    )
+    current_report_due_date = tables.Column(
+        verbose_name="Due Date", accessor="report_config__current_report__end_date"
+    )
+    current_report_last_notified_date = tables.Column(
+        verbose_name="Last Notified",
```
frjo, line 324: Please avoid title case. We have it in several places but are moving away from it.

hypha/apply/projects/urls.py
```diff
@@ -39,6 +40,7 @@
     path("all/", ProjectListView.as_view(), name="all"),
     path("statuses/", get_project_status_counts, name="projects_status_counts"),
     path("invoices/", InvoiceListView.as_view(), name="invoices"),
+    path("reporting/", ReportingView.as_view(), name="reporting"),
```
frjo: Moving this down to "reports/" might look nicer.

frjo (changes requested):
What is the main task that this table makes possible that can not be done with the main project table?

Could we extend the project table instead?

Select2 and fancybox are both packages we are moving away from. A bit hesitant to add new code that uses them.

frankduncan:
This table is specifically for monitoring the status of reports.  When projects get through the main phases and into the reporting phase, our client needed a way to manage, filter, and sort based on the status of those reports.  So this adds a number of columns not on the original project table specifically having to do with whether reports are up to date, etc.

We felt that this table should be separate, just for the user experience.  There would be a number of new columns that would need to be prefixed with "Report <X>" if we added them to the project table, rather than having a view into just the status of the various project reports.  Furthermore, the project table might get a bit busy.

frjo (approved):

wes-otf:
Works well! I might write up an issue to put this as a view under Reports thought so we don't have menu items for `Reports` & `Reporting`, I could see that being mildly confusing to users. Let me know what you think @frjo

frjo:
@wes-otf I agree that the two report items in the menu can be confusing. Please create a PR with your suggestion, sounds better.

PR #4059, opened 2024-07-17
Make background color consistent across the app

wes-otf:
The applicant dashboard (and various other views) used tailwind's `bg-light-grey` which seems inconsistent with the overall app and caused the menu to have a different standout color.

Screenshot 2024-07-17 at 11 39 32

I believe the grey background does add a nice contrast to things so I made that the system default. Open to making it all white, though.

## Test steps
- [ ] Ensure there are no inconsistencies with background color & menu/tab items


frjo:
Nice to get consistent colours!

I think the default should be white.

The light grey do look nice when there is big boxes like on the Dashboard.

wes-otf:
That sounds good. Should I put the `block body_class` tags (with `bg-light-grey`) back in but for all the dashboards or just leave it all white?

frjo:
Maybe make all white first and then we can consider?

wes-otf:
I totally forgot about this for a bit but white is now the default background, let me know if you like this more or if we should switch back to grey @frjo

frjo (approved):

frjo:
@wes-otf Needs a rebase after the the latest additions to main.

frjo (approved):

Issue #4060, opened 2024-07-17
Use pattern matching to invalidate certain `full_name`s


bickelj:
**Is your feature request related to a problem? Please describe.**

Originally this was related to an attack where valid emails were used with invalid full names, generating emails to people that included attacking websites. In addition to the changes in to add rate limiting, this will prevent this attack vector.

**Is your feature request related to an existing functionality? Please describe.**

Full name in profile.

**Describe the solution you'd like**

Allow setting regular expressions that are disallowed in `full_name`s.

**Additional context**

The original attack vector appears mitigated now by not allowing full name to be entered in the form, but in situations where the full name can be entered the feature may still be useful.

**Priority**

- Low priority (annoying, would be nice to not see)

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- Developers
- General public

PR #4061, opened 2024-07-17
Sanitize user `full_name` without roadblocks

bickelj:
Rather than impeding the flow of users setting full name or requiring a new variable to be set by administrators, clean up users' given full names. This means removing all HTML tags and removing colons and slashes. It may be unfriendly to people such as "David-Wynn: Miller" but this is the cost of nerfing spam URLs in full names.

Issue #4060

## Test Steps

 - [ ] Create a user with the full name `<a href="https://myspam.com">Somebody: somewhere</a>`
 - [ ] The resulting user name should be `Somebody somewhere` (no HTML tags, no colons, no slashes).

### DIFF
--- a/hypha/apply/users/forms.py
+++ b/hypha/apply/users/forms.py
@@ -1,5 +1,6 @@
 from re import match
 
+import nh3
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -159,7 +160,10 @@ class CustomUserCreationForm(CustomUserAdminFormBase, UserCreationForm):
 
 
 class ProfileForm(forms.ModelForm):
-    error_messages = {"invalid_full_name": _("Full Name is Invalid")}
+    error_messages = {
+        "invalid_full_name": _("Full name is invalid"),
+        "no_html": _("Full name may not contain HTML"),
+    }
 
     class Meta:
         model = User
@@ -204,6 +208,8 @@ class ProfileForm(forms.ModelForm):
 
     def clean_full_name(self):
         full_name = self.cleaned_data["full_name"]
+        if nh3.is_html(full_name):
+            raise forms.ValidationError(self.error_messages["no_html"])
         for regex in settings.INVALID_FULL_NAME_REGEXES:
             if match(regex, full_name):
                 raise forms.ValidationError(self.error_messages["invalid_full_name"])
--- a/hypha/apply/users/forms.py
+++ b/hypha/apply/users/forms.py
@@ -1,4 +1,3 @@
-import nh3
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -9,6 +8,7 @@ from django_select2.forms import Select2Widget
 from wagtail.users.forms import UserCreationForm, UserEditForm
 
 from .models import AuthSettings, GroupDesc
+from .utils import strip_html_and_nerf_urls
 
 User = get_user_model()
 
@@ -200,13 +200,7 @@ class ProfileForm(forms.ModelForm):
         return email
 
     def clean_full_name(self):
-        full_name = self.cleaned_data["full_name"]
-        # Remove all HTML tags. This prohibits HTML without creating hurdles.
-        cleaned_full_name = nh3.clean(full_name, tags=set())
-        # Remove all colons and slashes to nerf URLs regardless of HTML tags.
-        cleaned_full_name = cleaned_full_name.replace(":", "")
-        cleaned_full_name = cleaned_full_name.replace("/", "")
-        return cleaned_full_name
+        return strip_html_and_nerf_urls(self.cleaned_data["full_name"])
 
 
 class BecomeUserForm(forms.Form):
--- a/hypha/apply/users/forms.py
+++ b/hypha/apply/users/forms.py
@@ -1,3 +1,5 @@
+from re import match
+
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -157,6 +159,8 @@ class CustomUserCreationForm(CustomUserAdminFormBase, UserCreationForm):
 
 
 class ProfileForm(forms.ModelForm):
+    error_messages = {"invalid_full_name": _("Full Name is Invalid")}
+
     class Meta:
         model = User
         fields = ["full_name", "email", "slack"]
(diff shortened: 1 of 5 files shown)


frjo:
@bickelj Can we not run username through `nh3.clean()` instead?

bickelj:
@frjo Perhaps, but from what I see at https://nh3.readthedocs.io/en/stable/ and https://github.com/rust-ammonia/ammonia the library appears to be for "HTML sanitization" where the output string is expected to be HTML. In this case, I think we are looking for plain text both on input and on output. `nh3` might be useful additionally but the goal here is to disallow URLs completely.

Edit: I see `nh3.is_html` which we could use to reject any HTML tags, is that what you had in mind? But that would still allow non-HTML URLs to come through, correct?

bickelj:
ffb57c89e09d10c4e7d7871b6e6eed3f2bbf4d34 uses `nh3` to prohibit HTML content in a Full name.

frjo:
@bickelj It can clean out all html.

```
>>> import nh3
>>> nh3.clean("<b><a href='https://example.com'>Hello</a></b>", tags=set())
'Hello'
```

But that leaves all url schemas intact, the original problem needed to be solved.

Could we add a hardcoded feature that removes every ":" and "/"? That would break all links and they should not be in any real usernames in any case.

I'm open to doing this silently. If the user only adds tags etc. they end up with a empty username. No error message or anything. Will only effect people with bad intensions.

bickelj:
@frjo Ah, right. That makes sense. Let me try that.

bickelj:
@frjo Updated. It looks like it's in two places, now, though. I guess an API should be able to create users as well which might omit the form but would touch the model. Perhaps the sanitation should only be in the model. What do you think?

Edit: and obviously needs fixed tests now.

bickelj:
@frjo I found this test that clearly intends to allow weird names including HTML and such:
```python
    def test_doesnt_mess_with_name(self):
        full_name = "I have; <a> wei'rd name"
        self.submit_form(name=full_name)
        submission = ApplicationSubmission.objects.first()
        self.assertEqual(submission.user.full_name, full_name)
```

I suppose we should change the test to indicate the opposite policy.

bickelj:
The plot thickens: that test passes. Ah. That is on the application submission, not on the user profile, perhaps.

hypha/apply/users/forms.py
```diff
@@ -198,6 +199,15 @@ def clean_email(self):
...
         return email
 
+    def clean_full_name(self):
+        full_name = self.cleaned_data["full_name"]
+        # Remove all HTML tags. This prohibits HTML without creating hurdles.
+        cleaned_full_name = nh3.clean(full_name, tags=set())
+        # Remove all colons and slashes to nerf URLs regardless of HTML tags.
+        cleaned_full_name = cleaned_full_name.replace(":", "")
+        cleaned_full_name = cleaned_full_name.replace("/", "")
+        return cleaned_full_name
```
frjo: Looks good but can we break out this into a util function? Then we have this logic only in one place and can reuse it where needed.

hypha/settings/base.py
```diff
@@ -155,6 +155,7 @@
 # Require an applicant to view their rendered application before submitting
 SUBMISSION_PREVIEW_REQUIRED = env.bool("SUBMISSION_PREVIEW_REQUIRED", True)
 
+
```
frjo: Can we clear out this please.

frjo (changes requested):

hypha/apply/users/forms.py
bickelj: Good call. I left the function in the same file because this is its one use. If it needs to be moved elsewhere in future due to use outside this file or speculatively now that is also fine.

hypha/settings/base.py
bickelj: I did it in a later commit because I assume these get squashed. If I need to amend, let me know.

hypha/apply/users/forms.py
```diff
@@ -156,6 +157,15 @@ def __init__(self, register_view=False, request=None, *args, **kwargs):
         )
 
 
+def strip_html_and_nerf_urls(value: str):
+    # Remove all HTML tags. This prohibits HTML without creating hurdles.
+    cleaned_value = nh3.clean(value, tags=set())
+    # Remove all colons and slashes to nerf URLs regardless of HTML tags.
+    cleaned_value = cleaned_value.replace(":", "")
+    cleaned_value = cleaned_value.replace("/", "")
+    return cleaned_value
```
frjo: Can we move this to `hypha/apply/users/utils.py`?

hypha/apply/users/models.py
```diff
@@ -170,6 +171,13 @@ def get_or_create_and_notify(
             elif not user.is_active:
                 raise IntegrityError("Found an inactive account")
         else:
+            if kwargs.get("full_name", False):
+                # Remove all HTML tags. This prohibits HTML without creating hurdles.
+                cleaned_full_name = nh3.clean(kwargs["full_name"], tags=set())
+                # Remove all colons and slashes to nerf URLs regardless of HTML tags.
+                cleaned_full_name = cleaned_full_name.replace(":", "")
+                cleaned_full_name = cleaned_full_name.replace("/", "")
+                kwargs["full_name"] = cleaned_full_name
```
frjo: Please use `strip_html_and_nerf_urls()` here as well.

frjo (changes requested):

hypha/apply/users/models.py
bickelj: Doh! Done...

hypha/apply/users/utils.py
```diff
@@ -1,5 +1,6 @@
 import string
 
+import nh3
```
bickelj, line 3: My IDE complains that we do not have `nh3` in the project requirements. We have `django-nh3`. Are we importing a transitive dep here?

Yes:
```
$ grep nh3 requirements.txt 
django-nh3==0.1.1
$ pip show nh3
Name: nh3
Version: 0.2.18
...
Required-by: django-nh3
```

Maybe if we import it we should add it.

bickelj: Added in 3ef7a738daee00c0555c280734a09a1139de8295

frjo (approved):

wes-otf:
works well, thanks @bickelj!

PR #4064, opened 2024-07-19
Implemented `Remember me` for passwordless login

### LINKED ISSUE #4009: Implement `Remember me` for OAuth/Magic Link
Staff have asked if the "remember me" option could be implemented for OAuth as they have to re-login daily. This has had some small residual effects like staff losing a typed up determination that they tried to submit 12 hours later (which should have been drafted but nonetheless). More of an inconvenience than a mission critical addition.

wes-otf:
Fixes #4009. Implements a session extension for the magic link by utilizing an extra `remember-me` URL param. Also adds a configuration option to extend the session age to `SESSION_COOKIE_AGE_LONG` when OAuth is utilized for the login.

## Test Steps

**Testing Passwordless `Remember me`**
 - [ ] Login to Hypha utilizing the passwordless flow and do *not* select `Remember me`
 - [ ] Ensure the session age is set to `SESSION_COOKIE_AGE` (12 hours by default)
     - [ ] If user testing, I can assist with where to find cookie expiration times
 - [ ] Login to Hypha utilizing the passwordless flow and select `Remember me`
 - [ ] Ensure the session age is set to `SESSION_COOKIE_AGE` (2 weeks by default)

**Testing OAuth session extension**
 - [ ] Ensure the config var `SOCIAL_AUTH_USE_LONG_SESSION` is set to true
 - [ ] Login with OAuth/Org email
 - [ ] Ensure the session age is set to `SESSION_COOKIE_AGE` (2 weeks by default)


wes-otf:
had to override the OAuth `social:complete` view which was a little more hack-y than I hoped but that's where the session age gets written. only includes the URL override if the config is set for the session extension

originally had this logic run in the pipeline which was a lot cleaning but it'd get overwritten to the short age on `social:complete`

wes-otf:
@frjo is this ready to go to test?

hypha/settings/base.py
```diff
@@ -358,6 +358,10 @@
 SOCIAL_AUTH_GOOGLE_OAUTH2_KEY = env.str("SOCIAL_AUTH_GOOGLE_OAUTH2_KEY", "")
 SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET = env.str("SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET", "")
 
+# Use `SESSION_COOKIE_AGE_LONG` when a user authenticates with OAuth.
+# Will override `SOCIAL_AUTH_SESSION_EXPIRATION`
+SOCIAL_AUTH_USE_LONG_SESSION = env.bool("SOCIAL_AUTH_USE_LONG_SESSION", False)
```
frjo: Could we make it so Hypha always use the long session time for oauth logins instead?

We could add  text to `SESSION_COOKIE_AGE_LONG` explaining that it is also used for oauth logins.

wes-otf: that works, should be set now

PR #3932, opened 2024-05-20
Rebuild communication and activity feed + assign a comment as task to user

theskumar:
Fixes #3048 
Fixes #4029

- [x] Introduce machine generated avatars based on the user's name (in-future allow for gravatar or similar service)
- [x] Color Code the user avatar based on their roles 🆕 
- [x] Render the staff roles as label (#3916)
- [x] Reduce the size of the rendered comment box.
- [x]  Merge activities with communication in a single timeline. 🆕 
- [x] Use htmx for comment edit function, the js version seems to break some functionality.
- [x] Use different icons for different activities where possible. 
- [x] Render paginated comment and activities 🆕 
- [x] Cleanup/remove JSON APIs related to comments
- [x] Use the same logic for rendering of comments on the projects module
- [x] ~Async comment add (won't be done in this PR)~
- [x] Deep-link for individual comments
- [x] fix testcases 

## Work in progress screenshots

Screenshot 2024-09-22 at 11  26 34@2x

Screenshot 2024-09-23 at 6  56 56@2x

## Test Steps

 - [ ] Checkout the activity feed and communication

PR #4065, opened 2024-07-21
Implement 'Add manual task' feature

### LINKED ISSUE #3872: Create a "Add task" feature
Staff would like to have a way to alert other staff members (or themself) about a submission with a short message. An idea is to add this alert as a task.

Some ideas around functionality.

- [ ] Add task funktion with short message on each submission. Staff members can add a task to themself or to other staff. The task will link to the submission.
- [ ] The tasks gets listed on the users task list with a way to manually mark them as done (i.e. remove them).
- [ ] Notifications should be sent when a task is added (unless a person adds it to their own list).
- [ ] Add task in the same way to projects as well? Invoices? Contracts?
- [ ] Should a generic "Add task" be added on the task list itself?

* What should we name the feature?
* Icon and label for this on the task list?

sandeepsajan0:
Fixes #3872 
It creates a task for a user via communication, if a comment is assigned to a user(staff). If assigned_to is empty then it will just be a comment, same like before.

Comment form:
image

Comment listing:
image

Task list: 
image

## Test Steps

 - [ ] …


sandeepsajan0:
@frjo @theskumar @wes-otf  Is it looking good?

image

frjo:
@sandeepsajan0 Looks really good!

sandeepsajan0:
The listing is also updated like this. "Assigned as task to" => "<b>Assigned to</b>" 
image

frjo:
@sandeepsajan0 This can now be rebased on top of main.

sandeepsajan0:
@frjo @theskumar @wes-otf Does it look better? Instead of having label and input in the same line now it is the same as visibility.
image

sandeepsajan0:
@frjo The latest Ui is also ready for testing.

frjo:
Deployed to test.

hypha/apply/activity/templates/activity/include/listing_base.html
```diff
@@ -20,6 +20,9 @@
                     <span class="js-last-edited text-fg-muted text-sm" data-tippy-content="{{ activity.edited|date:"SHORT_DATETIME_FORMAT" }}">{% trans "edited" %}</span>
                 {% endif %}
             </p>
+            {% if request.user.is_apply_staff %}
+                <p hx-get="{% url 'activity:assigned' pk=activity.id %}" hx-trigger="revealed" hx-target="this" class="feed__meta-item"></p>
+            {% endif %}
```
theskumar, line 26: This will cause a lot of http request, would be better to better to hx-multi-swap, or render them during initial render.

hypha/apply/todo/models.py
```diff
@@ -28,6 +28,8 @@ class Task(models.Model):
     )
     related_object_id = models.PositiveIntegerField(blank=True, null=True)
     related_object = GenericForeignKey("related_content_type", "related_object_id")
+    # for manually added tasks
+    message = models.CharField(max_length=250, null=True, blank=True)
```
theskumar, line 32: Pull the message from the related item (i.e. the comment/activity object) itself? Maybe I'm missing something but it would better not to de-duplicate the message, as they can be edited.

sandeepsajan0: I think comment edited case should be handled but a separate task message can provide us flexibility in changing the language of the task message, it might be helpful later or in another kind of tasks. And duplicity in terms of storage shouldn't be an issue as these tasks are temporary and will be removed regularly.

hypha/apply/activity/templates/activity/include/listing_base.html
sandeepsajan0: It was a bit complex for the initial render(might effect load time), and I thought we would use htmx soon for a communication tab similar to activity feed then we can include it in that?
I'll check the hx-multi-swap and other methods if work well with it.

sandeepsajan0: Subquery with outerref worked better in this case but I am not sure about its performance. I'll push that logic soon.

frjo (approved):

wes-otf (approved):

frjo:
This PR will be merged in to https://github.com/HyphaApp/hypha/pull/3932 since that PR redesigns the communication tab.

theskumar:
closing in favor of #3932 . This will be included in that PR

PR #4078, opened 2024-08-08
Fix submit contract document category permission with required remaining categories instead of remaining categories

sandeepsajan0:
Fix bug for submit contract document by applicant.

Right now, there is an option in wagtail to set a contract document category optional/required but in Ui for applicants we are just checking for remaining contract document categories for submit contract document permissions while it should be checking for required only because others are optional.

## Test Steps

 - [ ] Look for the project in the contracting state where it is waiting for contract document submission from the applicant.
 - [ ] Look in wagtail Project>contract document categories and mark any of them optional.
 - [ ] Check if applicant can see the submit contract button even without uploading the optional document.


theskumar (approved):
Nice!

frjo (approved):

wes-otf:
worked well in testing!

PR #4080, opened 2024-08-08
New submission table view

theskumar:
Changes

- [x] Replaces the old view with new table view, that is powered by the filtering and rendering logic build for the new list view.
- [x] removes round detail view and round export, reuse the submission all page.
- [x] Add rounds/labs menu to primary nav, makes it easy to list and view status of different rounds, for staff.
- [x] Update my flags view on the dashboard page.
- [x] Implement `SUBMISSION_LIST_EXCLUDE_FIELDS` on both list and table view (in progress)
- [x] Expose old table view at an accessible url

## Test Steps

 - General regression of the submission view for different roles.


frjo:
@theskumar  This can now be rebased on top of main.

frjo (approved):

theskumar:
> @theskumar This can now be rebased on top of main.


@frjo done.

frjo:
@theskumar Test failing, looks like a url issue.

Really looking forward to have staff test this one!

theskumar:
I've updated the tests @frjo

theskumar:
SUBMISSIONS_TABLE_EXCLUDED_FIELDS has be implement with all supported exclusions that were present in the settings. 

> Possible values are: fund, round, status, lead, reviewers, screening_statuses, category_options, meta_terms, organization_name

wes-otf:
passed this off to the user testing group but in my testing I noticed small gaps on the right hand side when the table is used
Screenshot 2024-08-28 at 10 51 52
Screenshot 2024-08-28 at 10 52 16

frjo:
@wes-otf The gap is for the scrollbar due to "overflow-y: scroll;" on the div that wraps the table. I think that should be "overflow-x: scroll;". Deploying a fix to test now.

wes-otf:
ahh makes sense! that seemed to work, test instance looks great!

wes-otf:
@frjo @theskumar here's some initial feedback from staff:

> 1. Thank you so much for working on making this cleaner for all of us! We appreciate it :smile:
> 1. I find the difference between the "Sort" function and the "Updated" function to be a bit unclear.
> 1. I would really love an option to filter out draft apps - I think that would be more helpful than some of the other dropdowns. For example, we don't really need "Terms" at this time. We don't ever use that but that could be subject to change.
> 1. I like the colour associated with each phase but it could be a bit more streamlined. For example, "RFI" and "Dismissed" both being red is a bit deceptive.

frjo:
@wes-otf 

1. Great to hear that we are on track!
2. "Update" filter submissions based on last updated date and "Sort", well sort. Can we name them differently to make it clearer?
3. Maybe add a "Drafts" toggle like the one we have for "Archived". Then staff can show hide draft easily. I also think we could use `SUBMISSIONS_TABLE_EXCLUDED_FIELDS` to hide filters as well as columns.
4. The colours can easily be adjusted. They are different already but not enough.

frjo:
@wes-otf Does this mean that there are no big issues with the new table? Only minor adjustments like above?

wes-otf:
@frjo yeah I agree with all your suggestions! for sort maybe we can use a different icon entirely, like  instead of ? I also like the suggestion for filtering drafts.

This is the only feedback I've gotten, expecting a little more but not expecting it to be huge - should all be minor like this.

frjo:
@wes-otf I have made a stab at implementing the issues above and deployed them to test now. Please review at https://test-apply.hypha.app/apply/submissions/all/.

Sort has a different icon, there is a draft toggle and status colours are updated.

I think these are all good changes that improve the view.

Make sure staff knows how much the dev team appreciate quality feedback like this!

theskumar:
Great and positive feedback! Adjusted the icon size a bit, the up/down was looking too big compared to chevron-down. All icons are now 16px instead of 20px earlier.

PR #4083, opened 2024-08-09
Invoice status updating improvements

### LINKED ISSUE #3208: Improve usability of picklist on project page
### Description
When there are less options(<4), the interface should show the list of options upfront instead of a dropdown.

image

wes-otf:
Fixes #3208. Implemented an improvement for the picklist where radio buttons will be shown when there are less than 4 options. Otherwise the prompt will be like it was before.

This PR also implements HTMX for the invoice updating prompt, allowing invoices in table to be updated w/o refreshing the pages

## Test Steps

 - [ ] …

### DIFF
--- a/hypha/apply/activity/adapters/utils.py
+++ b/hypha/apply/activity/adapters/utils.py
@@ -133,4 +133,4 @@ def get_users_for_groups(groups, user_queryset=None, exact_match=False):
             user_queryset = user_queryset.filter(groups__name=groups.pop().name)
         return get_users_for_groups(groups, user_queryset=user_queryset)
     else:
-        return user_queryset
+        return user_queryset if user_queryset is not None else set()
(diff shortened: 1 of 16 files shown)


theskumar (commented):
@wes-otf instead of rendering each of the one by one during initial page load, either you can load the invoices from server as it was getting loaded before and only when the status changed update the row data. 

Or if you want to save up some queries and load the invoices section async, then I would suggest load all the rows at once, and then when a particualar invoices is update either re-render all the invoices row or just trigger fetch and update of a single row.

frjo:
@wes-otf You can update invoice status on the invoice detail page as well. I think that function also need to be updated to use htmx, getting some errors when using it now.

frjo:
@wes-otf When you are on the invoice detail template can you make it so the show/hide for activities only show when there are more than 1 activity?

wes-otf:
thanks for the catch @frjo, totally missed that. I think all should work now - didn't update the batch prompt because that seems a little out of scope but that'll probably be in line next

frjo (approved):
Nice improvement!

frjo:
Some merge conflicts in `invoices.html` with other PRs on test so waiting a bit.

PR #4095, opened 2024-08-15
Assign tasks to project leads instead of staff

### LINKED ISSUE #4081: Only project leads should have project tasks added to their task list
Pertaining to the comment [here](https://github.com/HyphaApp/hypha/pull/4026#issuecomment-2274301760), when a new project is created only the staff lead of that project should get the task, rather than all staff.

sandeepsajan0:
Fixes #4081 

## Test Steps

 - [ ] …


wes-otf:
works great!

PR #4102, opened 2024-08-19
Update htmx to 2.x

### LINKED ISSUE #4094: Reviews block on submission detail view gets stuck on loading if not in viewport
Caught by a staff member - with the new HTMX implementation sometimes the reviews card gets stuck on the loading animation and will never populate reviews
Screenshot 2024-08-14 at 14 54 39

sometimes refreshing the page will fix this

wes-otf:
Fixes #4094. ~~For some reason, when combining other triggers with `revealed` the revealed trigger itself breaks while others remain fine. Swapping for the `intersect once` trigger resolved this but I'm not exactly sure why. Also not sure if moving forward for consistency we should use `intersect` instead of `revealed`. @theskumar any thoughts?~~

Updated htmx from v1.9.12 -> v2.0.2 to fix the usage of multiple tags breaking `revealed`. Tested locally and there doesn't seem to be any major breaking changes.

## Test Steps

 - [ ] Ensure all other htmx functionality works as expected.
 - [ ] Ensure when reviews block is out of viewport then is scrolled in the block loads.


frjo:
Seems to be a htmx bug that is resolved in v2.0.2, we are running v1.9.12 at the moment.

https://github.com/bigskysoftware/htmx/issues/2233

frjo:
For me dialogs etc. open quite a bit slower with htmx 2.0.2.

@wes-otf @theskumar Are you noticing this as well or is it something with my local setup?

The "htmx-ext-multi-swap.min.js" extensions is complaining that you should not run version 1.x with htmx 2.x.

wes-otf:
@frjo which dialogs? I just tested on partners/reviewers updating and they still seemed pretty snappy to me.

I just noticed the `make copy-npm-scripts` command which I didn't know existed, I just copy & pasted htmx.min.js before so htmx-ext-multi-swap.min.js should now be updated

frjo:
@wes-otf @theskumar All of the dialogs, the flag buttons etc. A small but noticeable lag on every htmx feature.

frjo:
@wes-otf @theskumar My bad, v2.0.2 seems to work just fine after I restart everything.

theskumar:
Just checked and it works fine for me!

frjo (approved):

PR #4103, opened 2024-08-20
Fix js error when related sidebar is not displayed.

frjo:
When there is no related submissions the script generates an error. This PR fixes that.

Also removes one old unused js file.


theskumar (approved):
Looks good to me and works fine!

PR #4106, opened 2024-08-21
Added links & icons to wagtail form view `Used By` columns

wes-otf:
Adds icons & edit links to the Wagtail forms views (Application Forms, Determination Forms, Review Forms, etc.) "Used by" column, making it easier to differentiate what's used where at a glance . Sometimes rounds & funds take the same name so it's not possible to know which is which.

Also added an enum class to utils to allow for consistency across the interface with icon usage rather than hardcoding the same values over and over. I would've used `StrEnum` but it felt strange to have repeat values for different enums. Open to better solutions, though.

## Screenshots

*before*
Screen Shot 2024-08-21 at 15 06 02

*after*
Screen Shot 2024-08-21 at 15 03 08

## Test Steps

 - [ ] Ensure funds, rounds & labs that use a form are properly linked & have their respective icons in the "Used By" column of the form views

### DIFF
--- a/hypha/apply/categories/admin.py
+++ b/hypha/apply/categories/admin.py
@@ -1,6 +1,8 @@
 from django.urls import re_path
 from wagtail.contrib.modeladmin.options import ModelAdmin
 
+from hypha.apply.utils.admin import AdminIcon
+
 from .admin_helpers import MetaTermButtonHelper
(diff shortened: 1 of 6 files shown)


frjo:
Good feature!

frjo (approved):

theskumar (approved):

PR #4109, opened 2024-08-25
Update makefile, auto-help and better serve

theskumar:
- The current make -j parallelization doesn't kill the remaining process when one of them errors out, the causes confusion if let's css/js is not building.

- This PR also refractors the help target to use comment based generator which is more maintainable.

- Update the node scripts to not require `collectstatic` during development, the runserver -> static server can find all the all files from packages and serve them on demand.
- `make serve` now also installs python & node packages and builds the static_src if they are outdated.


frjo (approved):

theskumar:
@frjo thanks for the quick review, I've update the PR with a bigger refractor to 

- Not require collect static during development, the runserver -> static server can find all the all files from packages and serve them on demand. 
- `make serve` now also installs python & node package and builds the static_src if they are outdated.

Makefile
```diff
@@ -59,55 +38,61 @@ lint:
...
+
+.cache/py-packages: requirements-dev.txt requirements-docs.txt  ## Install python packages
+	@mkdir -p $$(dirname $@)
+	pip install -r requirements-dev.txt -r requirements-docs.txt
+	@touch $@
+
+
+.cache/npm-packages: package.json  	## Install node packages and copy javascript files to vendor directory
+	@mkdir -p $$(dirname $@)
+	NODE_ENV=developement npm install
```
frjo: developement -> development

Makefile
```diff
@@ -1,55 +1,34 @@
...
+	@grep -E '^[0-9a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "  \033[0;36m%-12s\033[m %s\n", $$1, $$2}'
+	@echo ""
+
 
 .PHONY: serve
-serve:
-	$(MAKE) -j3 watch serve-django serve-docs
+serve: .cache/tandem .cache/py-packages .cache/dev-build  ## Run Django and docs preview server, also watch and compile frontend changes
+	@.cache/tandem \
+		'python manage.py runserver 0.0.0.0:$(DJANGO_PORT) --settings=$(DJANGO_SETTINGS_MODULE)' \
```
frjo: Switch to `runserver_plus`? We already have "Werkzeug" installed. Nice error handling!

Should we also switch to `127.0.0.1` instead of `0.0.0.0`? Do devs ever access the server from other devices on the LAN?

frjo (changes requested):

Makefile
theskumar: We can switch to 127.0.0.1, generally it's useful to test on phone, and I had that as default. I rarely use it, so I think it's safe to switch.

theskumar: For runserver_plus I think we can.

Makefile
theskumar: Done.

Makefile
theskumar: Updated both

frjo (approved):

frjo:
@theskumar I merged this is but added a small change as well. Please test to make sure it works.

I added a address to the port for runserver_plus. This makes the test server link in the terminal clickeble, I find this convenient.

`http://127.0.0.1:9001/` instead of `http://[127.0.0.1]:9001/`

frjo:
Also added a way to pick up address set as env var, `DJANGO_SERVE_ADDRESS`.

PR #4114, opened 2024-08-28
Improve cloned forms naming

### LINKED ISSUE #3546: Improvement to naming of cloned forms
When a form is cloned in Hypha it automatically get a new name. Something like "Original name (name of round)". A cloned form can be it turn cloned etc. so the naming becomes quite convoluted.

Ideas it improve the algoritm that does the naming so the forms always get names that make sense to the users?

wes-otf:
Fixes #3546. Improves the naming of cloned forms by changing the format to `<original form name> (Copied on <%Y-%m-%d %H:%M:%S.%f>)`, ie. cloning `Test Form` would result in a duplicated form with the name `Test Form (Copied on 2024-08-27 17:34:32.004)`. If a form name with a timestamp already exists on a form it will be subbed via RegEx matching.

## Test Steps

 - [ ] Ensure when cloning a form (either manually or via the creation of a new round) that the form's name gets `(Copied on <%Y-%m-%d %H:%M:%S.%f>)` appended to it
 - [ ] Ensure if a timestamp string already exists on a cloned form, it is replaced with an updated timestamp

### DIFF
--- a/hypha/apply/funds/utils.py
+++ b/hypha/apply/funds/utils.py
@@ -164,7 +164,7 @@ def is_filter_empty(filter: filters.FilterSet) -> bool:
     return any(reduce(iconcat, [param[1] for param in query.lists()], []))
 
 
-def get_copied_form_name(orignal_form_name: str) -> str:
+def get_copied_form_name(original_form_name: str) -> str:
     """Create the name of the form to be copied
 
     By default, takes the orginal forms name and adds `(Copied on %Y-%m-%d %H:%M:%S.%f)`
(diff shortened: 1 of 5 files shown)


frjo (approved):

theskumar (approved):

hypha/apply/funds/utils.py
```diff
@@ -159,3 +162,29 @@ def is_filter_empty(filter: filters.FilterSet) -> bool:
 
     # Flatten the QueryDict values in filter.data to a single list, check for validity with any()
     return any(reduce(iconcat, [param[1] for param in query.lists()], []))
+
+
+def get_copied_form_name(orignal_form_name: str) -> str:
```
theskumar: can i suggest adding some quick unit test for this util. esp. the regex pattens I don't trust them too much.

hypha/apply/funds/utils.py
```diff
@@ -159,3 +162,29 @@ def is_filter_empty(filter: filters.FilterSet) -> bool:
...
+    """
+    copy_str = _("Copied on {copy_time}")
+    copy_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S.%f")[:-4]
+    date_reg = r"(\d{2,4}-?){3} (\d{2}(:|.)?){4}"  # match the strftime pattern of %Y-%m-%d %H:%M:%S.%f
+    name_reg = r" \({}\)$".format(
+        re.escape(copy_str).replace(r"\{copy_time\}", date_reg)
+    )
+
+    # If a copied timestamp already exists, remove it
+    new_name = re.sub(name_reg, "", orignal_form_name)
```
theskumar: typo. it should be `original_form_name`

theskumar (dismissed):

hypha/apply/funds/utils.py
wes-otf: yes totally!

wes-otf:
@theskumar unit tests added in the last commit - was trying to find a way to patch/mock `django.utils.translation.gettext` to test the usage of the regex w/ translated strings but was struggling to get it to work. let me know if you have any solutions for that

PR #4122, opened 2024-09-03
Remove all Finance 2 features

frjo:


hypha/apply/api/v1/permissions.py
```diff
@@ -20,18 +20,10 @@ def has_object_permission(self, request, view, obj):
 
 class IsFinance1User(permissions.BasePermission):
```
theskumar, line 21: This func should rename as well?

frjo: All finance1 stuff needs updating, planing that for a separate PR.

wes-otf (approved):

PR #4126, opened 2024-09-05
Fix 'Reminder Created' duplicate message, restrict duplicate toast messages

### LINKED ISSUE #4082: Reminder alert showed twice
The reminder alert showed twice when crating new reminder on submission detail page.
 
Screenshot 2024-08-08 at 10 46 29

sandeepsajan0:
Fixes #4082 

## Test Steps

 - [ ] Create a reminder and check there should be one Reminder Created toast with info icon only.

### DIFF
--- a/hypha/apply/funds/views.py
+++ b/hypha/apply/funds/views.py
@@ -1105,9 +1105,7 @@ class ReminderCreateView(View):
             return HttpResponse(
                 status=204,
                 headers={
-                    "HX-Trigger": json.dumps(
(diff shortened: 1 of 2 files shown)


theskumar:
Should the duplicate message dispatch be removed, instead of handling it after it's sent. 

The duplicate seems to be because the django messages framework also creates a entry, and it's that's is happening only one should send the message, either the view function or django messages framework.

frjo (approved):

theskumar:
@frjo I think this one is good merge now?

PR #4127, opened 2024-09-10
Update homepage to not used wagtail pages

### LINKED ISSUE #4116: Rebuild homepage to not be a wagtail page

theskumar:
Use standard django views to show hompeage instead of piggy backing on
the wagtail. This allows to remove the "Pages" menu from the wagtail
admin and have the homepage settings be at a single "System Settings"

- update the hompeage to use Settings > System Settings
- add data migration to move the properties from ApplyHompeage model to
  SystemSettings
- hide "Pages" menu from wagtail
- Refractor the homepage template

Fixes: https://github.com/HyphaApp/hypha/issues/4116


hypha/home/models.py
```diff
@@ -1,28 +1,7 @@
...
-
-from hypha.apply.funds.models import ApplicationBase, LabBase
 
 
 class ApplyHomePage(Page):
     # Only allow creating HomePages at the root level
     parent_page_types = ["wagtailcore.Page"]
     subpage_types = ["funds.FundType", "funds.LabType", "funds.RequestForPartners"]
-
-    strapline = models.CharField(blank=True, max_length=255)
```
frjo, line 14: Removing this line makes the code in `copy_homepage_title_and_strapline` not run since the model no longer have a strapline.

Would be really good to get this working making the deployment of this release seamless.

Maybe we can remove the strapline field in a separate release?

frjo (changes requested):

hypha/home/models.py
theskumar: Right, `hasattr` will always be false. I think we can remove it later after couple of releases.

theskumar: Updated the code so that it keeps the strapline on the ApplyHome model.

frjo (approved):

wes-otf (approved):

Issue #4129, opened 2024-09-11
Link Projects more tightly to Submissions


bickelj:
**Is your feature request related to a problem? Please describe.**

When using the Projects capability, the seemingly hard separation of Projects from Submissions can be confusing.

**Is your feature request related to an existing functionality? Please describe.**

Projects feel disconnected from their respective Submissions.

**Describe the solution you'd like**

At least some users have staff and reviewers continuing to monitor the grant all the way from start to finish. In this case, there is not a hard distinction between "Submission" and "Project" (other than the obvious "a Project happens when a Submission is accepted"). So these users refer to Submissions by their integer ID for convenience. The Project does not have that same ID so there is friction with IDs when going between Project and Submission. There are some other links that could be added too, some of which is captured in issue #4117.

We and they would like the same ID, especially given that there is already a one-to-one relationship from Project back to Submission (not vice versa). Unless there are specific plans for multiple projects derived from a single grant, there should be no issue with re-using the same ID for the Project as the Submission.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Staff
- Reviewers

frjo:
Divided up in to several issues.

Issue #4117, opened 2024-08-29
Add "Project" and "Submission" tabs on the submission and project detail pages


frjo:
A quick and obvious way to switch between related submission/project. Makes them feel more contacted.

- [ ] Only show Project tab when there is a project.
- [ ] The Activity tab can be removed.
- [ ] Remove Submission simplified view, link to normal submission detail view instead.

PR #3944, opened 2024-05-28
Update public facing id of project to be the id of the submission

theskumar:
Update the project urls to use it's submission id instead of the project's db id

The reason for the update is that there are two main entities in the system, one submission and other project.
The ids of the submission is linkable in the communiction and also used for referencing at different places. 

Considering a project is an extension of application submission, if we have the project publically available id same as that of it's submission it always to link and reference the project by the same id. 

## Test Steps

 - [ ] General regression test and make sure all the urls in the project section resolve and link to the right pages

PR #4130, opened 2024-09-11
Project and submission links

bickelj:
Issue #4117
Issue #4129

## Test Steps

 - [ ] As staff
 - [ ] Visit the Projects page `apply/projects/all/`
 - [ ] Visit one of the projects e.g. Project 3 (associated with Submission 5).
 - [ ] Verify that the link to Project 3 shows the Submission ID in the URL, e.g. `apply/submissions/5/project/`
 - [ ] Verify that the Communications tab is actually a link to the Submissions#communications tab. Click it.
 - [ ] Verify that a Project tab/link shows up on the Submission page and that it leads back to the associated Project.
 - [ ] Verify that other links to the Project or Submission work, e.g. from Invoices.

### DIFF
--- a/hypha/apply/funds/templates/funds/applicationsubmission_detail.html
+++ b/hypha/apply/funds/templates/funds/applicationsubmission_detail.html
@@ -74,6 +74,11 @@
                             {% trans "View message log" %}
                         </a>
                     {% endif %}
+                    {% if object.project and PROJECTS_ENABLED %}
+                        <a class="tab__link" href="{% url 'apply:submissions:submission_project_detail' submission_pk=object.id rest='' %}">
+                            {% trans "Go to Project" %}
+                        </a>
+                    {% endif %}
                 </div>
             </div>
         </div>
--- a/hypha/apply/funds/templates/funds/includes/project_block.html
+++ b/hypha/apply/funds/templates/funds/includes/project_block.html
@@ -2,6 +2,6 @@
 <div class="sidebar__inner">
     <h5>{% trans "Project" %}</h5>
     <p>
-        <a href="{% url 'apply:projects:detail' pk=object.project.id %}">{{ object.project.title }}</a>
+        <a href="{% url 'apply:submissions:submission_project_detail' submission_pk=object.id rest='' %}">{{ object.project.title }}</a>
     </p>
 </div>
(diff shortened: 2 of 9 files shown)


frjo:
I see two big changes in this PR. 

1. Making project URLs a part of submission URLs.
2. Merging project and submission communication tabs.

If we go with 1 I think we should update URLs  for invoices and reports as well. We should also remove the "project/[0-9]+" URLs.

If we go with 2 there is no need to specify "Submission communications". It would just be "Communications".

@wes-otf What does OTF think of these two changes

@bickelj Why is the URL change important for your clients? My experience is that users do not look in the address-bar. Many browsers even actively hides the full address and only show the domain to help users combat fake sites.

wes-otf:
Thanks @bickelj for upstreaming this! A few notes from me:

- URLs
    - I really like the new format & from the convos I've had OTF would be on board with that too. I think we have a few power users that may notice/utilize it.
    - Agreed with @frjo that it makes sense to refactor invoice & report URLs to be the same as the regular project URL
    - The old URL format should redirect to the new one (ie. `/apply/projects/<PROJECT ID>/` -> `/apply/submissions/<SUBMISSION ID>/project/`)
- Projects table view
    -  I'm a fan of showing the corresponding submission in the table but it makes more sense to me to present it like the old table view did (ie. `<Submission Name> (#<ID>)` with a link to the submission as opposed to just showing the ID.
- Submission detail view
    - `Go To Project` tab formatting should be the same as the `View Message Log` in terms of hover states & text formatting.
    - Should the `Project` sidebar item be kept if we have a `Go To Project` tab? Could we show the project name in a tooltip on hover?
- Project detail view
    - If we have a `Go To Project` tab on the submission detail view it makes sense to me to have a `Go To Submission` tab on the project
    - I'm a fan of the combined comms tab but I couldn't get it to show up when running locally. Only `Details` & `Activity Feed` tabs showed.

This was after a brief look so feel free to let me know if I overlooked anything. Cheers!

theskumar:
https://github.com/HyphaApp/hypha/pull/3944 attempts something similar to what's done here. Essentially allows you to reference the submission or it's project by the same id. 

So if this feature is going in then #3944 can be closed.

frjo:
The PR https://github.com/HyphaApp/hypha/pull/3944 handles all the URLs (reports/invoices). I might also prefer that approach.

@bickelj Can you take a look at that PR and see how it would work out for you?

bickelj:
@wes-otf @frjo @theskumar I think, as we discussed in the maintainers' meeting, that #3944 and the merging of communications into a single feed would satisfy what we are looking for here.

bickelj:
Superseded by #3944 and related.

PR #4133, opened 2024-09-13
Added command to delete drafts that haven't been modified in specified days

wes-otf:
Adds a management command to delete all drafts older than two years. Intended to be used with a cron job to consistently clean up drafts.

## Test Steps

 - [ ] Ensure all drafts older than 2 years get removed after command is run

### DIFF
--- a/hypha/apply/funds/management/commands/drafts_cleanup.py
+++ b/hypha/apply/funds/management/commands/drafts_cleanup.py
@@ -1,3 +1,4 @@
+import argparse
 from datetime import timedelta
 
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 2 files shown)


frjo (approved):

frjo:
Looks good. Maybe two years is a bit overly cautious?

wes-otf:
@frjo I talked with @Techslammer about this and we figured it might make more sense to make this a customizable time rather than doing something hard coded as different instances needs/sensitivities will probably differ. So now the command format is: 
```
python3 manage.py drafts_cleanup <drafts older than days>
```
I also added a confirmation which can be skipped with `--noinput`/`--no-input`, ie.
```
python3 manage.py drafts_cleanup 365 --noinput
```

hypha/apply/funds/management/commands/drafts_cleanup.py
```diff
@@ -0,0 +1,80 @@
...
+
+    if ivalue < 0:
+        raise argparse.ArgumentTypeError(
+            f'"{value}" is an invalid non-negative integer value'
+        )
+    return ivalue
+
+
+class Command(BaseCommand):
+    help = "Delete all drafts that are older than 2 years"
```
frjo: Works perfectly, very neat! Just update this help text and we can merge it in.

frjo (changes requested):

hypha/apply/funds/management/commands/drafts_cleanup.py
frjo: Maybe we should add a section in the docs about neat cron commands that can be used? We have this one and the account cleanup at least.

wes-otf: great catch! updated the help text & wrote a new configuration page in the docs for cron jobs. let me know if you think I missed anything

frjo (approved):

PR #3990, opened 2024-06-14
Report reminders

bickelj:
Fixes #3989

## Test Steps

Set up a fake SMTP server or use file-based email.

- Configure reports for projects in Wagtail Admin
- See that report reminders go out as configured

PR #4136, opened 2024-09-17
Hide contracting category documents from restricted users

### LINKED ISSUE #4119: Hide tax forms/banking info from staff, only showing finance/admins
Today finance had requested that sensitive forms like tax forms & banking info when uploaded in the project documents only be visible to them. It would be nice if PMs and other staff could still see that status (uploaded/not uploaded), just not the document itself.

sandeepsajan0:
Fixes #4119 

image

## Test Steps

 - [ ] Add restricted users' groups in Contractcategory section in admin.
 - [ ] Check if those users have access to view the document.


frjo:
Adding all the roles with checkboxes is great for flexibility. But seeing it I start to think it would be better to reverse the logic.

"**Allow document access to these groups:**"

"Only selected groups will have document access."

Can we also list only Staff, Staff Admin, Finance and Contracting as options? Maybe it would be good to add this list of groups to `hypha/apply/users/groups.py` as "GROUPS_ORG_FACULTY" so we can reference it in other places?

By default we can then have all groups in this list checked.

sandeepsajan0:
image

image

image

sandeepsajan0:

frjo:
Migration conflict with https://github.com/HyphaApp/hypha/pull/3990, as soon as that one is approved we can test this one.

PR #4142, opened 2024-09-25
Submission data on determination and review form page

### LINKED ISSUE #3971: View Submission data while drafting a determination

frjo:
Fixes #3971

## Test Steps

- Add or edit a determination (e.g. `/apply/submissions/2/determination/2/edit/`)
- Click "Toggle submission" button in the header to see the submission.
- Same thing for reviews


hypha/apply/determinations/templates/determinations/base_determination_form.html
```diff
@@ -54,22 +62,24 @@ <h2>{{ value }}</h2>
...
-                <button class="button button--primary"
-                        hx-get="{% url 'funds:submissions:partial-answers' submission.id %}"
-                        hx-trigger="click once"
-                        hx-swap="innerHTML"
-                        hx-target="#submission-summary"
-                        @click="submissionVisible = ! submissionVisible">
-                    {% trans "View/Hide Submission" %}
-                </button>
-                <section id="submission-summary" x-bind:style="submissionVisible ? 'display: block' : 'display: none'">
+            <aside :class="showSubmission ? 'flex-1 ps-4' : ''">
```
bickelj: Why the aversion to `id`?

hypha/apply/determinations/templates/determinations/base_determination_form.html
```diff
@@ -54,22 +62,24 @@ <h2>{{ value }}</h2>
...
+                <section
+                    class="p-4 overflow-y-auto bg-white shadow-xl max-h-screen"
+                    x-show="showSubmission"
```
bickelj: Ah, of course!

hypha/apply/determinations/templates/determinations/base_determination_form.html
```diff
@@ -54,22 +62,24 @@ <h2>{{ value }}</h2>
...
+                    x-transition:enter="transition ease-out duration-250"
+                    x-transition:enter-start="transform opacity-0 translate-x-full"
+                    x-transition:enter-end="transform opacity-100 translate-x-0"
+                    x-transition:leave="transition ease-in duration-150"
+                    x-transition:leave-start="transform opacity-100 translate-x-0"
+                    x-transition:leave-end="transform opacity-0 translate-x-full"
```
bickelj: Thanks for this polish!

hypha/apply/determinations/templates/determinations/base_determination_form.html
```diff
@@ -54,22 +62,24 @@ <h2>{{ value }}</h2>
                     {{ message }}
                 </div>
             {% endfor %}
-            <aside id="submission-flyout"
-                   style="margin: 1rem"
-                   x-bind:style="submissionVisible ? 'flex-basis: 50%; margin: 1rem' : 'flex-basis: 2rem; margin: 0'">
-                <button class="button button--primary"
-                        hx-get="{% url 'funds:submissions:partial-answers' submission.id %}"
-                        hx-trigger="click once"
```
bickelj: The button seems to make a request on every click (whether on show or hide). I wonder if the removal of `hx-trigger="click once"` is related. Is this desired behavior? I would expect the request to happen exactly once for the page. Barring that, I would expect it to happen only when the submission gets shown, right?

bickelj (commented):
This looks much better. Perhaps the "Toggle submission" button is too subtle and off-to-the-side now, though. I am getting an error in the console that I plan to look further into:
```
Uncaught TypeError: proposal_form_field is null
    <anonymous> https://hypha.test/static/js/determination-template.js:53
    EventListener.handleEvent* https://hypha.test/static/js/determination-template.js:51
    <anonymous> https://hypha.test/static/js/determination-template.js:57
```

I see a declaration of `once` in an `hx-trigger` but the partial page keeps loading every click. I plan to look into this too.

Edit: I get the Uncaught TypeError on main at a5d3a4f5e8257af713683cfcbe3e37719d4fa152 as well, so I think that's unrelated to this PR.  So I created #4144 for this issue.

bickelj:
I poked around at the multiple requests for the partial page issue by peppering some more `once` declarations in the `hx-trigger` and on the button but that didn't prevent the partial (submission) page from loading multiple times. Hmm.

hypha/apply/determinations/templates/determinations/base_determination_form.html
frjo: Good catch, will test this. I copied most of this from the mobil nav menu.

hypha/apply/determinations/templates/determinations/base_determination_form.html
frjo: High specificity and need to be unique. Good when you need to target some js or link form elements with id/for, but otherwise I rarely use it.

hypha/apply/determinations/templates/determinations/base_determination_form.html
frjo: I think it works better now, removed "intersect". Appreciate some testing to confirm that is only loads once.

bickelj: Yes, it worked! I made the same change on the other form and it seems to work there too (removing `intersect` and the comma).

bickelj:
Those test failures look like a temporary GitHub hiccup, I edited the commit message and re-pushed.

bickelj:
@frjo Thanks for embracing this one and helping carry it along.

frjo:
No mention. I think OTF staff will really like this one.

The current look will work well for your clients as well?

bickelj:
@frjo I think there is a good chance. Rather than assuming "yes" I made a little video and asked.

A couple things I noticed while making the video:
1. When the window is a bit smaller (say, 1024x576 pixels), I see scrollbars appear on the submission summary even though the submission doesn't reach to the bottom of the review or determination form. I'm not sure what's causing that but it seems like a problem.
2. I think we missed the review "Edit" form (even though it's on the "Add" form).

There is a little bit of weirdness during the animation as well, but I think that's less important than the above two.

frjo:
1. Overflow scroll does not always act perfectly in all browsers. Also an issue with setting the height of the submission fly out section. It is now set to screen height and that is a bit much. We could test doing -20% or something like that. Would like to avoid calculate it with some extra js.
2. Is this features needed when editing reviews? There are other ui elements on that view that makes it a bit more work to add so I skipped it for now.

bickelj:
> 1. Overflow scroll does not always act perfectly in all browsers. Also an issue with setting the height of the submission fly out section. It is now set to screen height and that is a bit much. We could test doing -20% or something like that. Would like to avoid calculate it with some extra js.

When I experimented with the simplest possible `flex` container with two items and lorem ipsum text, the columns were auto-adjusted as expected to match height, which is I think what we would ideally want (except for some min and max boundaries, perhaps). But with this page I could not get it to behave as expected. Even when I hid all the stylesheets from my browser's view, I couldn't quite get it to behave properly with the same CSS as the simple example. That is why I ended up with the `flex-basis` hack to try to approximate what we want. I looked at the TinyMCE on the left side and those appear to be `iframe`s for each editing box. I wonder if that has something to do with the behavior but I didn't look further. (Aside: sometimes I wonder if all the CSS voodoo is better than the relatively simple `<table>` of yesteryear.) 

>     2. Is this features needed when editing reviews? There are other ui elements on that view that makes it a bit more work to add so I skipped it for now.

I expect users who see and use it on the "Add" page will ask where that button went on the "Edit" page. But I don't think that should stop us from shipping what we have now if that's a more difficult thing to do.

In any case, I think this is an improvement that no one will object to, except perhaps one of our customers who might ask for setting the default to show the submission.

wes-otf:
This is so slick! I'm pitching the user tests to staff tomorrow - I think this is going to be very well received

bickelj:
@wes-otf @frjo Feedback I received suggested that the button be "Show application" and extrapolating that to then be "Hide application" too. I think it's easy enough to change it to "Toggle application" but it looks like more of a pain to change the text every time. Would you be OK with "Application" instead of "Submission" on this button?

frjo:
@bickelj I plan to suggest a renaming of "Submission" -> "Application" throughout Hypha

bickelj:
@frjo OK. At least for this one button I made the change. Was the name change well-received for the rest of the UI?

frjo:
Yes, added an issue about that so we will rename everywhere.

bickelj:
:tada: :clinking_glasses:

PR #4150, opened 2024-09-27
OTS Extension for Wagtail configuration of messages

frankduncan:
Like #4149 , this adds an extension to hypha.  There's some overlap between the two PRs (some of the base setup).  This one differs in that it uses django signals to facilitate communication with an extension.

This extension adds a wagtail configuration for email/slack messaging, overriding the developer required code.  It includes simplistic find/replace for various variables depending on the object leading to the message.

Like the #4149, it includes tests and migrations that live apart from the codebase, and therefore should not cause migration conflicts.  Also, it includes a number of commits that we need to be included (and tests for them) to declare the signal, even if the meat of the extension is not something you're looking for.

Those are: 9b5012df1, c9fcaed55


frjo:
We will take a look at this implementation.

Where mush like the idea of allowing easy customisation of messages.

We definitely want to make e-mail the default notification method for all users. The current Slack notifications for staff will remain as an option.

frankduncan:
I think you may be confused by the terminology.  "Default" in this extension dictates whether the notifications get sent by default.  The idea is letting the administrator decide whether they want emails to be off unless otherwise specified, and whether they want slack messages to be off unless otherwise specified.

We ran into this because the client wanted us to have an allow list of what slack messages, and emails, to send.  Rather than just blocking ones the don't.  So the use case is if someone wants to only send an email message for determination, and only send a slack message for submission creation, they say "default off" for both slack and email messages, but then add an overriding "turn on for this message only" to both.

This prevents messages going out in the case of an upgrade in hypha adding new ones.

frjo:
I was unclear, the default e-mail part refers to the work we plan to do, hopefully later this year, to have all staff notifications go via e-mail by default.

frankduncan:
Ahh!  After this PR gets reviewed/accepted, there will be another which connects to Zulip (dependent on this one), which uses the hypha slack infrastructure.  I expect this extension well be helpful to keep that enabled after your changes!

bickelj:
@frjo Like #4149 I think the first commits here are the extension capability and then the last commits are an example of an extension using that infrastructure.

extensions/ots/message_configuration/models.py
```diff
@@ -0,0 +1,100 @@
...
+        super().__init__("", "messaging_help.html", classname="messaging_help_panel")
+
+    class BoundPanel(HelpPanel.BoundPanel):
+        def __init__(self, **kwargs):
+            super().__init__(**kwargs)
+            self.email_messages = self.panel.email_messages
+
+
+@register_setting
+class MessagingSetting(BaseSiteSetting, ClusterableModel):
```
frjo, line 41: We are moving away from Wagtail sites so `BaseGenericSetting` is better to use.

See https://github.com/HyphaApp/hypha/blob/a027eb04090f0dd7986f4de10dd6f35efd648773/hypha/core/models/system_settings.py#L14-L18

extensions/ots/message_configuration/static/js/message_configuration_wagtail_admin.js
```diff
@@ -0,0 +1,34 @@
...
+        });
+    }
+
+    $(function () {
+        $("#id_messaging_settings-ADD").click(function () {
+            add_update_events();
+        });
+        add_update_events();
+    });
+})(jQuery);
```
frjo, line 34: VanillaJS or AlpineJS, we are retiring jQuery.

extensions/ots/message_configuration/templates/messaging_help.html
```diff
@@ -0,0 +1,53 @@
...
+<p>
+  {% trans "Click 'Add' at the bottom to configure a template for an outgoing message type." %}
+</p>
+<p>
+  {% trans "Variables that will get replaced follow.  Use the uppercase name (before the colon) in any subject, email message, or slack message." %}
+</p>
+
+<p>{% trans "When the submission is available, the following are available:" %}</p>
+
+<ul>
```
frjo, line 15: Add class `w-mb-4` to add some margins to the lists.

extensions/ots/message_configuration/templates/messaging_help.html
```diff
@@ -0,0 +1,53 @@
...
+  {% trans "Here you can configure message templates for different messages hypha sends out.  You can configure the footer and header (which will apepar on each email), as we well as configure whether hypha sends out messages by default." %}
+</p>
+<p>
+  {% trans "Click 'Add' at the bottom to configure a template for an outgoing message type." %}
+</p>
+<p>
+  {% trans "Variables that will get replaced follow.  Use the uppercase name (before the colon) in any subject, email message, or slack message." %}
+</p>
+
+<p>{% trans "When the submission is available, the following are available:" %}</p>
```
frjo, line 13: Maybe wrap these in `<strong>`? Makes it easer to see the sections of the text.

extensions/ots/message_configuration/templates/messaging_help.html
```diff
@@ -0,0 +1,53 @@
...
+<ul>
+  <li>{% trans "ORG_EMAIL: " %}</li>
+  <li>{% trans "ORG_SHORT_NAME: " %}</li>
+  <li>{% trans "ORG_LONG_NAME: " %}</li>
+  <li>{% trans "ORG_URL: " %}</li>
+  <li>{% trans "ORG_GUIDE_URL: " %}</li>
+</ul>
+
+{% for id, default in self.email_messages.items %}
+  <div class="is-hidden" attr-email-id="{{ id }}">
```
frjo, line 50: Can we just use "hidden" attribute instead of adding a class and a custom css file?

extensions/ots/message_configuration/models.py
```diff
@@ -0,0 +1,100 @@
...
+        on_delete=models.CASCADE,
+        related_name="messaging_settings",
+    )
+    message_type = models.CharField(
+        choices=MESSAGES.choices,
+        max_length=50,
+    )
+    email_subject = models.TextField(null=True, blank=True)
+    email_message = models.TextField(null=True, blank=True)
+    slack_message = models.TextField(null=True, blank=True)
```
frjo, line 89: The two messages fields should be textareas I believe.

frjo (changes requested):
Making messages configurable will be an excellent improvement of Hypha. This implementation however feels "plastered on".

If we in the future want to change messages we will need to update in several places from what I can see. 

Can we not build this in to the activities app directly. Making this the way to handle e-mail notifications?

Are any of your clients using Slack? My guess is that only OTF uses it so we can skip Slack notifications for now.

frankduncan:
> Making messages configurable will be an excellent improvement of Hypha. This implementation however feels "plastered on".

Yes!  The architecture of the messaging was a bit hard to puncture.  Our first attempt was to add it directly into the activities app.  However, there was a lot of business logic, both in the templates themselves, as well as in the actual calls (for instance, https://github.com/HyphaApp/hypha/blob/main/hypha/apply/activity/adapters/base.py#L56-L63 adds a lot of complexity)

We were able to get as far as our client needed with the first pass of configuration in the activities app, as a lot of the logic in the base hypha activities app was able to be bypassed (they really just wanted far more simplistic messaging around a small subset of messages).  However, building out something that could reasonably reproduce what hypha is currently doing with configuration, turned out to be a much bigger project than we had budget for.  If you guys can do that, that would be awesome!  In the meantime, we settled for saying "hey, add a hook into the base messaging app, and let us do a reasonable job of using that hook for our clients"

> If we in the future want to change messages we will need to update in several places from what I can see.

Yes and no.  As long as you make the base tests pass from the activities app, then you can safely ignore the extension.  We just need the contract maintained of turning email on and off in the default case, as well as being able to replace the message generated by the activities app with one of our own.

> Can we not build this in to the activities app directly. Making this the way to handle e-mail notifications?

See above!  It looked like a pretty massive project to take what's currently in hypha to that method.  If you guys can come up with a good way, that would be even better than our extension.  One thing we did decide we needed for our client was no logic in the configuration.  Simple string replacement only.

> Are any of your clients using Slack? My guess is that only OTF uses it so we can skip Slack notifications for now.

Yes!  We actually subclassed the slack adapter to add zulip (coming to the hypha codebase after we work through this PR).  They use it for general channel things (not targeted messages) for general events, like a submission happening.  This lets people in their organization stay aware of general stuff in hypha without having to subscribe to events, or even have accounts on the hypha instance.

extensions/ots/message_configuration/models.py
frankduncan: I can see that!

extensions/ots/message_configuration/templates/messaging_help.html
frankduncan: Sure!

extensions/ots/message_configuration/templates/messaging_help.html
frankduncan: I don't think we have a lot of opinions on the styling of the help text, so whatever you guys think looks good.

extensions/ots/message_configuration/static/js/message_configuration_wagtail_admin.js
frankduncan: Not as familiar with those, so probably won't have a chance to do that.

extensions/ots/message_configuration/templates/messaging_help.html
frankduncan: Good call!  I'll make the change depending on how the overarching PR discussion goes :)

extensions/ots/message_configuration/models.py
frankduncan: Cool!  I'll tackle this depending on how the overall PR discussion goes :)

frankduncan:
Overall, a potential successful merge here is to say "we don't want the extension, but we will accept the changes to the base hypha app that enable your extension."  That way, we can move forward on our end and let you guys put together a project that is more in line with how you'd want this feature to work in mainline hypha.

If that's what you end up doing, we can create another PR with just the changes we need to enable our extension to work in isolation :)

PR #4151, opened 2024-09-27
Add rolespermissions module, refractor the group and delete submission to use it

theskumar:
Updates group management:

- **Centralized Group Definition:** Instead of creating groups within migration files, which can be cumbersome to maintain, group definitions are now managed through the AbstractRoles model. These roles are synchronized using the `python manage.py sync_roles` command. This ensures no existing groups or their associated permissions are deleted.

- **Module Renaming**: The `hypha.apply.users.groups` module has been renamed to `hypha.apply.users.roles` to reflect the shift from group-based to role-based permissions. This aligns with upcoming changes utilizing the `rolepermissions` module.

- **Streamlined Group Descriptions**: The GroupDesc model is removed. Instead, help text can be directly defined within the role itself. This simplifies management and allows for translation of group descriptions.

**This is the first of a series of pull requests aimed at refactoring the permissions system.**

As a sample implementation, converts the delete submission to use this role-permissions. See: https://github.com/HyphaApp/hypha/pull/4151/commits/8239c2172ab9150cdba0db8d95f13e1681f176cd

## Test Steps

 - make sure the migrations run fine.
 -  groups are create correctly via `python manage.py sync_roles`, it should also keep the existing groups. 
 -  groups description is rendered in the admin.
 - ensure delete submission/application is working as expected


frjo (approved):
Notice no changes at all in how Hypha works so that is good. Running the "sync_roles" does not add/delete/change any of the existing roles. Deleting an existing role and running "sync_roles" restores it.

theskumar:
On another thought, let's hold on to merging this, I'm reading and trying out few things with some of the use-cases we have the project and make a final decision on going with `django-rolepermissions` package.

esp. i'm evaluating `django-rules` for object level permissions, while the `django-rolepermissions` is quite good for group based permissions it doesn't seem to provide much when it comes to object level permissions. 

Looking at the use case, hypha seems to have majority object level permissions.

theskumar:
Implemented a sample permission and it's checks for the delete submissions, it looks straightforward. I was not confident that the `can` template tag and `HasPermissionMixin` will work with seamlessly with `@register_object_checker`, but it does.

https://github.com/HyphaApp/hypha/pull/4151/commits/8239c2172ab9150cdba0db8d95f13e1681f176cd

theskumar:
@frjo @sandeepsajan0 @wes-otf  If the delete submission implementation looks good, let's try get this merged so rest of the followup PRs will be mostly converting existing permission check to use the `rolepermissions` package, instead of custom functions we using.

frjo (changes requested):
A failing test to fix. I also think we need to add the `sync_roles` to the `Procfile` so it runs on Heroku deployments.

frjo (approved):

theskumar:
Update the test to match new test settings where `SECURE_SSL_REDIRECT` is set to false by default. 

> I also think we need to add the sync_roles to the Procfile so it runs on Heroku deployments.

This should covered already, let me know if you face issue when deploying this branch.

PR #4157, opened 2024-10-07
Improvment to the submisison's related section

theskumar:
- Update the data displayed - add current status and round/lab
- Update the look and feel of it
- Limit the title display to 2 lines.
- Refractored the JS to be jquery independent and make it collapsing
  behaviour generic, some concepts borrowed from https://rstacruz.github.io/rsjs/ for structuing the custom js code

Before: 
Screenshot 2024-10-07 at 8  56 32@2x

After: 
Screenshot 2024-10-07 at 8  56 09@2x


wes-otf:
this looks fantastic! I'll do a deeper dive tomorrow but seems like a really solid improvement.

frjo (approved):
Works well.

PR #4161, opened 2024-10-10
Remove 'drf-yasg' swagger docs viewer

theskumar:
Since the APIs are all depreciated, and APIs are not used by anyone,
this PR starts the depreciation phase and tries to remove the dead load.


frjo (approved):

theskumar:
On the reasons for removing this is pip install with `uv` fails b/c of this package as it requires some outdate internal pkg_resources. 

I'm also trying to switch to `uv` instead of `pyenv` for development.

theskumar:
If it looks good, can get this merged in sooner than later! @frjo

PR #4162, opened 2024-10-10
Updated ESLint to v9.12.0 & migrated to a flat config file

wes-otf:
In some of the translation features I was using optional chaining & eslint would toss errors saying it was invalid despite it working. This led me to exploring the config to find the [yaml config we used was depreciated](https://eslint.org/docs/latest/use/configure/configuration-files-deprecated). This PR updates to a newer version of eslint & migrates the config to the [flat file version](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file) (and now optional chaining is recognized 🥳).

Some code changes that came with the updated linting:
- [As of ES6](https://262.ecma-international.org/6.0/#sec-strict-mode-code) modules are always in strict mode, thus `"use strict"` isn't needed
- Removed unused var
- Removed unneeded `eslint-disable` comment

There are also a ton of JSDoc warnings now - I was going to remove them but I think JSDoc is actually a weak point in our code documentation so maybe the in-terminal annoyances will be a good reminder to do a quick write up

## Test Steps

 - [ ] Ensure JavaScript linting happens as expected, allowing for updated ES syntaxes


frjo:
Good work on getting this updated.

Is it needed to switch conf format? I find the yaml version far easier to read and understand.

eslint.config.mjs
```diff
@@ -0,0 +1,154 @@
...
+                    words: true,
+                    nonwords: false,
+                },
+            ],
+
+            "spaced-comment": ["error", "always"],
+            strict: ["error", "function"],
+            yoda: ["error", "never"],
+            "max-nested-callbacks": ["warn", 3],
+            "jsdoc/require-description": "warn",
```
theskumar, line 151: If we don't have a strong preference I think probably remove this? I see bunch of warning via the GithHub actions on almost all the files.

theskumar (approved):
Left a minor comment rest looks good to me. 

Given the yaml format is depreciated, i think the js version will be fine for us.

eslint.config.mjs
wes-otf: I leaned toward keeping it to encourage more JS documentation, especially maybe a nudge for us to revisit some of the old jquery stuff but can totally toss if you think it's counter productive/too annoying! I documented some low hanging fruit too so there should be less warnings now

frjo (approved):

PR #4163, opened 2024-10-12
Add "templates_custom" folder to tailwind search path

theskumar:
Right now, the custom templates put in the recommended "template_custom" doesn't checked by tailwindcss. So, the custom templates are not styled by tailwindcss.


frjo (approved):
Nice fix!

PR #4164, opened 2024-10-15
Speed up CI 

theskumar:
Check the individual commits for more info, for now.


Makefile
```diff
@@ -4,6 +4,9 @@ DJANGO_SETTINGS_MODULE = hypha.settings.dev
 JS_VENDOR_DIR = ./hypha/static_src/javascript/vendor
 JS_ESM_DIR = ./hypha/static_src/javascript/esm
 
+# check if pip install present, else try to use uv pip
+PIP := $(shell command -v pip 2> /dev/null || echo 'uv pip')
```
frjo: Will not `command -v pip` be true as default on most systems? Can we check for `uv` instead and use `pip` only if it is not installed?

frjo (changes requested):
Looks really neat! One minor comment.

Makefile
theskumar: It make sense, on my machine, the pip was missing instead pip3 was available as global. `pip` get available only in the virtualenv. 

This check won't be needed when/if we move to using `uv` with dependencies in pyproject. At that point we would always have `uv` and then run `uv sync` all the time.

frjo (approved):

frjo:
From a reset test run on current hypha-ci:

Skärmavbild 2024-10-20 kl  11 48 53

From the new hypha-ci in this PR on the test branch:

Skärmavbild 2024-10-20 kl  11 49 39

I know from experience that the times to run these GitHub actions varies quite a bit but a two minuts save, -30%, is nice to see indeed.

theskumar:
I agree, they are very good numbers and on a regular run it should finish around 3mins (-50%), which is similar to the time taken by CodeQL. 

This allows to write more tests without worrying about test timing for a while.

PR #4169, opened 2024-10-17
Running total of score on review form

### LINKED ISSUE #3963: Display the total score on Review Forms

Review form enhancement.

Display the total score on Review Forms so that Reviewers can see the summed total score as they complete (and prior to submitting) their reviews.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Reviewers

frjo:
Live score label added to edit review form

Fixes #3963

## Test Steps
- As a Reviewer, add a review
- Change the values in the drop-downs see the live score updates on the right.
- As a Reviewer, edit a review
- Change the values again, see the live score updates again.

### DIFF
--- a/hypha/apply/review/templates/review/review_edit_form.html
+++ b/hypha/apply/review/templates/review/review_edit_form.html
@@ -26,7 +26,7 @@
                 x-clock
                 x-show="showScore"
             >
-                {% trans "Score:" %} <span x-text="formattedScore">-</span>
+                {% trans "Score:" %} <span x-text="totalScore">-</span>
             </div>
         </div>
 
@@ -74,7 +74,7 @@
                 {% endif %}
 
                 <div class="text-fg-muted" x-clock x-show="showScore">
-                    {% trans "Total Score:" %} <span x-text="formattedScore"></span>
+                    {% trans "Total Score:" %} <span x-text="totalScore"></span>
                 </div>
             </div>
         </form>
--- a/hypha/apply/review/templates/review/review_edit_form.html
+++ b/hypha/apply/review/templates/review/review_edit_form.html
@@ -1,6 +1,12 @@
 {% extends "base-apply.html" %}
-{% load i18n %}
+{% load i18n static %}
 {% block title %}{{ title }}{% endblock %}
+
+{% block alpine_scripts %}
+    {{ block.super }}
+    <script defer src="{% static 'js/behaviours/review-score.js' %}"></script>
+{% endblock %}
+
 {% block content %}
 
     {% adminbar %}
@@ -12,8 +18,19 @@
 
     {% include "forms/includes/form_errors.html" with form=form %}
 
-    <div class="wrapper wrapper--medium wrapper--inner-space-medium">
-        <form class="form form--with-p-tags form--scoreable" action="" method="post" novalidate>
+    <div class="my-8 relative" x-data="reviewScore">
+        <!-- Review score sticky display -->
+        <div class="absolute h-full right-0 top-0 hidden lg:block">
+            <div
+                class="bg-arsenic text-white font-semibold inline-block px-3 py-1.5 mt-1 text-center sticky top-1"
+                x-clock
+                x-show="showScore"
+            >
+                {% trans "Score:" %} <span x-text="formattedScore">-</span>
+            </div>
+        </div>
+
+        <form class="form form--scoreable max-w-3xl" action="" method="post" novalidate>
             {{ form.media }}
             {% csrf_token %}
 
@@ -22,7 +39,7 @@
             {% endfor %}
 
             {% for field in form.visible_fields %}
-        {# to be replaced with better logic when we use stream form #}
+                {# to be replaced with better logic when we use stream form #}
                 {% ifchanged field.field.group %}
                     {% for key, value in form.titles.items %}
                         {% if key == field.field.group %}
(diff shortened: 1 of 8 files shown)


frjo:
@bickelj What do you think of this version?

bickelj:
It looks good to me, here is how it looks for me:
running_total_20241018

bickelj:
Maybe just like the previous change for "Hide application" and "Show application" should be made here though to do "Tot" button and "Avg" button - although that would be confusing because am I seeing a Total or asking for a Total? Hmm.

hypha/static_src/javascript/review-form-score.js
```diff
@@ -0,0 +1,51 @@
...
+        let score = total;
+
+        if (use_average) {
+            score = total / count;
+        }
+
+        // Update the text in .form--score-box with the current score
+        document.querySelector(".form--score-box > .score-number").textContent =
+            formatter.format(score);
+    }
```
theskumar:
```suggestion
    /**
     * Calculate the score based on the selected values
     */
    function calculate_score() {
        const formatter = new Intl.NumberFormat({
            maximumSignificantDigits: 4,
        });
        const validValues = selectors
            .map((selector) => parseInt(selector.value))
            .filter((value) => !isNaN(value) && value !== 99);
        const total = validValues.reduce((sum, value) => sum + value, 0);
        const score = use_average ? total / validValues.length : total;
        document.querySelector(".form--score-box > .score-number").textContent =
            formatter.format(score);
    }
```

Add jsDoc and update to use more functional approach for a cleaner code.

hypha/static_src/javascript/review-form-score.js
```diff
@@ -0,0 +1,51 @@
+(function () {
+    let use_average = false;
+
+    // grab all the selectors
+    const selectors = Array.prototype.slice.call(
+        document.querySelectorAll("select")
+    );
```
theskumar: While the js is included just on the review form, I think this select is too wide open and should be limited to a named class or `data-**` selector.

theskumar:
```suggestion
    const selectors = [...document.querySelectorAll(".form--scoreable select")];
```

frjo: Good point, fixed.

frjo: When testing this I noticed it selected the recommend field as well so the average was wrong. Added `data-score-field` to score field with and without text so we are sure we get the correct fields.

hypha/static_src/javascript/review-form-score.js
frjo: Nice, implemented.

hypha/static_src/javascript/review-form-score.js
theskumar: yes. I noticed it too, it picks up all the values, whether it's a score or not and since the form is dynamic it would produce errors.

theskumar:
> Maybe just like the previous change for "Hide application" and "Show application" should be made here though to do "Tot" button and "Avg" button - although that would be confusing because am I seeing a Total or asking for a Total? Hmm.

I think the same. I think there shouldn't be a button at all there, no-one will go and click it. Esp. since it doesn't remember the choice, it will get very annoying. 

I propose, calculate and display both and remove any interaction with it. Something like this maybe?

Screenshot 2024-10-21 at 5  09 27@2x

theskumar:
Also, thoughts on not having the `REVIEW_SHOW_RUNNING_TOTAL` configuration at all. Instead, if we thing some org will use `Avg` and other will use `total` then we have should have a configure what review scoring method to use, instead. 

Also, in the submission detail page. The avg. is displayed against the all the reviews submitted and not against the individual reviews. How is avg of the each of the questions/answer in a particular review helpful.

frjo:
> How is avg of the each of the questions/answer in a particular review helpful.

Very good point. Unnecessary feature request from me. I did not think through the average of all reviews vs average of one review part.

I think we should remove the average part and only show the total if the setting is true.

bickelj:
@theskumar @frjo Thanks for all the fixes and improvements on this capability!

hypha/static_src/sass/components/_form.scss
```diff
@@ -39,6 +39,12 @@
         }
     }
 
+    &--score-box {
+        background-color: variables.$color--dark-grey;
+        color: variables.$color--white;
+        font-weight: variables.$weight--bold;
+    }
```
theskumar: This can be easily replaced with twailwind classes on the element itself. 

I think this should roughly translate to `bg-dark-grey text-white font-semibold`

hypha/apply/review/templates/review/review_edit_form.html
```diff
@@ -10,37 +10,49 @@
...
-            {% if not object.id or object.is_draft %}
-                <button class="button button--submit button--top-space button--white" type="submit" name="{{ form.draft_button_name }}">{% trans "Save draft" %}</button>
-            {% endif %}
-            <button class="button button--submit button--top-space button--primary" type="submit" name="submit">{% trans "Submit" %}</button>
-        </form>
+                <button class="button button--submit button--top-space button--primary" type="submit" name="submit">{% trans "Submit" %}</button>
+            </form>
+        </section>
+        {% if show_running_total %}
+            <div class="form--score-box inline-block px-4 py-2 mt-1 text-center sticky top-1" hidden>
```
theskumar:
```suggestion
            <div class="form--score-box bg-dark-grey text-white font-semibold inline-block px-4 py-2 mt-1 text-center sticky top-1" hidden>
```

theskumar (dismissed):
Small suggest on using tailwind classes instead of sass for this one of element.

hypha/apply/review/templates/review/review_edit_form.html
```diff
@@ -10,37 +10,49 @@
...
+                {% trans "Score:" %} <span class="score-number pe-1">0</span>
```
theskumar:
```suggestion
                {% trans "Score:" %} <span class="score-number">0</span>
```

padding left is not needed now.

frjo:
@theskumar Fixed a bug that I introduced to this form on main so please rebase.

theskumar:
Updates: 

- Use alpinejs to keep track of total score and display
- Hide sticky score on mobile viewports.
- Add the score along side the submit button, so you can see them while submitting (even on smaller screen)

Screenshot 2024-10-23 at 10  54 37@2x

theskumar:
@frjo Thanks for fixing the decimal, the sum will always be integer a believe, in that case. The `formattedScore` can be removed and the template could use the `score` directly.

bickelj:
Wonderful, thank you!

PR #4184, opened 2024-10-23
Added tag to generate existing submission view params for pagination

### LINKED ISSUE #4183: Filters are lost in both list & table views when navigating to next page
**Describe the bug**

When filtering the list/table views that have enough results to be paginated, selecting `Next`/`Previous`/`Last`/`First` at the bottom of the view will result in navigating to the selected page of All Submissions rather than the filtered criteria

**To Reproduce**

1. Navigate either the list or table view
2. Use a broad filter to cause pagination (ie. `No Screening`)
3. Scroll down to the page nav buttons at the bottom (ie. Next`/`Previous`/`Last`/`First`)
4. Select any of the buttons & notice page has been changed but filters are gone

**Expected behavior**

Filters are preserved while the page is changed

**Priority**

- Medium (slowing you down)
**Screenshots**

**Additional context**

Manually entering the page query param in the URL causes the filter to be preserved while navigating to the intended page.

wes-otf:
Fixes #4183. Preserves query params when utilizing pagination options in list/table views.

## Test Steps

 - [ ] Set a broad filter on the list/table submissions view to cause pagination
 - [ ] Ensure user pagination buttons at the bottom of the list & table views preserve active filters while changing pages


theskumar:
@wes-otf After consulting with @frjo on the call today, I went ahead update to code to the existing `modify_query` template tags which essentially does the same thing that you built, but no new code.

theskumar:
And added a bit of stylistic update.

frjo (approved):

wes-otf:
perfect! I had no idea the `modify_query` tag existed so this works well

PR #4187, opened 2024-10-26
Add new workflow for Request with same time review

frankduncan:
This workflow is like Request with External Review, in that both internal and external reviewers can review submissions.  However, it has a workflow more like Request in that there is only one Review step.

## Test Steps

Best way to test/experiment with is to create a fund with this workflow and move through the process, using both internal and external reviewers.


hypha/apply/funds/workflow.py
```diff
@@ -1285,7 +1423,7 @@ def phases_matching(phrase, exclude=None):
     },
     "internal-review": {
         "name": _("Internal Review"),
-        "statuses": phases_matching("internal_review"),
+        "statuses": phases_matching("rev_review") + phases_matching("internal_review"),
```
frjo: Can we not rename `rev_review` to `rev_internal_review` instead?

hypha/apply/funds/workflow.py
```diff
@@ -380,6 +382,134 @@ def make_permissions(edit=None, review=None, view=None):
...
+            "permissions": applicant_edit_permissions,
+        },
+    },
+    {
+        "rev_review": {
+            "transitions": {
+                "rev_post_review_discussion": _("Close Review"),
+                INITIAL_STATE: _("Need screening (revert)"),
+            },
+            "display": _("Internal Review"),
```
frjo: Maybe only name this stage "Review" since there is only one step and it is both internal and external.

frjo:
I think this is a good addition, one where internal and external review happens at the same stage. This has a good chance of being useful to many organisations.

Happy to see that it is fairly easy to add a new workflow.

The plan is to make it even easier by splitting them up in separate files.

frjo (changes requested):
Some minor comments around naming things but overall look good I think.

hypha/apply/funds/workflow.py
frankduncan: Well, it's not really internal review.  It's external AND internal review.

The problem this solves is that there's a banner on submissions page (maybe gone in future versions) that showed a count of "Internal Reviews".  So we lumped this in there rather than creating a new summary tab that would always show up for other workflow types.

Not really sure what's the best way to do this.

frjo: Since you want it to `phases_matching` with `*_internal_review` I think it makes sense to name it `rev_internal_review`. This is a machine name and not displayed anywhere in the UI.

frjo (approved):

frjo:
"Request with review" is not the best name for this.

We already have "Request" that do have reviews but only internal ones. Then there is "Request with external review" that has internal och external reviews in series.

This new workflow also has internal och external reviews, but in parallel. Or in the same phase; step etc.

So what makes this new workflow unique is that the external and internal reviews are done at the same time. The name should give a clue about that I think.

Any suggestions?

frjo:
The workflow itself works without issues when I test it. Fix the naming and we can merge it I think.

frankduncan:
As they say, naming is one of the two hardest problems in computer science.  I have no idea what to call it.  If I were starting over, I would have three:

* Request
* Request with Only Internal Review
* Request with Separated External Review

Or, in other words.  This one feels like the standard, the current workflows seem like deviations from it, if that makes sense.

frjo:
What about "Request with same time review"?

RequestRev -> RequestSame

rev_more_info -> same_more_info

etc.

frankduncan:
Update with name changes.  I think I didn't miss any.

frankduncan:
Oh, whoops.  Thanks for the catch!!

wes-otf:
So in testing with 2 different accounts assigned (one reviewer role, one staff role) when in the review stage I can only see that it's ready for review as a staff member. I get permission denied when trying to access it as the reviewer role, despite getting an email that it's ready for review. 

Let me know if there's a detail/config option I missed!

frjo:
@wes-otf It was the Reviewer settings that was the issue. See https://test-apply.hypha.app/admin/settings/funds/reviewersettings/2/.

I have deactivated the settings now so Hypha reverts back to the default.

wes-otf:
Ah great that works well! Maybe we could document that somewhere too if that makes sense to you @frjo?

PR #4196, opened 2024-11-05
Add application and project tabs.

### LINKED ISSUE #4117: Add "Project" and "Submission" tabs on the submission and project detail pages

frjo:
Fixes #4117 

## Test Steps

 - [ ] Check that the "Application" and "Project" tabs work and show up when they should.
 - [ ] Check that all links to the application from project side works as they should.
 - [ ] Is it ok to remove the simplified submission view?


theskumar:
Should these be updated as well? 
Screenshot 2024-11-06 at 5  33 21@2x

Navigation Menu
Screenshot 2024-11-06 at 5  34 33@2x

On the submission all page:

Screenshot 2024-11-06 at 5  39 24@2x

frjo:
Yes, we should likely update these strings as well for consistency.

There is an issue that this will break a number of translations that have been done in to Russian and Chinese.

I think best to tackle that in a separate PR.

theskumar:
> Yes, we should likely update these strings as well for consistency.
> 
> There is an issue that this will break a number of translations that have been done in to Russian and Chinese.
> 
> I think best to tackle that in a separate PR.

make sense.

wes-otf:
so the only feedback from staff was that it might make more sense for the tabs to remain in the same order, so keep it `Application`, `Conversations`, `Project` in both application and project views:
Screenshot 2024-12-02 at 5 28 26 PM
Screenshot 2024-12-02 at 5 28 17 PM

I think especially as we move into merging the comms tab this is going to make the most sense

wes-otf:
it also seems like the simplified view can be removed

frjo:
@wes-otf I have updated the tab order now on project detail pages, good change I think. Do you want to retest on test?

wes-otf:
I think it's good to go! I think it's a nice addition

PR #4200, opened 2024-11-06
Enhance look and feel of the submission results page

theskumar:
## Before

image

## After

image

### DIFF
--- a/hypha/apply/funds/templates/funds/submissions_result.html
+++ b/hypha/apply/funds/templates/funds/submissions_result.html
@@ -9,59 +9,88 @@
         {% slot sub_heading %}{% trans "Track and explore submission results" %}{% endslot %}
     {% endadminbar %}
 
-
     <div class="wrapper wrapper--large wrapper--inner-space-medium">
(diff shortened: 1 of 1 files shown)


frjo (changes requested):
Overall looks good, some minor issues on smaller screens/windows.

Skärmavbild 2024-11-06 kl  14 57 37

1. I prefer left aligned headers but lets see what other people feel about that.
2. On smaller screens large numbers overflow. Maybe take the size down a notch?
3. Can we remove "Your" to make it shorter and make sure it does not wrap?
4. Decimals are not really needed here I think, we can round the numbers instead. That would help with the overflow issue in 2.

theskumar:
Updating to accept a fairly large numbers now. Updating the decimal formatting is tricky unless we make a hard choice of not using babel to convert into locale based formatter. 

https://github.com/user-attachments/assets/cf19e33e-47de-4cc1-b40e-3267b2512a53

frjo (approved):
Looks really good!

PR #4201, opened 2024-11-08
Implement DJP: Django Plugins.

frjo:
Read up on DJP at https://djp.readthedocs.io/

### DIFF
--- a/docs/references/plugins.md
+++ b/docs/references/plugins.md
@@ -1,0 +1,9 @@
+Hypha implements DJP: Django Plugins. A plugin system for Django.
+
+See https://djp.readthedocs.io/ for more information.
+
+## Some tips
+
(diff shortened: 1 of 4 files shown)


theskumar:
Looks good.

theskumar:
You can also add folder based discovery, so if someone drop a plugin a specific folder it will be automatically picked up. 

Benefit I see you don't do installed them.

theskumar:
> You can also add folder based discovery, so if someone drop a plugin a specific folder it will be automatically picked up.
> 
> Benefit I see you don't do installed them.

I thought it needed any configuration in host application, but looks like nothing needs to be done. 

https://djp.readthedocs.io/en/latest/installing_plugins.html#loading-plugins-from-a-directory

wes-otf:
Going to do some studying up on this in the next coming weeks but this looks good!

PR #4219, opened 2024-11-14
Add the ability to translate submissions (redux)

wes-otf:
This builds on the #4134 PR that initially introduced machine translations into Hypha. This isolates the translation behavior; putting pip dependencies in a separate `requirements-translate.txt` and will not attempt any translate imports unless the setting for it is true.

Other small changes are also a full docs page explaining how to install language packages & changing the setting once again from `SUBMISSION_TRANSLATIONS_ENABLED` to `APPLICATION_TRANSLATIONS_ENABLED` to reflect the system wide shift away from submission terminology.

## Test Steps

 - [ ] Ensure Hypha functions completely the same without translations enabled and without argostranslate installed
 - [ ] Ensure that when enabled & packages/languages are installed, translations work as expected.


wes-otf:
To use hypha the `argostranslate` package does not need to be installed but right now for tests to pass it needs it (despite all it's functionality being mocked). We could add the install to the CI or I could try to get a functional mock/conditional set up, whatever you think @frjo

frjo:
Maybe we can use `@skipif`/`@skipunless` so the tests only run when `APPLICATION_TRANSLATIONS_ENABLED ` is true?

hypha/apply/funds/forms.py
```diff
@@ -24,6 +25,9 @@
 from .widgets import MetaTermSelect2Widget, Select2MultiCheckboxesWidget
 from .workflow import get_action_mapping
 
+if settings.APPLICATION_TRANSLATIONS_ENABLED:
+    pass
+
```
frjo: Is this left over test code?

hypha/apply/funds/templates/funds/includes/rendered_answers.html
```diff
@@ -44,3 +58,30 @@ <h5 class="text-base">{% trans "Organization name" %}</h5>
...
+        margin-top: 0.5rem;
+        margin-bottom: 0.5rem;
+        border-width: 1px;
+    }
+
+    #translate-card-loading.htmx-request.htmx-indicator span {
+        display: inline-block;
+        height: 1rem;
+    }
+</style>
```
frjo, line 79: Could we use tailwind classes for this instead?

frjo:
Overall it looks like a smart way to implement this. We have the feature easily available without putting a burdon on all installs. Nice work!

Will try it out locally and see if I find any edge cases.

hypha/apply/funds/templates/funds/includes/rendered_answers.html
wes-otf: the `htmx-request` class gets added while the request is processing so I that bit of CSS will need to stay but I moved everything else to tailwind

wes-otf:
@frjo your `skipUnless` solution works well! should we still add the install of argostranslate to the CI so that logic can be tested too or leave it as is? Again all of argos is mocked in my tests so that wouldn't slow it down, it'd just add time on the pip install

theskumar:
I think it's good candidate for #4201 or have it implemented in the client/server mode then use it's translator api.

Advantage of the client/server mode, it will allow someone to easily replace it with another API and/or allow it to run without blocking the main application server.

frjo:
@wes-otf I think we can run the tests for this in the CI. The pip installs are cached so the effect should be minimal.

wes-otf:
@frjo CI looks good to go!

wes-otf:
@theskumar I totally agree! I think with how isolated the logic is now especially once we're ready to do that implementation I could totally extract it into a plugin

PR #4220, opened 2024-11-16
Fix submission with "invite to proposal" status stuck in draft if language is not english

### LINKED ISSUE #4211: Proposal stuck in draft when language code is set
When `LANGUAGE_CODE` is set to another language every application gets stuck in draft.

A separate issue is that the menu does not show if the applicant only have draft applications.

**To Reproduce**

1. Apply for a proposal
2. Fill the form
3. Click submit
4. The proposal stock in draft

**Expected behavior**

The form should go to review, even though a language is set on env vars
**Priority**

**Affected roles**

- Applicants

theskumar:
The current implementation tries to find the transition based on the `name` property of the transitions available. The name of the transition is a translated string so while submitting a proposal the key "submit" from the `request.POST` could not be found in the available transition.

This PR updates the logic to fetch the transition target, by searching through the newly added marker on when to auto transition if the application is submitted.

Fixes #4211 

See https://github.com/HyphaApp/hypha/pull/4220/commits/e6656d62f6db57426bb15f298127c106ea4f6397 for patch.

## Test Steps

 - Update the settings `LANGUAGE_CODE` to `sv`
 - For a fund with "Concept and Proposal" workflow, create a submission and approve the concept.
 - Now, logic as applicant and try to submit the "Proposal" while you application is in "Invited for Proposal" stage
 - The application should go through
 - If this patch is not applied, the status won't update.

### DIFF
--- a/hypha/apply/funds/views.py
+++ b/hypha/apply/funds/views.py
@@ -1260,13 +1260,6 @@ class BaseSubmissionEditView(UpdateView):
 
     model = ApplicationSubmission
 
-    @property
-    def transitions(self):
-        transitions = self.object.get_available_user_status_transitions(
-            self.request.user
-        )
-        return {transition.name: transition for transition in transitions}
-
     def render_preview(self, request: HttpRequest, form: BaseModelForm) -> HttpResponse:
         """Gets a rendered preview of a form
 
@@ -1327,6 +1320,25 @@ class BaseSubmissionEditView(UpdateView):
             return assigned_fund.open_round
         return False
 
+    def get_on_submit_transition(self, user):
+        """Gets the transition that should be triggered when a form is submitted.
+
+        Checks all available status transitions for the current user and returns the first
+        one that has trigger_on_submit=True in its custom settings.
+
+        Returns:
+            dict: The transition configuration dictionary with trigger_on_submit=True,
+                or None if no matching transition is found.
+        """
+        return next(
+            (
+                t
+                for t in self.object.get_available_user_status_transitions(user)
+                if t.custom.get("trigger_on_submit", False)
+            ),
+            None,
+        )
+
     def form_valid(self, form: BaseModelForm) -> HttpResponse:
         """Handle the form returned from a `SubmissionEditView`.
 
@@ -1379,19 +1391,18 @@ class BaseSubmissionEditView(UpdateView):
                 related=revision,
             )
 
-        action = set(self.request.POST.keys()) & set(self.transitions.keys())
-        try:
-            transition = self.transitions[action.pop()]
-        except KeyError:
-            pass
-        else:
-            self.object.perform_transition(
-                transition.target,
-                self.request.user,
-                request=self.request,
-                notify=not (revision or submitting_proposal)
-                or self.object.status == DRAFT_STATE,  # Use the other notification
-            )
+        if "submit" in self.request.POST:
+            if transition := self.get_on_submit_transition(self.request.user):
+                notify = (
+                    not (revision or submitting_proposal)
+                    or self.object.status == DRAFT_STATE,
+                )
+                self.object.perform_transition(
+                    transition.target,
+                    self.request.user,
+                    request=self.request,
+                    notify=notify,  # Use the other notification
+                )
 
         # Required for django-file-form: delete temporary files for the new files
         # uploaded while edit.
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -286,7 +286,7 @@ def transition_id(target, phase):
 class AddTransitions(models.base.ModelBase):
     def __new__(cls, name, bases, attrs, **kwargs):
         for workflow in WORKFLOWS.values():
-            for phase, data in workflow.items():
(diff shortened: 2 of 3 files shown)


hypha/apply/funds/views.py
```diff
@@ -1388,12 +1380,33 @@ def form_valid(self, form: BaseModelForm) -> HttpResponse:
...
             )
+            # Get the transition corresponding to the object's current phase name
+            transition = next(
+                (t for t in transitions if t.source == self.object.phase.name), None
+            )
+
+            if transition:
+                # Only notify for initial submission or drafts
+                notify = (
+                    not (revision or submitting_proposal)
```
frjo, line 1397: `submitting_proposal` will always be true inside `if submitting_proposal:`

frjo (changes requested):
Excellent progress! This fixes the invite for proposal draft.

It does not seem to fix an initial draft, when e.g. an applicant apply for a lab or content note and begin with saving as draft.

hypha/apply/funds/views.py
theskumar: Nice catch.

theskumar:
> Excellent progress! This fixes the invite for proposal draft.
> 
> It does not seem to fix an initial draft, when e.g. an applicant apply for a lab or content note and begin with saving as draft.

It is a oversight from my side, the latest should work. https://github.com/HyphaApp/hypha/pull/4220/commits/def0e76759a2eec577d57cdb8218a2fdb402bb8b

theskumar:
Test are failing, I'll take a deeper look.

frjo (approved):

theskumar:
After carefully reviewing, I realized that the application auto-advances to the next `phase` at different phases, and there could be multiple phases it can transition to when the application is at a specific phase, say `more info required` phase. Therefore, we need an explicit marker to decide:

1. Whether to auto advance or not, if the user submits the application. 
2. which target phase to pick when the user submit a application. 

The current fix, while functional, has an issue where it will always transition to the first available transition at that Phase.

I think we can solve this we need to update the workflow spec itself, that enables to deterministically know above two:  

1. Remove the translation marker,  on the "Submit" display name, I believe it's not displayed in the UI anywhere.
2. Introduce a new custom property on the target phase that indicates whether to auto-advance to this phase if the user submits the application at the source phase.

Screenshot 2024-11-18 at 3  32 04@2x

@frjo @wes-otf Any preference on the (1) or (2). My take would be to go for an explicit custom marker, even though it's little verbose, for the shake of clarity.

sandeepsajan0:
@theskumar I am not sure about the best approach but what do you think about having a predefined method that can return both types of keys as translated(in same lang) and we can call that method wherever in code we are comparing these keys like [here](https://github.com/HyphaApp/hypha/pull/4220/files#diff-cf22f20205f278f19b4b22cd6e5decb2f0a9548aba7e3a779b7ff9e0b1d117faL1406). 
Ex:  
```
def translated_keys(post_keys, transition_keys):
    ......
    return translated_post_keys, translated_transition_keys

```

theskumar:
> @theskumar I am not sure about the best approach but what do you think about having a predefined method that can return both types of keys as translated(in same lang) and we can call that method wherever in code we are comparing these keys like [here](https://github.com/HyphaApp/hypha/pull/4220/files#diff-cf22f20205f278f19b4b22cd6e5decb2f0a9548aba7e3a779b7ff9e0b1d117faL1406). Ex:
> 
> ```
> def translated_keys(post_keys, transition_keys):
>     ......
>     return translated_post_keys, translated_transition_keys
> ```

Thank for the input, initial tried getting the untranslated version of the `display` but it looks by the time it reaches here as a key of the dictionaries, this is already translated, and it make sense as the translated string will be needed for hashing in in the dict. 

I'm in a opinion that we should not be doing comparison on the basis of translated strings. 

Also, since the `post.keys` are never marked as translatable, you can't get equivalent translated string reliably.

frjo:
I think option two is best.

I agree that we should never compare strings that might be translated.

I see no other places but in this `form_valid()` function that we do this check. Correct?

Progressing when saving a submission should only happen when submitting a draft and when submitting a "More information is needed".

Having something explicit in the workflows for these two statuses would make it more clear what is happening.

theskumar:
> I see no other places but in this `form_valid()` function that we do this check. Correct?

Yes, that's when we check if the "Submit" action is present in the targets and then `perform_transition` to that phase.

I'll check other places as well, but it doesn't seem it should have been used. (If we are using translated string for comparisons at other places, that would probably be another ticket for us to go over)

---

For the custom route, I'll need to modify some parts of how the phases and transitions are initialized, but it shouldn't be too complicated. I'll be updating the this PR. 

_PS: Down the road, we must split this workflow file into different module. I see two `Request` object defined in the same file, one related to Stage and another for Workflow._

theskumar:
Updated the logic to add new marker in the workflow definition indicating this particular phase to pick when the user submits the application and then auto transition to it.

frjo (approved):
Works nicely for me.

wes-otf:
worked great, really nice work @theskumar!

PR #4188, opened 2024-10-28
Remove fancybox modal for invoicing and reporting

sandeepsajan0:
Fixes #ISSUEID

- [ ] Use htmx/alpine for Invoice status update modal
- [ ] Use htmx/alpine for report frequency update modal
- [ ] Use htmx/alpine for Update project status modal( Closing and Complete)
- [ ] Use htmx/alpine for lead update modal
- [ ] Use htmx/alpine for project title modal
- [ ] Remove fancybox for report save, sabmit and skip confirmation.
- [ ] Remove fancybox imports from project code

## Test Steps

 - [ ] …

PR #4226, opened 2024-11-25
Remove old submissions table related code and removed fancybox

### LINKED ISSUE #4224: Remove the old submission all table
This tables batch actions in one of the last places we use the fancybox library.

The new submission all table is the default since some time and feedback say it completely replaces the old view, with many improvements to boost.

sandeepsajan0:
Fixes #4224 

- [ ] Removed old submissions all table code
- [ ] Removed old submission table batch actions
- [ ] Use htmx for invoice batch status update action

## Test Steps

 - [ ] All batch actions are working fine for new table.
 - [ ] Invoice status batch update is working fine.


frjo:
@sandeepsajan0 Added task to remove obsolete CSS, notices that when I created https://github.com/HyphaApp/hypha/pull/4230.

sandeepsajan0:
When I was looking at its failed test([user/test_middleware](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/users/tests/test_middleware.py#L35)), I found an interesting issue that is somehow working fine in UI and passed the tests as well. 
Actually, [this url](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/funds/urls.py#L94) was added as a particular submission success page that has to be exempted from 2FA so we added it [here in middleware](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/users/middleware.py#L21) but didn't add the submission id as it has to be dynamic. And it is working with middleware/UI because we are checking URLs by using ['startswith'](https://github.com/HyphaApp/hypha/blob/main/hypha/apply/users/middleware.py#L80) but in the above test we are trying to access the same url without submission id that was unintentionally matching with path 'submissions/\<str:status>\' and then auth(redirect) and passing the test. As that path has been removed in this PR, test failed so for now I just added a submission and used that submission id to match success url.

But I think for the long-term instead of accessing the url we should only be testing the middleware code to check if it is going through or not? @frjo @theskumar

frjo:
@sandeepsajan0 I have no merged https://github.com/HyphaApp/hypha/pull/4188 in to main.

frjo:
Did some minor lint fixes.

Notices this "+131 −1755", that is a lot of code lines we no longer need to maintain!

frjo (approved):

frjo:
@wes-otf I believe @sandeepsajan0 fixed a bug regarding batch actions in the new submission all table as well in this PR.

wes-otf:
seems like all worked pretty well! like @frjo said definitely nice to have that much to not maintain! Regarding the first test step, the only issue I saw with batch updating was I tried to do a batch action of dismissing [these 6](https://test-apply.hypha.app/apply/submissions/all/?format=table&query=wes%40opentech.fund&status=need-screening) and got a [server error](https://otf.sentry.io/issues/6138319823/?project=5823154&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=90d&stream_index=0). @sandeepsajan0 / @frjo if that feels out of scope for this PR I can write up another issue

frjo:
@wes-otf @sandeepsajan0 The submissions (only three show up on test today with this search) use different determination forms. This makes it impossible to dismiss them together in one batch run.

For status changes that require a determination (Accept, Dismiss and Invite for proposal) the batch status dropdown needs to check that the selected submissions use the same determination form. If easier we can check that they belong to the same round/lab.

The error message is not correct I believe, "Submissions expect different forms - please contact admin". The problem is that the forms are **not** the same from what I understand.

Issue #4239, opened 2024-12-04
Create a unified reports table


wes-otf:
Currently there are two views for looking at reporting, neither of which seems to give the user the ability to easily filter the reports by criteria of date, status, project, etc. as one would need to bounce between the report table & report list to do so. The difference between the two can also be confusing

It seems like it would make sense to make a single table where this could be done that is simpler but similar to the all submissions view.

PR #4240, opened 2024-12-04
Combine the reports/reporting nav items

wes-otf:
Building off of the conversation had in #4051, this adds the navigation to the reporting table to be under the submitted reports view rather than being in the main nav bar

This makes things a little simpler and explicit for the user but I think this is a temp fix, with #4239 being a permanent solution 

## Test Steps

 - [ ] Ensure the reporting table is accessible via `Reports` in the nav menu -> `View all reports` in the heading

### DIFF
--- a/hypha/apply/projects/templates/application_projects/report_list.html
+++ b/hypha/apply/projects/templates/application_projects/report_list.html
@@ -9,7 +9,10 @@
 
     {% adminbar %}
         {% slot header %}{% trans "Submitted Reports" %} ({{ table.rows|length }}){% endslot %}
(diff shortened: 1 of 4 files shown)


frjo:
@wes-otf Good improvement I think.

If the current text links is to subtle we could use tabs instead. Maybe in the style we use for the Table/List view on the submission all view.

frjo (approved):
Works well in my testing.

wes-otf:
> Maybe in the style we use for the Table/List view on the submission all view.

@frjo I like this idea, would make everything seem more cohesive - potentially something to consider in #4329

PR #4251, opened 2024-12-05
Updated python packages December 2024.

frjo:
The flow to update are as follow:

1. Run `uv tree -d 1 --outdated` to get  list of packages with updates.
2. Update the package version in `pyproject.toml`
3. Run `uv sync`, this will install all the new packages and sync the `uv.lock` file.
    * The command will alert of any dependency conflicts, you need to resolve these.
4. Run `pre-commit run --all-files`, this will run the `uv` commands needed to update all the requirements* files. (Info is taken from the `uv.lock` file so it must be updated before this step.)


theskumar:
For just updating the lock file without doing anything else, `uv lock --upgrade` can also be used.  If pre-precomit is enabled, just adding a new commit should generate the update the `requirements/*.txt` files. 

`uv sync` is obviously the best command to do all the things, like upgrading python, removing any packages not mentioned in the lock file etc. 

If the tests passes and regression passes, this looks good to me, I'll spend sometime doing smoke test.

wes-otf:
I think this is ready, in all my testing today I didn't seem to have any issues

PR #4261, opened 2024-12-06
Separate SOW & PAF form editing

### LINKED ISSUE #4204: Project Form/SOW workflow can be confusing
When a Project Form & SOW exist on a fund, populating them in Hypha can be a bit confusing. From my experience, both are initially populated under `Project Form`. Once you select `Fill in`, you first populate the Project Form then the SOW immediately under it. Once you press `Save`, that's when the SOW field will appear under `Project Documents`. But, to edit it the user needs to select edit on `Project Form and scroll back down`

*before filling in PAF/SOW*
Screenshot 2024-11-08 at 11 44 01

*after hitting save while editing `Project Form`*
Screenshot 2024-11-08 at 11 57 20

I'm not sure if there's context I'm missing or if this is a bug but to me it'd make more sense to have the SOW be it's own line from the start, rather than having to go through Project Forms to edit it.

wes-otf:
Fixes #4204. Moves the SOW out into it's own form when editing. This also removes the `user_has_updated_details` attribute in favor of using a property for both the project form & SOW that checks the field_data on the respective form. This allows for tracking of the project form & SOW independently. 

Also a few small aesthetic changes bundled in, like margin additions & hiding of submission attachments sidebar when there's none.

## Screenshots

*appearance when first creating a project*
Screenshot 2024-12-06 at 16 07 24

*appearance after editing both forms*
Screenshot 2024-12-06 at 16 04 29

## Test Steps

 - [ ] When viewing the details of a project with both a Project Form & SOW, ensure that both appear under `Project documents`.
 - [ ] Confirm that both can be edited, saved, and viewed on their own
 - [ ] Confirm that when creating a project from a fund that has a Project Form & SOW, a todo list item is created for both


hypha/static_src/sass/components/_docs-block.scss
```diff
@@ -168,6 +167,7 @@
         &__actions {
             display: flex;
             align-items: center;
+            margin-left: auto;
```
frjo: Please use `margin-inline-start` instead.

hypha/apply/projects/views/project.py
```diff
@@ -44,12 +44,13 @@
     PAF_REQUIRED_CHANGES,
     PAF_WAITING_APPROVAL,
     PAF_WAITING_ASSIGNEE,
-    PROJECT_SUBMIT_PAF,
+    PROJECT_SUBMIT_PF,
```
frjo: As a machine name we use "PAF" everywhere so I think it is best to stick with that.

I do not mind changing it to "PF" or "PROJECT_FORM" etc. but it should be done everywhere.

hypha/apply/projects/templates/application_projects/project_sow_detail.html
```diff
@@ -24,10 +24,10 @@
     <div class="wrapper wrapper--large wrapper--tabs">
         <div class="wrapper wrapper--sidebar">
             <article class="wrapper--sidebar--inner">
-                <h4 class="mb-2">{% trans "Project scope of work (SOW)" %}</h4>
+                <h4 class="mb-2">{% trans "Scope of Work" %}</h4>
                 <div class="card card--solid">
                     {% if object.sow.output_answers %}
-                        <div class="rich-text rich-text--answers">
+                        <div class="simplified__paf_answers">
```
frjo: This class was removed in https://github.com/HyphaApp/hypha/pull/4196.

frjo (changes requested):
This is a good change, it was a bit of a hack having them merged in the first place.

Found some minor issues.

How will the SOW now fit in to the approval workflow?

wes-otf:
all that left is getting these tests passing - a bit of a battle because my flag uses the project's `form_data`, but that won't get written unless a valid FormFieldBlock wrapped in a StreamValue is passed in the `save` method kwargs. it's giving me a hard time when I try to mock it

https://github.com/HyphaApp/hypha/blob/acdbea36177c90adb5cf24a7a61dd6d527a33f9e/hypha/apply/projects/forms/project.py#L137-L143

hypha/apply/projects/templates/application_projects/project_sow_detail.html
frjo: Are not the "rich-text rich-text--answers" ok to use? We use them in most other places when we render streamfields.

hypha/apply/projects/views/project.py
frjo: Ping

hypha/static_src/sass/components/_docs-block.scss
frjo: ping

frjo:
@wes-otf How vital is the test? We could remove it.

wes-otf:
@frjo Yeah that sounds good - I don't think it's vital as it'd really just be checking if `form_data` is an empty dict. I tossed it and added your suggestions

frjo (approved):

wes-otf:
grabbed some final bugs but should be ready to be deployed to test now @frjo

frjo:
@wes-otf When I download a SOW at http://apply.hypha.test:9001/apply/projects/4/sow/ no fields are added to the exports. I guess something changes regarding getting the fields?

When I test the main branch locally it seems to work as it should.

wes-otf:
@frjo great catch! my logic was copy & pasted, checking `self` rather than `project` - also removed some mentions of the SOW in the PAF export.

frjo:
Deploying the latest to test now.

frjo:
Can confirm that the bug in sow exports are now fixed. Looks all good to me.

wes-otf:
@frjo thanks for testing - I think it's ready to go then! I did more testing on my end too and didn't see anything

PR #4266, opened 2024-12-09
Set dir auto via js on all p tags in applications.

### LINKED ISSUE #4265: Set dir auto on all p tags in applications so browser automatically set directionality correctly
By setting `dir="auto"` on all paragraphs (p-tags) browsers will set the directionality according to the first characters with strong directionality.

This attribute does not inherit and cannot be set via css "direction". It must be set via javascript since the p tags are user generated.

This attribute should only be used on user generated text where directionality is not known beforehand.

frjo:
Fixes #4265

With dir auto:

Skärmavbild 2024-12-09 kl  08 31 35

Without dir auto:

Skärmavbild 2024-12-09 kl  08 31 53

## Test Steps

 - [ ] Submit an application with some answers in ltr langiage and some in rtl languages (e.g. Arabic). The paragraphs with rtl languages should show up as rtl automatically.

### DIFF
--- a/hypha/static_src/javascript/submission-text-cleanup.js
+++ b/hypha/static_src/javascript/submission-text-cleanup.js
@@ -2,16 +2,20 @@
     // Select all rich text answers.
     const richtextanswers = document.querySelector(".rich-text--answers");
 
-    // Remove p tags with only whitespace inside.
+    // Select all rich text answers paragraphes.
     const richtextanswers_paras = richtextanswers.querySelectorAll("p");
     richtextanswers_paras.forEach(function (para) {
+        // Remove p tags with only whitespace inside.
         if (para.textContent.trim() === "") {
             para.remove();
         }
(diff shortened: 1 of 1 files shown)


theskumar:
Looks good. Thinking out loud:

1) Do we need to cover up for for field types?
2) ~Is there is way to apply it to parent level element, ie..g on the submission detail something i.e. applying to all paragraphs on the submissions. something equivalent to css property: `p > *`~ You already answered in the parent issue.

frjo:
The solution in this PR will only work for markdown and rich text textareas. Can be improved but should cover most of the texts.

theskumar (approved):

theskumar:
> The solution in this PR will only work for markdown and rich text textareas. Can be improved but should cover most of the texts.

Make sense to me. As next we should look into adding this other fields as well, otherwise you end up with mixed rendering.

PR #4272, opened 2024-12-10
Set default reporting frequency as disable

### LINKED ISSUE #4233: Set reporting frequency default to disabled
The default is monthly currently. This will likely suit few.

A better default would be to disable reporting by default.

Later on we could add a setting on funds/labs for the reporting frequency default.

sandeepsajan0:
Fixes #4233 

## Test Steps

 - [ ] …


theskumar (approved):

frjo:
@wes-otf What do you think OTF prefer.

To me it feels like disabling reports by default is the best option.

The current default is once per month and in most cases that will need to be changed in any case I assume.

A future improvement could be to add a report frequency default on funds/labs.

frjo:
This PR has a migration conflict with https://github.com/HyphaApp/hypha/pull/4261 so need to hold a bit.

PR #4273, opened 2024-12-10
Convert jquery to vanilla, remove unused js, fix minor bugs

frjo:
* Removed symbol-es6.min.js, not used and we do not support IE since a long time.
* Removed url-search-params.js, all browsers support URLSearchParams nowadays.
* Removed all-reviews-table.js, not used anywhere.
* Removed mobile styling/js toggle of sidebar actions. They are responsive as they are.
* Replace two-factor-focus.js with one line inline vanilla js.
* Avoid double adding of file-uploads.js.
* Avoid double adding of comment_form.media.*.
* Only load daterangepicker js where it is used on submission/all.
* Converted file-uploads.js to vanilla js.
* Use gap on wrapper--sidebar class instead of padding on wrapper--sidebar--inner.
* Fix bug that made status not show on projects on mobile.

## Test Steps

 - [ ] Test that all actions in sidebar on submissions, projects, invoices show up as they should on small screens.
 - [ ] Test that all file upload fields work as they should.
 - [ ] Test that form groups work as they should.
 - [ ] Test that comment forms work as they should.


hypha/static_src/javascript/csrf-token-ajax.js
```diff
@@ -1,30 +1,4 @@
...
-
-            if (previousValues === formValues) {
-                // Previously submitted - don't submit again
-                e.preventDefault();
-            } else {
-                $form.attr("data-django-form-submit-last", formValues);
-            }
-        });
-
     // Setting the CSRF token on AJAX requests.
```
frjo: @theskumar I believe I added this when I created the flag feature and used fetch/ajax to set/unset the flags. That is now converted in to htmx, do we need it?

At https://django-htmx.readthedocs.io/en/latest/tips.html#make-htmx-pass-the-csrf-token it seems csrftoken is needed when doing `hx-post`. Maybe we can remove the script and add the `hx-headers` stuff instead whenever we use `hx-post`?

frjo:
I will stop here, enough things to test as it is.

Will be a part 2.

hypha/static_src/javascript/csrf-token-ajax.js
theskumar: `hx-headers` is added already and works as expected. This code is good be removed if we are not making any fetch request.

wes-otf:
Only issue I was seeing on test was with group fields not working - looked like they were revealed from the start. Heres the [lab](https://test-apply.hypha.app/wes-test-lab/) that I was testing with along with the [form for that lab](https://test-apply.hypha.app/admin/funds/applicationform/edit/237/). Let me know if I'm missing something - nice changes overall though! excited to have less jquery

frjo:
@wes-otf You found a bug in form-group-toggle.js. I believe I have fixed it now, new version deploying to test now.

frjo:
@wes-otf There are more bugs, reverting to old jQuery version. Deploying to test now.

wes-otf:
Looks like it's fixed to me on the test instance!

PR #4287, opened 2024-12-17
More clean up in how we load js and css files

frjo:
- [x] Remove unused document-category.js.
- [x] Replaced toggle-payment-block.js with alpine code.
- [x] Load all *media.js|css that needs jQuery in block extra or directly in base.html. Move jQuery back to the bottom of the page.
- [x] Merge application-form-links-new-window.js in to application-form.js.
- [x] Load comment_form.media on submission and project detail templates directly.
- [x] Remove outdated block extras and move block extra css just above block extra js in templates.

## Test Steps

 - [ ] Test all roles dashboards so all needed js and css gets loaded.
 - [ ] Test submission and project pages so all needed js and css gets loaded.
 - [ ] Test that all file upload fields, especially in htmx dialog, work as they should.


hypha/templates/base.html
```diff
@@ -125,6 +122,11 @@
         {% include "includes/_modal-placeholder.html" %}
         {% include "includes/_toast-placeholder.html" %}
 
+        <!-- We still need jQuery for select2, daterangepicker and a few custom scripts. -->
+        <!-- Please do not add new features that require jQuery! -->
+        <script src="{% static 'js/vendor/jquery.min.js' %}"></script>
+
+        {{ form.media.js }}
```
frjo, line 135: @theskumar Is this ok, any drawbacks?

Seems to have no ill effect. On many pages there are no "form.media*" but then it just outputs nothing silently.

This is needed on many htmx loaded dialogs etc. Loading it here we make sure it is loaded after jQuery and available when needed.

theskumar: I don't see much of an issue. I would suggest wrapping it in a `if` condition. I probably won't happen, but do check for duplicate rendering, one via the `{{ form }}` itself or in the inline template and one in the base template.

For the dialog/htmx, the file upload widget and select2 might break, they don't reinitialize if new element is added dynamically.

wes-otf:
I didn't see any issues in my testing, seems pretty good!

frjo:
As @theskumar predicted this PR breaks file uploads in htmx dialogs.

hypha/static_src/javascript/file-uploads.js
```diff
@@ -1,26 +1,30 @@
...
       window.initUploadFields(form);
 
       // Hide wrapper elements for hidden inputs added by django-file-form
-      $("input[type=hidden]").closest(".form__group").hide();
+      document.querySelectorAll("input[type=hidden]").forEach(function (input) {
+        const closestFormGroup = input.closest(".form__group");
+        if (closestFormGroup) {
+          closestFormGroup.style.display = "none";
+        }
+      });
```
frjo, line 27: Do not think this part is needed anymore. Did not find a place where it had any effect. Let it be for now since we are changing so many other things.

hypha/static_src/javascript/file-uploads.js
```diff
@@ -1,26 +1,30 @@
-jQuery(function ($) {
+// We use htmx.onLoad() so it will initilise file uploads in htmx dialogs.
+htmx.onLoad(function () {
```
frjo, line 2: This is a really neat function!

wes-otf:
was using all dialogs quite a bit today and didn't see any issues jumping out at me so I think all is good unless there was something else specific you wanted me to look at @frjo

PR #4293, opened 2024-12-17
Removed unused code related to the v1 & Intacct API

wes-otf:
Removes unused/unmaintained code that was pertaining to the previous React frontend along with the attempted Intacct API integrations. The only RESTful endpoint left is `/api/v1/rounds/open`, which \(when [utilizing an API key created in Django Admin](https://florimondmanca.github.io/djangorestframework-api-key/guide/#creating-and-managing-api-keys)\) will return a JSON output of all available rounds.

## Test Steps

 - [ ] Ensure Hypha works as it did before (especially in regards to invoicing, as that's where most of the Intacct/deliverable code was actually implemented)
 - [ ] Confirm the `/api/v1/rounds/open` endpoint works as it did before


wes-otf:
wouldn't be hard to toss django-rest-framework altogether in this scenario as we talked about on Monday but wasn't sure if we wanted to do a manual API Key implementation or if there would be any additional APIs in the future that would need some framework. 

also added some documentation but if we do end up adding a lot more endpoints in the future it may be smart to leverage swagger or something

wes-otf:
unpublished rounds also show up as open in the endpoint output because of #3867, which will probably be my next focus

theskumar:
It would be great to handle the removal of drf-dependency in another PR. Manual API key is not too hard. We can use django-ninja, but I don't see the need for it now, simple JSONResponse would do. 

Regarding docs, simple Markdown-based docs will work best. That allows you to document any rate limit, authentication, and sample requests.

hypha/apply/projects/views/payment.py
```diff
@@ -229,13 +229,12 @@ def get_success_url(self):
...
 
-    def get_context_data(self, **kwargs):
-        invoice = self.get_object()
-        project = invoice.project
-        deliverables = project.deliverables.all()
-        return super().get_context_data(**kwargs, deliverables=deliverables)
+    # def get_context_data(self, **kwargs):
+    #     invoice = self.get_object()
+    #     project = invoice.project
+    #     return super().get_context_data(**kwargs)
```
theskumar: Remove this completely?

frjo:
I think we should remove it all. 

The "rounds/open" end point can be rebuilt and token is likely overkill for that endpoint. Just rate limiting it should be enough, the content you get is the same that is displayed on the front page in any case.

frjo:
A lot of work went it to this code but since we are not using it it's good to get out of the way.

I do not know of anyone using the "rounds/open" endpoint.

hypha/apply/projects/views/payment.py
wes-otf: ah great catch! I originally intended to, slipped by me

wes-otf:
I tossed the endpoint, could definitely see the utility of one like it in the future though

frjo:
@wes-otf  Do a `uv remove djangorestframework-api-key djangorestframework drf-nested-routers` (I think they are not needed anymore.)

That will update `pyproject.toml` and `uv.lock`.

Then attempt to commit and the pre-commit hook will notice that the `requirements/*.txt` needs updating and do that. Then you commit again and push. (or run `pre-commit run --all-files` before you commit.)

wes-otf:
@frjo thanks for that pointer - still learning the uv ropes! I believe all should be good to go now

frjo (approved):

PR #4294, opened 2024-12-18
Update the UI to switch user aka become user

### LINKED ISSUE #4131: Move the "Become" functionality to the user menu
Clicking "Become" in the user menu should open a dialog box with the become form.

theskumar:
Fixes #4131 

- Rename to "switch user"
- Can switch from the user menu
- Switch to choices.js from Select2
- Update the hijacked view, shows an alert at the top

Screenshot 2024-12-18 at 9  22 59@2x
Screenshot 2024-12-18 at 9  23 31@2x
Screenshot 2024-12-18 at 9  23 53@2x
Screenshot 2024-12-18 at 9  24 19@2x

### DIFF
--- a/hypha/apply/users/forms.py
+++ b/hypha/apply/users/forms.py
@@ -4,7 +4,6 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import AuthenticationForm
 from django.template.defaultfilters import mark_safe
 from django.utils.translation import gettext_lazy as _
-from django_select2.forms import Select2Widget
 from rolepermissions import roles
 from wagtail.users.forms import UserCreationForm, UserEditForm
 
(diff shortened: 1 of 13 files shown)


frjo:
This looks and works really nicely! Will be a pleasure to use when we get it deployed.

I have one issue. The function `get_become_user_choices()` is slow. I think `get_role_names()` takes up most of the time.

With 4000 users it takes around three (3) seconds. I think OTF has even more users. This should not be used in production but even test data can have a lot of users.

Possible to speed it up a bit, showing the role in the list is useful so good if we can keep that. Maybe caching the list would work?

hypha/apply/users/forms.py
```diff
@@ -216,11 +215,33 @@ def clean_full_name(self):
...
+    """Returns list of active non-superusers with their roles as choice tuples."""
+    active_users = User.objects.filter(is_active=True, is_superuser=False)
+    choices = []
+
+    for user in active_users:
+        role_names = user.get_role_names()
+        if role_names:
+            roles_html = ", ".join(
+                [
+                    f'<span class="inline-flex items-center px-2.5 py-0.5 rounded-full text-xs font-medium bg-gray-100 text-gray-800">{str(g)}</span>'
```
frjo: This seems not to show up in the dialog select  list?

frjo (changes requested):

theskumar:
> I have one issue. The function get_become_user_choices() is slow. I think get_role_names() takes up most of the time.

> With 4000 users it takes around three (3) seconds. I think OTF has even more users. This should not be used in production but even test data can have a lot of users.

Will look into it.

wes-otf:
This is super slick! I really like the banner as opposed to the pop up window that you need to dismiss every time you reload the page. Nice improvement!

hypha/apply/users/forms.py
theskumar: This doesn't seem be supported, removed it.

theskumar:
@frjo Updated the query to fetch the groups. It should always come down to 2 queries now. That said, if we face issue in the test site, we can switch to fetching the options as you type, via a search api.

theskumar:
The new implementation of `user.is_*` checks will improve the performance overall at multiple places.

frjo (approved):
Now the load time is more or less instant and works equally well in other respects.

wes-otf:
@frjo has this been deployed to test?

frjo:
Nope, forgot to push. Deploying now.

PR #4296, opened 2024-12-18
Update referrer-policy to strict-origin-when-cross-origin, remove django-referrer-policy packages, built in to Django 3 and later

frjo:


theskumar (approved):
💯

PR #4302, opened 2024-12-20
Remove public site apps and migrations

theskumar:
This is final step in the public site removal


theskumar:
It also reduce the test duration on the CI by another 20 to 25 sec. 🎉

frjo (approved):
Thanks for remembering this. Good time to get this removed.

PR #4304, opened 2024-12-24
Fix batch determination different forms issue

### LINKED ISSUE #4290: Batch determination breaks with different form submissions
**Describe the bug**

Batch determination raise Value error if selected submissions have different forms. https://github.com/HyphaApp/hypha/pull/4226#issuecomment-2544838609

There are other validations as well that might cause the similar issue.

**To Reproduce**

1. Select submission with different determination forms
2. Try to dismiss them.

**Expected behavior**

Problematic statuses shouldn't be available for batch actions, it should check the validation on selected submissions and hide those options.

Just for the safer side, try to show error message in UI without performing batch actions instead of raising the error.

sandeepsajan0:
Fixes #4290 

## Test Steps

 - [ ] Try to perform batch update status action on submissions
 - [ ] Try to update some batch determinations(dismissed, accept and more info), all should work fine.

### DIFF
--- a/hypha/apply/funds/utils.py
+++ b/hypha/apply/funds/utils.py
@@ -209,11 +209,17 @@ def check_submissions_same_determination_form(submissions):
     """
 
     same_form = True
-    # check form id
-    determination_form_ids = [
-        submission.get_from_parent("determination_forms").first().id
-        for submission in submissions
-    ]
+    try:
+        # check form id
+        determination_form_ids = [
+            submission.get_from_parent("determination_forms").first().id
+            for submission in submissions
+        ]
+    except Exception:
+        # if there is a form id error, handles old determination form issues
+        same_form = False
+        return same_form
+
     if any(d_id != determination_form_ids[0] for d_id in determination_form_ids):
         same_form = False
     return same_form
--- a/hypha/apply/funds/utils.py
+++ b/hypha/apply/funds/utils.py
@@ -198,3 +198,22 @@ def get_copied_form_name(original_form_name: str) -> str:
     # If a copied timestamp already exists, remove it
     new_name = re.sub(name_reg, "", original_form_name)
     return f"{new_name} ({copy_str.format(copy_time=copy_time)})"
+
+
+def check_submissions_same_determination_form(submissions):
+    """
+    Checks if all the submission have same determination forms.
+
+    We can not create batch determination with submissions using two different
+    type of forms.
+    """
+
+    same_form = True
+    # check form id
+    determination_form_ids = [
+        submission.get_from_parent("determination_forms").first().id
+        for submission in submissions
+    ]
+    if any(d_id != determination_form_ids[0] for d_id in determination_form_ids):
+        same_form = False
+    return same_form
--- a/hypha/apply/funds/views/partials.py
+++ b/hypha/apply/funds/views/partials.py
@@ -37,7 +37,7 @@ from ..utils import (
     get_statuses_as_params,
     status_and_phases_mapping,
 )
-from ..workflows.constants import PHASES_MAPPING
+from ..workflows.constants import DETERMINATION_OUTCOMES, PHASES_MAPPING
 
 User = get_user_model()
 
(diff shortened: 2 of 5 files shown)


hypha/apply/funds/views_all.py
```diff
@@ -355,11 +366,47 @@ def bulk_update_submissions_status(request: HttpRequest) -> HttpResponse:
 
     submissions = ApplicationSubmission.objects.filter(id__in=submission_ids)
 
-    redirect: HttpResponse = BatchDeterminationCreateView.should_redirect(
-        request, submissions, transitions
-    )
-    if redirect:
-        return HttpResponseClientRedirect(redirect.url)
+    # should redirect
```
sandeepsajan0: Moved `should_redirect` method directly in here to avoid raising an Error, instead showing the error to the user without performing the batch action. 

It is hard to find all the cases at once so I guess it is a better way to avoid 500. We can include the cases to validation as soon as it is reported.

frjo:
@sandeepsajan0 Can you take a look at the merge conflicts? I will put it on test when that is cleared up.

hypha/apply/funds/views_partials.py
```diff
@@ -327,6 +330,13 @@ def sub_menu_update_status(request: HttpRequest) -> HttpResponse:
         else []
     )
 
+    # hide determination actions if submissions have different forms
+    if not allow_determination:
+        determination_actions = ["Dismiss", "Request More Information", "Accept"]
```
frjo: Does this work with translations?

frjo (changes requested):
Needs a rebase.

hypha/apply/funds/views/partials.py
```diff
@@ -317,6 +320,13 @@ def sub_menu_update_status(request: HttpRequest) -> HttpResponse:
         else []
     )
 
+    # hide determination actions if submissions have different forms
+    if not allow_determination:
+        determination_actions = ["Dismiss", "Request More Information", "Accept"]
```
frjo: @sandeepsajan0 Will this work with translations?

frjo (changes requested):

wes-otf:
In testing this definitely was a lot smoother! @sandeepsajan0 I noticed when I selected a couple of lab submissions they all gave me the option to bulk dismiss, and the same with when I selected fund submissions, but combining them removed that option - is this because of the difference in determination forms? if so I think this is ready

sandeepsajan0:
>combining them removed that option - is this because of the difference in determination forms?

@wes-otf Yes, determination actions are not available for different forms now. Because for determination we need to render determination form and it is impossible to render different form for a single bulk action.

hypha/apply/funds/views/partials.py
sandeepsajan0: @frjo It is fixed now, used machine names(DETERMINATION_OUTCOMES) to compare.

hypha/apply/funds/utils.py
```diff
@@ -198,3 +198,22 @@ def get_copied_form_name(original_form_name: str) -> str:
...
+    We can not create batch determination with submissions using two different
+    type of forms.
+    """
+
+    same_form = True
+    # check form id
+    determination_form_ids = [
+        submission.get_from_parent("determination_forms").first().id
+        for submission in submissions
+    ]
```
frjo: There is an edge case where no determination form id exist. I believe this only affects OTF that still have some funds/labs that use the old hardcoded determination forms, before we made them into stream field forms.

Unsure how to best solve this issue without messing up the code more than needed.

@wes I assume all current funds and labs by OTF uses streamfield determination forms, can you confirm that?

We have a property to check if a submission use the old forms, see:

https://github.com/HyphaApp/hypha/blob/44febe59fe6d935e99933f32249400bf38050578/hypha/apply/funds/models/submissions.py#L603-L614

wes-otf: did a quick pass today and it doesn't seem like any of OTFs funds still use it actively. would still probably be worth it to handle it so it doesn't error out if there's an archive fund or something weird down the road. even if that's just ignoring it

frjo (approved):
Works well in my testing.

PR #4305, opened 2024-12-24
Fix report submit notification to stream field form flow

### LINKED ISSUE #4289: Report notifications not sent after converted to stream fields
Notifications should be sent in the form_valid() function but that functions seems not to be reached any longer.

https://github.com/HyphaApp/hypha/blob/eb2116f57a23562192b1723cf0c2dc589cc9b7c5/hypha/apply/projects/views/report.py#L173-L197

When reports were converted to stream fields a post() function was created that takes over.

https://github.com/HyphaApp/hypha/blob/eb2116f57a23562192b1723cf0c2dc589cc9b7c5/hypha/apply/projects/views/report.py#L159-L167

Does not look like other stream field forms are using post() so we overlocked something here.

sandeepsajan0:
Fixes #4289 

## Test Steps

 - [ ] Submit Report and notification should go out(right now it is going out to the applicant).

### DIFF
--- a/hypha/apply/projects/views/report.py
+++ b/hypha/apply/projects/views/report.py
@@ -161,6 +161,24 @@ class ReportUpdateView(BaseStreamForm, UpdateView):
         if form.is_valid():
             form.save(form_fields=self.form_fields)
             form.delete_temporary_files()
+
+            should_notify = True
+            if self.object.draft:
+                # It was a draft submission
+                should_notify = False
+            else:
+                if self.object.submitted != self.object.current.submitted:
+                    # It was a staff edit - post submission
+                    should_notify = False
+
+            if should_notify:
+                messenger(
+                    MESSAGES.SUBMIT_REPORT,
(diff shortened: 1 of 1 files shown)


wes-otf:
hey @sandeepsajan0, I tried to submit a couple different reports [here](https://test-apply.hypha.app/apply/projects/141/) and [here](https://test-apply.hypha.app/apply/projects/174/) but it didn't seem like either sent any notifications, is there something I missed?

frjo:
My fault, forgot to push to test. Being deployed now.

wes-otf:
Still doesn't seem to be working for me after the push, I tried submitting a report both as staff and as the applicant and on my end neither seemed to result in an email. The reports were overdue though so I'm not sure if that makes a difference?

wes-otf:
@frjo pointed out the notification came through slack - I had been looking for it via email for some reason. This should be good to go!

PR #4306, opened 2024-12-29
Make lead update modals UI same for both projects and applications

### LINKED ISSUE #4288: Changing Lead on applications and projects has different UI
Staff can update the Lead on both applications and projects but via different UI. Would be good to align them so they look and work the same.

Skärmavbild 2024-12-17 kl  09 03 00
Skärmavbild 2024-12-17 kl  09 03 34

sandeepsajan0:
Fixes #4288 

For projects:
image

For applications:
image

## Test Steps

 - [ ] …

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ npm-debug.log
 
 # IntelliJ
 out/
+.idea/*
 
 # JIRA plugin
 atlassian-ide-plugin.xml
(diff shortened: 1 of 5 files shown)


theskumar (approved):

wes-otf:
Nice addition! Feels super consistent between both project & application views now.

PR #4307, opened 2024-12-31
Replace eslint and prettier with Biome

### LINKED ISSUE #4295: Replace styelint, eslint and prettier with biomejs
### Discussed in https://github.com/HyphaApp/hypha/discussions/4283

<div type='discussions-op-text'>

<sup>Originally posted by **theskumar** December 16, 2024</sup>
## Pros
- faster
- Rust-based, Node.js not required
- Plays nicely with pre-commit, the current setup feel hacky
- Similar configuration, don’t have to work with three different tools 
- It’s built as a replacement for Prettier, ESLint, and the rules can be ported over
- Good editor support across VSCode

## Cons
- It’s a bit new
- While it supports almost all the rules, it doesn't support 100% of all the rules 

## Conclusion: 
Given this project is not FE heavy, we probably don't need a heavily customized tool to run linter and formatter; the current tools are quite slow, esp. This feels to me like the `ruff` of the Node.js world, and will help with the dev efficiency because we only have to work with a single tool that is easy to maintain, is faster, and provides better error reports. 

https://biomejs.dev/

cc: @frjo @wes-otf </div>

frjo:
Fixes #4295

Keeping stylelint for sass files since biome doesn't support them, yet.

The biome configuration is set up to induce minal changes to current codebase.

With time we should move towards using biome "recommended" configuration as is.


theskumar (approved):
Looks good to me, literally you have the config that doesn't update any files. 💯

PR #4308, opened 2024-12-31
Refractor fund views

theskumar:
Break down the views into more managable modules and move them all into the apply/funds/views/ folder and update the urls.py. No logical change is done. 

The views.py is getting too big


hypha/apply/funds/services.py
```diff
@@ -261,32 +260,3 @@ def annotate_review_recommendation_and_count(submissions: QuerySet) -> QuerySet:
         ),
     )
     return submissions
-
-
-def has_valid_str(tag: element.Tag) -> bool:
```
theskumar, line 266: Removed this as it's not used anywhere, there is duplicate of this in `translate.utils`

frjo (approved):

PR #4318, opened 2025-01-03
Refactor fund -> workflow module

theskumar:
- Refactor workflow.py to break it down into smaller modules
- Rename funds.workflow -> funds.workflows
- Improvement to Documentation 
- Remove unused constant declation`DETERMINATION_PHASES`


frjo (approved):
As intended it seems to have zero effect, everything works just as before.

This makes the workflows a lot more manageable. As discussed, there are more changes we should do to them, this is a first good step.

wes-otf (approved):

wes-otf:
In my other testing today I haven't seen anything deviate & code looks good!

PR #4322, opened 2025-01-07
Style fixes to Contracting documents

### LINKED ISSUE #4321: Project supporting document filenames buttons wrap weird
As the title says, when files are uploaded, the project document buttons can get a little weird (ie. the `View` button below on budget):
Screenshot 2025-01-07 at 12 36 45

It doesn't appear filename length changes this

frjo:
Fixes #4321

I think this fixes the correct list.

The styles in "_docs-block.scss" feels a bit overcomplicated.


theskumar:
I believe this is right but I didn't realize you had a create a fix for it. I took the opportunity to revamp the whole document category display, if you like my PR, this can be closed. 

https://github.com/HyphaApp/hypha/pull/4324

PR #4323, opened 2025-01-08
Remove any limit on rows exported when the submission is exported

### LINKED ISSUE #4320: All submissions export never export more than 60 items
The button export all the submissions, with export only the first page of all the submissions. Even if you move to other pages only the first page will be exported
If you switch to old view, you are able to export all the submissions

1. Go to 'All submissions'
2. Click on 'Export'
3. The file will include only the first 60 items
4. No error on console logs

**Additional context**

Hypha v5.16.0

theskumar:
Update the export to use the filtered queryset without pagination. Unless we have proper way to generate a very large dataset. I believe it whould be better to let the server generate as much data it can possibily generate.

Fixes #4320

## Test Steps

 - Goto submission list view, ensure export submission is enabled and you are logged in as admin
 - Try exporting a list of submisison with multiple pages, all the data should get exported not just initial page

### DIFF
--- a/hypha/apply/funds/views/all.py
+++ b/hypha/apply/funds/views/all.py
@@ -268,7 +268,7 @@ def submissions_all(
     if request.GET.get("format") == "csv" and permissions.can_export_submissions(
         request.user
     ):
-        csv_data = export_submissions_to_csv(page.object_list)
+        csv_data = export_submissions_to_csv(qs)
         response = HttpResponse(csv_data.readlines(), content_type="text/csv")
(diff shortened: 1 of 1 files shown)


frjo (approved):

wes-otf (approved):

wes-otf:
works well! definitely agree that moving forward we'll need some background worker to handle this (in the same way we need one for batch actions) - seems like the server even struggled a bit generating the 128 applications we have on the test instance

PR #4337, opened 2025-01-12
Update to release-drafter v6.

frjo:


theskumar (approved):
Looks like only nodejs version used is changed. 

🚢 💯

PR #4340, opened 2025-01-14
Clean out old migration code and drupal id migration field.

frjo:
This is code used by OTF when migrating from the old Drupal based system back in 2018.


frjo:
@theskumar We now have one migration that adds the drupal_id and then one that removes it.

Is that just how it is or should we clean up the old migrations?

frjo:
@wes-otf I do not see any benefit to OTF to keep these around, do you?

With all the cleaning of the code base you and others have made I remembered these.

theskumar:
> Is that just how it is or should we clean up the old migrations?

Yes. for now that's how it should be. squashing/removing migration will break the dependency graph for the migration files.

This blog post has some guidance on how to do it properly: https://jacklinke.com/squashing-django-migrations-the-easy-way. I have not tried it myself, but seems to me the right way to do it. TLDR; it's a long process.

wes-otf (approved):
I can't see need for anything drupal related in the future, would be sweet to be able to clean up these migrations too

PR #4342, opened 2025-01-14
Add django-model-info for devs.

frjo:
Adding https://github.com/OmenApps/django-model-info as dev dependency.


frjo:
Dependency issue with dslr, see https://github.com/mixxorz/DSLR/issues/23

That’s why uv added an old version of django-model-info.

frjo:
dslr have been updated.

PR #4349, opened 2025-01-17
Merge application and project communication tabs

### LINKED ISSUE #4249: Merge application and project communication tabs
They already use the same backend for rendering. The main issue to solve is to make sure we maintain the same visibility rules.

Different user roles should continue to see the same comments as now even when all communication is in a single tab.

wes-otf:
Fixes #4249. Migrates all project comments to the associated application, and loads the applications comments via htmx (merging in project activities). This was the best solution I could come up with until we potentially take a step a coupling the two closer (ie. sharing an ID (#3944), a URL space, and/or a detail view).

I had tried another version of this that did the same migration but loaded the project view via htmx from the application detail view, which made everything feel more consistent, but I had concerns about the user needed to see the status/stages bar when making a comment which is where the UI got hairy.

Would love any suggestions/critiques!

## Test Steps

Ensure...
 - [ ] ...the comments section on an application is identical to that of it's associated project
 - [ ] ...leaving a comment works as expected
 - [ ] ...applications without a project have unchanged communications tabs


hypha/apply/projects/templates/application_projects/project_detail.html
```diff
@@ -208,31 +208,19 @@ <h5>{% trans "Project form approvals" %}</h5>
...
-        <div class="tabs__content" id="tab-3">
-            <div
-                class="feed ms-3"
-                hx-get="{% url 'apply:projects:partial-activities' object.id %}"
-                hx-trigger="intersect once"
-            >
-                {% comment %} Loaded using the htmx via alpine's custom event "open-tab-3"{% endcomment %}
-                <p>{% trans "Loading…" %}</p>
-            </div>
-        </div>
```
frjo, line 233: We can delete this part from `applicationsubmission_detail.html` as well. I think we might even clear out all `partial-activities` urls and `partial_submission_activities()` unless they are still used somewhere else.

hypha/apply/projects/templates/application_projects/project_detail.html
```diff
@@ -208,31 +208,19 @@ <h5>{% trans "Project form approvals" %}</h5>
             <div class="feed">
                 {% if not object.is_archive %}
                     <h4 class="m-0 sr-only">{% trans "Add communication" %}</h4>
-                    {% include "activity/include/comment_form.html" %}
+                    {% include "activity/include/comment_form.html" with action=object.submission.get_absolute_url %}
```
frjo, line 209: This works but feels a bit, un neat. You go to `projects/p:n/#communications`, submit the form and end up on `submissions/s:n/#communications`. Not that normal users look at the address bar.

frjo (changes requested):
In my initial testing it seems to work well. Some minor comments.

hypha/apply/projects/templates/application_projects/project_detail.html
wes-otf: This is a nice catch, I didn't even notice that was the case with the redirect. Do you think it should be handled in htmx so the redirect doesn't happen?

frjo: Or does it make sense now to break out communications to its own view and url?

wes-otf: that's a good idea, I can explore that a bit

wes-otf: Another (more temporary) solution is to change up the `get_success_url` to resolve to the referrer URL:

So swapping
https://github.com/HyphaApp/hypha/blob/b643557c67fd287235a20031dd01b61a5acef400/hypha/apply/activity/views.py#L156-L157

with
```python
    def get_success_url(self):
        if (ref := self.request.META.get("HTTP_REFERER")) and (path := urlparse(ref).path):
            try:
                url_resolved = resolve(path)
                if url_resolved.url_name == "detail":
                    return f"{path}#communications"
            except Resolver404:
                # If it couldn't be resolved, use the default `get_absolute_url()`
                pass
        return f"{self.object.source.get_absolute_url()}#communications"
```

frjo: The redirect is not a big issue so I do not tink we need hack around it.

If creating a stand alone view/url is a bit big we can do it in a separate PR.

hypha/apply/activity/views.py
```diff
@@ -54,17 +54,19 @@ def partial_comments(request, content_type: str, pk: int):
         )
         editable = False if obj.status == "complete" else True
     else:
-        return render(request, "activity/include/comment_list.html", {})
+        return render(request, "activity/include/activity_list.html", {})
 
-    qs = services.get_related_comments_for_user(obj, request.user)
+    qs = services.get_related_activities_for_user(obj, request.user)
+    if hasattr(obj, "project") and obj.project:
```
theskumar: Alternatively, you can use the Q operator on the source. 

The base queryset for activities should be something on the line of 

If the class of source object is Application, then source=(source or source.project). 

If class of source object is Project then  query for source=(source or source.submission)

theskumar (commented):
I believe this works, there is another approach that you can try.

frjo (approved):

hypha/apply/projects/templates/application_projects/project_detail.html
wes-otf: Might be better to break the PRs up, I can take a stab at breaking it out into it's own view

hypha/apply/activity/views.py
wes-otf: Could you expand on that? In terms of being able to combine the two activity querysets I think the "or" throws me off

theskumar: I believe the end result might be same here. But this is what was trying to say. Bear with me with formatting as i am on phone. 


```
def get_activities(obj):
  qs = Activity.objects.all()  # old base queryset for comments and activities for a submission. 
  
  if hasattr(obj, "project"):
    filter_q = Q(source=obj) | Q(source__project=obj)
  if hasattr(obj, "submission"):
    filter_q = Q(source=obj) | Q(source__submission=obj)

  return qs.filter(filter_q)
```

theskumar: Activities for project are attached to project as source and not the main submission. We can consider updating this logic to store the activities against the submission in future. 

This way you only have a single common model to work with, that is closer to actual business logic.

wes-otf: Ahh okay I see! I think I might have to hack around a little but just because source is a GenericForeignKey attribute reversing that relationship in a filter can maybe get sticky from what I've seen. 

Regardless, I'll figure it out - go enjoy your time off! Thanks for the help!

frjo:
@wes-otf Trying to put this on test I get this error:

~~~~
  Applying application_projects.0097_move_project_comments_to_application...Traceback (most recent call last):
  File "/app/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/app/.heroku/python/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/app/.heroku/python/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/app/.heroku/python/lib/python3.12/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/app/.heroku/python/lib/python3.12/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/core/management/base.py", line 106, in wrapper
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/core/management/commands/migrate.py", line 356, in handle
    post_migrate_state = executor.migrate(
                         ^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/db/migrations/executor.py", line 135, in migrate
    state = self._migrate_all_forwards(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
    state = self.apply_migration(
            ^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/db/migrations/executor.py", line 252, in apply_migration
    state = migration.apply(state, schema_editor)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/db/migrations/migration.py", line 132, in apply
    operation.database_forwards(
  File "/app/.heroku/python/lib/python3.12/site-packages/django/db/migrations/operations/special.py", line 193, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/app/hypha/apply/projects/migrations/0097_move_project_comments_to_application.py", line 17, in migrate_project_comments_to_application
    application = Project.objects.get(id=comment.source_object_id).submission
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.heroku/python/lib/python3.12/site-packages/django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(
__fake__.Project.DoesNotExist: Project matching query does not exist.
~~~~

frjo:
Running it locally works it seems.

hypha/apply/activity/services.py
```diff
@@ -37,44 +39,39 @@ def edit_comment(activity: Activity, message: str) -> Activity:
...
+                source_object_id=obj.project.id, source_content_type=proj_content_type
+            )
+    else:
+        source_filter = Q(
+            source_object_id=obj.id, source_content_type=proj_content_type
+        )
+        if hasattr(obj, "submission") and obj.submission:
+            source_filter = source_filter | Q(
+                source_object_id=obj.submission.id, source_content_type=app_content_type
+            )
```
theskumar: I believe this can further simplified like this, 

logic to get the source filter. 
```python
    if hasattr(obj, "project") and obj.project:
        source_filter = Q(submission=obj) | Q(project=obj.project)
    if hasattr(obj, "submission") and obj.submission:
        source_filter = Q(submission=obj.submission) | Q(project=obj)
    else:
        source_filter = Q(submission=obj)
```

```suggestion
    if hasattr(obj, "project") and obj.project:
        source_filter = Q(submission=obj) | Q(project=obj.project)
    if hasattr(obj, "submission") and obj.submission:
        source_filter = Q(submission=obj.submission) | Q(project=obj)
    else:
        source_filter = Q(submission=obj)

```

theskumar:
We'll also need to update the `visibility_options` template tag, to account for an activity that could be on either a submission object or project instance. 

```diff
--- a/hypha/apply/activity/templatetags/activity_tags.py
+++ b/hypha/apply/activity/templatetags/activity_tags.py
@@ -84,7 +84,11 @@ def visibility_options(activity, user) -> str:
     Returns:
         A JSON string of visibility options
     """
-    submission_partner_list = activity.source.partners.all()
+    if hasattr(activity.source, "partners"):
+        submission_partner_list = activity.source.partners.all()
+    else:
+        submission_partner_list = activity.source.submission.partners.all()
+
     choices = activity.visibility_choices_for(user, submission_partner_list)
     return json.dumps(choices)
````

hypha/apply/activity/services.py
wes-otf: this worked! when I tried this before but I kept running into issues with the GenericForeignKey, not sure what I was trying before

wes-otf:
@frjo taking a look at the migration issues today

frjo:
@wes-otf Tried to deploy to test again but got the "__fake__.Project.DoesNotExist: Project matching query does not exist." error from above again.

hypha/apply/activity/services.py
theskumar: 👍🏼

wes-otf:
@frjo really weird, seems like some of the activities are orphaned. Normally when an application is deleted the deletion cascades to the project & the project's activities. I wrote a script that identified the following activities used to be associated with projects that no longer exist (on the test instance):
Screenshot 2025-02-04 at 14 53 49
should these comments just be deleted in the migration?

frjo:
I have messed a bit with the test db over the years, e.g. deleting all projects to start over. So this is likely due to a mistake by me. I think we can delete them manually on the test db.

wes-otf:
sweet, they should be deleted now so hopefully migration will succeed

wes-otf:
@frjo could give deploying this another try when you get a sec?

frjo:
@wes-otf There is a migration conflict with https://github.com/HyphaApp/hypha/pull/4375 so will need to wait for that to be approved.

wes-otf:
ah gotcha, capped the testing for that today!

frjo:
@wes-otf Now this PR can be rebased.

wes-otf:
When the status of things like PAFs & Invoices are updated with a comment, they're still going under the project and thus not visible, I'll fix this today

wes-otf:
the `get_related_activities_for_user`'s `source_filter` was overwriting itself in the logic statements, should be ready for redeploy @frjo

wes-otf:
@frjo this works in my testing, if you wanted to give it a test I think it's ready for merge

Issue #4358, opened 2025-01-23
Gracefully handle not getting valid input for an ID in query params


wes-otf:
When an invalid `id` is provided as a query param for the all submissions view (ie. `?rounds=aa&fund=fghj`), the server throws an error rather than notifying the user/ignoring the invalid input

wes-otf:
very much an edge case, just came up a few times when staff were trying to work around #4357

wes-otf:
Closed by #4360

PR #4360, opened 2025-01-23
Small fixes to the submissions table

### LINKED ISSUE #4357: Ability to select multiple rounds to filter on in all submissions view
Right now the filtering menu only allows a user to filter on one round, while manually using query params ie. `?round=123&round=456` is recognized by both the filtering table and the menu.

wes-otf:
Fixes #4357 & #4358. Multiple small fixes to the all submissions table, including:

- Allowing for the selection of multiple rounds, leads, and funds while filtering
- Updated display tags for filters to reflect multiple selections
- Added template tag to get display names of users in filter tag display (leads + reviewers)
- Added `scroll-auto` to remove the scroll bar gap on some of the submenus
- Caps normalizations
- Handling of non-int values in query params that use int ids
- Added more translatable strings

### DIFF
--- a/hypha/apply/funds/templates/submissions/all.html
+++ b/hypha/apply/funds/templates/submissions/all.html
@@ -1,7 +1,7 @@
 {% extends base_template %}
 
 {% load i18n static %}
-{% load querystrings heroicons %}
+{% load querystrings heroicons table_tags %}
 {% load humanize %}
 
 {% block title %}{% trans "Submissions" %}{% endblock %}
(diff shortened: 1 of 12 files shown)


wes-otf:
I added the new template tag to `table_tags` because that felt the most appropriate, let me know if you think it should be somewhere else @frjo

hypha/apply/funds/views/all.py
```diff
@@ -96,6 +96,13 @@ def submissions_all(
     can_view_archives = permissions.can_view_archived_submissions(request.user)
     can_access_drafts = permissions.can_access_drafts(request.user)
 
+    # Filter out anything that isn't an int
+    if selected_funds:
+        selected_funds = [id for id in selected_funds if id.isdigit()]
+
+    if selected_rounds:
+        selected_rounds = [id for id in selected_rounds if id.isdigit()]
+
```
frjo: Can we add a function that checks all the filters that only accepts int? I think it is almost all of them except status.

hypha/apply/funds/templates/submissions/all.html
```diff
@@ -205,43 +205,43 @@
 
             {% if selected_statuses %}
                 <span class="inline-flex items-center py-1 px-2 text-xs font-medium text-blue-800 bg-blue-100 rounded select-none me-2">
-                    status:{% for s in selected_statuses %}{{ s }}{% endfor %}
+                    Status: {% for s in selected_statuses %}{{ s }}{% endfor %}
```
frjo: Maybe a good opportunity to make these strings translatable as well?

frjo (changes requested):
Loks good and works well in my testing. Only found some minor things, see comments.

hypha/apply/funds/templates/submissions/all.html
frjo: Quite a few in this template is not translatable.

wes-otf: nice catch! I took a pass on them, pretty sure I was able to get most? lots were in aria attributes

frjo (approved):

frjo:
Nice improvements!

PR #4374, opened 2025-02-03
Unify messages styles

### LINKED ISSUE #4373: Three types of on page message styles
1. Old style on page messages, this is what you get when using e.g `messages.success()`:

Skärmavbild 2025-01-20 kl  14 28 58

2. Add `PROJECT_ACTION_MESSAGE_TAG` to e.g `messages.success()` and you get:

Skärmavbild 2025-01-20 kl  14 54 00

3. Adding `showMessage` to a `HX-Trigger` you get this style that also clears automatically after a short while:

Skärmavbild 2025-01-20 kl  14 14 39

Can we merge style 1 and 2 so we have only to types? Making them look a bit more alike would be good in any case. Consistent colours for error/info/success.</div>

frjo:
Fixes #4373

The htmx/toast messages and the django messages now use similar style and the same colours for error/success/info etc.

Skärmavbild 2025-02-05 kl  09 48 33
Skärmavbild 2025-02-05 kl  09 49 12
Skärmavbild 2025-02-05 kl  09 50 20

Also fixed some translations and related styling.

## Test Steps

 - [ ] Check that messages when saving draft, batch updating etc show in a nice and consistent way.
 - [ ] Check that messages for submissions and projects look the same.


hypha/templates/includes/message_item.html
```diff
@@ -1,21 +1,32 @@
...
-            <button
-                type="button"
-                class="transition-opacity hover:opacity-70 messages__button"
-                @click="show = false"
-            >
-                {% trans "Ok" %}
-            </button>
-        {% endif %}
-    </div>
+    {% heroicon_outline "check" stroke_width=2 class="inline-block w-4 h-4 me-1" aria_hidden=true %}
```
wes-otf: everything worked well, definitely feels much more consistent! super minor but the toast popup uses `check-circle`, not sure if it matters to make the two the same

https://github.com/HyphaApp/hypha/blob/88ccdc8ab37fc9bb5304a60d0a99536ae29b0dbd/hypha/templates/includes/_toast-placeholder.html#L120-L122

PR #4379, opened 2025-02-05
Implement ViewTransition for htmx and tabs

frjo:
We added ViewTransition between page loads in #4359.

You can also add it to any javascript that change things on screen. In this PR I added it to tabs.js. It gives a subtle and nice animations when switching tabs.

htmx has built in support for it and this PR turns it on for most hx-swap actions. (I did not turn it on for the review modals on submission all, some minor bug made them pop in/out.)

Read up on @view-transition her:

https://developer.mozilla.org/en-US/docs/Web/CSS/@view-transition

For hands on examples and explanations see https://www.youtube.com/watch?v=quvE1uu1f_I


theskumar:
works great!

theskumar (approved):

PR #4380, opened 2025-02-06
Add author updated feature to submission

### LINKED ISSUE #4192: Allow staff to change applicant/author of an application
A new person in an organisation is taking over the application. The applicant has multiple accounts. There are a number of scenarios where this would be handy.

A select/search field that list all current applicants and a confirmation step.

We could add this to the standard edit view or create a specific modal for just changing applicant/author.

The later might be best since staff do not always have edit perms on applications.

sandeepsajan0:
Fixes #4192 

Staff can update the submission's author. Flow is similar to the Lead update but the button is a bit hidden in more actions.

## Test Steps

 - [ ] Staff should find an option to 'Change Author' on the submission detail page in sidebar(More action).
 - [ ] Staff can select any applicant and assign them as an author of that submission.
 - [ ] New applicant should be notified about new assignment.
 - [ ] A slack notification should be there, like which user has replaced {old applicant} from {new_applicant}.
 - [ ] The new applicant's name should be updated in all the places on submission.
 - [ ] A new activity item should be created in the activity feed indicating the author update


hypha/apply/funds/forms.py
```diff
@@ -324,6 +324,35 @@ def make_role_reviewer_fields():
     return role_fields
 
 
+class UpdateAuthorForm(ApplicationSubmissionModelForm):
+    author = forms.ModelChoiceField(
+        queryset=User.objects.applicants(),
+        label=_("Applicants"),
+        required=False,
+    )
+    author.widget.attrs.update({"data-placeholder": "Select..."})
```
frjo: Can we make this `author.widget.attrs.update({"data-placeholder": "Select...", "data-js-choices": ""})`?

hypha/apply/funds/templates/funds/modals/update_author_form.html
```diff
@@ -0,0 +1,37 @@
...
+    const selectElements = document.querySelectorAll('#author_form select');
+
+    // add choices to all select elements
+    selectElements.forEach((selectElement) => {
+        new Choices(selectElement, {
+            removeItemButton: true,
+            allowHTML: true,
+        });
+    });
+</script>
```
frjo: With setting `data-js-choices` above I believe this script snippet is not needed.

hypha/apply/funds/templates/funds/modals/update_author_form.html
```diff
@@ -0,0 +1,37 @@
...
+    <div>
+        <dl class="mb-4">
+            <dt class="font-semibold">{% trans "Current Author" %}</dt>
+            <dd class="truncate">{{ object.user }} &lt;{{object.user.email}}&gt;</dd>
+        </dl>
+    </div>
+
+    {% url 'funds:submissions:change_author' pk=object.pk as author_update_url %}
+    {% include 'funds/includes/dialog_form_base.html' with form=form value=value %}
+</form>
```
frjo: The "lead" modal is much simpler, no form in form. Here there seems to be two htmx calls, is that needed?

I see that the partner modal looks like above, I suspect that is overcomplicated as well.

hypha/apply/funds/templates/funds/modals/update_author_form.html
```diff
@@ -0,0 +1,37 @@
...
+    hx-post="{% url 'funds:submissions:change_author' pk=object.pk %}"
+>
+    <div>
+        <dl class="mb-4">
+            <dt class="font-semibold">{% trans "Current Author" %}</dt>
+            <dd class="truncate">{{ object.user }} &lt;{{object.user.email}}&gt;</dd>
+        </dl>
+    </div>
+
+    {% url 'funds:submissions:change_author' pk=object.pk as author_update_url %}
```
frjo: I do not think this is used anywhere.

frjo (changes requested):
Works well but can be simplified I think.

wes-otf:
@frjo if you have the capacity can you review this? gotten lots of requests for this feature lately. Will likely add unit tests but otherwise feels pretty good to go. would eventually love to add emails to the choice options so staff doesn't only need to rely on the display name

Almost wondering if we should move the update author button to this `Edit` button instead as I find that could be a bit confusing
Screenshot 2026-05-18 at 14 09 01

hypha/apply/activity/templates/messages/email/author_updated.html
```diff
@@ -0,0 +1,16 @@
+{% extends "messages/email/applicant_base.html" %}
+
+{% load i18n %}
+
+{% block content %}{# fmt:off #}
+{% blocktrans with title=source.title %}You have assigned as an Applicant to submission "{{ title }}".{% endblocktrans %}
```
frjo: You have been assigned as an Applicant…  (missing "been")

hypha/apply/funds/forms.py
```diff
@@ -330,6 +330,35 @@ def make_role_reviewer_fields():
     return role_fields
 
 
+class UpdateAuthorForm(ApplicationSubmissionModelForm):
+    author = forms.ModelChoiceField(
+        queryset=User.objects.applicants(),
+        label=_("Applicants"),
+        required=False,
```
frjo: "required=True" is more correct I think.

hypha/apply/funds/forms.py
```diff
@@ -330,6 +330,35 @@ def make_role_reviewer_fields():
...
+    def __init__(self, *args, **kwargs):
+        kwargs.pop("user")
+        super().__init__(*args, **kwargs)
+        current_author = self.instance.user
+
+        author_field = self.fields["author"]
+
+        # Removed current author from queryset
+        author_field.queryset = author_field.queryset.exclude(id=current_author.id)
+        author_field.initial = current_author
```
frjo: This has no effect since current_author is excluded on the line above.

frjo (changes requested):

hypha/apply/funds/views/submission_edit.py
```diff
@@ -639,6 +640,64 @@ def post(self, *args, **kwargs):
 
 
 @method_decorator(staff_required, name="dispatch")
+class UpdateAuthorView(View):
+    model = ApplicationSubmission
+    form_class = UpdateAuthorForm
+    context_name = "author_form"
```
frjo: I do not think this is actually used anywhere. Same for "UpdateLeadView".

hypha/apply/funds/forms.py
frjo:
```
kwargs.pop("user")
self.fields["author"].queryset = User.objects.applicants().exclude(
  id=self.instance.user.id
)
self.fields["author"].widget.attrs.update({"data-js-choices": ""})
```

hypha/apply/funds/views/submission_edit.py
wes-otf: As in the view itself isn't used anywhere? or the `context_name`

wes-otf:
thanks for the flags @frjo! couple of issues were existing that I glazed over when I revived the PR

hypha/apply/funds/views/submission_edit.py
frjo: The "context_name" part only.

wes-otf: confirmed it's not used, should be set now

frjo (approved):

frjo:
Regarding the  "Edit" link, could we make the Legal name itself into a link, for staff only, to the wagtail admin user page? Then we can remove the separate "Edit" link.

PR #4386, opened 2025-02-07
WIP: Replacing comment assignments with a mentions-like UI

wes-otf:
This is a proof of concept of replacing the current `Assign To` interface with something that is potentially a little more intuitive and allows multiple users to be mentioned in the same comment with context.

Being that this is more of a PoC there is lots of sloppier code that needs to be fixed, along with edge cases to handle to don't look *too* hard just yet. I more just wanted to open this for discussion before any more work was done on it. Lots of notes were taken from Github, Signal & [this blog post](https://web.archive.org/web/20240625214619/https://phuoc.ng/collection/mirror-a-text-area/add-autocomplete-to-your-text-area/)

Staff have mentioned multiple users being assigned to the same comment would be nice and they take lots of meeting notes in comments so being able to contextualize why user(s) got assigned could be beneficial

Next steps if we decide to move forward with this:
 - [ ] Move more vanilla `comment_form` JS logic into alpine
 - [ ] Ensure staff can't accidentally leak PII via changing comment visibility to non-staff
 - [ ] Handle staff changing email addresses
 - [ ] Adding unit tests
 - [ ] Mentioning formatting to also appear on the user's `My Tasks`

## Demo

https://github.com/user-attachments/assets/ceca4ee1-0526-4b85-84af-b19ce88a4920

## Test Steps

 - [ ] …


wes-otf:
would love to hear any thoughts @frjo @theskumar @sandeepsajan0!

frjo:
The UI looks neat!

Some nit picking around workflow and language:

Being "assigned" and that becoming a task seems logical to me. Someone "mentioning" me and I get a task less so.

If more than one person is "mentioned" who is responsible for the task?

Maybe OTF does not see this being used to hand out actual tasks? Perhaps more of mention people and they getting notified?

I think it would be better to have "tasks" be only actual tasks, things you need to do.

PR #4395, opened 2025-02-12
Add django_hooks for template hooks

frankduncan:
This is pulled from https://github.com/nitely/django-hooks with everything except the template hooks removed.

This is the first of three PRs, adding the infrastructure needed to add django style hooks inside templates.  The hooks in templates look like:

```
{% hook 'hook_name' %}
```

where `hook_name` is defined somewhere else via:

```
from django.template.loader import render_to_string

# Code that loads a template
def my_extension_for_hook_name(context, *args, **kwargs):
    return render_to_string(
        "my_extension_for_hook_name.html",
        context.flatten(),
    )

class MyApp(AppConfig):
    def ready(self):
        from extensions.django_hooks.templatehook import hook
        hook.register("hypha_extension_head", my_extension_for_hook_name)
```

PR #4396, opened 2025-02-12
Extension user field addition

frankduncan:
Adds template hooks in necessary places to allow adding fields to the User object.  The first commit is an example replacement for the `{% include "includes/head_end.html" %}` in `hypha/templates/base.html`, with an improvement that if you have multiple extensions that want to add things to the `<head>` of every hypha page, they won't overwrite each other and can remain uncoupled.

This also adds in hooks into the user create/update forms in wagtail, as those aren't built on the `WAGTAIL_CUSTOM_FIELDS` like the main profile update page is.

This depends on #4395

### DIFF
--- a/hypha/apply/users/templates/wagtailusers/users/create.html
+++ b/hypha/apply/users/templates/wagtailusers/users/create.html
@@ -1,5 +1,6 @@
 {% extends "wagtailusers/users/create.html" %}
 {% load  static %}
+{% load hooks_tags %}
 
 {% block fields %}
     {% if form.separate_username_field %}
@@ -7,6 +8,8 @@
     {% endif %}
     {% include "wagtailadmin/shared/field_as_li.html" with field=form.email %}
     {% include "wagtailadmin/shared/field_as_li.html" with field=form.full_name %}
+
+    {% hook 'wagtail_user_edit' %}
 
     {% comment %}
         First/last name hidden input with dummy values because.. Wagtail admin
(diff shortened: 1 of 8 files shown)


frjo:
@frankduncan The template hooks feature are now merged in. Can you make this PR only add the hooks in the templates, thanks!

Did you have plans to add tests to make sure the various hooks still exist?

hypha/apply/users/templates/wagtailusers/users/create.html
```diff
@@ -8,6 +9,8 @@
     {% include "wagtailadmin/shared/field_as_li.html" with field=form.email %}
     {% include "wagtailadmin/shared/field_as_li.html" with field=form.full_name %}
 
+    {% hook 'wagtail_user_edit' %}
```
theskumar, line 12: Shouldn't this be called `wagtail_user_create`, thoughts? I believe the context for edit and create will be different. 

For a specific plugin, if it's needs to the same thing at the both the places, it can implement both the hooks, but those plugins that need to insert different code at different places, might not get an option to do so. 

Also, this could introduce bug due to it's non-explicit nature of the hook name. 

My 2 cents!

frankduncan: That makes sense.  I'll update when I make the PR.

frankduncan:
@frjo Yeah, I'll add tests and clean up this branch.  Thanks!

PR #4397, opened 2025-02-12
Add django hook in activity

frankduncan:
This is a django signal that can override various parts of the messaging infrastructure (email, slack, etc).  See updated documentation for information about the signal, and what it expects/provides.

That documentation file (signals.md) was removed on main, but not sure where to put it.


frjo:
@frankduncan I have discussed this with the dev team. We will not be accepting this PR.

We have a plan to rebuild the notifications system that will improve it and make it more flexible.

Likely it will be something like this:

1. Move out e-mail and slack from activities in to separate notification classes.
2. Allowing implementors to add their own notification classes in djp plugins.
3. A setting will decide what notification classes are active.

Next step will likely be to make each activity a signal, you can opt in to send any kind of notification/web hook etc. on these signals.

frankduncan:
That sounds great!  We should be able to adapt this to that system in a pretty straightforward manner.  One idea to keep in mind is if you can make it easy to subclass the email adapter to pull message from a non file system template place, and to have the list of which messages get emails sent also be dynamic/settable in the subclass, that would meet our needs very well.  A quality of life improvement there would be to move the business logic out of the templates and into methods that then call into a overridable function that return the message.

PR #4399, opened 2025-02-13
Allow users to select Hypha language

### LINKED ISSUE #4398: Multilingual support

An organization wants to serve Hypha to a community that includes 6 different language populations.  Currently, there is no easy way to (a) select a user language preference and (b) switch to that language for system strings, templates, applications, emails.  This makes Hypha a poor fit for their needs.

There are several potential ways for a user to indicate a language preference.  (click on a flag, a profile interface, etc.)  But once indicated, to have that language preference reflected throughout Hypha's interaction with that user.  Of course translated strings would have to be available (or we might fall back to machine-translated strings).

We have had some success simply displaying two languages in the submission form (e.g. English followed by Arabic).  This leaves the rest of the Hypha system in English, which is not ideal.  It also doesn't make sense for supporting six languages.

We might also use Google's "translate this page" feature, but that also is less than ideal for privacy and accuracy.

**Additional context**

**Priority**

This missing feature is preventing an org from using Hypha at all.  And they are not the only one.  They're just the one that is mentioning it today.

**Affected roles**

Potentially all roles affected, but really this is about applicants.

**Ideal deadline**

Ideally?  Yesterday.  Realistically? If this can change within the next three months, they might choose Hypha.

Thanks!

frjo:
Fixes #4398

- [x] Add django.middleware.locale.LocaleMiddleware
- [x] Set LANGUAGE_COOKIE_NAME to hypha_language
- [x] Add a language switcher that sets the language cookie
- [x] Make the language switcher accessible.

With "django.middleware.locale.LocaleMiddleware" Hypha will look for language preference in the following order:

1. Language set in the language cookie
2. Language set in users browser (Accept-Language HTTP header)
3. Language set in "LANGUAGE_CODE"

Skärmavbild 2025-02-20 kl  14 55 09

## Test Steps

 - [ ] Set LANGUAGE_SWITCHER to True
 - [ ] Confirm that the language switcher gets added to the header.
 - [ ] Confirm that switching language is possible and that the "hypha_language" cookie gets set.

### DIFF
--- a/hypha/core/context_processors.py
+++ b/hypha/core/context_processors.py
@@ -23,4 +23,5 @@ def global_vars(request):
         "SENTRY_PUBLIC_KEY": settings.SENTRY_PUBLIC_KEY,
         "SUBMISSIONS_TABLE_EXCLUDED_FIELDS": settings.SUBMISSIONS_TABLE_EXCLUDED_FIELDS,
         "HIJACK_ENABLE": settings.HIJACK_ENABLE,
+        "LANGUAGE_SWITCHER": settings.LANGUAGE_SWITCHER,
     }
--- a/hypha/core/context_processors.py
+++ b/hypha/core/context_processors.py
@@ -23,5 +23,6 @@ def global_vars(request):
         "SENTRY_PUBLIC_KEY": settings.SENTRY_PUBLIC_KEY,
         "SUBMISSIONS_TABLE_EXCLUDED_FIELDS": settings.SUBMISSIONS_TABLE_EXCLUDED_FIELDS,
         "HIJACK_ENABLE": settings.HIJACK_ENABLE,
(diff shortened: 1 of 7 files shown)


frjo:
@jvasile @wes-otf Django and Wagtail contains translations for many languages but Hypha is only completely translated for English, Simplified Chinese and Russian.

I opted for limiting LANGUAGES setting to these languages when LANGUAGE_SWITCHER is on. When it is off, the default, LANGUAGES setting only contains English.

If this is not done "django.middleware.locale.LocaleMiddleware" will switch to any of the Django supported languages. This will result in a site that is 95% in English with a few generic terms translated here and there, words like "Submit", "Cancel" etc.

What are your thoughts on this?

wes-otf:
still going to be doing more testing but I'm really impressed by my initial pass! this seems like a really nice addition

PR #4402, opened 2025-02-17
Fix alpinejs discrepencies, typos in review score

frankduncan:
alpinejs, when using x-show, doesn't need the element to be hidden, as it will take care of that.  Also, x-clock should be x-cloak

Found this bug when I was merging our codebases together, and the score floating widget wasn't showing up.  The one next to the submit buttons was though.

The floating score does interact somewhat poorly with the application when "Show Application" is viewed.

I didn't create an issue for this, let me know if you want me to!


frjo:
@frankduncan Thanks for the bug fix!

PR #4410, opened 2025-02-20
The runtime.txt file is deprecated by Heroku, it has been replaced by the more widely supported .python-version file

frjo:


theskumar:
Nice!

PR #4411, opened 2025-02-21
Move to specify python version without patch version. This way we get patch and security releases automatically

frjo:


frjo:
uv picks version from cache if that satisfies requirements so skipping this.

PR #4414, opened 2025-02-24
Update communication view to be an independent page

### LINKED ISSUE #4355: Create standalone url and view for comments
Currently comments are `POST`ed to the submission view (`/apply/submissions/<pk>/`) despite #4349 combining the communication between projects. Serving communications from it's own view would make more sense from a user perspective, especially when a comment posted under `/apply/projects/<pk>/#communications` redirects to `/apply/submissions/<pk>/#communications`

theskumar:
Depends on: 
- https://github.com/HyphaApp/hypha/pull/3944
- https://github.com/HyphaApp/hypha/pull/4413

What's changed: 
- project and submission to use the same communication page
- update urls to point to the new page
- update urls in email. 
- Cleanup the old comment view page code
- new permission check added "view_comments" -- anyone with either access to project or submission will have access to it's comments page. 

Other updates:
- Global activity feed has updated text for the comments
- New function based view is use to show and add comment
- Remove the use of tab.js on the submisison/project detail page, all the tabs are server-rendered full-page now. 

Fixes https://github.com/HyphaApp/hypha/issues/4355


frjo:
@theskumar Thinking about what URL to use for comments.

We have `/apply/submissions/1234/` and `/apply/projects/1234/`. Maybe comments should be `/apply/comments/1234/`?

theskumar:
> @theskumar Thinking about what URL to use for comments.
> 
> We have `/apply/submissions/1234/` and `/apply/projects/1234/`. Maybe comments should be `/apply/comments/1234/`?

The `/apply/comments/1234/` feels like linking to a specific comment with 1234. I believe this is counter intuitive and the fix should be done to update the project url as follows `/apply/submissions/1234/project` instead and then rest of the project related urls could follow the convention `projects/.../invoices`, `projects/.../reports`, 

There are also urls like `/apply/submissions/1234/determination-outcome` etc. 

Alternatively, we can also consider using simply

```
/applications/<id>/detail
/applications/<id>/comments
/applications/<id>/project
/applications/<id>/determinations
/applications/<id>/invoices
/applications/<id>/reports
```

level two urls can be then: 

```
/reports/<id>/
/reports/<id>/edit
/invoices/<id>/
/projects/<id>/ .... same as /applications/<id>/project
```

with a redirection in place for old urls.

frjo:
> The /apply/comments/1234/ feels like linking to a specific comment with 1234. 

This makes sense.

I like this pattern:

> /applications/<id>/detail
> /applications/<id>/comments
> /applications/<id>/project
> /applications/<id>/determinations
> /applications/<id>/invoices
> /applications/<id>/reports

theskumar:
This should be good to review now.

frjo (approved):

theskumar:
Rebased.

PR #4419, opened 2025-02-25
Show and link to associated project in submisison list view

theskumar:
Add a non-intrustive way to display that a submission/application has
been converted to project and also link to the project tab directly from
it.

I considered showing the status of the project as well, but it felt little
too much of info, so just displaying the text - "Project"

Screenshot 2025-02-25 at 3  39 13@2x


frjo (approved):
Nice.

PR #4422, opened 2025-02-26
Use same style for notification view as used on communication tab.

### LINKED ISSUE #4387: Make the activity/notifications view unified like the application communication view

frjo:
Fixes #4387 

Skärmavbild 2025-02-27 kl  16 06 05

### DIFF
--- a/hypha/apply/activity/templates/activity/include/activity_list.html
+++ b/hypha/apply/activity/templates/activity/include/activity_list.html
@@ -6,9 +6,9 @@
         <div class="timeline">
             {% for activity in activities %}
                 {% if activity.type == "comment" %}
-                    {% include "activity/ui/activity-comment-item.html" with activity=activity notifications_page=notifications_page %}
+                    {% include "activity/ui/activity-comment-item.html" with activity=activity %}
                 {% else %}
-                    {% include "activity/ui/activity-action-item.html" with activity=activity notifications_page=notifications_page %}
+                    {% include "activity/ui/activity-action-item.html" with activity=activity %}
                 {% endif %}
             {% endfor %}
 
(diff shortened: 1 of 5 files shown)


theskumar:
This is great improvement!  

Given across different funds these activities can be quite a lot, thought on converting it to be a paginated list with auto load similar to how it's done in the communication tab?

theskumar:
Should we update the title to save "Activity Feed" instead? same as it's on the user's menu. 

Screenshot 2025-02-26 at 2  56 01@2x

theskumar:
I see there is an issue with the new feed, it's not possible to identify which project/application this particular activity belongs to.

frjo:
Good catch, that I missed. Need to look over this more.

theskumar:
> I see there is an issue with the new feed, it's not possible to identify which project/application this particular activity belongs to.

On the possibility you can explore is showing the activity in the format. 

- [noun] [verb] on [target] on [source] [timestamp]

and use the same format for the comment as well, so for the comment it will say something like. 

- X left a commented on FundX 2 days ago
- X updated status to approve on FundX on Feb 12, 2024

theskumar:
> > I see there is an issue with the new feed, it's not possible to identify which project/application this particular activity belongs to.
> 
> On the possibility you can explore is showing the activity in the format.
> 
> * [noun] [verb] on [target] on [source] [timestamp]
> 
> and use the same format for the comment as well, so for the comment it will say something like.
> 
> * X left a commented on FundX 2 days ago
> * X updated status to approve on FundX on Feb 12, 2024

Or the id/name of application/project can be prefix in all the activities and update the comment item to use the activity item with a `comment`

frjo:
Now added a fix so submission/project is visible for each comment/activity.

frjo:
Grouped the notification by submission/project after suggestion from @theskumar. I think that makes the view more useful in general. Looks good as well.

PR #4425, opened 2025-02-27
Use new filters UI with django filters

### LINKED ISSUE #4345: Implement submission all filter UI on all Table2 filters
We replaced select2 with choices.js in #4319. This is the next step that unifies the filter UI.

sandeepsajan0:
Fixes #4345 

Projects all page:
image

image

image

Reports page:
image

image

Result page:
image

image

## Test Steps

 - [ ] Look around for all tables pages except submission all page and try filters there, all should be working fine and UI should be similar to the submission al page or to the screenshots above.


frjo (changes requested):
A bit of margin top would be good here:

Skärmavbild 2025-03-17 kl  14 24 14

Code related to the old filters can be removed:

https://github.com/HyphaApp/hypha/blob/f904ade86aa6f87b15c3606b0e5c5b4c8b40e555/hypha/apply/funds/templates/funds/includes/table_filter_and_search.html#L60-L70

And:

https://github.com/HyphaApp/hypha/blob/f904ade86aa6f87b15c3606b0e5c5b4c8b40e555/hypha/apply/funds/templates/funds/includes/table_filter_and_search.html#L97-L101

I also think we should be able to remove `submission-filters.js` completely. If I remove it now the filters break so some small part is active.

wes-otf:
This is a huge UI improvement! really nice to no longer have the filtering inconsistencies

hypha/apply/projects/templates/application_projects/project_list.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base-apply.html" %}
 
 {% load render_table from django_tables2 %}
-{% load i18n static %}
+{% load i18n static heroicons querystrings %}
```
theskumar: I believe these two new template tags are unused.

hypha/core/templatetags/querystrings.py
```diff
@@ -61,6 +61,32 @@ def modify_query(context, *params_to_remove, **params_to_change):
     )
 
 
+@register.simple_tag(takes_context=True)
+def dup_modify_query(
+    context, param_to_remove, param_key_to_change, param_value_to_change
+):
```
theskumar: Kindly modify the template tag's name to be more descriptive or include an appropriate docstring detailing its purpose and uses.

sandeepsajan0: Updated

frjo:
@sandeepsajan0  There are some remaning uses of "submission-filters.js" that can be removed and then we can delete the file itself.

I believe we can delete "_filters.scss" as well, after we removed all uses for the "filters*" classes.

`{% include "funds/includes/table_filter_and_search.html" with filter_form=filter_form search_term=search_term use_search=True filter_action=filter_action filter_classes="filters-open" can_export=can_export %}`

In these included we can delete both `can_export=can_export` that is not in use anymore from what I can see and `filter_classes="filters-open"`

frjo:
I think this PR will end up deleting more code than it adds, plus it makes the filtering a lot better.

sandeepsajan0:
@frjo It is now ready for testing. I have removed the unused code related to filters(submissions-filters.js, filters.scss, base_submission_table.html).

frjo (approved):
Nice!

PR #4427, opened 2025-03-02
Update docstrings of BaseStreamForm class

theskumar:


frjo (approved):

hypha/apply/stream_forms/models.py
```diff
@@ -22,11 +22,27 @@
 
 
 class BaseStreamForm:
+    """Base class for creating dynamic forms using Wagtail StreamField.
+
+    This class provides functionality for handling form field definitions,
+    form data serialization/deserialization, and form generation.
```
bickelj, line 28: Where is the serialization here? I see only deserialization on the surface. Genuine question.

PR #4428, opened 2025-03-03
Fix report form, allow saving as draft without filling required fields

theskumar:
Right now, the report form validates for the required fields when saving
as draft.

This updates the behaviour to not validate the required fields, similar
to application submision forms.

### DIFF
--- a/hypha/apply/projects/views/report.py
+++ b/hypha/apply/projects/views/report.py
@@ -73,19 +73,7 @@ class ReportUpdateView(BaseStreamForm, UpdateView):
     object = None
     form_class = None
     form_fields = None
-
-    def get_form_class(self, draft=False, form_data=None, user=None):
-        """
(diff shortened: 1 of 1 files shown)


hypha/apply/projects/views/report.py
```diff
@@ -73,19 +73,7 @@ class ReportUpdateView(BaseStreamForm, UpdateView):
...
-        Expects self.form_fields to have already been set.
-        """
-        # This is where the magic happens.
-        fields = self.get_form_fields(draft, form_data, user)
-        the_class = type(
-            "WagtailStreamForm",
-            (ReportEditForm,),
-            fields,
-        )
-        return the_class
```
theskumar, line 88: 📓  This is not needed as the Parent class "BaseStreamForm" implement this function that makes use of `submission_form_class` class property.

frjo (approved):

PR #4433, opened 2025-03-07
Cleanup alert scss file, only used on report detail/form

theskumar:
This is only use in few places, update to use tailwindcss classes.

I plan to create either dedicated component or classes for different
alert/info/notice boxes.

Screenshot 2025-03-07 at 6  40 30@2x


hypha/apply/projects/templates/application_projects/report_detail.html
```diff
@@ -17,14 +17,14 @@
...
-                </p>
-            </div>
+            <p class="flex items-center py-2 px-3 mt-0 mb-8 text-sm border bg-[var(--color-light-blue-90)]">
+                {% heroicon_outline 'exclamation-circle' stroke_width=2 size=18 class="inline me-1 stroke-dark-blue" aria_hidden=true %}
+                <span>
+                    {% trans "This report is for the period" %} <strong class="font-semibold">{{ object.start_date }}</strong> {% trans "to" %} <strong class="font-semibold">{{ object.end_date }}</strong>
+                </span>
+            </p>
 
-            <div class="card card--solid">
```
theskumar, line 27: Screenshot 2025-03-07 at 6  42 33@2x

PR #4435, opened 2025-03-08
Python package upgrades. Mostly to get the file extension fix in  django-file-form 3.8.1

### LINKED ISSUE #4434: Allow case insensitive file upload extension matching
**Describe the bug**

File upload stream field is case insensitive on the web page.

**To Reproduce**

1. Add a file upload field to the application template
2. Start filling out an application, notice that `.jpeg` files are OK (see `hypha/settings/base.py`)
3. Add a file with extension `.Jpeg` instead of `.jpeg`

**Expected behavior**

Files with name `.Jpeg` or `.JPEG` should be allowed in addition to `.jpeg`. I think the backend allows this but the front-end JS library `django-file-form` is to blame. I created https://github.com/mbraak/django-file-form/issues/1018 in that library's repository to track this issue. This issue is a reminder to include the newer version when it's available and make sure it works as expected.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Staff
- Applicants
- Reviewers
- Partners
- Developers
- General public

**Screenshots**

Image

 - OS: Trisquel 11.0.1 Linux kernel 6.12.17-gnu
 - Browser: Abrowser 135.0.1 (64-bit) (like Firefox)

**Additional context**

We have a workaround whereby we include several globs such as `[Jj][Pp][Ee][Gg]` but this is pretty ugly especially when it is displayed to the user in the stream form widget.

frjo:
Fixes #4434


theskumar (approved):
LGTM

PR #4442, opened 2025-03-11
Load field form assets always since we have file fields in modals/htmx.

### LINKED ISSUE #4441: Unable to upload contracts in the contracting phase
When trying to upload a contract file, after selecting the file it seems like nothing happens/is uploaded and the prompt cannot be completed:

https://github.com/user-attachments/assets/a4645d27-a1cd-4366-a4d0-4c02c4da6212

frjo:
Fixes #4441

## Test Steps

 - [x] Test that file field work in dialogs, and everywhere else.


wes-otf:
works like a charm!

PR #4448, opened 2025-03-12
Indent all scss files with 2 and not 4 spaces.

frjo:


theskumar:
how did you do it? i mean the batch processing. i tried with styelint in the previous PR, without success.

theskumar (approved):

frjo:
Sublime has a good multi-file search and replace where you can filter on things like *.scss.

PR #4452, opened 2025-03-12
Block multiple form submits on application form.

### LINKED ISSUE #4348: Duplicate applications being created
I haven't been able to recreate it yet or find a specific pattern, but on OTF live staff keep reporting duplicate applications. Staff shared two slack notifications from last week showing 2 applications from the same application seemingly submitted in the same minute. When reviewing the two applications, they appeared identical besides the latter application being stripped of file attachments. This seems to be happening semi-frequently.

I'll keep digging on this and update if I find anything interesting

frjo:
Fixes #4348

Before this js snippet it was possible on a slow network to post a application form multiple times by rapidly clicking the submit button multiple times.

Would be neat to use aplinejs for this everywhere but it interfere with the current "unlockApplicationForm" feature that we use to block some spam.

## Test Steps

 - [ ] Set your browser to mimic a slow network and then post a application form by clicking many times in rapid succession.

### DIFF
--- a/hypha/apply/funds/templates/funds/application_preview.html
+++ b/hypha/apply/funds/templates/funds/application_preview.html
@@ -38,12 +38,12 @@
             </div>
         </form>
 
-        <form id="preview-form-edit" class="form application-form" action="{% url 'funds:submissions:edit' object.id %}">
+        <form id="preview-form-edit" class="form application-form" action="{% url 'funds:submissions:edit' object.id %}" x-data="{ isFormSubmitting: false }" x-on:submit="isFormSubmitting = true">
             {% csrf_token %}
         </form>
 
         <div class="form__group">
-            <button class="button button--primary" form="preview-form-submit" name="submit" type="submit">{% trans "Submit for review" %}</button>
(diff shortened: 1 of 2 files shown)


wes-otf:
That's super strange, I tried network throttling in Firefox but couldn't replicate. Maybe I should also try it in Safari.

I think this solution is nice! Do you think any server side validation would be worthwhile to add a hurdle to some spam attacks? Seems like something like this would be easy to script just based off this bug

frjo:
I added the following to stop some spam years back:

https://github.com/HyphaApp/hypha/blob/31fb7e412e4a92576902b4317a84bca222db8bc4/hypha/static_src/javascript/application-form.js#L57-L80

This code stops some scripts from posting since you need to move mouse/press key/touch screen before you can submit form.

Do you have more ideas how to limit spam postings?

wes-otf:
I was thinking maybe a server side submit cooldown and a daily submission limit but that's out of scope of this PR, this fix seems good to me for this!

frjo:
Yes, this is a quick fix.

For logged in users we could definitely add a check to warn/block them from submitting more than one application to each round/lab per hour or so.

We should also alert users if they have a draft belonging to a round/lab when they visit the application form.

PR #4456, opened 2025-03-18
Small improvements for displaying closed funds

wes-otf:
A couple simple additions:
 - Added a new message (as a customizable system setting) that will display when there are no open funds
 - Added call to action to closed round message along with a fix for displaying the verbose name
 - Small misc. fixes/additions (`Mobil` -> `Mobile` typo in system settings, docstrings in `util_tags.py`

## Test Steps

 - [ ] Ensure that the set system settings message displays on the home page when no funds are live
 - [ ] Ensure there is a link to the homepage in the message displayed when a closed round URL is accessed


wes-otf:
this test failure is super weird - no errors when running the server & it pulls the message from the system setting fine? But tests all error out saying the column doesn't exist. will dig deeper tomorrow

frjo:
@theskumar Can you figure out why the test fails like this? It's like the migration does not run before the tests.

The code looks fine to me, just another home page setting. The site works as intended with this new setting.

frjo (approved):
Something weird with the tests. The code works just as intended when I test.

theskumar:
> @theskumar Can you figure out why the test fails like this? It's like the migration does not run before the tests.
> 
> The code looks fine to me, just another home page setting. The site works as intended with this new setting.

Checking

hypha/home/templates/home/home.html
```diff
@@ -16,11 +16,17 @@ <h1 class="text-5xl font-bold text-balance">
...
+        {% if funds %}
+            <div class="divide-y">
+                {% for fund in funds %}
+                    {% include "home/includes/fund-list-item.html" with page=fund %}
+                {% endfor %}
+            </div>
+        {% else %}
+            <div class="mt-8">
+                {{ settings.core.SystemSettings.home_no_applications_msg|nh3|safe }}
+                <div>
```
frjo: I think this should be a closing div `</div>`

wes-otf: nice catch! not sure why I'm forgetting to properly close all my tags

wes-otf:
This testing failure is so weird - seems to happen when any new field gets added in a migration. Like I added a `test` attribute to the `Activity` model on a clean branch:
```
class Activity(models.Model):
    test = models.TextField(default="test")
    ...
```
which resulted in the same `django.db.utils.ProgrammingError: column activity_activity.test does not exist` despite migrations running fine on my local DB. can't find out how to debug pytest-django's migration running to see where/why it's not running if that's the case, maybe a previous migration is messing with things?

theskumar:
> This testing failure is so weird - seems to happen when any new field gets added in a migration. Like I added a `test` attribute to the `Activity` model on a clean branch:
> 
> ```
> class Activity(models.Model):
>     test = models.TextField(default="test")
>     ...
> ```
> 
> which resulted in the same `django.db.utils.ProgrammingError: column activity_activity.test does not exist` despite migrations running fine on my local DB. can't find out how to debug pytest-django's migration running to see where/why it's not running if that's the case, maybe a previous migration is messing with things?

By default, pytest-django creates a test database the first time a test requires it. This database is cached for subsequent tests, which helps improve performance by avoiding repeated setup, unless you provide `--reuse-db` parameter.

wes-otf:
> This database is cached for subsequent tests, which helps improve performance by avoiding repeated setup, unless you provide --reuse-db parameter

@theskumar did you mean `--create-db`? I was trying to get it remade each time just to see if that would change anything but was using create, maybe I should try with reuse if that's the case

theskumar:
The issue is that in the previous migration `core -> 0007`, `SystemSettings` is loaded. The Python version of the class in the codebase expects the `home_no_applications_msg` to be present, but this property is not available until the migration 0008 is run.

image


To reproduce, create a fresh database, update your `DATABASE_URL`, and try running `./manage.py migrate`.

wes-otf:
thanks @theskumar for saving me quite a bit of time! I was stumped

PR #4457, opened 2025-03-19
Required file field behaviour in stream field forms

### LINKED ISSUE #4401: Required file field in a stream form doesn't show error on submit when empty
If the file field is present in the stream forms and it's a required field, submitting the form doesn't validate the files and saves as normal.

Desired behavior: error should be shown for all the required file fields both single file field and multiple file field.

sandeepsajan0:
Fixes #4401 

## Test Steps

 - [ ] Try it as an applicant, and pick any fund/form that has the required file/image field.
 - [ ] It should show the error if the required file is not uploaded(previously it was silently failing).
 - [ ] It should work just fine as soon as all required files are uploaded.

### DIFF
--- a/hypha/apply/funds/templates/funds/application_base.html
+++ b/hypha/apply/funds/templates/funds/application_base.html
@@ -85,30 +85,6 @@
 {% endblock %}
 
 {% block extra_js %}
-    <script>
-        document.addEventListener("DOMContentLoaded", function () {
-            let fileInputs = document.querySelectorAll("input[type='file'][required]");
-
-            fileInputs.forEach(input => {
-                input.addEventListener("invalid", function (event) {
-
-                    event.preventDefault(); // Prevent default browser behavior
-
-                    // Find the closest dff-uploader wrapper and display an error
-                    let container = input.closest(".dff-uploader");
-                    if (container) {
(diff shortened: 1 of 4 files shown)


hypha/apply/funds/templates/funds/application_base.html
```diff
@@ -85,6 +85,30 @@ <h3>{% blocktrans %}Sorry this {{ page|verbose_name }} is not accepting applicat
...
+                            errorDiv.innerText = "This field is required. Please upload a file."; // Show error message
+                            errorDiv.style.color = "red";
+                        }
+                    } else {
+                        alert("Please upload a file."); // Fallback for direct file inputs
+                    }
+                });
+            });
+        });
+    </script>
```
frjo: Maybe we can place this code in the `file-uploads.js` file? That file is always loaded and also relates to file upload fields.

frjo (dismissed):
The test fails so needs some adjustments.

hypha/apply/funds/templates/funds/application_base.html
sandeepsajan0: Yes, that seems more isolated and consistent.

frjo (approved):
Nice!

PR #4458, opened 2025-03-19
Fix migration files

theskumar:
How to test. 

- create a fresh db
- update you DATABASE_URL
- run `./manage.py migrate

issue: https://github.com/HyphaApp/hypha/pull/4456#issuecomment-2737780456

Strategy: 
- fetch only the required columns, and update data only if previous system setting exist.


wes-otf (approved):
Works locally and causes the migration in #4456 to work without issue 🥳

PR #4461, opened 2025-03-21
Fixes for bugs around exporting applications

### LINKED ISSUE #4460: Exporting some application groups results in `'str' object has no attribute 'items'`
When an address field is `-`, rather than the expected dict, the CSV export in the all submissions view fails with the error `'str' object has no attribute 'items'`

wes-otf:
Fixes #4460 along with another issue where a month duration somehow had a value of `36` (I believe it was an older application, so maybe some removed code?) so I added it to the duration block's constants. Once this gets merged I'm going to add translation strings for the duration block too as I noticed it was missing a few

## Test Steps

 - [ ] Ensure applications that have address fields containing `-`/no value don't result in an error when exporting

### DIFF
--- a/hypha/apply/funds/blocks.py
+++ b/hypha/apply/funds/blocks.py
@@ -176,6 +176,7 @@ class DurationBlock(ApplicationSingleIncludeFieldBlock):
         12: "12 months",
         18: "18 months",
         24: "24 months",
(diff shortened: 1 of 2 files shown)


theskumar (approved):
LGTM!

PR #4464, opened 2025-03-23
Update view link to use heroicon instead of SVG

theskumar:
Update the "View Determination" to use heroicons instead of svg.

Also, the svg sprite was missing.

## Before

Screenshot 2025-03-23 at 8  07 47@2x

## After
Screenshot 2025-03-23 at 8  06 55@2x


frjo:
Good fix, I missed this one when I removed all the svg sprites. Did a quick search and this seems to be the last one.

PR #4466, opened 2025-03-24
Modernize the task-list UI

theskumar:
### Notable changes:

- Show tasks in chronological order
- Add relative timestamp against each tasks
- Review view button to save some spaces and make the row clickable
- Update the tasks copy to include the fund name and application id
- admin bar sub-heading truncates now for submissions with long title

### Worth considering (up for discussion):

- should be update and make use of `[fund_name] #{application_id}` as generic indentier for applications instead of the current "title" only.
- When it comes to selecting a main fund name, should use priorities name of the "round" or name of the "fund" from which the round is derived.

## Internal changes

Included `fund_name` and `application_id` on both the submission and project instance to easily access them, while abstracting the complexity around fallback on submission.id when the `submission.public_id` is missing

Screenshot 2025-03-24 at 8  59 02

### DIFF
--- a/hypha/apply/activity/templates/activity/notifications.html
+++ b/hypha/apply/activity/templates/activity/notifications.html
@@ -35,7 +35,7 @@
                             <div class="relative timeline-item" id="communications#{{ activity.id }}">
                                 {% ifchanged activity.source.id %}
                                     <div class="py-0.5 mt-4 bg-gray-100 ps-5 pe-2 border-s-2 border-slate-300">
-                                        <span class="text-sm font-semibold text-fg-muted">{{ activity.source_content_type.name|source_type }}</span> <a href="{{ activity.source.get_absolute_url }}">{{ activity.source.fund_name }} #{{ activity.source.application_id }}: {{ activity.source.title|capfirst|truncatechars:50|truncatechars:50 }}</a>
+                                        <span class="text-sm font-semibold text-fg-muted">{{ activity.source_content_type.name|source_type }}</span> <a href="{{ activity.source.get_absolute_url }}">{{ activity.source.fund_name }} #{{ activity.source.application_id }}: {{ activity.source.title|capfirst|truncatechars:50 }}</a>
                                     </div>
                                 {% endifchanged %}
                                 <div
--- a/hypha/apply/activity/templates/activity/notifications.html
+++ b/hypha/apply/activity/templates/activity/notifications.html
@@ -35,7 +35,7 @@
                             <div class="relative timeline-item" id="communications#{{ activity.id }}">
                                 {% ifchanged activity.source.id %}
                                     <div class="py-0.5 mt-4 bg-gray-100 ps-5 pe-2 border-s-2 border-slate-300">
-                                        <span class="text-sm font-semibold text-fg-muted">{{ activity.source_content_type.name|source_type }}</span> <a href="{{ activity.source.get_absolute_url }}">{{ activity.source.title|capfirst }}</a>
+                                        <span class="text-sm font-semibold text-fg-muted">{{ activity.source_content_type.name|source_type }}</span> <a href="{{ activity.source.get_absolute_url }}">{{ activity.source.fund_name }} #{{ activity.source.application_id }}: {{ activity.source.title|capfirst|truncatechars:50|truncatechars:50 }}</a>
                                     </div>
                                 {% endifchanged %}
                                 <div
--- a/hypha/apply/determinations/templates/determinations/includes/determination_button.html
+++ b/hypha/apply/determinations/templates/determinations/includes/determination_button.html
@@ -1,6 +1,9 @@
 {% load i18n determination_tags workflow_tags %}
 {% if request.user|show_determination_button:submission %}
-    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width {{ class }}">
+    <a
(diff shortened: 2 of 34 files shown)


hypha/settings/base.py
```diff
@@ -63,9 +63,9 @@
     "SUBMISSIONS_ARCHIVED_ACCESS_STAFF_ADMIN", True
 )
 
-# Possible values are: "public_id" and "title"
+# Possible values are: "application_id", "fund_name" and "title"
 SUBMISSION_TITLE_TEXT_TEMPLATE = env(
-    "SUBMISSION_TITLE_TEMPLATE", default="{title} (#{public_id})"
+    "SUBMISSION_TITLE_TEMPLATE", default="{fund_name} #{application_id} - {title}"
```
theskumar: I find this format more easy to digest, what your thoughts on it? Keep or revert? @frjo @wes-otf

hypha/apply/todo/options.py
```diff
@@ -62,23 +62,23 @@
     # :todo: actions for multiple stages of submission
     SUBMISSION_DRAFT: {
         "text": _(
-            'A Submission draft [<span class="truncate inline-block max-w-32 align-bottom ">{related.title}</span>]({link} "{related.title}") is waiting to be submitted'
+            '<span class="text-xs">{related.fund_name} #{related.application_id}</span><br>A draft application <span class="truncate inline-block max-w-32 align-bottom ">"{related.title}")</span> is waiting to be submitted'
```
frjo: `")</span>` This close parenthesis is a left over it looks like.

theskumar: yep

theskumar:
@frjo Following our discussion yesterday, where we noted that `title` is more frequently used than `Round/Lab name` and the application id, I've changed the default `SUSUBMISSION_TITLE_TEXT_TEMPLATE` back to `{title} (#id)`.

I've retained the fund and id in the tasks list for a gradual, non-intrusive trial and adoption of the new format. If the staff finds it beneficial to include ids and fund names as a straightforward way to identify applications, we can consider implementing it in additional areas.

hypha/apply/activity/templates/activity/notifications.html
```diff
@@ -35,7 +35,7 @@
                             <div class="relative timeline-item" id="communications#{{ activity.id }}">
                                 {% ifchanged activity.source.id %}
                                     <div class="py-0.5 mt-4 bg-gray-100 ps-5 pe-2 border-s-2 border-slate-300">
-                                        <span class="text-sm font-semibold text-fg-muted">{{ activity.source_content_type.name|source_type }}</span> <a href="{{ activity.source.get_absolute_url }}">{{ activity.source.title|capfirst }}</a>
+                                        <span class="text-sm font-semibold text-fg-muted">{{ activity.source_content_type.name|source_type }}</span> <a href="{{ activity.source.get_absolute_url }}">{{ activity.source.fund_name }} #{{ activity.source.application_id }}: {{ activity.source.title|capfirst|truncatechars:50|truncatechars:50 }}</a>
```
frjo: One to many `truncatechars`?

hypha/apply/determinations/templates/determinations/includes/determination_button.html
```diff
@@ -1,6 +1,9 @@
 {% load i18n determination_tags workflow_tags %}
 {% if request.user|show_determination_button:submission %}
-    <a target="_blank" href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}" class="button button--primary button--full-width {{ class }}">
+    <a
+        href="{% url 'apply:submissions:determinations:form' submission_pk=submission.id %}"
+        class="button button--primary button--full-width {{ class }}"
+    >
```
frjo, line 6: I'm not fond of `target="_blank"` but it makes some sense here (but less now that we have the option to view the submission in place). It has also been in place for years for determinations and reviews so staff is used to it.

If we make a change it should be in both places.

frjo (changes requested):
Good changes overall! Like the `application_id()` function.

Some minor items needs a look.

hypha/apply/todo/options.py
```diff
@@ -220,11 +222,17 @@ def get_task_template(request, task, **kwargs):
         "link": link_to(related_obj, request),
     }
     if task.code == COMMENT_TASK:
-        # Replace all newlines with spaces
-        template_kwargs["msg"] = " ".join(related_obj.message.splitlines())
+        # Replace all newlines with spaces and truncate to 60 characters
+        message = " ".join(related_obj.message.splitlines())
+        if len(message) > 60:
+            template_kwargs["msg"] = message[:57] + "..."
```
frjo: I prefer using `…` instead of `...`.

hypha/apply/activity/templates/activity/notifications.html
theskumar: yep

hypha/apply/todo/options.py
theskumar: sure

hypha/apply/determinations/templates/determinations/includes/determination_button.html
theskumar: I believe, any person if they really want to have two tabs they would know how to do it, and in most cases they might be just reading the determination and in that case they simply want to view it. 

I'll make an update to reviews.

theskumar: Removed from review button

frjo (approved):

PR #4467, opened 2025-03-24
Updated to latest major versions of almost all python packages.

frjo:
Didn't update "django-filter" since Wagtail 5 depends on an older version.

`wagtail==5.2.8 depends on django-filter>=23.3,<24`


pyproject.toml
```diff
@@ -25,40 +25,40 @@ dependencies = [
...
-    "django-storages~=1.14.2",
-    "django-tables2~=2.7.0",
+    "django-storages~=1.14.5",
+    "django-tables2~=2.7.5",
     "django-tinymce~=4.1.0",
     "django-two-factor-auth~=1.17.0",
     "django-web-components~=0.2.0",
     "django~=4.2.0",
-    "django-stubs~=4.2.0",
+    "django-stubs~=5.1.3",
```
theskumar: ideally this should follow the version of Django

frjo: Fixed

PR #4468, opened 2025-03-26
Refactor "Reports" into its own app

theskumar:
- The app is named "project_reports," and currently, the table names are hardcoded.
- It is currently located under "hypha.apply.projects.reports," but in the future, I plan to move the entire "projects" app/module as "hypha.projects."
- The database tables will stay unchanged; the migration state will show that the models have been relocated to a different app.
- All URLs will stay the same.

This is the initial step among other refactors/fixes for the reporting module.

Additionally, once this is approved, I will proceed to refactor the invoicing module.


frjo:
@theskumar Would it be a big lift to move "apply/project/reports" to "apply/reports"?

This would be more in line with e.g. "apply/determinations" and "apply/review".

theskumar:
> @theskumar Would it be a big lift to move "apply/project/reports" to "apply/reports"?
> 
> This would be more in line with e.g. "apply/determinations" and "apply/review".

It won't be more work as the migrations, tables and apps are isolated now. 

Given that reports are directly tied to projects, I believed that the namespace `projects.reports` is more logical. I'm open to changing it to `apply.reports` or `hypha.reports` if you have a different opinion.

I know Django conventions is to keep a very flat app structure, but I've also seen deviations where it make sense. 

Ref: https://github.com/taigaio/taiga-back/tree/main/taiga
image

theskumar:
If `reports` had a direct dependency on `submission` then `hypha.apply.reports` made more sense. 

Thinking out loud, reports could be been built to skip the project as the middleman, and could have been a direct dependency of `application/submission`

frjo:
> Given that reports are directly tied to projects, I believed that the namespace `projects.reports` is more logical.

Make sense.

theskumar:
If you agree, we should go with two main namespaces for applications and projects, and the supporting global stuff like activities

```
hypha.apply (or hypha.applications)
- hypha.apply.reviews
- hypha.apply.determinations
                
hypha.projects.
- hypha.projects.invoices
- hypha.projects.reports
- hypha.projects.paf

hypha.activities.
hypha.todo.
```

theskumar:
Missed one model responsible for the report form, I'll move them here as well.

theskumar:
> Missed one model responsible for the report form, I'll move them here as well.

The missing one is the `ProjectReportForm` that inherits the StreamForm, i think it's fine in the `projects` module. So no furthers updates to this PR.

frjo:
@theskumar After merging in other PRs this needs a rebase now.

theskumar:
> @theskumar After merging in other PRs this needs a rebase now.

Oh no! :( 

PS: No worries on it now! :)

frjo (changes requested):
Looking good, just need a rebase after merging in some other PRs.

frjo (approved):

PR #4473, opened 2025-03-31
Fix the updated at filter not auto submitting on submission list page

### LINKED ISSUE #4469: Updated (date) filter on submission all does not run automatically
Selecting any date does populate the search field but does not run the query.

Clicking in the search field and pressing enter activated the search and it all works as it should.

theskumar:
Fixes https://github.com/HyphaApp/hypha/issues/4469


frjo (approved):
Works nicely!

PR #4479, opened 2025-04-02
Fix fund_name attribute error for determination

sandeepsajan0:
Fixes Todo list latest updates misses

Draft determination message uses fund_name attribute while determination doesn't have one. So, fixed it by using determination.submission.fund_name. 
Also make it consistent for project, use only project fund name property for project related items.

## Test Steps

 - [ ] …


theskumar (approved):
Nice catch!

frjo (approved):

PR #4480, opened 2025-04-02
Use similar name(with 's' in end) for all file in projects views, forms, and models

sandeepsajan0:
Updated projects' views, models, and forms filenames from project-> projects, payment-> payments, report -> reports to make it similar like other filenames.

## Test Steps

 - [ ] …


hypha/apply/projects/migrations/0069_contractpacketfile_contractdocumentcategory.py
```diff
@@ -72,7 +72,7 @@ class Migration(migrations.Migration):
                     "document",
                     models.FileField(
                         storage=django.core.files.storage.FileSystemStorage(),
-                        upload_to=hypha.apply.projects.models.project.contract_document_path,
+                        upload_to=hypha.apply.projects.models.projects.contract_document_path,
```
sandeepsajan0, line 75: Will these create any issues with existing documents?
I checked for a few documents, seems to be working fine.

wes-otf (approved):

PR #4488, opened 2025-04-08
Remove obsolete proposed_start and proposed_end fields from projects.

frjo:


hypha/apply/projects/reports/views.py
```diff
@@ -239,6 +239,11 @@ def get_due_report_data(self):
         report_data = {}
         if not self.object.disable_reporting:
             project_end_date = self.project.end_date
+            import logging
+
+            logger = logging.getLogger("hypha")
+            logger.debug(self.project)
+            logger.debug(self.project.end_date)
```
theskumar: This can be removed. 
```suggestion
```

hypha/apply/projects/models/project.py
```diff
@@ -372,13 +361,7 @@ def start_date(self):
...
     @property
     def end_date(self):
-        # Aiming for the proposed end date as the last day of the project
-        # If still ongoing assume today is the end
-        if self.proposed_end:
-            return max(
-                self.proposed_end.date(),
-                timezone.now().date(),
-            )
+        # Assume today is the end
```
theskumar, line 364: I think it would be great to include a little explanation on why `today` is chosen as the end date.

theskumar (changes requested):

hypha/apply/projects/reports/views.py
frjo: Oops.

hypha/apply/projects/models/project.py
frjo: Good point.

frjo:
The `proposed_end` date is used to set when repeated reports are to end. Can't find it bing used anywhere else.

@wes-otf Any idea if OTF have any other interest in setting project start and/or end dates?

@theskumar @sandeepsajan0 If we only need an end date for reports should we perhaps make it part of "ReportConfig"?

theskumar:
> [...] 
> @theskumar @sandeepsajan0 If we only need an end date for reports should we perhaps make it part of "ReportConfig"?

Yes, I believe the project start date is picked up from the contract approval, which I think is a little odd, and now that we can skip contracting stages, this could be a problem. If needed, both project start and end dates should be explicit properties.

I vote for the report end time to be stored in the ReportConfig.

wes-otf:
I talked to the program team today and they brought up that it would be nice to have end dates displayed in projects without me even mentioning this PR haha - they specifically wanted to be able to set end dates then filter on them when looking at all the projects to be able to better track projects at a glance. They also requested that this show up in the `Project Information` part of the project view.

I was thinking maybe they could be prompted in that initial project creation dialogue to set the end date (which could be guided by the `Project Duration` field if it exists on the application, suggesting a date 3 months in advanced if duration was set to 3 months) and we could assume "today" as the start date

PR #4489, opened 2025-04-09
Fix duplicate file issue on every save/submit

### LINKED ISSUE #4484: When saving drafts, uploaded files get duplicated on the filesystem until a "File name too long" error starts happening
**Describe the bug**

When an application has a uploaded file, each draft save appends to the filename, until the filename is too long for the filesystem, causing an error.

**To Reproduce**

1. Create an application with a file upload (I used multi file upload, and didn't confirm that single file upload had the same issue)
2. Create an application, uploading a file as part of it, and save it as draft
3. When continuing to edit, hit save draft over and over again
4. See 500 error (also, see the filenames get longer and longer in the media directory)

**Expected behavior**

No error.

**Priority**

- Medium (slowing you down)

**Affected roles**

- Applicants

 - OS: Debian Linux Server

**Additional context**

We run into one of these every month or so.  It seems like the normal workflow for some applicants is to get the files uploaded, and then edit the draft a lot over time until they get it all done.

Also, of note, this really adds to the disk space, as some of these attachments are large and keep getting duplicated.

sandeepsajan0:
Fixes #4484 

## Test Steps

 - [ ] Try to upload files in draft submissions, files name shouldn't get longer every time or file shouldn't be duplicated.
 - [ ] Try file upload for other places as well just to check that it is not breaking anything else.

### DIFF
--- a/hypha/apply/funds/models/mixins.py
+++ b/hypha/apply/funds/models/mixins.py
@@ -1,3 +1,5 @@
+import hashlib
+
 from django.core.files import File
 from django.utils.safestring import mark_safe
 from django_file_form.models import PlaceholderUploadedFile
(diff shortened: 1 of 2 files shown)


frjo (approved):

wes-otf:
worked great!

sandeepsajan0:
@wes-otf We tried a simpler solution in #4499  that seems to be working fine.

PR #4495, opened 2025-04-10
Set project start date to now and add field in create project form to set the end date

### LINKED ISSUE #4494: Make the proposed start & end dates on projects more usable and prominent
Related to the comment made [here](https://github.com/HyphaApp/hypha/pull/4488#issuecomment-2790838568), OTF staff would like to be able to have project start/end dates easily viewable in the project table. This would also include giving staff the option to set & edit these dates if deadlines had changed. Currently this is all being done in a google sheet, which has a column of end dates that staff filter on to see what projects are nearing their completion date.

The staff member should be prompted for these dates when creating the project, maybe it can also be configured on a per fund basis to say if it's optional or not as well? Curious to hear any thoughts on this

frjo:
Fixes #4494

- [x] Set project start date to now and add field in create project form to set the end date
- [x] Show dates on project detail view.
- [x] Make it possible to edit the start and end dates on the project detail view.
- [x] Expose dates in project table.


frjo:
@sandeepsajan0 Can you see why the create project test is failing? Likely something obvious but I can't see it.

hypha/apply/funds/tests/test_views.py
```diff
@@ -329,6 +329,7 @@ def test_can_create_project(self):
                 "project_create_form": "",
                 "project_lead": self.user.id,
                 "project_initial_status": CONTRACTING,
+                "project_end": timezone.now(),
```
wes-otf:
```suggestion
                "project_end": timezone.now().strftime("%Y-%m-%d"),
```
From what I saw locally only using `timezone.now()` was causing the form's `DateField` to be invalid

wes-otf (changes requested):

hypha/apply/projects/tables.py
```diff
@@ -254,10 +255,10 @@ class Meta:
             "fund",
             "reporting",
             "last_payment_request",
-            "created_at",
+            "end_date",
         ]
         model = Project
         orderable = True
-        order_by = ("-created_at",)
+        order_by = ("-end_date",)
```
wes-otf:
```suggestion
        order_by = ("end_date",)
```
I think in their current spreadsheet staff sort by the ascending end dates so this may be a bit closer to that

hypha/apply/projects/templates/application_projects/partials/project_information.html
```diff
@@ -0,0 +1,47 @@
...
+        {% if request.user.is_apply_staff %}
+            <a class="transition-opacity hover:opacity-70 is-active"
+               href="{% url 'apply:projects:project_dates_update' object.submission.id %}"
+               hx-get="{% url 'apply:projects:project_dates_update' object.submission.id %}"
+               hx-target="#htmx-modal"
+            >
+                {% heroicon_solid "pencil-square" class="inline mt-2 ms-1" aria_hidden=true %}
+                <span class="sr-only">{% trans "edit dates" %}</span>
+            </a>
+        {% endif %}
```
wes-otf: should we also put this button on the start date? even though it opens the same modal it might be more intuitive if they wanted to open the start date

hypha/apply/projects/templates/application_projects/modals/project_dates_update.html
```diff
@@ -0,0 +1,6 @@
+{% load i18n %}
+{% modal_title %}{% trans "Update project dates" %}{% endmodal_title %}
+
+<div class="p-4">
+    {% include 'includes/dialog_form_base.html' with form=form value=value %}
+</div>
```
wes-otf, line 6: Not an important detail at all but the date pickers are different widths - not sure if this can be fixed here because of the include template though

Screenshot 2025-04-17 at 09 42 49

wes-otf:
worked really well! a few small comments/thoughts but it should be ready to go!

frjo:
@wes-otf Good changes I believe. I implemented them and deployed to test.

hypha/apply/projects/templates/application_projects/modals/project_dates_update.html
frjo: There was custom css to change the width of the last date field, been there for years.

Removed it and we will see if it mess something up, likely not.

wes-otf (approved):

wes-otf:
seems to work great! I think this is exactly what they were asking for, I'll tag this for user testing to confirm if that's cool with you

frjo:
Test fails because they compare date with time and date without time.

frjo:
Rebased after report app refractor.

wes-otf:
@frjo I have a few more small tweaks coming based on a some of my code changes, might add a few more unit tests too

wes-otf:
I added an error message for the end date being before the start date in the modal along with some info regarding when the start date will be set if `PROJECTS_START_AFTER_CONTRACTING` is enabled - should be about ready
Screenshot 2025-04-21 at 18 02 26

PR #4499, opened 2025-04-16
Easy fix for duplicate file issue

### LINKED ISSUE #4484: When saving drafts, uploaded files get duplicated on the filesystem until a "File name too long" error starts happening

sandeepsajan0:
Fixes #4484   Alternative of #4489 

field_id-uploads have path as file id for already uploaded file but for a new uploaded file it only have the uuid. So, either we can check for the path like instance_type/id/... or we can just look for the file name. For now, I checked only for name but we may check for both name and path like pattern. What do you think @frjo @theskumar

uploads data for existing file `[{'id': 'submission/268/509bee14-f03a-4f3a-bea4-66ad17e97635/20230915-testing-draft-with-specific-1_6.pdf', 'name': '20230915-testing-draft-with-specific-1 (6).pdf', 'size': 3345, 'type': 'placeholder'}]`
but for a new file id is just a uuid 'ef65de0a-e43b-4ebe-93d0-db9784083af'

## Test Steps

 - [ ] …

### DIFF
--- a/hypha/apply/funds/models/mixins.py
+++ b/hypha/apply/funds/models/mixins.py
@@ -1,7 +1,7 @@
 import json
+import uuid
 
 from django.core.files import File
-from django.core.files.storage import default_storage
 from django.utils.safestring import mark_safe
 from django_file_form.models import PlaceholderUploadedFile
 
@@ -98,7 +98,8 @@ class AccessFormData:
     def process_file_data(self, data, latest_existing_data=None):
         for field in self.form_fields:
             if isinstance(field.block, UploadableMediaBlock):
-                # check for the new file in uploads data
+                # {field_id}-uploads contains all files data for that field
+                # get all uploaded files(existing + new)
                 uploaded_files = data.get(field.id + "-uploads", "")
 
                 try:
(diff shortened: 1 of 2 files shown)


theskumar:
I think this approach works well. The `id` corresponds to the ID in the temporary table. Once the file is uploaded to its final destination, that ID is removed from the temp table, and the `path` is used in the JSON data in the destination table.

frjo (approved):

hypha/apply/funds/models/mixins.py
```diff
@@ -92,19 +95,66 @@ def process_file(cls, instance, field, file):
...
+                    self.form_data[field.id] = new_stream_file
+                else:
+                    self.form_data[field.id] = existing_file
+
+    def _is_valid_uuid(self, value):
+        try:
+            uuid.UUID(value)
+            return True
+        except ValueError:
+            return False
```
theskumar, line 157: This can be moved to a common util file at `hypha/core/utils.py` and renamed to `is_valid_uuid`

wes-otf (approved):

PR #4504, opened 2025-04-17
Do not cache application_id and fund_name since that result in "None" values in e-mail and slack notification.

### LINKED ISSUE #4500: Emails show `None` where submission ID should be for drafted applications
It seems that when an email is sent regarding an application being drafted it will display the submission ID as `#None`. Haven't dug deep into what's causing this though.

Image

frjo:
Fixes #4500

### DIFF
--- a/hypha/apply/activity/templates/messages/email/submission_confirmation.html
+++ b/hypha/apply/activity/templates/messages/email/submission_confirmation.html
@@ -3,7 +3,7 @@
 {% load i18n %}
 
 {% block content %}{# fmt:off #}
-{% blocktrans with title=source.title %}We appreciate your {{ title }} application submission to the {{ ORG_LONG_NAME }}.{% endblocktrans %}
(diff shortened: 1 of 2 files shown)


theskumar (approved):

frjo:
@wes-otf We use `@cached_property` for `application_id` switch that to a plain `@property` seems to fix the issue on my local install. Deploying to test now.

I assume it cache a early version that do not have public_is set.

theskumar:
It would have been really nice to the able to cache the `application_id` as it's used at multiple places and often queried as fk relation. I think we should merge this is in now, caching is hard anyway.

frjo:
Perhaps it is better to use a custom solution for the e-mails and keep `@cached_property` since that seems to work in all other instances.

theskumar:
> Perhaps it is better to use a custom solution for the e-mails and keep `@cached_property` since that seems to work in all other instances.

What is the `id` is referenced directly in the email, not via `application_id`? with a added comment on why it's done this way instead of using `application_id`.

PR #4510, opened 2025-04-19
Fix preview & submit button on application form

### LINKED ISSUE #4497: When `SUBMISSION_PREVIEW_REQUIRED` is enabled, selecting `Preview and submit` doesn't preview the application
When `SUBMISSION_PREVIEW_REQUIRED` is enabled, selecting `Preview and submit` will cause the application to immediately submit rather than displaying a preview to the user first

theskumar:
If the submit button is disabled before submit event is processed by the borwser, the name/value on the button is not included in the form value.

This PR fixes that by using `setTimeout`.

There are other minor updates to js, e.g. iterating on submit buttons instead of processing only the first submit button.

Some formatting update html is also there - helps keep html readiable and for better git diff in future.

Fixes #4497

## Test Steps

 - Hitting the "Preview and Submit" works as expected.

### DIFF
--- a/hypha/static_src/javascript/application-form.js
+++ b/hypha/static_src/javascript/application-form.js
@@ -1,7 +1,7 @@
 (function () {
   const form = document.querySelector(".application-form");
   const links = form.querySelectorAll("a");
-  const button = form.querySelector("[type=submit]");
+  const submitButtons = form.querySelectorAll("[type=submit]");
   const required = form.querySelectorAll("input[required]");
   const groups = form.querySelectorAll(".form__group");
   const errors = form.querySelectorAll(".form__error");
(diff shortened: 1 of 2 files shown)


hypha/static_src/javascript/application-form.js
```diff
@@ -51,12 +51,24 @@
...
+    // before the buttons are disabled, allowing their values to be included
+    setTimeout(function () {
+      submitButtons.forEach(function (button) {
+        button.setAttribute("disabled", "disabled");
+        if (button.textContent) {
+          button.dataset.originalText = button.textContent;
+          button.textContent = "Submitting...";
+        }
+      });
+    }, 0);
```
theskumar, line 64: This is the primary fix for form not sending the `preview` key/value.

hypha/static_src/javascript/application-form.js
```diff
@@ -65,9 +77,11 @@
   // 3. tab or enter key pressed
   document.body.addEventListener("mousemove", unlockApplicationForm, {
     once: true,
+    passive: true,
```
theskumar, line 80: Minor optimization.

hypha/static_src/javascript/application-form.js
```diff
@@ -51,12 +51,24 @@
...
   // Block multiple form submits.
   form.addEventListener("submit", function () {
-    button.setAttribute("disabled", "disabled");
+    // Use setTimeout with 0 delay to ensure form submission begins
+    // before the buttons are disabled, allowing their values to be included
+    setTimeout(function () {
+      submitButtons.forEach(function (button) {
+        button.setAttribute("disabled", "disabled");
+        if (button.textContent) {
+          button.dataset.originalText = button.textContent;
```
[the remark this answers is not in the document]
theskumar: The page is always refreshed, so this logic is not needed.

frjo (approved):

PR #4512, opened 2025-04-19
Fix typos in docs, code and comments and "typos" to pre-commit

theskumar:
This package is quite fast and has very low false positives.

Given the number typos worth keeping it was able to find, I think it will be quite helpful for it to be part of our pre-commit and lint pipeline. It's quite easy to add new spellings or regex-based ignore rules.

The exclusion and typo rules can be add to `pyproject.yml`. 

Read more at https://github.com/crate-ci/typos


frjo:
I get this error:

`django.db.migrations.exceptions.InconsistentMigrationHistory: Migration application_projects.0043_remove_project_approval_form is applied before its dependency funds.0089_remove_project_approval_form_references on database 'default'.`

theskumar:
@frjo can you give it a try again?

frjo:
@theskumar Issues fixed.

It's satisfying to see these spelling fixes!

Nice addition to pre-commit so we catch the misspellings early.

frjo (approved):

theskumar:
I was a bit surprised by how many issues it was able to find—no false positives, except at the obvious paths that are excluded now.

The PR is rebased now.

PR #4502, opened 2025-04-17
Refactor project reports permissions

theskumar:
UI Update: 

- Always show the reporting frequency text, use `update_report_config` to control the configuration update button. 

Code changes: 
- Update to use `register_object_checker` and move them to the reports
module
- Add inline docs

PR #4513, opened 2025-04-21
Correct path to import ReportFactory in todo app.

frjo:


theskumar:
Also, fixed at https://github.com/HyphaApp/hypha/pull/4502

PR #4518, opened 2025-04-23
Reporting is disabled by default so adjust text for that.

frjo:
Adjusting the text for disabled reporting since it is now the default.


frjo:
@wes-otf Does this looks ok?

PR #4526, opened 2025-04-30
Update PROJECTS_ENABLED to use middleware for url blocking

### LINKED ISSUE #4514: Running test locally complain that project is not a registered namespace in apply
If I make sure this part always loads the errors goes away.

https://github.com/HyphaApp/hypha/blob/38cd629ba3f9084ea0e8e50e952cf5430aa8cb04/hypha/apply/funds/urls.py#L285-L288

In settings/test.py we have this:

https://github.com/HyphaApp/hypha/blob/38cd629ba3f9084ea0e8e50e952cf5430aa8cb04/hypha/settings/test.py#L14

But it seems it is not loaded for some reasons.

This issue does not affect hypha-ci action on GitHub.

theskumar:
This PR introduces a runtime approach to controlling access to the
Projects feature based on the `PROJECTS_ENABLED` setting. Rather than
conditionally registering URL patterns, a new middleware is implemented
to handle access control to project-related URLs at request time. This
change improves the reliability of the feature toggle system and avoids
potential issues with module loading order and settings configuration.

## 💡 Implementation Details

The middleware works by resolving the current URL path and checking if
it belongs to a "projects" namespace. When the `PROJECTS_ENABLED`
setting is `False` and a user attempts to access a projects-related URL,
they'll receive a 404 error instead of seeing the projects
functionality.

This approach improves upon the previous implementation by:

1. Separating the concerns of URL registration and feature availability
2. Avoiding potential race conditions or initialization issues
3. Providing a more consistent user experience when the feature is
disabled

## 👀 Areas for Review Attention

- Middleware ordering in the settings file - is this the optimal
position?
- Performance impact of URL resolution for every request
- Handling of nested namespaces containing "projects"

Closes https://github.com/HyphaApp/hypha/issues/4514

### DIFF
--- a/hypha/apply/funds/urls.py
+++ b/hypha/apply/funds/urls.py
@@ -1,4 +1,3 @@
-from django.conf import settings
 from django.urls import include, path
 from django.views.generic import RedirectView
 
@@ -280,9 +279,5 @@ rounds_urls = (
 urlpatterns = [
     path("submissions/", include(submission_urls)),
     path("rounds/", include(rounds_urls)),
+    path("projects/", include(projects_urls)),
 ]
-
-if settings.PROJECTS_ENABLED:
-    urlpatterns += [
-        path("projects/", include(projects_urls)),
-    ]
--- a/hypha/apply/projects/middleware.py
+++ b/hypha/apply/projects/middleware.py
@@ -15,14 +15,14 @@ class ProjectsEnabledMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
+        # Skip processing if PROJECTS_ENABLED is true
+        if settings.PROJECTS_ENABLED:
+            return self.get_response(request)
+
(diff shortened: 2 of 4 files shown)


frjo (approved):

frjo:
I just tested the issue in #4514 and at the moment I cannot replicate it. Do other's have that issue from time to time?

theskumar:
> ...Do other's have that issue from time to time?

Yes. I've been facing this very frequently recently. I did narrow it down the PROJECTS_ENABLED and the root cause seems to be that once the urls.py is loaded it's not loaded again.

wes-otf:
I've seen this a bit in my testing too but always figured it was something wrong with my environment as it'd go away once I did a full reload - glad the annoyance is gone!

wes-otf (approved):

PR #4531, opened 2025-05-03
Fix/handle htmx requests for private page on session expiry

### LINKED ISSUE #4530: Login page is rendered in htmx request section
If a user(applicant) is logged out and a tab(dashboard) is already open and applicant click on any submission it swaps login HTML in submissions section only, while it should redirect instead of swap html.

Image

theskumar:
When a user's session expires and they click on a submission in the
dashboard, the login page is rendered within the submissions section of
the page (via HTMX swap) instead of redirecting to the login page. This
creates a broken user experience as shown in the screenshot in the
issue.

Created a new middleware `HtmxAuthMiddleware` that:
1. Intercepts HTMX requests that would normally redirect to the login
page
2. Instead of allowing HTMX to swap the login page content, it issues a
client-side redirect to the login page
3. Preserves the original path in the 'next' parameter so the user
returns to the right page after login

Tested by running unit tests on the middleware that verify:
1. HTMX requests that get redirected to login are properly converted to
client-side redirects
2. Non-HTMX requests are not affected
3. HTMX requests with non-auth redirects are not affected

Fixes #4530


frjo (approved):

wes-otf (approved):

wes-otf:
great fix! worked really well

PR #4533, opened 2025-05-06
Render richtext help text on login pages.

frjo:


hypha/apply/templates/forms/includes/field.html
```diff
@@ -58,7 +58,7 @@
         {% endif %}
 
         {% if field.help_text %}
-            <div class="form__help prose prose-sm">{{ field.help_text }}</div>
+            <div class="form__help prose prose-sm">{{ field.help_text|safe }}</div>
```
frjo: Is this an ok way to solve this? Help text should only come from trusted sources from what I can see.

theskumar: Use nh3 to be on safer side?

frjo: Got a "TypeError: argument 'html': 'lazy.<locals>.__proxy__' object cannot be converted to 'PyString'" and realised I'm doing it wrong.

Shifted to using "mark_safe()" for this specific help_text instead of messing with rendering of all help_text.

I think this is a more correct approach.

theskumar: I agree.

theskumar (approved):

PR #4541, opened 2025-05-14
Fix fund & round getting put together for applicants

wes-otf:
**Before**
Screenshot 2025-05-14 at 12 15 57

**After**
Screenshot 2025-05-14 at 12 15 28


theskumar (approved):
Nice catch!

PR #4546, opened 2025-05-19
Add max length option on Titleblocks that sets maxlength attribute.

### LINKED ISSUE #4544: Limiting application title length
Staff have gotten lots of applications with massive titles lately and have voiced that it'd be nice to set a 60-100 character limit on them

**Priority**

frjo:
Fixes #4544


hypha/apply/templates/forms/includes/field.html
```diff
@@ -14,6 +14,14 @@
          {% if field.field.word_limit %}
              data-word-limit="{{ field.field.word_limit }}"
          {% endif %}
+         {% if field.field.max_length %}
+             data-maxlength="{{ field.field.max_length }}"
+             x-data="{
+                     maxlength: $el.dataset.maxlength,
+                     curlength: 12345
```
frjo: @theskumar Any tips on best way to get the input fields current char count/length here?

Issue #4475, opened 2025-03-31
Replace daterangepicker with easepick  (no-jquery version)


theskumar:
Consider replacing the existing [daterangepicker](https://www.daterangepicker.com/), which relies on jQuery, with Easepick incorporating the `preset` plugin. 

Should you have any suggestions, kindly share them. 

- https://easepick.com/
- https://easepick.com/packages/preset-plugin.html

frjo:
I like the look of it, zero dependencies is a big plus.

frjo:
Also take a look at https://daisyui.com/components/calendar/

PR #4549, opened 2025-05-20
Better error handling for dates, added submitted filter criteria

### LINKED ISSUE #4548: Typing incorrect date into the query/search bar results in 500 error
Modifying dates in the query bar and submitting them in an invalid format (ie. `submitted:>=205-04-01 submitted:<=2025-04-30` instead of `submitted:>=2025-04-01 submitted:<=2025-04-30`) results in a 500 error

wes-otf:
Fixes #4548. This handles an invalid date the same as invalid/non-existent query params rather than erroring out. ~~In the long term, it would be better to indicate to the user that something wasn't right with their query though or no results were brought.~~ When an invalid query is entered, it'll show the no results prompt!

I also added the submitted drop down filter, as staff have been using that a bit recently

## Test Steps

 - [ ] Ensure putting an incorrectly formatted date in the query bar doesn't result in a 500 error
     - ie. `submitted:>=205-04-01 submitted:<=2025-04-30`
 - [ ] Ensure applications can be filtered by submitted date using the new drop down

### DIFF
--- a/hypha/apply/funds/templates/submissions/all.html
+++ b/hypha/apply/funds/templates/submissions/all.html
@@ -365,12 +365,17 @@
             <nav x-show="!showSelectedSubmissions"
                  class="flex flex-wrap gap-2 items-center menu-filters"
             >
-                <div id="filterupdated" aria-label="Filter by Updated" class="flex items-center">
+                <div id="filtersubmitted" aria-label="Filter by Submitted" class="flex items-center" data-query="submitted">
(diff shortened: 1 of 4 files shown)


wes-otf:
As we move away from jQuery it might be nice to check out a vanilla JS datepicker like https://air-datepicker.com too

Edit: this was already mentioned in #4475

frjo (approved):
Works well in my testing. Nice new submitted filter.

PR #4492, opened 2025-04-10
Add Co-applicants feature to submissions

sandeepsajan0:
Fixes #3919

Applicant UI, default open
image

Invite co-applicant, (on click invite button).
image

Update roles, after accept
image

Staff UI, default collapsed.
image

Co-applicant UI, invite landing page
image

Invalid invite page

image

## Test Steps

 - [ ] …

PR #4552, opened 2025-05-24
Expand co-applicant feature to projects

sandeepsajan0:
Extension of #4492 

Added project permission to co-applicant.
image
image

## Test Steps

 - [ ] …


frjo:
Works well in my testing. A few observations.

1. Should this notification maybe say "Co-applicant invited"?

Skärmavbild 2025-05-27 kl  13 52 05

----

2. The delete button is a bit easy to click by mistake. Maybe make it a text link so it is a smaller traget? Other ideas?

Skärmavbild 2025-05-27 kl  10 18 16

----

3. Do applicants understand what "Project" is? Do we need these fine grained permissions? @wes-otf?

Skärmavbild 2025-05-27 kl  10 18 56

sandeepsajan0:
> The delete button is a bit easy to click by mistake. Maybe make it a text link so it is a smaller traget? Other ideas?

Text link is a good option, but not alarming enough. Red color button seems a bit more alarming, and having it with a confirm stage/popup might prevent deletion by mistake.
image

>Do applicants understand what "Project" is? Do we need these fine grained permissions?

They might not know the whole project thing/process, but they are aware of the new 'Project' tab, and i guess we also send them a mail about project creation. 
Actually, that's the reason I used page sections for project permission instead of statuses(even permission names match with section names), because applicants would see all those sections separately, so if they want to show/hide permission of any section, they can just select accordingly. And also added a descriptive help text of how both permissions would work together. It would be great if we can make it more simpler.
image
image
image

frjo (approved):

wes-otf:
I'll be testing this afternoon! these comments make sense to me though, if we're already presenting with `Project` verbiage in the tab I think it makes sense to keep it that way, if we felt more clarity was needed we could just say `Project Management` or something

sandeepsajan0:
@wes-otf I have renamed Read Only to View as you suggested, and keeping the comment for now, we may wait to see the responses, if applicants want everyone to comment then we may later make it default with View.

@frjo I think it is ready to be put on the test.

wes-otf:
This looks good so far! Testing it locally I had a few quick notes but this is great:
- The co-applicants card should also appear in the projects tab
- Maybe it would also be useful to add notifications of reports being due/frequency changes to co-applicants with report edit permissions \(via a custom function [here](https://github.com/HyphaApp/hypha/blob/6ec7aa22d879c731f97a0bc123997ac40572952c/hypha/apply/activity/adapters/emails.py#L84)\)
- I could also see it beneficial to staff to show what user submitted a report/invoice if multiple are on a project

sandeepsajan0:
@frjo I have added the co-applicant block to the projects tab as well. We may move forward with this one.

wes-otf:
I'll test this on test tomorrow! Looks pretty good from what I can see though, not expecting any issues

wes-otf:
Testing went well! Last question is if the project already exists can `project_permission`s be included in the invite form too? Just so applicants don't have to invite then go back & edit co-applicant permissions once the user accepts if they want them to have project permissions.

Other than that this is ready for merge I think!

Issue #4566, opened 2025-06-03
PDF download of submission fails when projects are not enabled


bmispelon:
**Describe the bug**

Downloading an application's PDF doesn't work when projects are not enabled.

**To Reproduce**

1. Set up a hypha instance that doesn't use projects (`settings.PROJECTS_ENABLED = False`)
2. Create an application
3. As an admin, visit the application's page
4. Click the button Actions to take > More actions > Download PDF

**Expected behavior**

I would expect a PDF file to be downloaded. Instead, a 404 page is displayed.

**Priority**

The bug itself seems of pretty high priority since it breaks an essential (for our client anyway) feature.

**Affected roles**

- Staff

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

I've narrowed down the issue to the implementation of `SubmissionDetailPDFView.get_object()`:
```python
    def get_object(self, queryset=None):
        obj = super().get_object(queryset)

        if not hasattr(obj, "project"):
            raise Http404

        return obj
```

Removing the `if not hasattr(obj, "project")` block (actually the whole `get_object()` custom method) makes the PDF download work again.

Looking at the git history, it seems this code has been there since [the PDF download feature was introduced](https://github.com/HyphaApp/hypha/commit/17d292f9e9775e37e4144b1a1ee7028d8fac29b0). But I couldn't find an explanation as to why it was needed in the first place.

frjo:
Thanks for the report! Yes, it is this code that trigger this bug:

https://github.com/HyphaApp/hypha/blob/dbc62f03708604cd352367fc772f853b1617787c/hypha/apply/funds/views/submission_detail.py#L275-L285

@sandeepsajan0 Can you see any reason for the "project" check? Looks like we could just remove these lines to fix the bug.

We have had other similar functions and this was likely copied from there. Surprised that this bug was not discovered earlier.

sandeepsajan0:
@frjo Yes, It feels like copied from somewhere and there is no use of the project in this view. But permission for the view is `staff_or_finance_required` and finance has nothing to do with it until a project is created. I guess it was added for finance to look into/download budget-related things. 

I would say instead of removing the whole code we may add a condition for finance check. If user is finance then project must be there otherwise don't check for the project.

[...]

PR #4567, opened 2025-06-03
Fix overlapping lines in header of submission PDF

bmispelon:
I noticed that when the name of the lead on an application is quite long, the "meta" text in the PDF goes over several lines and ends up overlapping (see attached screenshots).

I tracked it down to what I believe is a missing line in `apply/utils/pdfs.py` (compared to the similar logic in handling `split_title` in the same function).

This seemed like a minor change so I didn't create an issue, please let me know if you'd prefer I did that.

## Screenshots 

<details>
<summary>Before</summary>

Screenshot 2025-06-03 at 15-30-51 asdf - asdf-11 pdf

</details>

<details>
<summary>After</summary>

Screenshot 2025-06-03 at 15-30-34 asdf - asdf-12 pdf

</details>

## Test Steps

1. Set up a hypha instance (with `USE_PROJECTS=True` for PDF download to work, see #4566)
2. Log in as an admin user (so you have access to PDF download)
3. Change your display name to something fairly long (~80 characters, with spaces)
4. Go to a submission's detail page and download the PDF (Actions to take > More actions > Download PDF)

### DIFF
--- a/hypha/apply/utils/pdfs.py
+++ b/hypha/apply/utils/pdfs.py
@@ -349,6 +349,7 @@ def draw_title_block(canvas, doc, page_title, title, meta, page_width, page_heig
     pos -= meta_size * 2
 
     for line in split_meta:
+        pos -= meta_size
         canvas.drawString(
             doc.leftMargin + FRAME_PADDING,
             pos,


sandeepsajan0 (approved):
Looks good!
Nice catch on the meta text wrapping. It was happening when any of the meta fields like workflow, round name, lead name, etc., were long enough to wrap it in multiple lines.

PR #4569, opened 2025-06-04
Fix submission PDF download when project do not exist

### LINKED ISSUE #4566: PDF download of submission fails when projects are not enabled

frjo:
Fixes #4566

### DIFF
--- a/hypha/apply/funds/views/submission_detail.py
+++ b/hypha/apply/funds/views/submission_detail.py
@@ -19,7 +19,6 @@ from django.views.generic.detail import SingleObjectMixin
 from hypha.apply.activity.messaging import MESSAGES, messenger
 from hypha.apply.activity.views import ActivityContextMixin
 from hypha.apply.users.decorators import (
-    staff_or_finance_required,
     staff_required,
 )
 from hypha.apply.utils.models import PDFPageSettings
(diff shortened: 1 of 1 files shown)


hypha/apply/funds/views/submission_detail.py
```diff
@@ -272,16 +271,13 @@ def should_redirect(cls, request, submission):
             )
 
 
-@method_decorator(staff_or_finance_required, name="dispatch")
+@method_decorator(staff_required, "dispatch")
 class SubmissionDetailPDFView(SingleObjectMixin, View):
     model = ApplicationSubmission
 
     def get_object(self, queryset=None):
```
sandeepsajan0: We can remove the whole method.

frjo: We check access somewhere else for this?

frjo: Or you mean get_object()?

frjo: Yes, I see, that is redundant now.

theskumar (approved):
💯

PR #4572, opened 2025-06-04
Fix submission filters queryset issue

### LINKED ISSUE #4558: When filtering by reviewer in the submissions table by someone who has never left a review, all submissions appear.
**Describe the bug**

When filtering by reviewer in the submissions table of someone who has never left a review, all submissions appear.

**To Reproduce**

1. Go to All Submissions
2. Filter by a reviewer who has never reviewed anything
3. See all submissions

**Expected behavior**

No submissions should show up

**Priority**

**Affected roles**

- Staff

sandeepsajan0:
Fixes #4558 

All the code removed in this PR were used for previous/old filter patterns. It was used to update filter datasets dynamically but now we are treating it in a different way and handling those things in submission all view so it is not needed here. It was also messing up with filtered queryset for current implementation. 
Now it should show/add/remove 0 submission if filtered query doesn't belong to any submission.

## Test Steps

 - [ ] Just test/play around the filters of submission all page.

### DIFF
--- a/hypha/apply/funds/tables.py
+++ b/hypha/apply/funds/tables.py
@@ -235,30 +235,13 @@ def get_used_rounds(request):
     return Round.objects.filter(submissions__isnull=False).distinct()
 
 
-def get_used_rounds_from_dataset(dataset):
-    return Round.objects.filter(id__in=dataset.values("round")).distinct()
-
-
 def get_used_funds(request):
     # Use page to pick up on both Labs and Funds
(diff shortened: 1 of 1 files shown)


frjo (approved):
Works perfectly in my testing, only remove code and fix bug!

I noticed that we have the same issue with the Category filter, show all items if no match. But that is a separate issue for another PR I think.

wes-otf (approved):

PR #4577, opened 2025-06-08
Use xhtml2pdf inplace of reportlab for submission pdf export

### LINKED ISSUE #4255: Make use of xhtml2pdf instead of reportlab
Evaluate the performance and layout different and consider replacing reportlab with xhtml2pdf for pdf generation. 

Reason: working with html and css is much easier and maintainable compared to reportlab.

sandeepsajan0:
Fixes #4255 

Sample pdf:

image

## Test Steps

 - [ ] …


sandeepsajan0:
Updated to this pattern, where the header can be used in other exports as well.
image

theskumar:
Based the discussion the the final layout: 

image

frjo:
@sandeepsajan0 Did you have some more things to add here or are we good to merge?

sandeepsajan0:
@frjo Just updated the header to make it in a single line.  Nothing like breaking changes so we can move forward.
image

PR #4580, opened 2025-06-10
Fix report file path issue and duplicate file issue, Fix file path issue for Project and SOW Form

### LINKED ISSUE #4574: Trying to access a file uploaded in a report results in a 404
Some of the OTF vendors submit their report in the form of a PDF/ODT. When trying to download/view the file, Hypha throws a 404. This can be seen here: https://test-apply.hypha.app/apply/projects/reports/3578/ when trying to click on the uploaded file.

sandeepsajan0:
Fixes #4574 

As reports are project documents so we added a different path for its files but we used ReportVersion instance checks and missed Report instances so reports files and its version files were having different path and were creating issues. We could solve it in that way as well but I realise as projects are also the extension of submission and we have been trying to merge them so keep everything similar with same url pattern.

Now it uses submission id to generate path and render the file similar to other submission files.

Now, it also handles the duplicate file issue on every upload. It was resolved for the application form but not for the report forms.

## Test Steps

 - [ ] Upload a file to report form.
 - [ ] Check if it is accessible.
 - [ ] Edit and submit form without uploading new file, file name shouldn't be changed like it used to `filename_<some hex code>.ext`

### DIFF
--- a/hypha/apply/funds/files.py
+++ b/hypha/apply/funds/files.py
@@ -22,7 +22,7 @@ class PrivateStreamFieldFile(StreamFieldFile):
 
     def get_entity_id(self):
         from hypha.apply.funds.models import ApplicationRevision
-        from hypha.apply.projects.reports.models import ReportVersion
+        from hypha.apply.projects.reports.models import Report, ReportVersion
 
         entity_id = self.instance.pk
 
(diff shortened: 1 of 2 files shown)


wes-otf (approved):

sandeepsajan0:
As @frjo mentioned [here](https://github.com/HyphaApp/hypha/issues/4574#issuecomment-2948455288), this issue exists with all other stream forms that have a file field, including Project Form and SOW form. It is now fixed for all other existing forms like project form, Sow, and report. 
@frjo we may put its latest version to test.

wes-otf:
Works well, the report files are accessible!

> Edit and submit form without uploading new file, file name shouldn't be changed like it used to `filename_<some hex code>.ext`

Did you mean that the file name shouldn't have hex appended at all or it shouldn't update it after resubmitting? It did append hex originally but after resubmitting via edit that hex stayed the same

Screenshot 2025-06-16 at 13 11 54

Should be ready to go if that was intended though!

frjo:
The file name will have the hex but it should not change on each edit unless the file itself has changed.

PR #4586, opened 2025-06-17
Add Author to report

### LINKED ISSUE #4565: Add author to reports
With the new co-applicant feature it would be good to have author visible on reports so staff knows what applicant/co-applicant has submitted it.

The author is shown in activities but let's add it to report detail view as well.

Would likely be best to add a user_id field to reports for this feature.

sandeepsajan0:
Fixes #4565 

image

## Test Steps

 - [ ] …


frjo:
Rebased from themes branch and made minor adjustment to report template.

Looks good!

PR #4590, opened 2025-06-20
Better handling for users entering long names

### LINKED ISSUE #4589: Users can submit names longer than the field maximum
When `FORCE_LOGIN_FOR_APPLICATION = False`, submitting a new application while not logged in allows you to enter a name longer than the Postgres char max for that field, which is 255 characters.

wes-otf:
Fixes #4589. This adds a global max character limit to the `full_name` field, working both in the application form and the account settings. There's a few other fixes also bundled in this:
 - Better handling for displaying long names in application details, submission all (table & list views), etc
 - In account view, the slack max input field not showing character count
 - Added translation to a few strings

## Test Steps

 - [ ] Set `FORCE_LOGIN_FOR_APPLICATION = False`
 - [ ] Ensure when submitting an application and not logged in, you cannot submit a name longer than 255 chars
 - [ ] Ensure character count updates on typing for both name & slack fields in the account view


theskumar:
Can we do this pull request against the themes branch please?

wes-otf:
@theskumar wasn't thinking when I made the branch originally - should be set now though

hypha/apply/funds/blocks.py
```diff
@@ -123,6 +124,7 @@ def prepare_data(self, value, data, serialize):
 class FullNameBlock(ApplicationMustIncludeFieldBlock):
     name = "full_name"
     description = "Full name"
+    widget = forms.TextInput()
```
theskumar: I think this is not needed. I need to remove from the title block as well.

theskumar (changes requested):

theskumar:
Minor comment, otherwise looks good to me. I think you can ahead and merge after that minor change.

hypha/apply/funds/blocks.py
wes-otf: it wasn't! I also removed it from the title block too so you didn't have to

PR #4591, opened 2025-06-23
Updated python packages June 2025.

frjo:


frjo:
@wes-otf Can you test and see if the updates to the translations packages works as they should?

wes-otf:
yeah I can do that today

PR #4594, opened 2025-06-28
Make timestamp consistent for projects

### LINKED ISSUE #4528: Making timestamps through Hypha more consistent
As it stands, timestamps seem to use a mix of UTC and the user's timezone without specifying which is which.

For example, when hovering over comments/activities that have an abbreviated date, the user's timezone is displayed:

Image

while the determinations block (same activity as the event above) uses UTC without indicating it:

Image

This could potentially be misleading and is prevalent in quite a few places throughout Hypha, especially the projects side. Maybe we could use the `relative-time` element more like was done in the first example to normalize timestamps?

sandeepsajan0:
Fixes #4528 

In projects, we use only dates(no timestamp) in most places. I found the timestamp only at two places that too from activities. One of them is in invoice_detail page that has updated in this PR, another is in the pdf_invoice_approved_page.html, which is in a generated PDF so relative time won't make sense here.

## Test Steps

 - [ ] Check timestamps around project(including invoices, reports) and it should be relative time everywhere.


theskumar:
This is already fixed in the themes branch.

theskumar:
closing this now.

PR #4599, opened 2025-07-05
Fix Project lead visibility for org memeber like finance

sandeepsajan0:
Fixes Project lead visibility for finance and other org faculty if Hide staff identity is True.
Project Lead shouldn't be visible to Applicant but visible to finance even if Hide Staff Identity is True.

## Test Steps

 - [ ] …


sandeepsajan0:
I am thinking of merging it into themes after getting it into the main because it is different from themes.

PR #4604, opened 2025-07-11
Removed nested htmx call causing triggering of submission export.

frjo:
This got lost in the themes merge.

### DIFF
--- a/hypha/apply/funds/templates/submissions/all.html
+++ b/hypha/apply/funds/templates/submissions/all.html
@@ -38,134 +38,136 @@
 
 
 {% block content %}{% spaceless %}
-    <form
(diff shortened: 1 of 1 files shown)


frjo:
@wes-otf Does this look correct?

theskumar:
Is it currently happening? In the themes, it's converted to link instead of button and I can't reproduce it.

frjo:
On the test site I saw this issue earlier today, now this PR is on test so no more issue.

You only see it when celery/async is in place.

theskumar:
okay, make sense i didn't had that enabled locally

theskumar (approved):

wes-otf:
worked for me!

PR #4620, opened 2025-09-05
Fixed broken user permission in multi-language setup (#4619)

### LINKED ISSUE #4619: Switching languages breaks user permission checks
**Describe the bug**

I run a Hypha install that uses multiple languages and lets the user switch between them (`settings.LANGUAGE_SWITCHER == True`).
We started noticing a strange behavior where **some** (not all) reviewers who were assigned to a submission didn't have the "[Review]" button available.
Stranger still, logging in as those users using the hijack feature (`settings.HIJACK_ENABLE == True`) the site appeared to work correctly (ie the "[Review]" button was present).

After a few hours or trying various things, we finally figured out that it had to do with the user's active language: switching to the site's default language in the language switcher (top right) made the "[Review]" button appear. Switching again to another language made the button disappear.

**To Reproduce**

1. Create a user X that has both "Staff" and "Reviewer" roles
2. Submit an application, get it to the "internal review" stage
3. Assign the reviewer X to that application
4. Log in as X and view the application, observe that the "[Review]" button is there
5. Switch language
6. Observe that the "[Review]" button is now gone.

Alternatively, here's a failing test case (I put it in a new `hypha/apply/users/tests/test_models.py` file but it should work anywhere I think):
```
from django.test import TestCase
from django.utils import translation

from hypha.apply.users.tests.factories import StaffFactory


class UserModelTestCase(TestCase):
    @classmethod
    def setUpTestData(cls):
        cls.user = StaffFactory()

    # Requires compiling the translations first:
    # uv run --with-requirements=requirements/dev.txt python -m django compilemessages --ignore=.venv/* --locale=ru
    def test_is_apply_staff(self):
        self.assertTrue(self.user.is_apply_staff)
        translation.activate("ru")
        del self.user.is_apply_staff  # Clear cached value
        self.assertTrue(self.user.is_apply_staff)
```

**Priority**

**Affected roles**

We noticed it for Reviewers, but I'm pretty sure this bug affects most roles.

**Additional context**

The source of the issue is that (some) permission checks are done by comparing group names. A user's groups come from the database where the group names are translated using the site's default language (`settings.LANGUAGE)`.
But when the language is switched using the language switcher, the constants in `hypha.apply.users.roles` end up being translated too, and they don't match the values stored in the database anymore, making the permission checks fail.

bmispelon:
Fixes #4619

See issue for details. I'm not convinced that this fixes ALL the translated-related issues in the codebase, but it should at least fix all the ones that involve `user.is_XXX`.

### DIFF
--- a/hypha/apply/users/models.py
+++ b/hypha/apply/users/models.py
@@ -254,7 +254,7 @@ class User(AbstractUser):
         return [g.name for g in self.groups.all()]
 
     @cached_property
-    @override(language=settings.LANGUAGE_CODE)
+    @defaultlocale
     def is_apply_staff(self):
         return STAFF_GROUP_NAME in self.roles or self.is_superuser
 
--- a/hypha/apply/users/models.py
+++ b/hypha/apply/users/models.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib.auth.hashers import make_password
 from django.contrib.auth.models import AbstractUser, BaseUserManager
 from django.core import exceptions
@@ -8,6 +9,7 @@ from django.urls import reverse
 from django.utils.crypto import get_random_string
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
+from django.utils.translation import override
 from wagtail.admin.panels import FieldPanel, MultiFieldPanel
 from wagtail.contrib.settings.models import BaseGenericSetting, register_setting
 from wagtail.fields import RichTextField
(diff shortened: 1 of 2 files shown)


wes-otf:
This is huge, thanks for this PR @bmispelon! I'll be reviewing & testing today

hypha/apply/users/models.py
```diff
@@ -244,34 +254,42 @@ def roles(self):
         return [g.name for g in self.groups.all()]
 
     @cached_property
+    @override(language=settings.LANGUAGE_CODE)
```
wes-otf:
```suggestion
    @defaultlocale
```

wes-otf:
All looks good to me! Was able to recreate, just super small comment to use the created decorator. Thanks again for raising this!

PR #4624, opened 2025-10-02
Functionality for the deleting application attachments

### LINKED ISSUE #4622: Application attachments aren't deleted with the application
When an application is deleted, all attached files are retained in the system. These files should be deleted with the associated application

wes-otf:
Fixes #4622. Deletes application attachments when an application is deleted and includes a migration to retroactively delete orphaned attachments.

Also removes a debug statement accidentally included in a PR

## Test Steps

 - [ ] Ensure that attached files are deleted with their associated application
 - [ ] Ensure that previously deleted applications that had attachments still in the system have been deleted


wes-otf:
There's a few more things left on this PR:

- [ ] Deleting of empty directories after an application is deleted (migration handles this)
- [ ] Test passing! That's been the biggest hold up, I can't pin why the `test_gets_draft_on_edit_submission` test fails but feel close

hypha/apply/funds/signals.py
```diff
@@ -0,0 +1,41 @@
...
+    Before the deletion of any sub class of AccessFormData, ensure the files associated with it are deleted too.
+
+    This can include things like ApplicationSubmission & ApplicationRevision objects
+    """
+
+    # TODO: This solution is not ideal but due to our unit tests writing to the filesystem
+    #       this can cause files belonging to the dev's local server to be deleted. Until
+    #       these can be better isolated, this signal will do nothing when pytest is running
+    if "pytest" in sys.modules:
+        return
```
wes-otf, line 25: This solution isn't ideal but I think until we can take a larger pass at our tests and making them more isolated it makes sense? When trying to run actual tests the files for applications on my local server were getting deleted, along with some [weird stuff coming up in the unit tests](https://github.com/HyphaApp/hypha/actions/runs/18226322488/job/51898419828) that I couldn't recreate in a running instance.

wes-otf:
All worked well on test! I've tested both on local storage and AWS, all deleted the attachments as they were supposed to. Gonna merge this

PR #4631, opened 2025-11-06
Update to Django 4.2.26 plus other python packages minor updates.

wes-otf:
Originally from @frjo, I accidentally rebased to main


wes-otf:
Thanks for being so quick with this @frjo! Hopefully my botched merge didn't blow up your email

PR #4633, opened 2025-11-18
Fix invalid end tag in email template

bmispelon:
The `hypha/apply/activity/templates/messages/email/project_final_approval.html` template incorrectly uses `{% blocktrans %}...{% blocktrans %}` instead of `{% blocktrans %}...{% endblocktrans %}`.

I found this while trying to run `makemessages` to generate the translatable strings for our project.


frjo:
Thanks for the fix!

PR #4636, opened 2025-11-24
Fix display of [View all] button for reviews on submission page

bmispelon:
After upgrading Hypha to the latest available version, we noticed that the "Reviews & assignees" > [View all] button on the application page (Admin) was missing.

Looking at the commit history, this appears to have changed in 9cfc791538a13aa222c59996666b00ba64ab09af.


wes-otf:
@bmispelon This is a great catch! A staff member here flagged this for me but I didn't get a chance to look at it yet. Thanks for this!

PR #4637, opened 2025-11-25
Fix incorrect logic in review detail template

bmispelon:
The `{% if %}` logic was erroneously inverted in 9cfc791538


bmispelon:
This is suspiciously similar to #4636 which I filed yesterday. Was this the work of an LLM?

wes-otf:
Also a good catch! @theskumar any idea how these could've made it in? anywhere else we should check?

PR #4638, opened 2025-11-26
Fix missing org name (short and long) in email templates

bmispelon:
I noticed that the signature of the application notification emails was off:
```
Kind Regards,
The  Team

--

http://example.com
```

While investigating I found out that the `base.html` email template was using `org_long_name` instead of `ORG_LONG_NAME` (which is what the `global_vars` [context processor](https://github.com/HyphaApp/hypha/blob/d26405a02bf8ef8ac348992c3a12f6c100f82bab/hypha/core/context_processors.py#L6) is providing).
This issue appears to be widespread, and it's not helped by the fact that some emails correctly receive the `org_long_name` variable (it's passed manually in the context).

For this PR, I went with the exhaustive approach of replacing every single instance of `org_long_name`, `org_short_name`, and `org_email` with their uppercase counterparts:
```
git grep -lw org_long_name | xargs sed -i 's/\borg_long_name\b/ORG_LONG_NAME/'
git grep -lw org_short_name | xargs sed -i 's/\borg_short_name\b/ORG_SHORT_NAME/'
git grep -lw org_email | xargs sed -i 's/\borg_email\b/ORG_EMAIL/'
```

### DIFF
--- a/hypha/apply/activity/adapters/emails.py
+++ b/hypha/apply/activity/adapters/emails.py
@@ -150,11 +150,11 @@ class EmailAdapter(AdapterBase):
             else:
                 try:
                     subject = source.page.specific.subject or _(
-                        "Your application to {org_long_name}: {source.title_text_display}"
-                    ).format(org_long_name=settings.ORG_LONG_NAME, source=source)
+                        "Your application to {ORG_LONG_NAME}: {source.title_text_display}"
+                    ).format(ORG_LONG_NAME=settings.ORG_LONG_NAME, source=source)
                 except AttributeError:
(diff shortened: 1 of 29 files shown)


bmispelon:
I just realized this breaks `compilemessages` because it introduces duplicate message definitions. Converting this PR to draft while I figure out how to fix that.

**Edit**: I ended up fixing the duplicate messages manually by opening the `po` files with `poedit` and letting it clean them up.

wes-otf:
Hey this looks good, thanks for putting it up! I think I'm going to open another PR after merging this that'll automatically add the `global_vars` context to Markdown emails too so we can toss a lot of that context code, gets redundant.

PR #4641, opened 2025-11-27
Restore colors for word limit warning message in tinymce widget

bmispelon:
The colors went missing with the daisyUI refactor 9cfc791538a13aa222c59996666b00ba64ab09af

The use of `!important` is not great, but it's already widespread inside this file (because the tinymce stylesheet is loaded after Hypha's and therefore takes precedence).

## Screenshots

<details><summary>Before</summary>

Screenshot 2025-11-27 at 16-26-51 Max words

</details> 

<details><summary>After</summary>

Screenshot 2025-11-27 at 16-29-04 Max words

</details>

### DIFF
--- a/hypha/static_src/tailwind/components/tinymce.css
+++ b/hypha/static_src/tailwind/components/tinymce.css
@@ -45,14 +45,14 @@
 
   .tox-statusbar__wordcount.word-count-warning,
   .tox-statusbar__wordcount.word-count-warning-2 {
-    background-color: var(--color-warning);
-    color: var(--color-warning-content);
-    font-weight: bold;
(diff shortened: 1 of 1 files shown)


frjo:
Looks good.

Thanks for finding and fixing all these bugs that creeped in!

The daisyUI refactor was pushed through faster than normal since OTF was forces to cut back on development.

bmispelon:
> Thanks for finding and fixing all these bugs that creeped in!

I should say that my work was made possible thanks to a contract between my employer (Torchbox) and our client (DigitalHub.sh). A great example of companies contributing back to open source, I'd love to see more of that 💚

PR #4666, opened 2026-01-10
Update to Django 5.2 LTS  and Wagtail 7 LTS

### LINKED ISSUE #4250: Upgrade to Django 5.2 LTS
Hypha are now on Django 4.x LTS. That version is supported until Q1 2026. Before that Hypha will be upgraded to run on Django 5.2 LTS that will be supported until Q1 2028.

See https://www.djangoproject.com/download/#supported-versions

Alongside this we should also upgrade to Wagtail 7.x LTS release that supports Django 5.2.

See https://github.com/wagtail/wagtail/releases/tag/v7.0

frjo:
Fixes #4250

These are big updated so everything needs to be tested.

Things to test:

- [ ] The complete login procedure.
    - [x] With enforce 2FA
    - [ ] For staff via Google
    - [x] Without password
    - [x] With password
    - [x] Forgot password
    - [x] Change password
- [x] Add and remove 2FA
- [x] The complete submission process for new and for old users.
- [x] The complete review process for reviewers and for staff.
- [x] Alla kind of file uploads and downloads. Do they work and do they have the correct permissions.
- [x] Permissions for all important views.
- [x] Notifications via slack, e-mail etc.


hypha/apply/utils/blocks.py
```diff
@@ -162,17 +162,19 @@ def add_error_to_child(self, errors, child_number, field, message):
...
+    #     """
+    #     This allows historic data to still be accessible even
+    #     if a custom field type is removed from the code in the future.
+    #     """
+    #     # If the data type is missing, fallback to a CharField
+    #     for child_data in value:
+    #         if child_data["type"] not in self.child_blocks:
+    #             child_data["type"] = "char"
+
+    #     return StreamValue(self, value, is_lazy=True)
```
frjo, line 177: This needs investigation.

wes-otf: @frjo so what you're saying is there was an change to StreamChild that broke this?

wes-otf: In my testing I have yet to hit any snags while submitting & rendering stream fields

wes-otf:
so far everything has been working great & as expected!

hypha/apply/utils/blocks.py
frjo: Yes, need to compare the overridden funktion in Wagtail 5 vs 6/7.

wes-otf:
I found admin broke on the test instance until I deleted https://github.com/HyphaApp/hypha/blob/main/hypha/core/wagtail_hooks.py

it didn't locally for some reason though

frjo:
Removed that file now. I removed all the wagtail css override files but missed this hook. It worked locally since the file was already in `static`.

wes-otf (approved):
worked well for me so far! did a lot of testing on the test instance last week too as I was giving team demos, I think it's ready to go!

PR #4679, opened 2026-01-26
Fix application status in confirmation email context (4678)

### LINKED ISSUE #4678: Application confirmation email incorrect states that application is in draft
**Describe the bug**

When submitting a drafted application, the confirmation email sent to the applicant incorrectly states that the application is still in draft.

**To Reproduce**

1. Open application form, fill in required data
2. Press [Preview and submit]
3. Press [Submit for review]
4. Observe that the confirmation email contains the sentence "Please note that it is not submitted for review because it's still in draft."

**Expected behavior**

The confirmation email should instead say "We will review and reply to your submission as quickly as possible."

**Priority**

**Affected roles**

- Applicants

bmispelon:
Fixes #4678

### DIFF
--- a/hypha/apply/funds/models/submissions.py
+++ b/hypha/apply/funds/models/submissions.py
@@ -1054,6 +1054,7 @@ class ApplicationSubmission(
                     code=SUBMISSION_DRAFT, user=by, related_obj=instance
                 )
                 # notify for a new submission
+                instance.status = target
                 messenger(
                     MESSAGES.NEW_SUBMISSION,
                     request=request,
--- a/hypha/apply/funds/tests/test_views.py
+++ b/hypha/apply/funds/tests/test_views.py
@@ -5,6 +5,7 @@ import wagtail.blocks
 from bs4 import BeautifulSoup
 from django.conf import settings
 from django.contrib.auth.models import AnonymousUser
+from django.core import mail
 from django.http import Http404
 from django.test import RequestFactory, TestCase, override_settings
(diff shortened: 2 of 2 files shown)


hypha/apply/funds/models/submissions.py
```diff
@@ -1054,6 +1054,7 @@ def log_status_update(self, descriptor, source, target, **kwargs):
                     code=SUBMISSION_DRAFT, user=by, related_obj=instance
                 )
                 # notify for a new submission
+                instance.status = target
```
sandeepsajan0, line 1057: I am not sure, but it might cause issues at another place because it is mentioned in [Line 1042-43](https://github.com/HyphaApp/hypha/pull/4679/changes#diff-90e3dfb9fa211df3a499c1a05432971e818a96aa9081329ae07a9207ca313b0eR1042-R1043) that "instance status here would be the old phase and we have to pass new phase as an arg" like it is done in [Line 1070 ](https://github.com/HyphaApp/hypha/pull/4679/changes#diff-90e3dfb9fa211df3a499c1a05432971e818a96aa9081329ae07a9207ca313b0eR1070). You may need to look for edge cases if it is not causing any other issue then it is fine otherwise i would suggest to use similar approach as we used below in MESSAGES.TRANSITION at line 1070.

frjo:
@bmispelon Can you test and see if https://github.com/HyphaApp/hypha/pull/4693 solves this issue as well? I believe it should.

PR #4693, opened 2026-02-05
Move save to log_status_update so e-mail get sent.

### LINKED ISSUE #4692: Email not sent to reviewers when processing to External Review status
**Describe the bug**

When in an external review workflow, no email is sent when transitioning to external review.

**To Reproduce**

1. Create a request with external review submission
2. Assign a reviewer
3. Progress the submission to External Review
4. See that no email is sent to reviewers
5. Progress the submission to Close Review
6. See an email sent to reviewers

**Expected behavior**

An email is sent when transitioning to the External Review step
**Affected roles**

- Reviewers

**Additional context**

It is most like related to this change setting instance to the state before the transition happens, which causes the messaging pipeline to not fire off since it only finds reviewers: https://github.com/HyphaApp/hypha/commit/60ce996617076fd6ac3a25618f4cfe4901cb0750

See the line `@status_field.on_success()`.  That fires before the instance is saved, and so all the messaging framework is working with a submission before it's actually been updated.  Of note in the messaging framework, in email.py, in `def reviewers`, there's `source.phase.permissions.can_review(reviewer)` which will fail if the submission is not currently in a review phase.  Interestingly enough, this is why the email gets fired when the submission moves OUT of the review phase (either reverting, or moving forward)

This may be the cause of #4678.

frjo:
Fixes #4692

## Test Steps

 - [ ] Test that notification go out to reviewers. Slack for internal and e-mail for external.
 - [ ] Test that status updated go out via e-mail to applicant and look ok when a submission is moved forward.

### DIFF
--- a/hypha/apply/activity/adapters/activity_feed.py
+++ b/hypha/apply/activity/adapters/activity_feed.py
@@ -221,7 +221,7 @@ class ActivityAdapter(AdapterBase):
             )
         )
 
-    def handle_transition(self, new_phase, source, old_phase=None, **kwargs):
+    def handle_transition(self, old_phase, source, **kwargs):
         def wrap_in_color_class(text):
             color_class = PHASE_BG_COLORS.get(text, "")
             return f'<span class="rounded-full inline-block px-2 py-0.5 font-medium text-gray-800 {color_class}">{text}</span>'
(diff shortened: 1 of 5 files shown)


frankduncan:
LGTM!  Thanks for the quick turnaround :)

wes-otf (approved):
Nice! works great!

PR #4705, opened 2026-02-18
Allow staff to delete their own comments.

### LINKED ISSUE #4606: Add the ability for staff to delete comments
Currently staff can edit their own comments, but can't delete them. If they can edit it makes sense that they could also delete

frjo:
Fixes #4606

Skärmavbild 2026-02-18 kl  21 57 39

## Test Steps

 - [ ] Check if staff, and only staff, can delete their own comments.


hypha/apply/activity/views.py
```diff
@@ -78,6 +78,31 @@ def edit_comment(request, pk):
...
+        raise PermissionError("You can only delete your own comments")
+
+    if request.method == "DELETE":
+        activity = services.delete_comment(activity)
+
+        return render(
+            request,
+            "activity/partial_comment_message.html",
+            {"activity": activity, "success": True},
+        )
```
wes-otf, line 103: What are your thoughts on this:
```suggestion
        response = render(
            request,
            "activity/partial_comment_message.html",
            {"activity": activity, "success": True},
        )
        response.headers["HX-Trigger"] = json.dumps({"commentsUpdated": None})
        
        return response
```

Combined with a new listener in [comments.html](https://github.com/HyphaApp/hypha/blob/baa9a9d7c886dec2e92e8f4580f3947de8e5debe/hypha/apply/funds/templates/funds/comments.html#L106)

```
 hx-trigger="load, intersect once, commentsUpdated"
```

Just so the comment auto-updates and tosses the delete/edit buttons on delete

hypha/apply/activity/templates/activity/ui/activity-comment-item.html
```diff
@@ -64,20 +63,32 @@
...
                             </a>
                         {% endif %}
+
+                        {% if editable and activity.user == request.user  and not activity.deleted and request.user.is_apply_staff %}
+                            <a
+                                hx-delete="{% url 'activity:delete-comment' activity.id %}"
+                                hx-target="#text-comment-{{activity.id}}"
+                                hx-confirm="{% trans "Are you sure you want to delete this comment? This action cannot be undone." %}"
+                                title="{% trans "Delete comment" %}"
+                                class="btn btn-error btn-sm btn-soft btn-square"
```
wes-otf:
```suggestion
                                class="btn btn-error btn-sm btn-ghost btn-square"
```
super unimportant but just matches the edit button a little better

wes-otf (commented):
looks & works great! just a few super small things

hypha/apply/activity/views.py
frjo: Good point! I have now set target to comment item instead of comment text to get this result. This has the same result but we do not need to reload all activities.

hypha/apply/activity/templates/activity/ui/activity-comment-item.html
frjo: Made the change, the trash icon then becomes black but on hover you get a red background.

hypha/apply/activity/migrations/0082_change_staff_pii_visibility.py
```diff
@@ -27,6 +27,7 @@ def change_updatelead_visibility(apps, schema_editor):
         ActivityAdapter.messages[MESSAGES.UNARCHIVE_SUBMISSION],
     ]
 
+    Activity = apps.get_model("activity", "Activity")
```
frjo, line 30: Never use the model directly in a migration. If you later on add fields etc. the migrations will fail. Instead use `get_model()` like above.

wes-otf (approved):
Looks great!

PR #4732, opened 2026-03-02
Remove summary part of result view to improve performance.

### LINKED ISSUE #4731: Improving performance of the results view
This view becomes slow when there are large amounts of historic submissions.

frjo:
Fixes #4731

Skärmavbild 2026-03-02 kl  14 50 29

### DIFF
--- a/hypha/apply/funds/templates/funds/submissions_result.html
+++ b/hypha/apply/funds/templates/funds/submissions_result.html
@@ -13,97 +13,7 @@
 
 {% block content %}
     <div class="flex flex-col gap-8 my-8 mx-auto">
-        <section>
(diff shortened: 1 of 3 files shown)


frjo:
@wes-otf This improves performance quite a bit. My guess is that it is the filtered table that has the interesting numbers for staff in any case.

If the summary is interesting as well we can have it in a separate view/tab perhaps?

wes-otf:
Yeah I think this makes sense! As we get the user interviews in for this I think we'll find more we could populate the space with but this works for now

Will also likely make things easier for me integrating the anonymize applications into this view

wes-otf (approved):

PR #4761, opened 2026-03-17
Add missing __str__ to flag and task models.

frjo:


wes-otf:
@frjo where do these get displayed? if the end user could see it should we add i18n?

if it's just django admin or debug stuff I'm not too worried about it otherwise

frjo:
Only in error messages as it is now.

PR #4767, opened 2026-03-19
Make the Requested amount field accepts locale formated values

### LINKED ISSUE #4628: Not possible to input requested amount values with thousand separator
The "Requested amount" field only accepts values written as int or as decimal value with a period as separator.

1. 9999
2. 9999.99

All other format are invalid.

This leads to a bad user experience for people used to write values with a thousand separator:

1. 9.999
2. 9.999,99
3. 9 999
4. 9 999,99

Hypha will do the right thing when displaying these values. With a German locale the value "9999.99" will display correctly as "9.999,99".

But if a German user enter "9.999" it will display as "10" (rounded up) while expecting to get "9.999.00".

Hypha should somehow allow users to enter values as they are used to and the clean them before saving.

For normal fields Django has this build it it seems, see https://docs.djangoproject.com/en/4.2/topics/i18n/formatting/#locale-aware-input-in-forms

But this is a stream field so we are likely needing to build something custom.

frjo:
Fixes #4628

Add custom `to_python()` function to allow localised number inputs to Requested amount field. Add custom `prepare_value()` to make the widget use the locale correct decimal separator.

## Test Steps

 - [ ] Try saving all number format you know about. 10000 can in different countries be written as "10 000", "10000.00", "10 000,00", "10.000,00", "10,000.00". They should all be saved as "10000" in the database.
 - [ ] Try to set the site to different locales and combine with different number format. All combinations should work.

### DIFF
--- a/hypha/apply/funds/blocks.py
+++ b/hypha/apply/funds/blocks.py
@@ -32,7 +32,8 @@ class LocalizedFloatField(forms.FloatField):
 
     def to_python(self, value):
         if value not in self.empty_values:
-            value = str(value).strip()
+            # Remove all spaces.
+            value = str(value).strip().replace(" ", "")
             has_dot = "." in value
             has_comma = "," in value
 
--- a/hypha/apply/funds/blocks.py
+++ b/hypha/apply/funds/blocks.py
@@ -1,6 +1,7 @@
 import json
 
 from django import forms
+from django.utils.formats import get_format
 from django.utils.translation import gettext_lazy as _
 from wagtail import blocks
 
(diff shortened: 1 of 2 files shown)


frjo:
I used up a lot of Claude code tokens for this 😄. So many nitty gritty details but after hours of testing I believe (hope) it's working.

Really needs testing by other people, important that this does not mess up users data.

If we can be sure it works I think it will be a nice ux improvement, allowing users to enter values the way they are used to and Hypha takes care of the rest.

frjo:
@bmispelon I heard you created a solution for this already. Something similar to above or did you find a cleaner fix for it?

wes-otf:
pretty cool how intelligent this field is now! initial testing is looking pretty good, going to keep probing a little deeper later this afternoon for edge cases

wes-otf:
In my testing numbers with spaces for separators don't validate - it was in the test steps but not the unit tests so not sure if it was intended to work or not
Screenshot 2026-03-25 at 12 29 10

wes-otf:
I think it's ready other than that! would be cool to get input from folks that actually will be utilizing this though

frjo:
Will see about allowing spaces on input.

frjo:
A small change makes it possible to add all the spaces you like on input, they are removed when user submit form.

frjo:
I have been testing this and I believe it is solid. For the case where it creates errors they should be obvious, very small or very large numbers.

wes-otf:
Agreed! Let's merge it!

PR #4772, opened 2026-03-24
Add support for passkeys to Hypha

### LINKED ISSUE #4563: Reimplement 2FA to create a smother user flow and add support for passkeys
Goals are:

- [ ] Make 2FA as smooth as possible for users. Especially important for organisations that enforce 2FA.
- [ ] Replace wizard multistep with, if possible, single page solution.
- [ ] Implement passkeys as a stand alone loggin method. The UI should encourage people to switch to it. (With passkeys no 2FA is needed even if enforced.)
- [ ] Allow for multiple passkeys.
- [ ] Allow users to remove their password. 
- [ ] New 2FA should if possible not force people to reset their current 2FA codes.

frjo:
Fixes #4563

This implementation uses [duo-labs/py_webauthn: Pythonic WebAuthn](https://github.com/duo-labs/py_webauthn) directly implementing its own Django wrapper. This is so passkeys are used as a stand alone login method and not as a 2FA option.

The interesting parts are in `passkey_views.py` and `passkeys.js`.

## Test Steps

 - [ ] Test that setting up and logging in with passkeys works on Mac, Windows, iPhone, Android and Linux. 
 - [ ] Using built in OS support, using usb keys etc.
 - [ ] Test that ENFORCE_TWO_FACTOR are bypassed for passkey users. Passkeys are more secure than 2FA.
 - [ ] Audit the implementation for any issues.


wes-otf:
This is exciting! looking forward to reviewing it in the next day or two!

wes-otf:
pushing this to test now - is it possible to test locally? getting domain errors trying it on my setup

frjo:
Test build will likely fail due to migration conflict. Fixed it just now in this branch. More migration conflicts will come, the translation PR have a lot migrations e.g.

If we push this to test and then do not merge this in before translate PR we will have to reset the test db a bit.

frjo:
> is it possible to test locally?

Yes, if you set base_url to "localhost". Then it will not require TLS either.

IP 127.0.0.1, hypha.test etc. will not work due to how browsers handle this for security.

This should work:

`python manage.py runserver_plus localhost:9001`

hypha/apply/users/templates/users/partials/list.html
```diff
@@ -0,0 +1,75 @@
...
+                            type="text"
+                            name="name"
+                            value="{{ passkey.name }}"
+                            class="flex-1 input input-xs input-ghost"
+                            aria-label="{% trans 'Passkey name' %}"
+                            data-tippy-content="{% trans 'Click to edit name' %}"
+                            @input="dirty = true"
+                            required
+                        />
+                        <button type="submit" class="btn btn-ghost btn-xs" :class="dirty && 'text-success'" title="{% trans 'Save name' %}">
```
wes-otf:
```suggestion
                        x-data="{ pkcurrent: '{{ passkey.name }}', pkinput: '{{ passkey.name }}' }"
                        hx-post="{% url 'users:passkey_rename' passkey.pk %}"
                        hx-target="#passkey-list"
                        hx-swap="outerHTML"
                        class="flex flex-1 gap-2 items-center"
                    >
                        {% csrf_token %}
                        <input
                            type="text"
                            name="name"
                            x-model="pkinput"
                            class="flex-1 input input-xs input-ghost"
                            aria-label="{% trans 'Passkey name' %}"
                            data-tippy-content="{% trans 'Click to edit name' %}"
                            required
                        />
                        <button type="submit" class="btn btn-ghost btn-xs text-success" :class="pkcurrent == pkinput && 'btn-disabled invisible'"  title="{% trans 'Save name' %}">
```

Such a nitpick but might make a little more sense to only allow users to submit when the input changes - as it stands the button is still enabled & will make the request even if the key name hasn't changed

wes-otf:
left a super minor comment but otherwise this is totally ready to go!

wes-otf (changes requested):

wes-otf (approved):

frjo:
@wes-otf Good change. Looking at it I came up with some more ways to polish it. The green check is there but when you start typing in the name field the check turns blu with a border so it looks like a button. Clicking it saves and makes it green without a border again.

frjo:
Added some new stuff:

* The db field is varchar 255 but the length of the name was never checked before saving. Cap it at 128 now.
* When trying to add another key for the same device, catch the error so it looks nicer for the user.
* Make all user facing strings translatable. Keep missing this for new code.
* Get CSRFToken from hxHeaders instead of cookie parsing.
* Use transaction.atomic for passkeys loggin in the small chance someone tries to use the same key from two devices.
* Logging errors so admins can track issues.

wes-otf:
@frjo I think I accidentally force pushed earlier when I intended to force push to the test branch - sorry about that!

> The green check is there but when you start typing in the name field the check turns blu with a border so it looks like a button

it looks like the original issue is still there where you can still make the request (by clicking the button) even if the name hasn't been changed. I think it makes things a little less intuitive and could result in an unnecessary request but probably isn't big in the scheme of things.

Everything else looks good to me! Happy to get this merged whenever you feel it's good.

frjo:
@wes-otf Easy to restore the branch so no worry. I think I managed to make the green check not behave like a button now. When you change the name the check turns blue and behave like a button.

hypha/apply/users/templates/users/includes/passkey_login_button.html
```diff
@@ -0,0 +1,20 @@
...
+    class="btn btn-secondary btn-outline"
+    data-passkey-ui
+    data-begin-url="{% url 'users:passkey_auth_begin' %}"
+    data-complete-url="{% url 'users:passkey_auth_complete' %}"
+    data-next-url="{{ redirect_url|default:'' }}"
+    hidden
+    onclick="hypha.passkeys.authenticate()"
+>
+    {% heroicon_micro "key" class="inline size-4" aria_hidden=true %}
+    {% trans "Sign in with passkey" %}
```
wes-otf: Wondering if we should make this a little more similar to the "Login with password" button in both verbiage and icon.

Like choosing to user either "Sign in" vs "Log in" and the password button uses the heroicon_mini: `{% heroicon_mini "key" size=18 class="opacity-80" aria_hidden=true %}`

hypha/apply/users/templates/users/login.html
```diff
@@ -73,9 +73,10 @@ <h1 class="mb-4 text-h1">{% blocktrans %}Log in to {{ ORG_SHORT_NAME }}{% endblo
 
                     <section class="flex-wrap card-actions">
```
wes-otf, line 92: should we make this flex col too? the button wrap kinda weird

wes-otf:
extremely small UI notes I didn't notice in times prior to wrap this up but once those are handled this should be completely set! works and looks great!

hypha/apply/users/templates/users/login.html
frjo: I moved the goggle button last and made the whole page wider so on desktop all three buttons is on one line. I think that look ok.

Locally I do not have GOOGLE_OAUTH2 active so only see two buttons and missed this issue.

hypha/apply/users/templates/users/includes/passkey_login_button.html
wes-otf: What do you think about these points @frjo?

wes-otf:
button layout is nice!

hypha/apply/users/templates/users/includes/passkey_login_button.html
frjo: Mush better, fixed it now.

wes-otf:
does this need retesting @frjo?

frjo:
@wes-otf Do not think so, just a rebase.

wes-otf:
works nice on test! I think it's ready to go

frjo:
The last thing I put it was that passkeys is disabled in production until WEBAUTHN_RP_ID is set. I think that is the right approach. Then organisations can turn it on when they are ready.

PR #4775, opened 2026-03-25
Prevent linkbots from consuming tokens in one time links

### LINKED ISSUE #4535: Workaround for Outlook previewing login links and expire them prematurely
MS Outlook mail system does a preview of links in all e-mails. This works badly with our login links since they can only be used once for security reasons.

When outlook access the link Hypha expire it and when the user clicks the link it will no longer work.

Having Cloudflare (or another filtering proxy) in front of Hypha can block these attempts. Likely this is the reason OTF has not noticed this issue.

`52.167.144.233 - - [07/May/2025:08:57:03 +0000] "GET /this/is/the/link/path HTTP/2.0" 404 527 "-" "Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; MicrosoftPreview/2.0; +https://aka.ms/MicrosoftPreview) Chrome/100.0.4896.127 Safari/537.36"`

I assume other does this previewing as well, Gmail? Yahoo?.

frjo:
Fixes #4535

OBS! `PASSWORD_RESET_TIMEOUT` have a new sane default of 15 minutes. A new timeout setting `PASSWORD_ACTIVATION_TIMEOUT` was added that also default to 15 minutes. 

`PASSWORD_RESET_TIMEOUT` used to default to 3 days to allow new users some time to activate their account. Organisation that still want the old default for activation mails can set `PASSWORD_ACTIVATION_TIMEOUT = 259200`.

This is mainly a fix for MS Outlook mail system habit of doing a preview of links in all e-mails. This preview expires the one time links so when a user tries to login or reset their password it does not work.

The solution is to show an extra confirmation screen with a login button, so users need to click one extra time. This solves the MicrosoftPreview issue and should work for any similar issues as well.

During the work I found a number of inconsistencies in various login related templates that I also attempted to fix. That is the reson so many files are changes in this PR. 

## Test Steps

 - [ ] Use the signup and passwordless login and confirm that you by clicking the link in the email come to a confirmation screen. Clicking the button there log you in to the site.


hypha/apply/users/templates/users/passwordless_login_signup.html
wes-otf, line 1: super nitpick but should we also add `{% block body_class %}bg-base-200{% endblock %}` after the title block? I like how that makes the card pop on the confirm template:
https://github.com/HyphaApp/hypha/blob/7da74b772363c4661b48f14049d8da6c5219b175/hypha/apply/users/templates/users/activation/confirm.html#L5

hypha/apply/users/templates/users/password_reset/done.html
```diff
@@ -1,21 +1,28 @@
...
+{% load i18n heroicons %}
 
 {% block title %}{% trans "Password Reset" %}{% endblock %}
 {% block body_class %}bg-base-100{% endblock %}
 
 {% block content %}
+    <div class="flex flex-col justify-center items-center min-h-[60vh]">
+        <section class="w-full max-w-2xl card shadow-xs bg-base-100 md:card-lg">
+            <div class="items-center card-body">
+                <span class="flex justify-center items-center mb-4 rounded-full bg-primary/10">
```
wes-otf, line 11: this is so slick! I love the shadowy circle behind the icon

wes-otf (approved):
Overall this looks and works fantastic! This is the solution we needed - you were also so fast about this!

hypha/apply/users/templates/users/passwordless_login_signup.html
frjo: Will look at that, we use the same style in many places so I make sure they all look the same.

frjo:
Been thinking about a solution for weeks and worked on it since last week, so not so fast 😃.

hypha/apply/users/templates/users/passwordless_login_signup.html
frjo: The background colour was missing from three templates, fixed now. Looks a lot nicer with it in place!

wes-otf:
great stuff, thanks @frjo !

PR #4784, opened 2026-03-27
Fix the rounds dropdown menu search on the all view.

frjo:
* Add unpublished and not started filter to round admin list in wagtail.
* Add unpublished  filter to rounds view.
* Fix headers in filter dropdown on rounds view.
* Fix unstyled pagination on rounds view.

Instead of tabs make open/closed look like this to make the filter in the dropdown work:

Skärmavbild 2026-03-27 kl  17 53 51

### DIFF
--- a/.github/workflows/hypha-ci.yml
+++ b/.github/workflows/hypha-ci.yml
@@ -26,15 +26,6 @@ jobs:
       - name: Run pre-commit
         uses: pre-commit/action@v3.0.1
 
-  login:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Login to Docker Hub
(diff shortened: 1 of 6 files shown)


frjo:
@wes-otf Does this look ok, to skip the tabs in rounds dropdown?

Found some bugs on the rounds view and added more filter options as well.

wes-otf:
@frjo in testing I don't see unpublished rounds - are they supposed to show in this? I think this format is nice!

frjo:
I added the "unpublished" filter here, and in wagtail admin rounds view.

Skärmavbild 2026-03-31 kl  19 49 10

Issue #4752, opened 2026-03-13
Javascript files include non-translated user-facing strings


bmispelon:
**Describe the bug**

Javascript files include non-translated user-facing strings.

Here are all the instances I found (by using `git ls-files '*.js' ':!*.min.js'` to find all non-minified js files then auditing them manually):

* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/application-form.js#L58
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/batch-actions.js#L67
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/batch-actions.js#L87
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/file-uploads.js#L42
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/report-calculator.js#L124-L127
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/report-calculator.js#L154-L161
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/submission-form-copy.js#L46
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/submission-form-copy.js#L58
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/submission-form-copy.js#L87
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/submission-form-copy.js#L91
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/tinymce-word-count.js#L26
* https://github.com/HyphaApp/hypha/blob/df928ec93cf261e428b6df9dca2efb2a18847677/hypha/static_src/javascript/tinymce-word-count.js#L31
[...]

PR #4789, opened 2026-03-31
Add missing translated strings

### LINKED ISSUE #4753: Model and form fields should have a translated verbose_name/label

Most model fields do not define a `verbose_name` (though a few do), which means they appear untranslated if they somehow end up displayed on a page (for example via a `ModelForm`). The same goes for form fields and their `label`.

I would suggest adding an explicit `verbose_name` for all model fields, and a `label` for all form fields, and if possible a linter rule that prevents new untranslated fields from being added to the codebase.

**Priority**

- Medium (slowing you down)

**Affected roles**

Arguably all roles are affected, but this is more of an issue for applicants who are the ones most likely to see those untranslated strings, and who might not know English.

frjo:
(This contribution was sponsored by [DigitalHub.sh](https://digitalhub.sh/))

Fixes #4753

I've split this PR over several commits which are mostly independent of each other. I'm happy to open separate PRs if you think that'd be easier to review.

### Mark missing strings for translation

After a few unsuccessful attempts at scripting the identification of untranslated strings, I ended up reviewing all Python files manually.

The command I used to identify the relevant files was:
```
git ls-files -z '**.py' ':!**/migrations/*.py' ':!**/urls.py' ':!hypha/settings/*.py' ':!docs/*' ':!**/test*.py' | find -files0-from - -not -empty
```
(That's all Python files, but excluding stuff like migrations, settings, ... (and empty files))

This change should have no impact for a site running the English version of Hypha.

### Add missing strings for translations (templates)

Similar to the previous commit, I ended up reviewing all templates manually (`git ls-files '*.html'`), adding `{% trans %}`, `{% blocktrans %}`, or `_(...)` where needed.

Again, this change should have no impact.

### Add trimmed option to blocktranslate

This removes irrelevant whitespace in source strings, making it a bit nicer.

I used `git grep blocktrans | grep -vw trimmed | grep -v endblocktrans` to identify uses of `{% blocktrans %}` without the `trimmed` argument, excluding lines that also included `{% endblocktrans %}` on the same line.

This is technically a breaking change, but it only introduces whitespace changes to HTML templates, which is irrelevant.

### Use `{% blocktranslate trimmed %}` to break down long lines

This one just makes templates a little bit more readable as it breaks very long lines into more manageable pieces.

Again, it can introduce whitespace changes to the HTML, but should not have any effect to the source strings.

### Added `verbose_name` and `verbose_name_plural` to all models

Some models already has a translated `verbose_name` or `verbose_name_plural`, but it was inconsistent (and quite rare). This change adds these attributes to all models.

I used `git ls-files '**/models/*.py'` and ` git ls-files '**/models.py'` to find all models, then went through the files manually to add the missing `verbose_name` and `verbose_name_plural`.

I decided to omit the plural form for model classes inheriting from `BaseSiteSettings` since it didn't seem to make sense.

I also opted to use lower case (like `_("foo bar")` for model `FooBar`) to match Django's own default behavior, but I used `Foo Bar` in a few instances where the model already had a `verbose_name` that used titlecase.

## Next steps

I ran out of time to complete all my objectives and I've still got some translation-related tasks I wanted to get to. I'm happy to open issues for these if you think they're relevant:

* #4752 Fix missing translations in javascript files: I found about 10 instances but fixing them would require either using the [Django javascript translation functions,](https://docs.djangoproject.com/en/6.0/topics/i18n/translation/#internationalization-in-javascript-code) or moving the strings to the HTML (data attributes for example).
* #4753 Add explicit `verbose_name` to all declared model fields (and maybe form fields as well). This can probably be scripted somewhat, ideally with a linter that would catch the introducing of future untranslated strings.
* Review cases where `gettext` is used at the module level instead of `gettext_lazy` as this could indicate some possible bugs (visible in a multi-language setup where the user is not using the default language).
* Fix translations that build up a sentence word by word, as this doesn't really work in practice. Something like `{% trans "Application" %} {{ application_id }} {% trans "updated on" %} {{ date }} {% trans "by" %} {{ author.name }}` should be rewritten to have the whole sentence in a single `blocktrans`, or when not possible small words like `by` should be given a context because they're likely to be translated differently for different sentences.


frjo:
I created a new PR on a Hypha repo branch so I could rebase it. Thanks @bmispelon for all the work on this one!

frjo:
@wes-otf Please have a look through the changes. The PR touches many files and creates many migrations (verbose_name on fields) but should not change anything in how Hypha works. There is one minor bugfix I noticed, correcting to `PermissionDenied`.

When you also looked at this and do not find anything that stands out you can merge it in.

This makes many more strings translatable in Hypha! Thanks to @bmispelon and DigitalHub.sh.

After this is in I will update all the po files so translations can be updated/created.

With the resent rapid improvements in LLM:s I think they could do a decent job of creating ruff translation of Hypha. I think it will be easier for a translator to correct an LLM translation that start from scratch and translate nearly 5000 strings.

wes-otf:
@frjo I did a rough check of it when it first dropped and things seemed pretty good but I'll heavily test it again today and make sure we're in the clear - definitely a fantastic improvement though!

frjo:
@wes-otf Do you think this is good to go in?

wes-otf:
@frjo from what I've seen I think it's ready yeah

PR #4794, opened 2026-04-01
When a submission uploaded file is missing, give nice error and not a 500

### LINKED ISSUE #4408: `An error occurred (404) when calling the HeadObject operation: Not Found` when loading an application that has specific files
Staff just let me know that this has been raised a few times on live, it seems that specific applications that have file uploads will raise the mentioned `An error occurred (404) when calling the HeadObject operation: Not Found`, which is from an underlying `'PrivateStreamFieldFile' object is not subscriptable`. I think it may have something to do with AWS?

Two examples of this on sentry are:
- https://otf.sentry.io/issues/5487435437/events/aede66a7826b49158b03742f7e7f4751/
- https://otf.sentry.io/issues/5487435437/events/c585a4838acc4b12a338c934016bc090/

As it stands this prevents staff from opening applications that have this issue

frjo:
Fixes #4408

I mostly notice this in development but I believe it will also fix the issue above.

## Test Steps

 - [ ] Ensure that uploading and edit files in submissions works as before.

### DIFF
--- a/hypha/apply/funds/views/__init__.py
+++ b/hypha/apply/funds/views/__init__.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import UserPassesTestMixin
+from django.http import Http404
 from django.shortcuts import get_object_or_404, render
 from django.utils.decorators import method_decorator
 from django_filters.views import FilterView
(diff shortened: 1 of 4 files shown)


wes-otf (approved):
looks & works great! deleted some stuff in S3 and verified all appeared as expected

PR #4795, opened 2026-04-01
Lower memory usage in submissions and projects

### LINKED ISSUE #4520: Minimise memory usage of Hypha
According to Scout it is the "SubmissionDetailView" that consumes most memory.

Especially the template "stream_forms/includes/file_field.html". Does this template somehow load the files themself?

https://github.com/HyphaApp/hypha/blob/b34608c179dbb183bb9a913477b943d72bfa4afb/hypha/apply/stream_forms/templates/stream_forms/includes/file_field.html#L1-L20

frjo:
Fixes #4520

This are not dramatic changes but many small things adds up. In testing the savings are small but I think that on large live environments they can be noticeable.

* Defer loading to have fields like search_data, search_document, form_data and form_fields only load when needed.
* Use submission_id instead of submission.id to avoid a db query when the model is not loaded/needed.
* Implement select_related() to avoid N+1 queries.
* Cache get_object() so it is not loaded multiple times in one request.

### DIFF
--- a/hypha/apply/activity/models.py
+++ b/hypha/apply/activity/models.py
@@ -236,7 +236,7 @@ class Activity(models.Model):
     def get_absolute_url(self):
         # coverup for both submission and project as source.
         submission_id = (
-            self.source.submission.id
+            self.source.submission_id
             if hasattr(self.source, "submission")
             else self.source.id
         )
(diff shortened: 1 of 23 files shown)


frjo:
@wes-otf I have put this on test now so we can make sure it doesn't break anything.

PR #4802, opened 2026-04-07
Update the hook system to use Python 3 and Django 5 patterns

frjo:
## Test Steps

 - [ ] Confirm that existing plugins work as expected.


frjo:
@frankduncan Updated the hook system a bit, can you confirm that it still works as before with your existing plugins?

PR #4790, opened 2026-03-31
Remove partner role

wes-otf:
Closes #4719. Migrates all existing partners to coapplicants with comment privileges and removes the partner role from Hypha. This also handles swapping partner comment visibility to applicant

## Test Steps

- [ ] All partner views are removed (update partner, add partner, etc)
- [ ] Comment visibility options show up as expected

**After migrations are ran confirm...**
 - [ ] all active partners are coapplicants w/ comment privileges
 - [ ] the partner role is removed from Hypha
 - [ ] activity visibility that previously specified partners (ie. `partners` or `applicant partners`) are now `applicant`

PR #4804, opened 2026-04-07
Add missing migration for Partners & AssignedReviewers

wes-otf:
Related to #4790. I missed an edge case where an `AssignedReviewer` was also a Partner, in which the migrations would fail as the `type` attribute is protected in the model. This moves all types of partner to reviewer.

## Test Steps

 - [ ] Ensure `AssignedReviewers` with type of partner get migrated to reviewer and that all migrations pass successfully.


frjo:
Running this:

```
git switch main
uv run python manage.py migrate
git switch fix/add-assignedreviewers-migration
uv run python manage.py migrate
```

I get this error:

```
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration users.0028_remove_partner_group is applied before its dependency funds.0134_change_partner_assigned_reviewers_role on database 'default'.
```

frjo:
I think we can make changes to `0028_remove_partner_group.py‎` directly.

The few system that have applied it likely do not have this edgecase so do not need fixes.

frjo (changes requested):
See comments above.

wes-otf:
Yeah I had to manually remove 0028_remove_partner_group from the database to get this in but this makes a little more sense, I think it should be good to go!

frjo (approved):

frjo:
@wes-otf The new function was never run so I did that minor fix. Also fixed so apps.get_model() always was used.

Works well in my testing now.

I think we can do a new release when this is merged in.

PR #4818, opened 2026-04-14
Retire jQuery

frjo:
We only have a few scripts using jQuery so time to retire it and go full vanilla js. 

This PR rebuilds the js part of the addressfield, merging all js in to a single file using modern vanilla js.

Also make strings in submission-form-copy.js translateble.

## Test Steps

 - [ ] Make sure the addressfield works just as before.
 - [ ] Test submission copy functionality, it had some bugs that are fixed in the PR.
 - [ ] Test that form groups works just as before.
 - [ ] Test batch action status change where a determination is created.
 - [ ] Test invoice batch actions.


hypha/static_src/javascript/submission-form-copy.js
```diff
@@ -1,116 +1,134 @@
...
+  }
+
+  const applicationForm = document.querySelector(".application-form");
+  if (!applicationForm) return;
+
+  const i18n = {
+    btnLabel: applicationForm.dataset.copyBtnLabel,
+    btnTitle: applicationForm.dataset.copyBtnTitle,
+    wordLimit: applicationForm.dataset.copyWordLimit,
+    selected: applicationForm.dataset.copySelected,
```
wes-otf, line 107: I don't think these are being properly set - I don't have any data attributes on my `.application-form` element and the copy button shows as an empty undefined button

Image

wes-otf (changes requested):
Small comment re: copy questions to clipboard

I was also only given an issue not populating specific required fields (mainly rich text) until I tried to submit after the preview. Fields like title & amount didn't let me make it to the preview without population which is what I expect of all required fields. Didn't dig too deep into what the issue was with that though.

All else seems to work pretty well though! bye bye jquery!!

hypha/static_src/javascript/submission-form-copy.js
frjo: They are only set on `applicationsubmission_form.html` and not on `application_base.html`. Will fix that.

frjo: Done.

frjo: Test is being updated in a few minuts.

frjo:
> I was also only given an issue not populating specific required fields (mainly rich text) until I tried to submit after the preview. Fields like title & amount didn't let me make it to the preview without population which is what I expect of all required fields.

Not sure I get this issue. Are the required field not working as they should in some way?

If you set `SUBMISSION_PREVIEW_REQUIRED=False` so users see a preview button they need to fill in all required fields before previewing. This does not change with this PR. Thinking about it we should perhaps add "formnovalidate" to the preview button as we do to the draft button.

wes-otf:
> Not sure I get this issue. Are the required field not working as they should in some way?

Some of the inconsistencies between fields that are validated client side vs server side (ie. project title vs rich text respectively) were tripping me up & manifesting in a strange way, but looks to be unrelated to this. All previous issues look to be fixed, thanks so much!

wes-otf (approved):

PR #4824, opened 2026-04-21
Notify users on login via e-mail

### LINKED ISSUE #3549: Make Hypha send an e-mail to the user on each successful login
Many services does this nowadays. When you login you get an e-mail stating that a "new login was made with [browser] on [plattform]".

The e-mail then has some info on IP the login was made from etc. and a text that says something like:

"If this was you, great! There's nothing else you need to do.

If you do not recognize this login, …"

Add a global setting so this can be turned of but enable it by default.

frjo:
Fixes #3549

## Test Steps

 - [ ] Check that user receives en e-mail on login.
 - [ ] Check that the e-mail content is correct and informative.

### DIFF
--- a/hypha/apply/users/signals.py
+++ b/hypha/apply/users/signals.py
@@ -1,0 +1,27 @@
+from django.conf import settings
+from django.contrib.auth.signals import user_logged_in
+from django.dispatch import receiver
+from django.utils import formats, timezone
+from django.utils.translation import gettext_lazy as _
+from wagtail.models import Site
+
+from hypha.core.mail import MarkdownMail
+
+
+@receiver(user_logged_in)
+def send_login_notification(sender, request, user, **kwargs):
+    if not settings.SEND_MESSAGES or not user.email:
+        return
+
+    email = MarkdownMail("users/emails/login_notification.md")
+    email.send(
+        to=user.email,
+        subject=_("Successful login to %(org)s") % {"org": settings.ORG_LONG_NAME},
+        from_email=settings.DEFAULT_FROM_EMAIL,
+        context={
+            "user": user,
+            "login_time": formats.date_format(timezone.now(), "SHORT_DATETIME_FORMAT"),
+            "site": Site.find_for_request(request) if request else None,
+            "ORG_EMAIL": settings.ORG_EMAIL,
+        },
+    )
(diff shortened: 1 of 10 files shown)


wes-otf:
This also works well & looks nice - is there a way to skip this when hijack/"switch user" is utilized? not a huge deal but would be nice to not send a notification for that

frjo:
> This also works well & looks nice - is there a way to skip this when hijack/"switch user" is utilized?

This I missed! No e-mail should be send in these cases I think, will take a look at that.

frjo:
> is there a way to skip this when hijack/"switch user" is utilized?

Fixed it now I believe. Should be on test site shortly.

hypha/apply/users/signals.py
```diff
@@ -0,0 +1,42 @@
...
+        subject = str(settings.EMAIL_SUBJECT_PREFIX) + str(subject)
+
+    email = MarkdownMail("users/emails/login_notification.md")
+    email.send(
+        to=user.email,
+        subject=subject,
+        from_email=settings.DEFAULT_FROM_EMAIL,
+        context={
+            "user": user,
+            "login_time": formats.date_format(timezone.now(), "SHORT_DATETIME_FORMAT"),
```
wes-otf: on the test instance I have `USE_TZ=true` & `TIME_ZONE='America/New_York'` but for some reason the time is showing as 6pm when EST time is 2pm. any ideas on if I configed that wrong?

 also thinking it could be beneficial to include the timezone abbreviation too for users of different timezones

wes-otf:
Switch user works as expected with no notification - nice work! I think this is ready to go, just had that small comment around timezones

hypha/apply/users/signals.py
frjo: Sounds like it is using server time and not user time, will take a look at that. Confusing for users and misleading so good catch.

frjo: @wes-otf I think I have fixed it now. Hypha in this PR now saved the user timezone to the session and uses that to format the datetime in the login notification e-mail. Also added the actual timezone.

This got me looking at how we display times in Hypha in general. Thanks to `<relative-time>` component we show local time in many places in Hypha. Will work on making that more complete in another PR.

frjo: @wes-otf Deploying the updated PR to test now.

PR #4837, opened 2026-05-07
Memory use improvements

frjo:
Lower memory usage for dashboards, submission all and csv export. All functionality should be unchanged.


wes-otf (approved):
Looks really slick! I'll get it deployed to test and play around but should be ready

PR #4843, opened 2026-05-13
Secure cookies on production, check next redirect, permission checks for partials

frjo:


wes-otf:
@frjo was there a reason secure cookies weren't the default before? I never noticed that

frjo:
@wes-otf We did not want it for local or test but the solution is not optimal since you need to set a setting to get secure cookies, it should be the default for production.

wes-otf (approved):

PR #4846, opened 2026-05-15
Set a min-release-age to 3 days for both npm and uv.

frjo:


wes-otf (approved):

wes-otf:
a good improvement with all the supply chain attacks as of late

PR #4856, opened 2026-05-21
Replace author `Edit` button with new window link icon

wes-otf:
Addressing the conversation [here](https://github.com/HyphaApp/hypha/pull/4380#issuecomment-4491706335). Making the link to the user's wagtail admin page a little less confusing now that the "Change Author" feature exists. Also made this open in a new tab. Feels a bit more intuitive 

*before*
Screenshot 2026-05-21 at 09 52 43
*after*
Screenshot 2026-05-21 at 09 51 55

### DIFF
--- a/hypha/apply/funds/templates/funds/includes/rendered_answers.html
+++ b/hypha/apply/funds/templates/funds/includes/rendered_answers.html
@@ -51,8 +51,8 @@
                 <dd class="flex gap-1 items-center">
                     {{ object.get_full_name_display }}
                     {% if request.user.is_apply_staff %}
-                        <a class="link" href="{% url 'wagtailusers_users:edit' object.user_id %}" target="_blank">
-                            {% heroicon_mini "arrow-top-right-on-square" class="inline align-text-bottom size-4" %}
+                        <a class="link" title="{% trans 'Edit user account' %}" href="{% url 'wagtailusers_users:edit' object.user_id %}" target="_blank">
+                            {% heroicon_mini "arrow-top-right-on-square" class="inline align-text-bottom size-4" aria_hidden=true %}
                         </a>
                     {% endif %}
                 </dd>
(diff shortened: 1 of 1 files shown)


hypha/apply/funds/templates/funds/includes/rendered_answers.html
```diff
@@ -48,11 +48,11 @@ <h3 class="pb-2 mb-2 font-medium border-b text-h3">
...
                 <dt class="font-semibold">{% trans "Legal Name" %}</dt>
-                <dd class="flex gap-2 items-center">
+                <dd class="flex gap-1 items-center">
                     {{ object.get_full_name_display }}
                     {% if request.user.is_apply_staff %}
-                        <a class="link" href="{% url 'wagtailusers_users:edit' object.user_id %}">
-                            {% trans "Edit" %}
+                        <a class="link" href="{% url 'wagtailusers_users:edit' object.user_id %}" target="_blank">
+                            {% heroicon_mini "arrow-top-right-on-square" class="inline align-text-bottom size-4" %}
                         </a>
```
frjo, line 56: That icon usually is used for external links. Can we not just make it underline?

Not to fond of `target="_blank"` but we use it in 25 other places already.

wes-otf: Was using some of the precedent [here](https://github.com/HyphaApp/hypha/blob/515dce8af67b67aa5bf682ddbbffe42bee2c34e1/hypha/apply/determinations/templates/determinations/batch_determination_form.html#L36) to convey to the staff member that this is going to remove them from their current workflow, as I can see a lot of them clicking it accidentally thinking it'll allow for filtering by applicant or something similar. Can make the name a link if you feel that's more intuitive though

frjo: Ok, let leave it as is. But we do need to add a "title" attribute to the link to make it accessible. Something like "Edit user account".

wes-otf: should be set now!

frjo (approved):

PR #4858, opened 2026-05-26
Added upload (and debug) to whitelisted URLs

### LINKED ISSUE #4857: Not having 2FA enabled causes file uploads to fail
When both `ENFORCE_TWO_FACTOR=True` and `FORCE_LOGIN_FOR_APPLICATION=True`, and a user does not have 2FA enabled the upload form field will fail when an upload is attempted. Once the file gets uploaded, instead of getting a 201 response as expected, a 200 is returned and the content of the response is the HTML page `Permission Denied: Two factor authentication required`.

wes-otf:
Fixes #4857, adds the upload URL prefix to whitelisted URLs to allow for uploading of files to complete a submission without needing to do 2FA. (also added `/__debug__/` as the toolbars were failing to load)

## Test Steps

 - [ ] With both `ENFORCE_TWO_FACTOR=True` and `FORCE_LOGIN_FOR_APPLICATION=True`, ensure uploading a file as a user without 2FA enabled works as expected.

### DIFF
--- a/hypha/apply/users/middleware.py
+++ b/hypha/apply/users/middleware.py
@@ -19,7 +19,11 @@ TWO_FACTOR_EXEMPTED_PATH_PREFIXES = [
     "/logout/",
     "/account/",
     "/apply/submissions/success/",
+    "/upload/upload/",
 ]
+
+if settings.DEBUG:
+    TWO_FACTOR_EXEMPTED_PATH_PREFIXES.append("/__debug__/")
 
 
 def get_page_path(wagtail_page):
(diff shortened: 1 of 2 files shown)


wes-otf:
this test is proving a little annoying to fix

wes-otf:
This works but I'm questioning how effective those tests actually are in general, from what I'm seeing a 200 is normally still returned but the HTML is the `Permission Denied: Two factor authentication required` page so I don't know if status code checking is enough/useful

frjo (approved):

frjo:
@wes-otf Nice work on finding this bug, a real edge case.

wes-otf:
@frjo it actually kept coming up for OTF and was driving us nuts, thankfully @Techslammer correlated 2FA to the error today

PR #4865, opened 2026-06-11
Correcting a __hash__ function, has no practical effect due to the unique_together constraints but it is still not correct

### LINKED ISSUE #4864: Some forms do not follow Python's `__hash__` contract
In another context (creation of a new one of these forms), an LLM (Qwen3.6) raised a concern about the Python contract around `__hash__` overrides.

See `hypha/apply/funds/models/forms.py` where this function is defined in several classes:

```python
    def __hash__(self):
        fields = [field.id for field in self.fields]
        return hash((tuple(fields), self.sort_order, self.pk))
```

The issue is the inclusion of `self.pk` which is a mutable field. It starts as `None` and then is assigned a value on save.

See https://docs.python.org/3/reference/datamodel.html#object.__hash__:
> If a class defines mutable objects and implements an __eq__() method, it should not implement __hash__(), since the implementation of [hashable](https://docs.python.org/3/glossary.html#term-hashable) collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

For a bit I thought Django itself violated this, but on closer inspection, Django follows the contract.

Do we need to simply remove `self.pk` from the args and `return hash((fields, self.sort_order))`? What bad thing would happen then?

frjo:
Fixes #4864

## Test Steps

 - [ ] Test that added/removed reviewers are are calculated correctly, shown in notifications.

### DIFF
--- a/hypha/apply/funds/models/assigned_reviewers.py
+++ b/hypha/apply/funds/models/assigned_reviewers.py
@@ -184,7 +184,7 @@ class AssignedReviewers(models.Model):
         verbose_name_plural = _("assigned reviewers")
 
     def __hash__(self):
-        return hash(self.pk)
+        return hash((self.reviewer_id, self.role_id))
 
     def __str__(self):
         return f"{self.reviewer}"


wes-otf:
If there's no effect should we keep it at all? I haven't tested to maybe I'm misunderstanding

frjo:
The code is working as is so we do not *need* to make any changes. This PR is about correctness.

PR #4867, opened 2026-06-12
Updated sandbox_db.dump with migrations

wes-otf:
Update the sandbox_db.dump with the latest migrations


wes-otf:
@frjo if this looks good I'll get the sandbox rebased and deployed

frjo (approved):

PR #4884, opened 2026-06-20
Ad hoc reports

### LINKED ISSUE #4529: Ad-hoc reporting
As it stands reports can either be one time or ran on a specific frequency. It has been requested to have the option to add reports as needed in somewhat of an ad-hoc fashion, as not all projects report on patterned basis.

AFAIK this isn't urgent to this round of testing as most of these projects report close to monthly, but all the staff knew projects with irregular reporting schedules

frjo:
Fixes #4529

Adds ad-hoc report management for projects, giving staff the ability to manually add/edit/delete report due dates rather than relying solely on automated scheduling.

## Test Steps

 - [ ] Confirm that added due dates work as expected.
 - [ ] Confirm that only staff can delete and edit due dates.
 - [ ] Confirm that notification are created when staff delete a report.


hypha/apply/projects/reports/forms.py
```diff
@@ -96,6 +96,42 @@ def save(self, commit=True, form_fields=dict):
         return instance
 
 
+class ReportAddDateForm(forms.Form):
+    end_date = forms.DateField(
+        label=_("Report end date"),
```
wes-otf: Might make more sense for this to be "due date" as it'd be in line with the edit prompt

wes-otf (changes requested):
This looks and works great! I just had a super small nitpick around due vs end date language

wes-otf:
Also maybe adding a description to the PR just for posterity when we do merge it in

hypha/apply/projects/reports/forms.py
frjo: Good catch.

wes-otf (approved):

PR #4890, opened 2026-06-23
Invoice tagging

### LINKED ISSUE #4874: Add indicator of international or domestic to invoices
Finance team voiced the need to differentiate between international or domestic contracts for logistical reasons. This could likely be a bool associated once at the project level then displayed on the invoice view & invoice table.

This also was OTF specific but could see it be utilized by other adopters, happy to hear any thoughts otherwise

frjo:
Fixes #4874

This feature allow staff to add any invoice tags hey need in wagtail admin and then use them to tag invoices. Tags will show up on the invoice detail page and in the invoice table.

## Test Steps

 - [ ] That staff can add/edit invoice tags in wagtail admin
 - [ ] That invoices can be tagged via modal .
 - [ ] That tags show up in detail view.
 - [ ] That tags show up in the invoice table.


wes-otf:
this looks good! I think my only feedback would be that if possible it'd be awesome to get the tags on the project invoice table in the project view - do you think that'd be too crowded?

frjo:
I will test that.

frjo:
@wes-otf  It is this table on the project detail you are thinking about?

Skärmavbild 2026-07-03 kl  10 00 16

To add the tags here I think we ned to remove something. Do both the dates need to be listed here?

wes-otf:
Lets actually just merge it for now then I'll come through with another PR if they give feedback otherwise - might be a minute before I can actually get feedback

wes-otf (approved):

PR #4892, opened 2026-06-23
Use cotton attachment card everywhere an attachment is shown

### LINKED ISSUE #4891: Show an error when an attachment can't be found
Likely a super edge case, but handling a missing attachment for invoices would be beneficial

wes-otf:
Closes #4891. I found myself duplicating the logic that was in `templates/stream_forms/includes/file_field.html` into the cotton templates. With the push we've made to cotton lately, it felt more logical to wrap that logic all into one component and have a uniform display for attachments throughout Hypha.

This also handles for missing attachments in comments & invoices as well as the existing submission & reports.

## Test Steps

 - [ ] Ensure that missing attachments are properly displayed for submissions/comments/reports/invoices
 - [ ] Ensure all attachments display uniformly across Hypha and the attachment card works as expected


hypha/apply/stream_forms/files.py
```diff
@@ -78,6 +78,9 @@ def url(self):
 
     @property
     def size(self):
+        if not self.exists:
+            # Prevent an error for when a file doesn't exist
+            return 0
```
wes-otf, line 83: Having this return 0 allows for editing of things like reports when the attachment has been deleted instead of it erroring out because it can't find the file

hypha/apply/stream_forms/templates/stream_forms/render_file_field.html
```diff
@@ -2,7 +2,12 @@
 {% block data_display %}
     {% if data.name %}
         <div class="w-full">
-            {% include "stream_forms/includes/file_field.html" with file=data %}
+            <c-card-attachment
+                href="{{ data.url }}"
+                :filename="data.filename"
+                :exists="data.exists"
+                :modification_time="file.modification_time"
```
frjo: Should be `data.modification_time` I think.

hypha/templates/cotton/card_attachment.html
```diff
@@ -2,12 +2,46 @@
...
+                    <relative-time datetime={{ modification_time|date:"c" }}>{{ modification_time|date:"SHORT_DATE_FORMAT" }}</relative-time>
+                </span>
+            {% endif %}
+        </span>
+        <span>
+            {% heroicon_mini "arrow-down" class="w-5 h-5" stoke_width=2 aria_hidden=true %}
+        </span>
+    </a>
+{% else %}
+    <div class="flex justify-between items-center py-2 px-3 mb-2 max-w-2xl font-medium rounded-sm border opacity-60 not-prose bg-base-200"
```
frjo: max-w-2xl -> max-w-xl to make it the same for if file exist or not?

frjo (changes requested):
Two minor things to look at, otherwise a solid improvement.

wes-otf:
Both should be resolved! Thanks for the flags @frjo!

frjo (approved):

PR #4894, opened 2026-06-24
CSV export for invoices

### LINKED ISSUE #4875: Exporting of invoice table to CSV/XSLX
A very similar flow to how we export the "All Submissions" tab, but just the ability to export specific fields of invoices that are being filtered in the invoice table to CSV has been requested

frjo:
Fixes #4875

Add CSV export for invoices, implemented in the same way as submission CSV export.

## Test Steps

 - [ ] Test that CSV exports work


wes-otf:
this works great! maybe we wait for #4890 to get merged then also serialize the tags into the CSV? I think it's ready to go otherwise

frjo:
@wes-otf I have now added invoice tags to the export.

hypha/apply/projects/tasks.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+        request_user = User.objects.get(pk=request_user_id)
+
+        if current := InvoiceExportManager.objects.filter(user=request_user):
+            current.delete()
+
+        export_manager = InvoiceExportManager.objects.create(
+            user=request_user, total_export=len(qs_ids)
+        )
+        export_manager.export_data = export_invoices_to_csv(qs.iterator())
```
wes-otf: After a ton of debugging as to why the `iterator` call seemed to quietly fail and result in a 404, I found this in the django docs:

> [`iterator()` is compatible with previous calls to `prefetch_related()` as long as chunk_size is given. Larger values will necessitate fewer queries to accomplish the prefetching at the cost of greater memory usage.](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#iterator)

The following (the usual `chunk_size` default when `prefetch_related` isn't used) seemed to work for me:
```suggestion
        export_manager.export_data = export_invoices_to_csv(qs.iterator(chunk_size=2000))
```

wes-otf (changes requested):
Small change needed that I left a comment for - I couldn't get past the [catch all 404](https://github.com/HyphaApp/hypha/blob/af7f73c5fa054247ddb3b4327592b81fc864885c/hypha/apply/projects/views/payment.py#L638) that was really stumping me when testing yesterday

wes-otf:
New tags all exported as expected though and the prefetch optimizations will definitely be helpful!

hypha/apply/projects/tasks.py
frjo: Good catch! Fixed now.

wes-otf (approved):
Works great!

PR #4897, opened 2026-07-01
Allow a submission to have multiple projects associated

### LINKED ISSUE #4869: Allow for submissions to be one to many with projects
It was identified recently that while an edge case, it could be possible for a submission being accepted to result in two different projects needed to be created to manage different contracts being executed. As it stands, Hypha is built to be 1:1 submissions to projects in both backend and UI which makes this a bit of an overhaul.

frjo:
Fixes #4869

## Test Steps

 - [ ] Tests that PROJECTS_ALLOW_MULTIPLE, default false, makes everything work as before, single project per submission.
 - [ ] Test that setting PROJECTS_ALLOW_MULTIPLE to true allow multiple projects to be created.
 - [ ] Test that multiple projects works with reports, invoices etc. Functionality within each project should be unchanged.
 - [ ] Test that all project url:s work as expected. They are now in the form "apply/submissions/1234/project/4/". The last part, project id, was added to handle multiple projects per submission.


hypha/apply/funds/templates/funds/includes/projects_tab.html
```diff
@@ -0,0 +1,53 @@
+{% load i18n heroicons project_tags %}
+{% comment %}
+Renders the "Project(s)" tab inside a submission/project tablist.
+
+Params:
+  submission       – the ApplicationSubmission
+  user             – the current user (for access filtering)
+  current_project  – the project being viewed, if any (for active item highlight)
+  active           – whether this tab is the active one
+{% endcomment %}
```
wes-otf, line 10: I like this format of kinda docstringing a template! super useful when trying to track down what the variables used are & where they come from

wes-otf (approved):
This is such an elegant solution! Did a pretty intensive test and it seems like everything is working as expected. The fact that I even changed `MULTIPLE_PROJECTS_ENABLED=1` → `MULTIPLE_PROJECTS_ENABLED=0` and nothing broke but preserved the multiple projects while not allowing multiples to be created moving forward was also really impressive. All the `x.submission.id` → `x.pk` changes look like they must've been painful hahaha.

Overall great work!

wes-otf (approved):
Works the same after rebases!

PR #4899, opened 2026-07-07
Fix `can_view_submission` ignoring ReviewerSettings configured in wagtail admin

wes-otf:
Fixed a permission bug that ignored the [ReviewerSettings](https://github.com/HyphaApp/hypha/blob/6b3aa1f72d6b30efa66641f968cff9813873e878/hypha/apply/funds/models/reviewer_role.py#L42) that were set in Wagtail admin. 

By default (no settings configured) a reviewer should have access to all submissions. But in Wagtail, setting can be configured to scope a reviewer's specific access to things like assigned applications, specific application outcomes, etc. This was not being considered by `can_view_submission`, which would give reviewers access to things like attachments on applications they wouldn't otherwise have access to.

The logic for ReviewerSettings was implemented on the `ReviewerSubmissionDetailView` alone so the detail view would account for these settings but not the `can_view_submission` function itself.

Thank you to https://github.com/Santoshkumarpuppala for discovering and responsibly disclosing this bug!

## Test Steps

 - [ ] Ensure that with no ReviewerSettings configured/enabled in Wagtail admin, reviewers are able to access all submissions (both in detail view and see them displayed in the All Submissions view)
 - [ ] Ensure that with ReviewSettings configured, all permissions act as expected (ie. if a submission cannot be viewed, confirmed associated files also return a 403)

### DIFF
--- a/hypha/apply/funds/permissions.py
+++ b/hypha/apply/funds/permissions.py
@@ -1,10 +1,13 @@
+from django.apps import apps
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.utils.translation import gettext as _
 from rolepermissions.permissions import register_object_checker
 
 from hypha.apply.funds.models.co_applicants import CoApplicant, CoApplicantRole
+from hypha.apply.funds.models.reviewer_role import ReviewerSettings
 from hypha.apply.funds.models.submissions import DRAFT_STATE
+from hypha.home.models import ApplyHomePage
 
 from ..users.roles import STAFF_GROUP_NAME, SUPERADMIN, TEAMADMIN_GROUP_NAME, StaffAdmin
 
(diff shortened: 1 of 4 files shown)


hypha/apply/funds/permissions.py
```diff
@@ -210,10 +213,21 @@ def can_view_submission(user, submission):
...
+            ).filter(pk=submission.pk).exists(), ""
+        else:
+            return True, ""
+
     if (
         user.is_apply_staff
         or submission.user == user
-        or user.is_reviewer
         or submission.co_applicants.filter(user=user).exists()
     ):
```
frjo: I think it would be most correct to move these above `if user.is_reviewer:`.

If e.g. a user is staff and reviewer the staff role should take precedence.

hypha/apply/funds/views/all.py
```diff
@@ -129,13 +129,11 @@ def submissions_all(
...
         qs = ApplicationSubmission.objects.current()
 
-    # Reviewers also have access to this view but should only see a subset of submissions.
+    # By default, reviewers can see all submissions. This can be configured in Wagtail Admin > Apply > Reviewer Settings
     if request.user.is_reviewer:
         reviewer_settings = ReviewerSettings.for_request(request)
         if reviewer_settings.use_settings:
             qs = qs.for_reviewer_settings(request.user, reviewer_settings)
-        else:
-            qs = qs.filter(reviewers=request.user)
```
frjo, line 138: This was good to get cleared up.

frjo (changes requested):

hypha/apply/funds/permissions.py
wes-otf: totally makes sense!

wes-otf:
should be fixed now @frjo !

frjo (approved):

PR #4905, opened 2026-07-13
Add a column and filter for Contractor in the project table.

### LINKED ISSUE #4873: Add contractor filtering to the project table
In the project table one should be able to filter on all projects for a specific contractor

frjo:
Fixes #4873


wes-otf (approved):
Works great!

PR #4906, opened 2026-07-15
Add contract number to projects

### LINKED ISSUE #4870: Inputting, tracking & displaying of contract numbers for projects
In OTF's workflow, a project will always have one contract associated with it. This should be able to be populated at the beginning of the project's creation and (again specific to OTF) would typically be populated with someone that has the role of Finance who would ideally get a notification when a project enters a draft phase. This number should also be displayed in the project table

It would be great to think through how this can be made a bit more general (or optional) to apply to other adopters, or if this should be something that happens on the OTF specific fork.

frjo:
Fixes #4870

This PR adds a contract number field to the project detail view. It can be edited in a modal. It can be shown in the project table but is excluded by default via the `PROJECTS_TABLE_EXCLUDED_FIELDS` setting.

Only staff, contracting and finance can edit the contract number but all who can view the project can see it.


hypha/apply/projects/tables.py
```diff
@@ -200,6 +200,7 @@ class BaseProjectsTable(tables.Table):
         verbose_name=_("Status"), accessor="get_status_display", order_by=("status",)
     )
     fund = tables.Column(verbose_name=_("Fund"), accessor="submission__page")
+    contract_number = tables.Column(verbose_name=_("Contract nr."), default="-")
```
wes-otf: How do you feel about this? 2 chars less and potentially a little more intuitive but maybe that's very english specific
```suggestion
    contract_number = tables.Column(verbose_name=_("Contract #"), default="-")
```

wes-otf (changes requested):
This works really nice! Left a super nitpick of a comment.

Do you think it could also be squeezed into the invoice table? That was a request that came from staff as well. Could see that being useful on CSV export.

hypha/apply/projects/tables.py
frjo: Change it to "Contract #" both in the table and on the project detail view now. Having it the same look better and less to translate.

wes-otf (approved):
Excellent!

PR #4934, opened 2026-08-05
Add coment visibility options for when a coapplicant exists on an application

wes-otf:
As it stands applicants do not have any visibility options and are always set to the default of "Applicant" - there may be scenarios where an applicant only wants to staff to see a comment, not the other coapplicants

## Test Steps

Ensure that with only the role of applicant, an application with...
 - [ ] ...no coapplicants does not show a visibility section on the comments form
 - [ ] ...one or more coapplicants does show a visibility section with the options of "Applicants" and "Staff only"


frjo (approved):
Works well in my testing.

PR #4944, opened 2026-08-12
Passwordless users can also use elevate

frjo:


wes-otf (approved):

wes-otf:
what do you do locally to test OAuth? or does it just need to be tested on the test apply site

approved because code looked good to me

PR #4952, opened 2026-08-19
Remove project id from Invoice urls

frjo:
This PR touches a lot of files and does not change any functionality. It makes the code neater in my opinion but is not strictly necessary.

Three changes are being made:

1. Invoice urls no longer contain the project is, aligning with how report urls are contracted.
2. Renaming the three `payment.py` files to `invoice.py`, they contain invoice models, views etc.
3. One bug fix managed to creep in, "c-modal" was misspelled in `invoice_confirm_delete.html`

The invoice url change would be a hassel to do when users are using the project part actively. So good to do now if we want the change.


hypha/apply/projects/templates/application_projects/modals/invoice_confirm_delete.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n %}
 
-<c-model.confirm title="{% trans 'Delete invoice' %}">
```
wes-otf: was this just broken before and we didn't notice?

wes-otf (approved):
good improvements! I remember it was annoying having to find invoices under payment when I first came up to speed with the codebase, definitely more intuitive.

hypha/apply/projects/templates/application_projects/modals/invoice_confirm_delete.html
frjo: Yes, we never tried to delete an invoice.

PR #4953, opened 2026-08-19
Nice row attributes that makes the row higlight on hover and make the whole row clickable

frjo:
Was in place for invoice table but overridden due to bug. Also added to more tables.


wes-otf (approved):
works nice!

PR #4958, opened 2026-08-20
Add django-upgrade and pyupgrade to pre-commit.

frjo:
Django upgrade, see https://django-upgrade.readthedocs.io/, is a neat tool that helps with fixing code when updating Django.

pyupgrade, see https://github.com/asottile/pyupgrade, same idea but for python.

They found a few instances of outdated code that it updated. No change in functionality.

I think these tools will be helpful the next time we update Django/Python. They will also help us avoid old style code.


wes-otf (approved):
Looks nice to me, especially like django upgrade's vaporwavey logo

PR #4974, opened 2026-09-01
Make category question widget render select input if more than 32 options for both single and multi select fields

frjo:
We have had this in place for multichoice fields for years but for some reason missed it for single choice fields.

Having e.g. 50 checkboxes or radioboxes on a field is just too many. I think 32 is a resonable limit to switch and it has been in place for multiselect for years.


wes-otf (approved):
Works great! I'd even be fine capping the options at something less than 32 but this works as is!

PR #4976, opened 2026-09-02
Add applicant determination access setting

frjo:
Adds a new setting `DETERMINATION_DETAILS_ACCESS_APPLICANT` that default to True. So no change in behaviour unless the setting is changed from the default.

Setting it to False will hide `detailed_data` from applicants.

Also fixed a view permission bug in ApplicantDeterminationDetailView.


hypha/apply/determinations/templates/determinations/determination_detail.html
```diff
@@ -49,28 +49,30 @@ <h2 class="pb-1 mb-2 font-medium border-b text-h3 border-base-300 question">
...
+                        <h2 class="section-header">{{ group.title|nh3 }}</h2>
+                    {% endif %}
 
-                {% for question, answer in group.questions %}
-                    <h3 class="pb-1 mb-2 font-medium border-b text-h3 border-base-300 question">{{ question }}</h3>
-                    {% if answer %}
-                        {% if answer == True or answer == False %}
-                            {{ answer|yesno:_("Agree,Disagree") }}
+                    {% for question, answer in group.questions %}
+                        <h3 class="pb-1 mb-2 font-medium border-b text-h3 border-base-300 question">{{ question }}</h3>
```
wes-otf: super nitpick & I don't think it was introduced in this PR but should we add a little more space between `detailed_data` questions? maybe even to match the `gap-8` of the other section separators? seems a little scrunched together

Image

wes-otf (changes requested):
It looks pretty good to me! Left a super small comment and also wondering if there's a way to indicate to staff what fields the applicants will see - while it seems intuitive in wagtail I could see staff accidentally writing something in the required `Message` field that applicants weren't supposed to see if it wasn't labeled correctly.

Open to thoughts on that though!

hypha/apply/determinations/templates/determinations/determination_detail.html
frjo: @wes-otf  How does this look?

Skärmavbild 2026-09-10 kl  09 01 25

wes-otf: This looks perfect! I was also thinking maybe a divider on the form itself so staff knew before submitting what fields would be seen. Let me know if you think that makes sense too? outside of that this is ready to go!

frjo: Will take a look at that.

frjo: Pushed a fix for the create/edit form now. Just a banner at the top. Tried to do it per field but it was a bit too complex.

Skärmavbild 2026-09-15 kl  09 46 41

wes-otf:
This works!

wes-otf (approved):

PR #4985, opened 2026-09-15
Handle `can_publish` seperately from the wagtail `CopyForm`

wes-otf:
Quick fix for copying rounds directly in wagtail - currently it will error out if copied due to wagtail not accepting `can_publish` as an argument

## Test Steps

 - [ ] Ensure rounds can be directly copied


wes-otf:
@frjo print statements removed 🤦

hypha/apply/funds/admin_views.py
```diff
@@ -22,10 +22,10 @@ def custom_admin_round_copy_view(request, page):
 
     # Check if the user has permission to publish subpages on the parent
     can_publish = parent_page.permissions_for_user(request.user).can_publish_subpage()
+    if not can_publish:
+        raise PermissionDenied(_("You do not have permission to publish this round."))
```
frjo: Is this part needed at all? I tried removing it with no apparent ill effect.

```
# Parent page defaults to parent of source page
parent_page = page.get_parent()

# Check if the user has permission to publish subpages on the parent
can_publish = parent_page.permissions_for_user(request.user).can_publish_subpage()
if not can_publish:
    raise PermissionDenied(_("You do not have permission to publish this round."))

```

wes-otf: I'm not sure - I also didn't see anything nasty happen when I removed them as well but wasn't sure if there was an edge case or reason that logic was there in the first place? happy to remove it though

frjo: I suspect it is code to handle old Wagtail behaviour.

wes-otf: should be set now!

PR #4987, opened 2026-09-15
Disbursements flow

bickelj:
Note: this is a draft, not ready for merge by any stretch of the imagination (I don't expect lint or tests to even pass yet). I cherry-picked the relevant commits into a first-pass on main here while the commit order was top of mind. There will be more commits added as the feature evolves downstream.

Without this change, there is quite a complex process for invoicing, approvals, and tracking of payments. With this change, there is a relatively straightforward way to associate money amounts with contracts and then disbursements with those contracts. It breaks the 1-to-1 project-to-contract relationship when a system setting `PROJECTS_PAYMENTS_FLOW` is set to `DISBURSEMENTS`. The standard flow is `INVOICING` and another is offered: `DISABLED`. The same `Contract` model is re-used in either flow, but after that, the models/forms/views diverge.


hypha/apply/projects/forms/project.py
```diff
@@ -33,9 +39,23 @@
...
+    "EUR",
+    "CNY",
+    "INR",
+    "JPY",
+    "RUB",
+    "IDR",
+    "BRL",
+    "MXN",
+    "NGN",
+]
```
frjo, line 57: I think this should be a setting. The values seem good as a default.

PR #4992, opened 2026-09-18
Update to biome v2

frjo:
Update to Biome v2, it has been out for a long time.


docs/references/notifications.md
```diff
@@ -49,7 +49,7 @@ def make_email_object(self, to: str | List[str], context, **kwargs):
     if not isinstance(to, (list, tuple)):
         to = [to]
 
-    lang = context.get('lang', None) or settings.LANGUAGE_CODE
+    lang = context.get("lang", None) or settings.LANGUAGE_CODE
```
wes-otf, line 52: Did you manually add this or should we manually exclude `*.md`\docs? could get messy if it's trying to lint markdown/python/bash in the docs

wes-otf:
one small comment but ready to go otherwise!

docs/references/notifications.md
frjo: Excluding the docs sounds like a good approach. As you say, this could get messy.

PR #4993, opened 2026-09-18
Determination message missing in batch runs

frjo:
A bug made determination messages not being added to the e-mails to applicants when doing batch determinations.


hypha/apply/determinations/views.py
```diff
@@ -430,13 +442,6 @@ def form_valid(self, form):
             return HttpResponseRedirect(self.submission.get_absolute_url())
 
         with transaction.atomic():
-            messenger(
-                MESSAGES.DETERMINATION_OUTCOME,
-                request=self.request,
-                user=self.object.author,
-                submission=self.object.submission,
-                related=self.object,
-            )
```
frjo, line 439: This was old code that has not done anything for a long time.

wes-otf (approved):

frjo:
@wes-otf OTF have not noticed this bug? Do they not use batch feature quite a lot?

